PR #59, opened 2018-09-24
Issue 57: Refactor repeated lines

### LINKED ISSUE #57: Refactor repeated lines on sdv.sampler.sample_rows
On `sdv.sampler.sample_rows`:
https://github.com/HDI-Project/SDV/blob/687d30a090bd2424abf675e66349bb516e4d6a5b/sdv/sampler.py#L52-L71
and https://github.com/HDI-Project/SDV/blob/687d30a090bd2424abf675e66349bb516e4d6a5b/sdv/sampler.py#L101-L117

 are basically identical. It could be a good idea to move that code to a separate method `sdv.sampler.transform_sampled_rows` to reduce duplicated code.

ManuelAlvarezC:
The repeated lines in the method `sdv.Sampler.sample_rows` have been moved to a separate method `sdv.Sampler.transform_synthesized_rows`.

Resolves #57


sdv/sampler.py
```diff
@@ -27,16 +74,6 @@ def sample_rows(self, table_name, num_rows):
...
-
-        # get primary key column name
-        primary_key = meta.get('primary_key')
-
-        if primary_key:
-            node = meta['fields'][primary_key]
-            regex = node['regex']
-            int_primary_key = (node['type'] == 'number') and (node['subtype'] == 'integer')
 
         for parent in parents:
```
csala: I would redo all of this in the following way:

Create a method that returns the random parent and a random row from it.

This method also validates that, if there are parents, all are sampled:

```
def _get_parent_row(self, table):
    parents = self.dn.get_parents(table_name)
    if not parents:
        return None

    parent_rows = dict()
    for parent in parents:
        if parent not in self.sampled:
            raise Exception()

        parent_rows[parent] = self.sampled[parent]

    random_parent, parent_rows = random.choice(list(dict().items())

    return random_parent, random.choice(parent_rows)
```

After this, in the `sample_rows` method you just do:

```
def sample_rows(self, table_name, num_rows):
    parent_row = self._get_parent_row(self, table_name)
    if parent_row:
        random_parent, parent_row = parent_row
        ....
    else:    # there is no parent
        ...
```

PR #87, opened 2019-02-05
Issue 81: enforce sampled values in range [0,1] for  categorical columns

### LINKED ISSUE #81: Enforce sampled values by model in range [0,1] for categorical data.
Fields that are labeled as `categorical` in the metadata, will be transformed using `rdt.transformers.CatTransformer`, that maps the set of categories into the interval [0, 1].

That means that when fitting the model, all the values for the given column are in such range, however, when sampling, the sampling values aren't always located in that interval, and are ignored by the categorical transformer and returned unchanged.

ManuelAlvarezC:
A method for rescaling values into the range [0, 1] in the case some are sampled out of it. These changes should fix the bug where some times, when sampling categorical values, numbers are returned on some rows instead of the expected categories.

Resolves #81

### DIFF
--- a/sdv/sampler.py
+++ b/sdv/sampler.py
@@ -48,11 +48,11 @@ class Sampler:
         scaled = scaler.fit_transform(column.values.reshape(-1, 1)).ravel()
         return pd.Series(scaled, name=column.name)
 
-    def transform_synthesized_rows(self, synthesized_rows, table_name, num_rows):
+    def transform_synthesized_rows(self, synthesized, table_name, num_rows):
         """Add primary key and reverse transform synthetized data.
 
         Args:
-            synthesized_rows(pandas.DataFrame): Generated data from model
+            synthesized(pandas.DataFrame): Generated data from model
             table_name(str): Name of the table.
             num_rows(int): Number of rows sampled.
 
@@ -70,8 +70,8 @@ class Sampler:
 
         if categorical_fields:
             for field in categorical_fields:
-                if ((synthesized_rows[field] < 0) | (synthesized_rows[field] > 1)).any():
-                    synthesized_rows[field] = self._rescale_values(synthesized_rows[field])
+                if ((synthesized[field] < 0) | (synthesized[field] > 1)).any():
+                    synthesized[field] = self._rescale_values(synthesized[field])
 
         if primary_key:
             node = meta['fields'][primary_key]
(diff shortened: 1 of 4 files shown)


sdv/sampler.py
```diff
@@ -48,6 +63,15 @@ def transform_synthesized_rows(self, synthesized_rows, table_name, num_rows):
         meta = self.dn.tables[table_name].meta
         orig_meta = self._get_table_meta(self.dn.meta, table_name)
         primary_key = meta.get('primary_key')
+        categorical_fields = [
```
csala: Instead of:
* look over the fields list once here
* evaluate an if for each one of them
* evaluate an if to see if there are any categorical fields
* loop over the categorical fields again

I would loop just once and have an "if categorical: check values and rescale if necessary" inside the loop.

sdv/sampler.py
```diff
@@ -48,6 +63,15 @@ def transform_synthesized_rows(self, synthesized_rows, table_name, num_rows):
...
+            field['name'] for field in orig_meta['fields']
+            if field['type'] == 'categorical'
+        ]
+
+        if categorical_fields:
+            for field in categorical_fields:
+                if ((synthesized[field] < 0) | (synthesized[field] > 1)).any():
```
csala: I'd capture the `synthesized[field]` in a variable to avoid having to access it every time we use it.

sdv/sampler.py
```diff
@@ -33,11 +34,25 @@ def reset_indices_tables(sampled_tables):
...
+        """Scale an array of values in the interval [0, 1].
+
+        Args:
+            column (pandas.Series): Column values to scale
+
+        Return:
+            pandas.Series: Column values scaled into [0,1]
+        """
+        scaler = MinMaxScaler()
+        scaled = scaler.fit_transform(column.values.reshape(-1, 1)).ravel()
```
csala: Since this scaling will be different at every call depending on the sampled values, aren't we altering probability distributions here?

ManuelAlvarezC: Yes, right. I changed the implementation.

sdv/sampler.py
```diff
@@ -405,6 +407,75 @@ def unflatten_model(self, parent_row, table_name, parent_name):
 
         return self.modeler.model.from_dict(model_parameters)
 
+    def __get_missing_valid_rows(self, synthesized, drop_indices, valid_rows, num_rows):
```
csala: Why is this method prefixed with a double underscore?

ManuelAlvarezC: This was a typo

PR #99, opened 2019-05-08
Issue 63: unicity primary keys

### LINKED ISSUE #63: Ensure unicity on primary keys on different calls
Currently, primary keys are generated using `exrex` module and the regex from the `meta.json` file.The way it's implemented, if we sample a single time, we are guaranteed that the primary keys will be unique, however, if we sample more than once, it's possible to obtain again keys that have been returned in the previous call.

Should we ensure uniqueness in this scenario?
Note that if we do this, we will only be able to sample as many rows as different matches the regex allows, afterwards we'll need a way to reset the database before sampling anything else.

For example, if we had a dataset consisting of a single table, with a single column, which is the primary key with regex `[1-5]{1}`, then the following could happen:

```python
>>>...
>>>first_samples = sampler.sample_all(num_rows=3)
>>>first_samples.T
   primary_key
0            1
1            2
2            3

# Then it's no guaranteed that if we sample a single row more, it's primary key will be neither 4 or 5
>>> second_sample = sampler.sample_all(num_rows=1)
>>> second_sample
   primary_key
0            3
```

ManuelAlvarezC:
Resolves #63

This PR contains the following changes:

- Raise the generator-related errors before the actual sampling.
- Add option to reset generators.
- Expose arguments controlling new behavior on `SDV` class.
- Update unittests to cover new behavior

### DIFF
--- a/sdv/sampler.py
+++ b/sdv/sampler.py
@@ -531,16 +531,19 @@ class Sampler:
             parents = bool(self.dn.get_parents(table_name))
             raise ValueError(MODEL_ERROR_MESSAGES[parents])
 
-    def sample_rows(self, table_name, num_rows):
+    def sample_rows(self, table_name, num_rows, reset_pks=False):
         """Sample specified number of rows for specified table.
 
         Args:
-            table_name (str): name of table to synthesize
-            num_rows (int): number of rows to synthesize
+            table_name(str): Name of table to synthesize.
+            num_rows(int): Number of rows to synthesize.
+            reset_pks(bool): Wheter or not reset the pk generators.
 
         Returns:
-            pd.DataFrame: synthesized rows.
+            pd.DataFrame: Synthesized rows.
         """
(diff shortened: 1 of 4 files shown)


sdv/sampler.py
```diff
@@ -505,17 +531,21 @@ def _sample_valid_rows(self, model, num_rows, table_name):
...
-    def sample_rows(self, table_name, num_rows):
+    def sample_rows(self, table_name, num_rows, reset_pks=False):
         """Sample specified number of rows for specified table.
 
         Args:
-            table_name (str): name of table to synthesize
-            num_rows (int): number of rows to synthesize
+            table_name(str): Name of table to synthesize.
+            num_rows(int): Number of rows to synthesize.
+            reset_pks(bool): Wheter or not reset the pk generators.
```
csala: Let's give this argument a more explicit name:  `reset_primary_keys`

PR #133, opened 2019-12-02
Issue 128 adapt evaluation

### LINKED ISSUE #128: Adapt evaluation to the different data types
The current `evaluate` function does not distinguish between the different field types and, most importantly, includes the id fields in the evaluation.

Id fields are not modeled and can be very different between the sampled and the real data, which means that they can produce a huge and meaningless impact on the produced scores.

The `evaluate` function should be changed to expect a `metadata` argument as input and adapt the evaluation to the different data types.

JDTheRipperPC:
Resolves #128 

Allow `evaluate` use only the synthesized data and load the real from csv.
Uses `Metadata.get_dtypes` to discard ids.


sdv/evaluation/__init__.py
```diff
@@ -78,67 +94,164 @@ def get_descriptor_values(real, synth, descriptor):
...
+
     described = list()
     for descriptor in descriptors:
         if isinstance(descriptor, str):
-            descriptor = DESCRIPTORS[descriptor]
+            descriptor, dtypes = DESCRIPTORS[descriptor]
+        elif isinstance(descriptor, tuple):
+            descriptor, dtypes = descriptor
+        else:
+            dtypes = ['int', 'float', 'object', 'bool', 'datetime64']
```
csala: Make this a module level constant variable: `DEFAULT_DTYPES`

sdv/evaluation/__init__.py
```diff
@@ -78,67 +94,164 @@ def get_descriptor_values(real, synth, descriptor):
...
+        real (dict[str, pandas.DataFrame] or pandas.DataFrame):
+            Map of names and tables of real data. Defaults to ``None``.
+        metadata (str, dict or Metadata):
+            String or dictionary to instance a Metadata object or a Metadata itself.
+            Defaults to ``None``.
         descriptors (list[callable]):
             List of descriptors.
         metrics (list[callable]):
             List of metrics.
+        root_path (str):
```
csala: I would put `root_path` right next to `metadata`:

```
def evaluate(synth, real=None, metadata=None, root_path=None, descriptors=...)
```

sdv/evaluation/__init__.py
```diff
@@ -78,67 +94,164 @@ def get_descriptor_values(real, synth, descriptor):
...
+
+        scores[name] = table_scores
+
+    scores = pd.DataFrame(scores).T
 
-    for metric in metrics:
-        scores[metric.__name__] = metric(real_descriptors, synth_descriptors)
+    if len(scores) > 1:
+        mean = scores.mean()
+        mean.name = '__mean__'
```
csala: Let's use `'mean'` instead of `__mean__`.
If the name `mean` already exists in the DF, this will duplicate it, but it's OK.

PR #145, opened 2020-01-16
Issue 134 add metadata validation

### LINKED ISSUE #134: Add Metadata Validation
Add methods to validate the metadata at two levels:
- Whether the metadata specification is valid and supported by SDV.
- Whether the metadata specification truly represents the given data.

JDTheRipperPC:
Resolves #134 

Added ``validation`` method and moved ``_validate_dataset_structure`` from sdv.py to metadata.py.

### DIFF
--- a/sdv/metadata.py
+++ b/sdv/metadata.py
@@ -120,112 +120,6 @@ class Metadata:
                         self._child_map[parent].add(table)
                         self._parent_map[table].add(parent)
 
-    def _validate_parents(self):
-        """Make sure that all the tables have at most one parent."""
-        for table in self.get_tables():
-            if len(self.get_parents(table)) > 1:
-                raise ValueError('Some tables have multiple parents, which is not supported yet.')
-
-    def _validate_table(self, table_name, table_meta, table_data=None):
-        """Validate metadata table.
-
-        Validate the type and subtype combination for each field in ``table_meta``
-        and in case a field of type ``id`` exists it must be the ``primary_key`` or
-        must have a ``ref`` entry.
-
-        If ``primary_key`` entry exists, check that it's an existing field
-        and its type is ``id``.
-
-        If ``table_data`` is provided, check all the data types for the table
-        correspond to each column and all the data types exists on the table.
-
-        Args:
-            table_name (str):
-                Table name to be validated.
-            table_meta (dict):
-                Table metadata to be validated.
-            table_data (pandas.DataFrame):
-                If it's provided, check the metadata types are valid and the
-                metadata fields exists in data columns.
-        """
-        dtypes = self.get_dtypes(table_name, ids=True)
-
-        # Primary key field exists and its type is 'id'
-        primary_key = table_meta.get('primary_key')
-        if primary_key:
-            pk_field = table_meta['fields'].get(primary_key)
-
-            if not pk_field:
-                raise MetadataError('Primary key is not an existing field.')
-
-            if pk_field['type'] != 'id':
-                raise MetadataError('Primary key is not of type `id`.')
-
-        if table_data is not None:
-            for column in table_data:
-                try:
-                    table_data[column].astype(dtypes[column])
-                    del dtypes[column]
-                except ValueError as ve:
-                    message = 'Invalid values found in column {} of table {}: {}'.format(
-                        column, table_name, ve)
-                    raise MetadataError(message) from None
-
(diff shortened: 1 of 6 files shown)


sdv/metadata.py
```diff
@@ -116,6 +116,76 @@ def _analyze_relationships(self):
                         self._child_map[parent].add(table)
                         self._parent_map[table].add(parent)
 
+    def _validate_dataset_structure(self):
```
csala: Rename this to something more meaningful, like `_validate_parents`

sdv/metadata.py
```diff
@@ -116,6 +116,76 @@ def _analyze_relationships(self):
...
+        if primary_key:
+            pk_field = table_meta['fields'].get(primary_key)
+
+            if not pk_field:
+                raise MetadataError('Primary key is not an existing field.')
+
+            if pk_field['type'] != 'id':
+                raise MetadataError('Primary key is not of type `id`.')
+
+        if tables:
```
csala: `if table_data is not None`

Also, let's be more verbose about the failing fields.
Try to capture a `ValueError` and, if so, raise a MetadataError indicating that field xxx has a a wrong type.
Otherwise the user will see something line `ValueError: invalid literal for int() with base 10: 'a'` and will have not clue about which column this comes from.

Do this by concatenating the ValueError message to the user-friendly message, and raising from None to avoid the extra long traceback:

```
except ValueError as ve:
    message = 'Invalid values found in column {} of table {}: {}'.format(
        column, table_name, ve)
    raise MetadataError(message) from None
```

sdv/metadata.py
```diff
@@ -116,6 +116,76 @@ def _analyze_relationships(self):
...
+            data = tables[table_name]
+
+            for column in data:
+                data[column].astype(dtypes[column])
+
+            # assert all dtypes are in data
+            for field in dtypes:
```
csala: Instead of looping a second time, during the first loop `pop` the values out of the `dtypes` dict.
When you reach this point, `dtypes` should be empty, so you can use:

```
if dtypes:
    raise MetadataError(
        'Missing columns on table {}: {}'.format(
            table_name, list(dtypes.keys()))
    )
```

sdv/metadata.py
```diff
@@ -116,6 +116,76 @@ def _analyze_relationships(self):
...
+
+        Args:
+            tables (bool, dict):
+                When is a ``bool`` if ``True`` load the data to validate each
+                field type from the metadata. When is a ``dict`` use this data
+                to validate. Defaults to ``None``
+        """
+        tables_meta = self._metadata.get('tables')
+        if not tables_meta:
+            raise MetadataError('Metadata missing "tables".')
```
csala: "tables" entry not found in Metadata

sdv/metadata.py
```diff
@@ -116,6 +116,76 @@ def _analyze_relationships(self):
...
+
+        data = None
```
csala: no need to create a new variable: reuse `tables`.

```
if tables and not isinstance(tables, dict):
    tables = self.load_tables()
```

sdv/metadata.py
```diff
@@ -116,6 +116,76 @@ def _analyze_relationships(self):
...
+        if isinstance(tables, dict):
+            data = tables
+        elif tables:
+            data = self.load_tables()
+
+        for table_name, table_meta in tables_meta.items():
+            self._validate_table(table_name, table_meta, data)
```
csala: Get `table_data` from `tables` (renamed from `data`).

Also do it safely, so if `table_name` is not in `tables` also raise a MetadataError:

```
if tables:
    table = tables.get(table_name)
    if table is None:
        raise MetadataError('Table `{}` not found in tables'.format(table_name))
else:
    table = None
```

sdv/metadata.py
```diff
@@ -116,6 +116,76 @@ def _analyze_relationships(self):
...
 
+    def _validate_dataset_structure(self):
+        """Make sure that all the tables have at most one parent."""
+        for table in self.get_tables():
+            if len(self.get_parents(table)) > 1:
+                raise ValueError('Some tables have multiple parents, which is not supported yet.')
+
+    def _validate_table(self, table_name, table_meta, tables=None):
+        """Validate metadata table.
+
```
csala: Can we add a short description here about what this method does?

PR #160, opened 2020-06-26
Issue 159 new evaluation

### LINKED ISSUE #159: Use SDMetrics for evaluation
Replace the current `sdv.evaluation` subpackage with a simpler module that uses [SDMetrics](/sdv-dev/SDMetrics).

The API should be as similar as possible to the current one and only the `overall` score should be returned.

csala:
Resolve #159 

Replace evaluation subpackage with new evaluation module that uses SDMetrics.

Also add new `EVALUATION.md` module with usage instructions

Also update the Makefile, add readme and tutorials tests and add github actions.


EVALUATION.md
```diff
@@ -0,0 +1,46 @@
+# SDV Evaluation
+
+After using SDV to model your database and generate a synthetic version of it you
+might want to evaluate how similar the syntehtic data is to your real data.
```
k15z, line 4: typo: syntehtic -> synthetic

EVALUATION.md
```diff
@@ -0,0 +1,46 @@
...
+
+SDV has an evaluation module with a simple function that allows you to compare
+the syntehtic data to your real data using [SDMetrics](https://github.com/sdv-dev/SDMetrics) and
```
k15z, line 7: typo: syntehtic -> synthetic

sdv/evaluation.py
```diff
@@ -0,0 +1,108 @@
...
+        meta_dict = {
+            table: metadata.get_table_meta(table)
+            for table in real.keys()
+        }
+        metadata = Metadata({'tables': meta_dict})
+
+    return synth, real, metadata
+
+
+def evaluate(synth, real=None, metadata=None, root_path=None, table_name=None, get_report=False):
```
k15z, line 74: I think this function signature needs to be updated, real/metadata should be required?

csala: No, the rationale here is that the only thing truly required is the synthetic data.
Then, you can either pass real data without metadata (and we build the metadata on the fly), or metadata without real data (then we load the real data from the metadata) or both.

This logic is embeded in the `_validate_inputs` method, which tries to build whatever is necessary to evaluate from as little input from the user as possible.

PR #183, opened 2020-08-10
Typo in documentation

ludovicc:


csala:
Good catch @ludovicc and thanks for fixing it!
Would you mind signing the CLA before we merge it?

ludovicc:
CLA signed

PR #322, opened 2021-02-19
Add support for conditional sampling 

### LINKED ISSUE #316: Conditional Sampling for tabular models
### Problem Description

This feature covers the functionality of sampling tabular data conditioned on some passed values.

The behavior of the conditioning should be as follows:

1. The `sample` method in the base tabular model should add a new `conditions` argument that can be passed as a `mapping-like` object (`dict`, `pd.Series` etc.) or as a `pd.DataFrame`.
2. The `conditions` should be validated inside the `BaseTabularModel.sample` method and passed down to the `self._sample_rows`.
3. Each subclass `_sample_rows` method should handle the `conditions` argument appropriately, raising a `NotImplementedError` in case they do not support conditioning, or a `ValueError` if the conditions are not valid (like, for example, if 2 or more condition columns are passed to CTGAN).

### Expected behavior

The following snippets would all be valid calls:

**Sample `num_rows` with the indicated condition value**

```python3
model = ATabularModel()
model.fit(data)
conditions = {
    column_name: column_value
}
model.sample(num_rows, conditions=conditions) 
```

**Sample 2 rows, each one with the indicated condition value**

```python3
model = ATabularModel()
model.fit(data)
conditions = pd.DataFrame({
    column_name: [column_value_1, column_value_2]
})
model.sample(conditions=conditions)
```

**Sample `num_rows`, conditioned on 2 values**

```python3
model = ATabularModel()
model.fit(data)
conditions = {
    column_name_1: column_value_1,
    column_name_2: column_value_2,
}
model.sample(num_rows, conditions=conditions)
```

### Current support

Currently CTGAN is the only model that supports conditioning, so overall the behavior should be as follows:

* GaussianCopula and TVAE would raise a NotImplementedError if any conditions are passed
* CTGAN and CopulaGAN would raise a NotImplementedError if more than one condition column is passed, or if conditioning is made in a non-categorical variable. If a `pd.DataFrame` of conditions is passed, CTGAN would be called multiple times, one for each unique value.

fealho:
Resolves #316.


sdv/tabular/base.py
```diff
@@ -185,6 +187,62 @@ def sample(self, num_rows=None, max_retries=100):
...
+
+        transformed_conditions = pd.concat(all_conditions, axis=1)
+        columns = transformed_conditions.columns
+        transformed_conditions["__condition_idx__"] = np.arange(len(transformed_conditions))
+        grouped_conditions = transformed_conditions.groupby(list(columns))
+
+        # sample
+        all_sampled_rows = list()
+        for index, dataframe in grouped_conditions:
+            one_condition = pd.DataFrame([index], columns=columns)
```
csala, line 234: I wonder if we should be passing down a dict (or Series) instead of DataFrame.
At this point all the condition columns in the DataFrame are constant, so the subclass model will implicitly be working on a single value combination at a time anyway.

sdv/tabular/base.py
```diff
@@ -185,6 +187,62 @@ def sample(self, num_rows=None, max_retries=100):
...
+                if len(transformed_conditions.columns) == 0:
+                    raise ValueError()
+                all_conditions.append(transformed_conditions)
+
+            except Exception:
+                raise ValueError(f'Cannot condition on {conditions}') from None
+
+        transformed_conditions = pd.concat(all_conditions, axis=1)
+        columns = transformed_conditions.columns
+        transformed_conditions["__condition_idx__"] = np.arange(len(transformed_conditions))
```
csala, line 228: Good solution! I was thinking about using a `reindex` to restore the order, but it seems like `reindex` would now work with duplicates, so I think this temporary indexes is the only option we have.

sdv/tabular/copulas.py
```diff
@@ -294,6 +294,8 @@ def _sample(self, num_rows):
             pandas.DataFrame:
                 Sampled data.
         """
+        if conditions is not None:
+            raise NotImplementedError("Copula-based models don't support conditional sampling.")
```
csala, line 298: Can we add a blank line below this one?

sdv/tabular/base.py
```diff
@@ -185,6 +187,62 @@ def sample(self, num_rows=None, max_retries=100):
...
+
+    def sample(self, num_rows=None, max_retries=100, conditions=None):
+        """Write this."""
+        if conditions is None:
+            return self._sample_conditioned_rows(num_rows, max_retries, None)
+
+        # convert conditions to dataframe
+        conditions = self._make_conditions_df(conditions, num_rows)
+
+        # transform conditions, will use RDT when that's ready
```
csala, line 213: RDT should be ready for it, but I suspect that we will need to also tackle the `Table.transform` and `Table.reverse_transform` methods to fully support this.
This comment can also be removed, since the loop will stay as it is.

sdv/tabular/base.py
```diff
@@ -135,13 +136,13 @@ def get_metadata(self):
         """
         return self._metadata
 
-    def _sample_rows(self, num_to_sample):
+    def _sample_rows(self, num_to_sample=None, conditions=None):
```
csala, line 139: I'm thinking that everything will be simpler if we do not make `num_to_sample` be ever `None`, and if `conditions` at this point is always a flat mapping (like a `dict` or a `Series).
That will also simplify a bit how the subclasses handle this input.

csala (approved):

PR #335, opened 2021-03-11
Tutorials review

fealho:
Grammar improvements in the tutorials.


csala (changes requested):
It seems that there are a couple of `markdown` cells that were changed to `raw` and now render wrongly, as well as an empty cell that has been added at the end of the CTGAN notebook.

Can you review this @fealho ?

fealho:
Fixed. I was unsure whether the `scrolled=True` and the `version=3.8.8` are a problem, so I left them there, but if you think it will cause issues I can change those back as well.

csala (approved):

PR #354, opened 2021-03-22
Metadata validation improvements

### LINKED ISSUE #348: Verbose metadata validation
### Problem Description

Current metadata validation stops at the first error found.

It would be interesting to change the behavior so that the validation continues and all the errors found are reported at the end.

### Expected behavior

Usage would be the same as before, but the text contained in the raised `MetadataError` would include a list of errors instead of just one message.

csala:
Resolve #348 

Now the validation errors are accumulated in a list and at the end all of them are included in the error message included in the `MetadataError` exception.


pvk-developer (approved):
LGTM!

PR #357, opened 2021-03-23
Fix categorical column after sequence_index column issue

### LINKED ISSUE #314: Categorical column after sequence_index column
Currently, if a dataset contains a categorical column immediately after the `sequence_index` column the `fit` breaks with the error `TypeError: cannot perform reduce with flexible type`. Below is an example that reproduces the issue:

```python3
from sdv.demo import load_timeseries_demo
from sdv.timeseries import PAR
import pandas as pd

data = load_timeseries_demo()
data.insert(loc=2, column="Categorical", value=["a","b"]*(data.shape[0]//2))
print(data)

model = PAR(
    entity_columns = ['Symbol'],
    context_columns = ['MarketCap', 'Sector', 'Industry'],
    sequence_index='Date'
)

model.fit(data)
```

fealho:
Resolve #314.


tests/integration/timeseries/test_par.py
```diff
@@ -47,3 +49,16 @@ def test_par():
...
+def test_column_after_date():
+    """Test that adding columns after the `sequence_index` column works."""
+    date = datetime.datetime.strptime('2020-01-01', '%Y-%m-%d')
+    daily_timeseries = pd.DataFrame({
+        'col': ['a', 'a'],
+        'date': [date, date],
+        'col2': ['hello', 'world'],
+    })
+
+    model = PAR(entity_columns=['col'], sequence_index='date')
```
csala, line 63: Maybe set `epochs=1` to avoid making tests slower?

sdv/timeseries/deepecho.py
```diff
@@ -67,7 +67,8 @@ def _fit(self, timeseries_data):
 
         data_types = list()
         context_types = list()
-        for field, meta in self._metadata.get_fields().items():
+        for field in self._entity_columns + self._data_columns:
+            meta = self._metadata.get_fields()[field]
```
csala, line 71: I would capture the `fields_metadata` in a variable before the loop to avoid having to call `self._metadata.get_fields()` at each iteration.

sdv/timeseries/deepecho.py
```diff
@@ -67,7 +67,8 @@ def _fit(self, timeseries_data):
 
         data_types = list()
         context_types = list()
-        for field, meta in self._metadata.get_fields().items():
+        for field in self._entity_columns + self._data_columns:
```
csala, line 70: I think this will possibly not work because the order of the columns will be altered, and also we would be missing the `context_columns`.

If the order of the key/value pairs from the `self._metadata.get_fields()` is the problem, maybe a possibility would be to just iterate over `self._output_columns` (which is the list of columns from the input data)?

Then, in order to solve the `sequence_index` problem, we could change line 74 (from the old code):

```python
                if field == self._sequence_index:
                    data_types.append('continuous')
```

to

```python
                if field == self._sequence_index:
                    data_types.extend(['continuous', 'continuous'])
```

And then just remove line 82 (from the old code).

tests/integration/timeseries/test_par.py
```diff
@@ -47,3 +49,16 @@ def test_par():
     assert sampled.shape == data.shape
     assert (sampled.dtypes == data.dtypes).all()
     assert (sampled.notnull().sum(axis=1) != 0).all()
+
+
+def test_column_after_date():
+    """Test that adding columns after the `sequence_index` column works."""
+    date = datetime.datetime.strptime('2020-01-01', '%Y-%m-%d')
+    daily_timeseries = pd.DataFrame({
```
csala, line 57: I think it would be worth to make this test slightly more complex, so there are multiple data types and both entity columns and context columns.

csala (changes requested):

sdv/timeseries/deepecho.py
```diff
@@ -67,7 +67,9 @@ def _fit(self, timeseries_data):
 
         data_types = list()
         context_types = list()
-        for field, meta in self._metadata.get_fields().items():
+        fields_metadata = self._metadata.get_fields()
```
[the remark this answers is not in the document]
csala: You are right @fealho I overlooked this

csala (approved):

pvk-developer (approved):
:shipit:

Issue #373, opened 2021-04-02
Request for sdv.constraint.ColumnFormula for a wider range of function


JetfiRex:
### Problem Description

Currently I have tried to use the 
```python
some_constraint = sdv.constraints.ColumnFormula(
    column="some_column",
    formula=(some lambda function),
    handling_strategy='reject_sampling'
)
```                        
It does not report an error for now. However, I then tried to have:
```python
model = CopulaGAN(... , constraints = [some_constraint], ...)
model.fit(real)
```
(The lambda function is checked and other attributes with the model is checked to be correct. If I changed the lambda function to equivalent function like:
```python
def function():
    return ...
```
It runs well
And it turns out to have:
```python
AttributeError                            Traceback (most recent call last)
<ipython-input-240-feec9f2fa8ef> in <module>
     30 
     31     # Fit the CopulaGAN
---> 32     model.fit(real)
     33 
     34     # Create 600 rows of data

...\conda\3\envs\py36\lib\site-packages\sdv\tabular\base.py in fit(self, data)
    108                      self._metadata.name, data.shape)
    109         if not self._metadata_fitted:
--> 110             self._metadata.fit(data)
    111 
    112         self._num_rows = len(data)

...\.julia\conda\3\envs\py36\lib\site-packages\sdv\metadata\table.py in fit(self, data)
    488 
    489         LOGGER.info('Fitting constraints for table %s', self.name)
--> 490         constrained = self._fit_transform_constraints(data)
    491         extra_columns = set(constrained.columns) - set(data.columns)
    492 

...\.julia\conda\3\envs\py36\lib\site-packages\sdv\metadata\table.py in _fit_transform_constraints(self, data)
    349                 constraint = constraint.to_dict()
    350 
--> 351             constraint = Constraint.from_dict(constraint)
    352             self._constraints[idx] = constraint
    353 

...\.julia\conda\3\envs\py36\lib\site-packages\sdv\constraints\base.py in from_dict(cls, constraint_dict)
    211                 constraint_class = subclasses[constraint_class]
    212 
--> 213         return constraint_class(**constraint_dict)
    214 
    215     def to_dict(self):

...\.julia\conda\3\envs\py36\lib\site-packages\sdv\constraints\base.py in __init__(self, *args, **kwargs)
     64                 self.__kwargs__.update(dict(zip(arg_names, args)))
     65 
---> 66             old__init__(self, *args, **kwargs)
     67 
[...]

PR #378, opened 2021-04-08
Sdv-issue-373: Request for sdv.constraint.ColumnFormula for a wider range of function

amontanez24:
#373

### DIFF
--- a/sdv/constraints/base.py
+++ b/sdv/constraints/base.py
@@ -21,9 +21,13 @@ def _get_qualified_name(obj):
     return module + '.' + obj_name
 
 
-def _is_function(obj):
-    """Return true if the object is a function (not including lambdas)."""
-    return callable(obj) and obj.__name__ != (lambda: None).__name__
+def _module_contains_callable_name(obj):
+    """Return if module contains the name of the callable object."""
+    if hasattr(obj, '__name__'):
+        obj_name = obj.__name__
+    else:
+        obj_name = obj.__class__.__name__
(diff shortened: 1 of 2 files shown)


sdv/constraints/base.py
```diff
@@ -21,6 +21,11 @@ def _get_qualified_name(obj):
     return module + '.' + obj_name
 
 
+def _is_function(obj):
+    """Return true if the object is a function (not including lambdas)."""
+    return callable(obj) and obj.__name__ != (lambda: None).__name__
```
amontanez24, line 26: If the function passed is the result of another function, then the problem still occurs. Not sure of a way to identify that scenario. ie.

```
def function_creator():
    def func(x):
        return x + 4
 ColumnFormula(
    column='column',
    formula=function_creator(),
    handling_strategy='transform'
)

File: sdv/constraints/base.py [hunk 2]
```diff
@@ -21,6 +21,15 @@ def _get_qualified_name(obj):
     return module + '.' + obj_name
 
 
+def _module_contains_callable_name(obj):
+    """Return if module contains the name of the callable object."""
+    if hasattr(obj, '__name__'):
+        obj_name = obj.__name__
+    else:
+        obj_name = obj.__class__.__name__
+    return obj_name in importlib.import_module(obj.__module__).__dict__
```
amontanez24, line 30: is there a better way to get module then importing it?

pvk-developer (commented):
LGTM!

pvk-developer (approved):

Issue #379, opened 2021-04-09
Not able to use constraints and conditions in the same time.


dyuliu:
### Environment Details

Please indicate the following details about the environment in which you found the bug:

* SDV version: 0.9.0
* Python version: Anaconda python 3.6.13
* Operating System: macOS 11.2.3

### Error Description

Not able to use constraints and conditions in the same time. 

### Steps to reproduce

Let’s take this tutorial notebook as example (https://github.com/sdv-dev/SDV/blob/master/tutorials/single_table_data/05_Handling_Constraints.ipynb), if you try to generated samples using conditions like the following, error came out (see the screenshot).

```python
sampled = gc.sample(10, conditions={'age': 40})
```

You will see the error message like the following

```bash
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-9-5410734405d8> in <module>
----> 1 sampled = gc.sample(10, conditions={'age': 40, 'age_when_joined': 30})

~/anaconda3/envs/sreg/lib/python3.6/site-packages/sdv/tabular/base.py in sample(self, num_rows, max_retries, max_rows_multiplier, conditions, float_rtol, graceful_reject_sampling)
    386                 raise ValueError(f'Invalid column name `{column}`')
    387 
--> 388             if len(self._metadata.transform(conditions[[column]]).columns) == 0:
    389                 raise ValueError(f'Conditioning on column `{column}` is not possible')
    390 

~/anaconda3/envs/sreg/lib/python3.6/site-packages/sdv/metadata/table.py in transform(self, data)
    515         LOGGER.debug('Transforming constraints for table %s', self.name)
    516         for constraint in self._constraints:
--> 517             data = constraint.transform(data)
    518 
    519         LOGGER.debug('Transforming table %s', self.name)

~/anaconda3/envs/sreg/lib/python3.6/site-packages/sdv/constraints/tabular.py in transform(self, table_data)
    164         # print(table_data[self._columns].values.tolist())
    165 
--> 166         lists_series = pd.Series(table_data[self._columns].values.tolist())
    167         table_data = table_data.drop(self._columns, axis=1)
    168         table_data[self._joint_column] = lists_series.str.join(self._separator)

~/anaconda3/envs/sreg/lib/python3.6/site-packages/pandas/core/frame.py in __getitem__(self, key)
   2910             if is_iterator(key):
   2911                 key = list(key)
[...]

PR #422, opened 2021-05-05
sdv-379-p1: Augment base constraint class to raise unique error for conditional sampling

amontanez24:
This PR does the following:

- Adds `_validate_constraint_columns` to base class
- Calls that method in `transform` and then calls subclass' `_transform`
- Adds `_transform` method to subclasses that need it
- Adds `_constraint_columns` variable to keep track of columns necessary to do constraint transformation
- Raises `MissingConstraintColumnError` if `_validate_constraint_columns` fails

This PR the first part of the work required for #379

### DIFF
--- a/sdv/constraints/base.py
+++ b/sdv/constraints/base.py
@@ -7,6 +7,8 @@ import logging
 
 import pandas as pd
 
+from sdv.constraints.errors import MissingConstraintColumnError
+
 LOGGER = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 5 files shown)


sdv/constraints/base.py
```diff
@@ -197,6 +213,22 @@ def filter_valid(self, table_data):
...
+        Args:
+            table_data (pandas.DataFrame):
+                Table data.
+
+        Returns:
+            bool:
+                Whether or not all column names in ``constraint_columns`` are present
+                in ``table_data``. Returns False if ``constraint_columns`` is None or empty.
+        """
+        if not self._constraint_columns:
```
csala, line 228: If we change the default value to a tuple we can skip this check and just do the `all` check below.

csala (approved):
This looks good @amontanez24 !

PR #426, opened 2021-05-11
sdv-379-p2: Catch MissingConstraintColumnError and skip conditioning on that constraint

amontanez24:
This PR is related to #379 

In this PR:

- The `constraint_columns` instance variable in the `Constraint` class is made public
- The metadata `Table.transform` method is changed to handle the constraint transforms differently. It now takes an optional parameter `on_missing_columns` to specify how to handle a `MissingConstraintColumnError`.
    - If the parameter is set to `error` then it will simply raise the error
    - If the parameter is set to `drop`, then it will drop all columns in the table related to the constraint
- The `sample` method is changed to now group the conditions based on the original conditions, not the transformed ones
- The sample method will pass `None` as the `transformed_condition` if the transform process drops all the columns

### DIFF
--- a/sdv/metadata/table.py
+++ b/sdv/metadata/table.py
@@ -10,6 +10,7 @@ import rdt
 from faker import Faker
 
 from sdv.constraints.base import Constraint
+from sdv.constraints.errors import MissingConstraintColumnError
 from sdv.metadata.errors import MetadataError, MetadataNotFittedError
 from sdv.metadata.utils import strings_from_regex
 
@@ -494,12 +495,28 @@ class Table:
         self._fit_hyper_transformer(constrained, extra_columns)
         self.fitted = True
 
-    def transform(self, data):
+    def _transform_constraints(self, data, on_missing_column='error'):
+        for constraint in self._constraints:
+            try:
+                data = constraint.transform(data)
+            except MissingConstraintColumnError:
+                if on_missing_column == 'error':
+                    raise MissingConstraintColumnError()
+                elif on_missing_column == 'drop':
+                    constraint_columns_in_data = data.columns.all(constraint.constraint_columns)
+                    data = data.drop(constraint_columns_in_data, axis=1)
+        return data
+
+    def transform(self, data, on_missing_column='error'):
         """Transform the given data.
 
         Args:
             data (pandas.DataFrame):
                 Table data.
+            on_missing_column (str):
+                If the value is error, then a `MissingConstraintColumnError` is raised.
+                If the value is drop, then the columns involved in the constraint that
+                are present in data will be dropped.
 
         Returns:
             pandas.DataFrame:
@@ -513,8 +530,7 @@ class Table:
         data = self._anonymize(data[fields])
 
         LOGGER.debug('Transforming constraints for table %s', self.name)
-        for constraint in self._constraints:
-            data = constraint.transform(data)
+        data = self._transform_constraints(data, on_missing_column)
 
         LOGGER.debug('Transforming table %s', self.name)
         return self._hyper_transformer.transform(data)
--- a/sdv/metadata/table.py
+++ b/sdv/metadata/table.py
@@ -502,10 +502,15 @@ class Table:
             except MissingConstraintColumnError:
                 if on_missing_column == 'error':
                     raise MissingConstraintColumnError()
+
                 elif on_missing_column == 'drop':
                     indices_to_drop = data.columns.isin(constraint.constraint_columns)
                     columns_to_drop = data.columns.where(indices_to_drop).dropna()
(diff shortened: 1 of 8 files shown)


sdv/metadata/table.py
```diff
@@ -494,12 +495,29 @@ def fit(self, data):
...
+            try:
+                data = constraint.transform(data)
+            except MissingConstraintColumnError:
+                if on_missing_column == 'error':
+                    raise MissingConstraintColumnError()
+                elif on_missing_column == 'drop':
+                    indices_to_drop = data.columns.isin(constraint.constraint_columns)
+                    columns_to_drop = data.columns.where(indices_to_drop).dropna()
+                    data = data.drop(columns_to_drop, axis=1)
+        return data
```
csala, line 509: Could we add a line break above this line?

sdv/metadata/table.py
```diff
@@ -494,12 +495,29 @@ def fit(self, data):
...
 
-    def transform(self, data):
+    def _transform_constraints(self, data, on_missing_column='error'):
+        for constraint in self._constraints:
+            try:
+                data = constraint.transform(data)
+            except MissingConstraintColumnError:
+                if on_missing_column == 'error':
+                    raise MissingConstraintColumnError()
+                elif on_missing_column == 'drop':
```
csala, line 505: Maybe we should add and `else` branch that raises a `ValueError`

sdv/tabular/base.py
```diff
@@ -385,24 +385,25 @@ def sample(self, num_rows=None, max_retries=100, max_rows_multiplier=10,
...
+                group = [group]
 
-            condition = conditions.loc[dataframe['__condition_idx__'].iloc[0]]
-            transformed_condition = dict(zip(condition_columns, index))
+            condition_index = dataframe['__condition_idx__'].iloc[0]
+            if transformed_conditions.empty:
+                transformed_condition = None
+            else:
+                transformed_condition = transformed_conditions.loc[condition_index].to_dict()
+            condition = dict(zip(condition_columns, group))
```
csala, line 406: Can we add a line break above?

tests/unit/constraints/test_base.py
```diff
@@ -284,7 +284,7 @@ def test_transform_invalid_table_data(self):
         # Run
         instance = Constraint(handling_strategy='transform')
         instance._transform = lambda x: x
-        instance._constraint_columns = ('a')
+        instance.constraint_columns = ('a')
```
csala, line 287: I think that this should be `('a', )`, otherwise it is equivalent to `'a'`, without the tuple.

tests/unit/metadata/test_table.py
```diff
@@ -18,7 +21,7 @@ def test__make_ids_fail(self):
         with pytest.raises(ValueError):
             Table._make_ids(metadata, 20)
 
-    def test_make_ids_unique_field_not_unique(self):
+    def test__make_ids_unique_field_not_unique(self):
```
csala, line 24: Good catch!

tests/unit/metadata/test_table.py
```diff
@@ -77,3 +80,113 @@ def test_make_ids_unique_field_index_out_of_order(self):
 
         assert new_data['item 1'].equals(data['item 1'])
         assert new_data['item 0'].is_unique
+
+    def test_transform_calls__transform_constraints(self):
+        """Test that the `transform` method calls `_transform_constraints` with right parameters"""
```
csala, line 85: Could we add some more details about what is being tested, like in the test below?

csala (changes requested):
This looks very good so far @amontanez24 !
I added a couple of style comments and a warning about a test that might not be completely right.
Also, I would add at least 1 small integration test that certifies that the functionality is properly covered, mocking the underlying `GaussianMultivariate` to make the test faster and also be able to check that the conditions that reach the model are the expected ones. For this test, I would send 2 condition columns, one which can be passed and another one that cannot (for example, create a UniqueCombinations on `city` and `state`, and condition on `age` and `state`, which should make only `age` reach the copula model).

csala (changes requested):
I think we are almost there :-)
I added 2 comments about an invalid variable name and another assert that I would like to add to the integration test.

tests/unit/metadata/test_table.py
```diff
@@ -108,7 +108,15 @@ def test_transform_calls__transform_constraints(self):
         Table.transform(table_mock, data, 'error')
 
         # Assert
-        table_mock._transform_constraints.assert_called_once_with(data, 'error')
+        expectedData = pd.DataFrame({
```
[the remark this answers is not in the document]
amontanez24: good catch! Old habits die hard

csala (approved):
Looks good now! It can be merged.

PR #427, opened 2021-05-12
Update benchmarking user guide for reading private datasets

katxiao:
Update benchmarking user guide for reading private datasets

### DIFF
--- a/docs/user_guides/benchmarking/run.rst
+++ b/docs/user_guides/benchmarking/run.rst
@@ -79,6 +79,10 @@ The ``sdgym.run`` function accepts the following arguments:
 -  ``output_path (str)``:: If an ``output_path`` is given, the generated
    leaderboard will be stored in the indicated path as a CSV file. The
    given path must be a complete path including the ``.csv`` filename.
+-  ``aws_key (str)``:: If an ``aws_key`` is provided, the given access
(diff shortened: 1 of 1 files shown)


docs/user_guides/benchmarking/run.rst
```diff
@@ -167,6 +171,23 @@ For example, if we want to evaluate our synthesizer function on only the
 
    In [15]: scores = sdgym.run(my_synthesizer_function, datasets=datasets)
 
+Running on Private Datasets
+~~~~~~~~
+
+If we want to run sdgym on datasets in a private s3 bucket, we can pass in
```
csala, line 177: I would rephrase this a bit to mention something like "in a private S3 bucket for which we do not have system-wide credentials configured", and link the corresponding part of the sentence to the AWS documentation that explains how to configure `.aws/credentials` or set environment variables.

csala (changes requested):
This looks good. I would just add a comment about being able to access private buckets without the need to pass credentials explicitly if they have been configured system-wide.

csala (approved):
Looks good @katxiao ! Let's leave this in standby while the corresponding changes are finished and merged in SDGym, and then we are ready to merge this.

PR #433, opened 2021-05-18
sdv-379-p3: Add columns model to constraints and sample from it when necessary

amontanez24:


sdv/constraints/base.py
```diff
@@ -112,31 +121,72 @@ def __init__(self, handling_strategy):
...
+
+        if len(missing_columns) == 0:
+            return table_data
+        if all(col in missing_columns for col in self.constraint_columns):
             raise MissingConstraintColumnError()
 
+        condition_columns = [c for c in self.constraint_columns if c in table_data.columns]
+        unrelated_columns = [c for c in table_data.columns if c not in self.constraint_columns]
+        conditions = self._hyper_transformer.transform(table_data[condition_columns])
+        sampled_data = conditions.apply(lambda d: self._columns_model.sample(
```
csala, line 184: I would do this in `for` loop instead of using an `apply/lambda`, and also use a `groupby` on the existing condition columns to avoid making a call to `sample` for each line.

sdv/constraints/base.py
```diff
@@ -112,31 +121,72 @@ def __init__(self, handling_strategy):
...
+        ``constraint_columns`` that ``table_data`` does contain. If ``table_data``
+        doesn't contain any of the ``constraint_columns`` then a
+        ``MissingConstraintColumnError`` will be raised.
 
         Args:
             table_data (pandas.DataFrame):
                 Table data.
         """
-        if any(col not in table_data.columns for col in self.constraint_columns):
+        if self.disable_columns_model:
```
csala, line 169: I would organize this method a bit differently in terms of `if` branches and do:

```
if "we have been given at least one column and there is at least one missing":
    if self._columns_model is None:
        raise Error
    else:
        table_data = "augment table data with the missing column"

return table_data
```

sdv/constraints/base.py
```diff
@@ -112,31 +121,72 @@ def __init__(self, handling_strategy):
...
+            if any(col not in table_data.columns for col in self.constraint_columns):
+                raise MissingConstraintColumnError()
+            return table_data
+
+        missing_columns = [col for col in self.constraint_columns if col not in table_data.columns]
+
+        if len(missing_columns) == 0:
+            return table_data
+        if all(col in missing_columns for col in self.constraint_columns):
```
csala, line 178: Bear in mind that if no column is given at all, this should not raise an error.

amontanez24: I think it should. Otherwise it is confusing as to how it gets resolved. If the error is thrown, it works back up to the metadata `_transform_constraints` where it will be ignored. If we return it here, the caller has to realize it is a special case and not continue to run the transformation

sdv/constraints/base.py
amontanez24: see other comment

sdv/constraints/base.py
```diff
@@ -112,30 +121,82 @@ def __init__(self, handling_strategy):
...
+            if not isinstance(group, tuple):
+                group = [group]
+
+            transformed_condition = self._hyper_transformer.transform(df).loc[0].to_dict()
+            sampled_row = self._columns_model.sample(
+                num_rows=df.shape[0],
+                conditions=transformed_condition
+            )
+            sampled_rows.append(sampled_row)
+        sampled_data = pd.concat(sampled_rows)
```
csala, line 167: Can we add a blank line above this one?

sdv/constraints/base.py
```diff
@@ -112,30 +121,82 @@ def __init__(self, handling_strategy):
...
 
         Args:
             table_data (pandas.DataFrame):
                 Table data.
         """
-        if any(col not in table_data.columns for col in self.constraint_columns):
-            raise MissingConstraintColumnError()
+        missing_columns = [col for col in self.constraint_columns if col not in table_data.columns]
+        if missing_columns:
+            all_columns_missing = all(c in missing_columns for c in self.constraint_columns)
```
csala, line 189: I think we could do this as `len(missing_columns) == len(self.constraint_columns)`, so we do not need to iterate again.

tests/integration/tabular/test_base.py
```diff
@@ -150,4 +152,66 @@ def test_conditional_sampling_properly_handles_constraints(gm_mock):
...
+    })]
+    gm_mock.return_value.sample.side_effect = sampled_numeric_data
+    model.fit(data)
+
+    # Run
+    conditions = {'age': 30, 'state': 'CA'}
+    sampled_data = model.sample(5, conditions=conditions)
+
+    # Assert
+    expected_states = pd.Series(['CA', 'CA', 'CA', 'CA', 'CA'], name='state')
```
csala, line 205: Why are we checking each series separately? I would just compare the two DataFrames, since we know the values for all the columns.

tests/integration/tabular/test_base.py
```diff
@@ -150,4 +152,66 @@ def test_conditional_sampling_properly_handles_constraints(gm_mock):
...
+    expected_ages = pd.Series([30, 30, 30, 30, 30], name='age')
+    sample_calls = model._model.sample.mock_calls
+    assert len(sample_calls) == 2
+    assert sample_calls[0][1][0] == 5
+    assert sample_calls[1][1][0] == 1
+    assert sample_calls[0][2]['conditions']['age'] == 30
+    assert sample_calls[1][2]['conditions']['age'] == 30
+    assert 'city#state' in sample_calls[0][2]['conditions']
```
csala, line 213: I do not understand this assert. Would `city#state` be expected to NOT be inside the `conditions` being passed to the sample method?

tests/integration/tabular/test_base.py
```diff
@@ -150,4 +152,66 @@ def test_conditional_sampling_properly_handles_constraints(gm_mock):
...
+    model.fit(data)
+
+    # Run
+    conditions = {'age': 30, 'state': 'CA'}
+    sampled_data = model.sample(5, conditions=conditions)
+
+    # Assert
+    expected_states = pd.Series(['CA', 'CA', 'CA', 'CA', 'CA'], name='state')
+    expected_ages = pd.Series([30, 30, 30, 30, 30], name='age')
+    sample_calls = model._model.sample.mock_calls
```
csala, line 207: If there are no array-like objects in the arguments, like in this case, I think it is better to just compare `calls`.

The code would be similar to this:
```
from unittest.mock import call
...
call_args_list = model._model.sample.call_args_list
assert call_args_list[0] == call(5, conditions={'age': 30})
assert call_args_list[1] == call(1, conditions={'age': 30})
```

sdv/constraints/base.py
csala: You are totally right! The code should actually raise the error to let the calling code handle the exception (so, the Table can skip this constraint)

csala (changes requested):
This is almost ready. I added a couple of comments about minor style and efficiency changes, and a couple of recommendations to improve the integration tests. After this is done I think we should be able to merge.

tests/integration/tabular/test_base.py
amontanez24: The call is not deterministic. There will be another condition: `{'city#state': 'some city value'}`

tests/integration/tabular/test_base.py
csala: My bad, I had the other case in my mind (constraint columns being dropped). You are totally right.

csala (approved):
Looks good now! This can be squashed and merged as soon as the CI finishes

PR #447, opened 2021-05-25
sdv-440: Conditionally sampling on variable in constraint should have variety for other variables

amontanez24:
This PR changes the way we sample conditions to not only group by the conditions passed by the user, but also by the transformed conditions that result from them. This is to ensure variety in the other columns that result in the transformed conditions. 

For example, if we have a `GreaterThan` constraint with a high and low column, and a condition on the high column, the sampled data should have varying values for the low column

### DIFF
--- a/sdv/tabular/base.py
+++ b/sdv/tabular/base.py
@@ -365,8 +365,8 @@ class BaseTabularModel:
                      f'be sampled within {max_retries} trials.')
 
         if len(sampled_rows) > 0:
-            sampled_rows[COND_IDX] = \
-                dataframe[COND_IDX].values[:len(sampled_rows)]
+            sampled_rows[COND_IDX] = dataframe[COND_IDX].values[:len(sampled_rows)]
+
         return sampled_rows
 
     def sample(self, num_rows=None, max_retries=100, max_rows_multiplier=10,
--- a/sdv/tabular/base.py
+++ b/sdv/tabular/base.py
@@ -335,8 +335,8 @@ class BaseTabularModel:
         return conditions.copy()
 
     def _conditionally_sample_rows(self, dataframe, max_retries, max_rows_multiplier,
-                                  condition, transformed_condition, float_rtol,
-                                  graceful_reject_sampling):
+                                   condition, transformed_condition, float_rtol,
+                                   graceful_reject_sampling):
         sampled_rows = self._sample_batch(
             len(dataframe),
             max_retries,
@@ -358,7 +358,7 @@ class BaseTabularModel:
 
             else:
                 warn(f'Only {len(sampled_rows)} rows could '
-                    f'be sampled within {max_retries} trials.')
+                     f'be sampled within {max_retries} trials.')
 
         if len(sampled_rows) > 0:
             sampled_rows['__condition_idx__'] = \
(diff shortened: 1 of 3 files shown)


sdv/tabular/base.py
```diff
@@ -387,8 +418,11 @@ def sample(self, num_rows=None, max_retries=100, max_rows_multiplier=10,
 
         transformed_conditions = self._metadata.transform(conditions, on_missing_column='drop')
         condition_columns = list(conditions.columns)
+        transformed_columns = list(transformed_conditions.columns)
         conditions.index.name = '__condition_idx__'
         conditions.reset_index(inplace=True)
+        transformed_conditions.index.name = '__condition_idx__'
```
csala, line 424: Since this `'__condition_idx__'` string is now being used in a lot of places, it may be worth to use a module level `CONSTANT` variable instead.
Also, in order to avoid any possible clashes with existing column names (even though `__condition_idx__` is an unlikely string), we can use an `uuid` generated on the fly for it:

```python3
COND_IDX = str(uuid.uuid4())
```

sdv/tabular/base.py
```diff
@@ -398,40 +432,36 @@ def sample(self, num_rows=None, max_retries=100, max_rows_multiplier=10,
...
+                    graceful_reject_sampling
+                )
                 all_sampled_rows.append(sampled_rows)
+            else:
+                transformed_conditions_in_group = transformed_conditions.loc[condition_indices]
+                transformed_groups = transformed_conditions_in_group.groupby(transformed_columns)
+                for transformed_group, transformed_dataframe in transformed_groups:
+                    if not isinstance(transformed_group, tuple):
+                        transformed_group = [transformed_group]
+                    transformed_condition = dict(zip(transformed_columns, transformed_group))
```
csala, line 454: Can we add a blank line above this one?

sdv/tabular/base.py
```diff
@@ -334,6 +334,37 @@ def _make_conditions_df(self, conditions, num_rows):
...
+        sampled_rows = self._sample_batch(
+            len(dataframe),
+            max_retries,
+            max_rows_multiplier,
+            condition,
+            transformed_condition,
+            float_rtol
+        )
+
+        if len(sampled_rows) < len(dataframe):
```
csala, line 349: VERY minor thing: we could capture the `len(sampled_rows)` and `len(dataframe)` in variables, so we do not need to compute them multiple times. It will also help a bit on the readability.

sdv/tabular/base.py
```diff
@@ -398,40 +432,36 @@ def sample(self, num_rows=None, max_retries=100, max_rows_multiplier=10,
...
+                    max_retries,
+                    max_rows_multiplier,
+                    condition,
+                    None,
+                    float_rtol,
+                    graceful_reject_sampling
+                )
                 all_sampled_rows.append(sampled_rows)
+            else:
+                transformed_conditions_in_group = transformed_conditions.loc[condition_indices]
```
csala, line 449: I wonder if this entire `else` block could be moved into its own method? I'm not completely sure about whether it would make this method more readable or the opposite, so I'm just thinking out loud. Your call @amontanez24

csala (changes requested):
I think the implementation is good!
I just added a few comments about small improvements that could be made, but in general it seems to be almost ready to go.

sdv/tabular/base.py
amontanez24: I'm torn as well. Moving it into another method would mean passing `all_sampled_rows` to that other method and appending there which I don't love because it would not be obvious within `sample` what is happening to that variable. The only real difference between this block and the other is grabbing the transformed groups and looping through them. I'm inclined to leave it

csala: I agree, let's leave it as it is right now.

csala (approved):
Looks good @amontanez24 !

PR #473, opened 2021-06-23
Limit memory consumption of the `GaussianMultivariate` when using constraints

fealho:


sdv/constraints/base.py
```diff
@@ -142,7 +143,9 @@ def fit(self, table_data):
                 'O': 'one_hot_encoding',
             })
             transformed_data = self._hyper_transformer.fit_transform(data_to_model)
-            self._columns_model = GaussianMultivariate()
+            self._columns_model = GaussianMultivariate(
```
amontanez24, line 146: maybe we can add a small test that makes sure this parameter is correct

amontanez24 (approved):
LGTM!

PR #474, opened 2021-06-23
One Hot Encoding Constraint

### LINKED ISSUE #303: OneHotEncoding Constraint
Problem Description

I am trying to generate this kind of data, using CTGAN.

as you can see from the above picture, categorical data saved as one-hot encoding.Sex_1, Sex_2 / race.ethnicity_1, race.ethnicity_2, race.ethnicity_3, race.ethnicity_4, race.ethnicity_5

0, 1 / 0, 0, 0, 0, 11,0 / 1, 0, 0, 0, 0There should be only one 1, and other variables should be 0. 

However, when I made data by using CTGAN some data shows 0,0, 1,1 0,-1 for Sex_1 and Sex_2. 

Expected behavior

I would like to use CustomConstraint  to solve this problem.1) I would like to set the output as binary. (data should be 0 or 1)2) also, I would like to set constraint likerace.ethnicity_1+race.ethnicity_2+race.ethnicity_3+race.ethnicity_4+race.ethnicity_5 = 1

married_1 + married_2 + married_3 + married_4 + married_5 + married_6 =1

Thank you so much for your help

fealho:
Resolve #303


docs/user_guides/single_table/constraints.rst
```diff
@@ -184,12 +184,50 @@ constraint by passing it:
...
+OneHotEncoding Constraint
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Another constraint available is the ``OneHotEncoding`` constraint.
+This constraint allows the user to specify a list of columns where each row 
+is a one hot vector. Then, the constraint will make sure that the output
+of the model is transformed so that the column with the largest value is
+set to 1 while all other columns are set to 0. 
+
+Since none of the columns of the demo dataset satisfy the requirements for this
```
csala, line 196: Rather than this, maybe it would be interesting to add a few one-hot-encoded columns to the demo data?

csala (commented):
This looks almost ready to go! I just added a comment about the documentation and adding one-hot columns to the demo data

sdv/constraints/tabular.py
```diff
@@ -379,3 +380,65 @@ def reverse_transform(self, table_data):
...
+    Args:
+        columns (list[str]):
+            Names of the columns containing one hot rows.
+        handling_strategy (str):
+            How this Constraint should be handled, which can be ``transform``
+            or ``reject_sampling`` (not recommended). Defaults to ``transform``.
+    """
+
+    def __init__(self, columns, handling_strategy='transform'):
+        self._columns = columns
```
csala, line 402: We should be adding the columns to `self.constraint_columns` too, to ensure that conditional sampling is properly handled.

csala:
This is a multi-column constraint, so we need to be careful about the conditional sampling.

One should be able to do:

```
ohe = OneHotEncoding(columns=['a', 'b', 'c'])

model = TabularModel(constraints=[ohe])
model.fit(data)
model.sample(conditions={'a': 1})
```

If I'm not mistaken, this right now would be a problem because the columns model would attempt to sample the other columns without being able to produce exact 0s, which would end up producing a crash.

To solve this, I suggest to overwrite the `_sample_constraint_columns` method to do the following:
1. If the input conditions are not valid (see below), raise a ValueError
2. If they are valid and contain a 1, set 0s to the columns that were not given
3. If only 0s are given without any 1, sample the rest of the columns and then apply the `reverse_transform` to make them valid inputs (and outputs)

WRT to invalid input, I can think of the following scenarions:
1. Values different than 0 and 1 are given
2. There is more than one column containing a 1
3. All the constrained columns are given as 0

Any thoughts @amontanez24 ?

docs/user_guides/single_table/constraints.rst
fealho: Done.

sdv/constraints/tabular.py
fealho: Done

csala (commented):
Everything looks good now, except for the `_sample_constraint_columns` thing.
What do you think @amontanez24 and @fealho ?
Should we overwrite the `_sample_constraint_columns` method as I described in the previous comment?

sdv/demo.py
```diff
@@ -333,7 +333,9 @@ def _load_tabular_dummy():
         'address': adresses,
         'age': age,
         'age_when_joined': age_when_joined,
-        'years_in_the_company': age - age_when_joined
+        'years_in_the_company': age - age_when_joined,
+        'joined_in_first_semester': [1.0, 0.0, 1.0, 1.0, 0.0, 0.0] * 2,
```
amontanez24, line 337: minor: the demo table is called `employees` right? I usually don't think of employees as having joined in semesters as that's usually a school related concept. Can we name it something else?

amontanez24:
> This is a multi-column constraint, so we need to be careful about the conditional sampling.
> 
> One should be able to do:
> 
> ```
> ohe = OneHotEncoding(columns=['a', 'b', 'c'])
> 
> model = TabularModel(constraints=[ohe])
> model.fit(data)
> model.sample(conditions={'a': 1})
> ```
> 
> If I'm not mistaken, this right now would be a problem because the columns model would attempt to sample the other columns without being able to produce exact 0s, which would end up producing a crash.
> 
> To solve this, I suggest to overwrite the `_sample_constraint_columns` method to do the following:
> 
> 1. If the input conditions are not valid (see below), raise a ValueError
> 2. If they are valid and contain a 1, set 0s to the columns that were not given
> 3. If only 0s are given without any 1, sample the rest of the columns and then apply the `reverse_transform` to make them valid inputs (and outputs)
> 
> WRT to invalid input, I can think of the following scenarions:
> 
> 1. Values different than 0 and 1 are given
> 2. There is more than one column containing a 1
> 3. All the constrained columns are given as 0
> 
> Any thoughts @amontanez24 ?

I agree with the suggestion to overwrite the `_sample_constraint_columns` method. Has this been tried at all? Does reject sampling not eventually succeed? I still think the method should be overwritten I'm just curious

sdv/demo.py
fealho: Honestly, I just had trouble coming up with a better idea :disappointed:, since I can only add two one hot columns to the demo (don't want to clutter it). Do you have any suggestions?

amontanez24: Maybe whether or not the employee was hired as a new grad? Or whether or not they were referred by another employee?

sdv/constraints/tabular.py
```diff
@@ -475,3 +476,96 @@ def reverse_transform(self, table_data):
...
+        self.constraint_columns = tuple(columns)
+        super().__init__(handling_strategy, fit_columns_model=True)
+
+    def _sample_constraint_columns(self, table_data):
```
csala, line 502: It would be interesting to add a docstring to this method explaining what it does

sdv/constraints/tabular.py
```diff
@@ -475,3 +476,96 @@ def reverse_transform(self, table_data):
...
+        table_data = table_data.copy()
+
+        condition_columns = [col for col in self._columns if col in table_data.columns]
+        is_one = table_data[condition_columns].values == 1.0
```
csala, line 506: I think you could use `table_data[condition_columns].isin([0.0, 1.0]).all(axis=1)` and skip the `is_one` and `is_zero`

sdv/constraints/tabular.py
```diff
@@ -475,3 +476,96 @@ def reverse_transform(self, table_data):
...
+        is_zero = table_data[condition_columns].values == 0.0
+        if not (is_one | is_zero).all():
+            raise ValueError('Condition values must be ones or zeros.')
+
+        if (table_data[condition_columns].sum(axis=1) > 1.0).any():
+            raise ValueError('Each row of a condition can only contain one number one.')
+
+        if (table_data[condition_columns].sum(axis=1) == 0.0).any():
```
csala, line 514: This `if` seems to be taken in case `any` row has all 0s.

I think this should not be the case, since there could be rows that have 1s in them. We should be treating them differently by sampling the missing values only for the rows that do not have 1, and simply populating with 0s the ones that do have a 1 in them. (Notice that with the current approach, the model could even sample a value greater than 1 and then force a change in the `hot` column!)
We can probably split the `table_data` in two batches, rows with 1s and rows with 0s, populating them, and then putting them back together, in order.

sdv/constraints/tabular.py
```diff
@@ -475,3 +476,96 @@ def reverse_transform(self, table_data):
...
+            proposed_table_data = super()._sample_constraint_columns(table_data)
```
csala, line 515: Before sampling we should check how many columns are left. It could be the case where there is only one column left, so then we directly know that the missing value must be 1

csala (changes requested):

tests/unit/constraints/test_tabular.py
```diff
@@ -1760,3 +1761,335 @@ def test_reverse_transform(self):
...
+        - Table data (pandas.DataFrame)
+        Output:
+        - Series of ``True`` and ``False`` values (pandas.Series)
+        """
+        # Setup
+        instance = OneHotEncoding(columns=['a', 'b', 'c'])
+
+        # Run
+        table_data = pd.DataFrame({
+            'a': [1.0, 1.0, 0.0, 1.0],
```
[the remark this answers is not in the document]
fealho: The one hot encoding is on the rows, not the columns. So since `a = 1.0`, `b = 0.0` and `c = 0.0`, the first row is valid.

sdv/constraints/tabular.py
```diff
@@ -475,3 +476,146 @@ def reverse_transform(self, table_data):
...
+                    c for c in self.constraint_columns if c in sub_table_data.columns
+                ]
+                grouped_conditions = sub_table_data[condition_columns].groupby(condition_columns)
+                all_sampled_rows = list()
+                for group, df in grouped_conditions:
+                    if not isinstance(group, tuple):
+                        group = [group]
+
+                    transformed_condition = self._hyper_transformer.transform(df).iloc[0].to_dict()
+                    sampled_rows = self._columns_model.sample(
```
fealho, line 558: This code was adapted from the base class. The version in the base class cannot be used directly since it uses rejection sampling (even when method='transform') which won't work for this constraint since it's virtually impossible to sample exactly 0.0 or 1.0. There are a few other ways to work around this that I tried but this seemed to be the best solution.

csala (dismissed):
I think we are almost there :-)
I added a few comments about the `_sample_constraint_columns` implementation, but most of it looks ready to go

csala (changes requested):
Added a few comments in the previous review which I mistakenly set as an approval

sdv/constraints/tabular.py
```diff
@@ -475,3 +476,146 @@ def reverse_transform(self, table_data):
...
+            sub_table_data = table_data.loc[~has_one, condition_columns]
+
+            if len(condition_columns) == len(self._columns) - 1:
+                proposed_table_data = sub_table_data.copy()
+                for column in self._columns:
+                    if column not in condition_columns:
+                        proposed_table_data[column] = 1.0
+
+            else:
+                condition_columns = [
```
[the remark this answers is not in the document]
fealho: Yep, I changed this in the previous commit (I think you looked at an older version of the code).

sdv/constraints/tabular.py
```diff
@@ -475,3 +476,146 @@ def reverse_transform(self, table_data):
...
+                    c for c in self.constraint_columns if c in sub_table_data.columns
+                ]
+                grouped_conditions = sub_table_data[condition_columns].groupby(condition_columns)
+                all_sampled_rows = list()
+                for group, df in grouped_conditions:
```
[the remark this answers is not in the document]
fealho: It's actually not necessary, because the grouped conditions are always 0.0. I changed the logic to remove the grouping in the latest version. 

**Note:** the constraint base class doesn't seem to implement any logic for keeping track of the order, I think you are referring to the tabular base class. After some exploring, I think we actually need to add the logic to the constraint base class, I don't see any test cases which test this specifically so it's possible it's currently broken...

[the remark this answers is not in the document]
csala: > It's actually not necessary, because the grouped conditions are always 0.0. I changed the logic to remove the grouping in the latest version.

Oh, of course! :-)

> **Note:** the constraint base class doesn't seem to implement any logic for keeping track of the order, I think you are referring to the tabular base class. After some exploring, I think we actually need to add the logic to the constraint base class, I don't see any test cases which test this specifically so it's possible it's currently broken...

Very good point. Maybe we can open an issue to just "review" this? If we look at it and it turns out it is not a problem we will end up closing it without doing anything.

csala (approved):
This looks good to go, only pending on resolving the conflicts with `master` branch

csala (approved):
This looks ready to go. Thanks @fealho !

PR #475, opened 2021-06-23
Use reject sampling to sample missing columns for constraints

### LINKED ISSUE #435: Use reject sampling to sample missing columns for constraints
### Problem Description:

Right now, the rows sampled by the constraint's `column_model` aren't guaranteed to be valid. For example, in the case of the `GreaterThan` constraint, if the smaller column is provided, the `column_model` will conditionally sample the larger column value. In some cases, it may sample a smaller value, which would mean that the conditions used are wrong and the conditional sampling will always fail.

amontanez24:
resolves #435

### DIFF
--- a/tests/integration/tabular/test_base.py
+++ b/tests/integration/tabular/test_base.py
@@ -248,7 +248,8 @@ def test_conditional_sampling_constraint_uses_columns_model_reject_sampling(gm_m
         low='age_joined',
         high='age',
         handling_strategy='transform',
-        fit_columns_model=True
+        fit_columns_model=True,
+        drop='high'
     )
     data = pd.DataFrame({
         'age_joined': [22.0, 21.0, 15.0, 18.0, 29.0],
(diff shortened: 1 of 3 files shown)


tests/integration/tabular/test_base.py
```diff
@@ -1,5 +1,7 @@
+from sdv.constraints.tabular import GreaterThan
```
csala, line 1: This should be further down, with the other `sdv` imports

tests/integration/tabular/test_base.py
```diff
@@ -214,3 +216,82 @@ def test_conditional_sampling_constraint_uses_columns_model(gm_mock):
...
+def test_conditional_sampling_constraint_uses_columns_model_reject_sampling(gm_mock):
+    """Test that the ``sample`` method handles constraints with conditions.
+
+    The ``sample`` method is expected to properly apply constraint
+    transformations by sampling the missing columns for the constraint
+    if ``fit_columns_model`` is True. All values sampled by the column
+    model should be valid because reject sampling is used on any that aren't.
+
+    Setup:
+    - The model is being passed a ``GreaterThan`` constraint and then
```
csala, line 232: This description of the test is great! However, the `Setup` section should be more about "What needs to be set up for this test to happen". So, in this case, rather than explaining what is expected to happen, I would mention that the setup involves a mock of the GM that will sample rows that do not satisfy the constraint to force the reject sampling to be used.

csala (changes requested):
I think this looks good already!
I added a couple of comments about a `lint` issue and a test docstring that could be refined, but I think it is almost ready to go.

tests/integration/tabular/test_base.py
amontanez24: The reason for this is that the model being mocked is not the one in the constraint but the one in sdv. I didn't mock the constraint's `_columns_model` since it is an integration test. I only mocked the other model so that we don't have to wait for it to fit

amontanez24: Update: I mocked the column model to force reject sampling

csala (approved):
Good to go

csala (approved):

PR #476, opened 2021-06-23
Change GreaterThan constraint's transform and reverse_transform methods to handle non-floats

### LINKED ISSUE #448: GreaterThan constraint's transform strategy fails on columns that are not float
### Environment Details

* SDV version: 0.10.0
* Python version: 3.x
* Operating System: Windows, MacOS

### Error Description

Related to #421 

When a `GreaterThan` constraint is provided `transform` as the `handling_strategy`, it will fail if the `high` column is not a float. This is because the `HyperTransformer` reverse transforms the columns before the constraint's reverse transformation takes place. If the high column was initially an int or datetime, it gets transformed by the `HyperTransformer`. The constraint then has a value in the wrong form when attempting to apply its reverse transformation.

### Steps to reproduce

```
from sdv.constraints import UniqueCombinations
from sdv.constraints import GreaterThan

employees = load_tabular_demo()

age_gt_age_when_joined_constraint = GreaterThan(
    low='age_when_joined',
    high='age',
    handling_strategy='transform'
)
constraints = [
    age_gt_age_when_joined_constraint
]

gc = GaussianCopula(constraints=constraints)
gc.fit(employees)

sampled = gc.sample(10, conditions={'age': 40})
```

amontanez24:
Resolve #448

### DIFF
--- a/sdv/constraints/tabular.py
+++ b/sdv/constraints/tabular.py
@@ -81,31 +81,6 @@ class UniqueCombinations(Constraint):
         super().__init__(handling_strategy=handling_strategy,
                          fit_columns_model=fit_columns_model)
 
-    def _valid_separator(self, table_data):
-        """Return True if separator is valid for this data.
-
-        If the separator is contained within any of the columns
-        or the column name obtained after joining the column
-        names using the separator already exists, the separator
-        is not valid.
-
-        Args:
-            table_data (pandas.DataFrame):
-                Table data.
-
-        Returns:
-            bool:
-                Whether the separator is valid for this data or not.
-        """
-        for column in self._columns:
-            if table_data[column].str.contains(self._separator).any():
-                return False
-
-            if self._separator.join(self._columns) in table_data:
-                return False
-
-        return True
-
     def _fit(self, table_data):
         """Fit this Constraint to the data.
 
@@ -121,7 +96,7 @@ class UniqueCombinations(Constraint):
                 Table data.
         """
         self._separator = '#'
-        while not self._valid_separator(table_data):
+        while not self._valid_separator(table_data, self._separator, self._columns):
             self._separator += '#'
 
         self._joint_column = self._separator.join(self._columns)
@@ -238,8 +213,11 @@ class GreaterThan(Constraint):
         """
         self._dtype = table_data[self._high].dtype
         self._diff_column = f'#{self._low}#{self._high}'
-        while self._diff_column in table_data.columns:
-            self._diff_column += '#'
+        separator = '#'
+        while not self._valid_separator(table_data, separator, self.constraint_columns):
+            separator += '#'
+
+        self._diff_column = separator.join(self.constraint_columns)
 
     def is_valid(self, table_data):
         """Say whether ``high`` is greater than ``low`` in each row.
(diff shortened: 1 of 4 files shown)


sdv/constraints/tabular.py
```diff
@@ -211,14 +211,21 @@ class GreaterThan(Constraint):
...
+    self_diff_column = None
+
     def __init__(self, low, high, strict=False, handling_strategy='transform',
-                 fit_columns_model=True):
+                 fit_columns_model=True, drop=None):
         self._low = low
         self._high = high
         self._strict = strict
         self.constraint_columns = (low, high)
+        self._diff_column = f'#{self._low}#{self._high}'
```
csala, line 227: This line may be unnecessary (see comment below)

sdv/constraints/tabular.py
```diff
@@ -230,6 +237,9 @@ def _fit(self, table_data):
                 The Table data.
         """
         self._dtype = table_data[self._high].dtype
+        self._diff_column = f'#{self._low}#{self._high}'
```
csala, line 240: Maybe we should be [finding a valid separator in a similar fashion as to what `UniqueConstraints` does](https://github.com/sdv-dev/SDV/blob/1dbbfdaca61b532ca85c9ddedde5c296b0cbccc7/sdv/constraints/tabular.py#L109)? We could move the logic to a `_find_valid_separator(table_data, columns` staticmethod at the `Constraint` class (or even `sdv.utils`), so all the constraints (and potentially other parts of SDV) can use it.
I would do either that or involve a `uuid` generated on the fly to make sure that the column name does not clash with an existing column.

sdv/constraints/tabular.py
```diff
@@ -294,13 +308,32 @@ def reverse_transform(self, table_data):
                 Transformed data.
         """
         table_data = table_data.copy()
-        diff = (np.exp(table_data[self._high]).round() - 1).clip(0)
-        low_column = table_data[self._low]
+        diff = (np.exp(table_data[self._diff_column]).round() - 1).clip(0)
 
-        if pd.api.types.is_datetime64_ns_dtype(low_column):
-            diff = pd.to_timedelta(diff)
+        if self._drop == 'high':
```
csala, line 313: There seems to be too much duplicated code in this `if/else` block. I would attempt to move common parts to another method

sdv/constraints/tabular.py
```diff
@@ -294,13 +308,32 @@ def reverse_transform(self, table_data):
...
+            low_column = table_data[self._low]
+            if pd.api.types.is_datetime64_ns_dtype(low_column):
+                diff = pd.to_timedelta(diff)
+
+            invalid = ~self.is_valid(table_data)
+            new_high_values = low_column.loc[invalid] + diff.loc[invalid]
+            table_data[self._high].loc[invalid] = new_high_values.astype(self._dtype)
 
-        table_data[self._high] = (low_column + diff).astype(self._dtype)
+        table_data = table_data.drop(self._diff_column, axis=1)
```
csala, line 336: a `del table_data[self._diff_column]` would also do the trick, but I do not know which one is faster and has less memory overhead. @amontanez24 could we time this and also research a bit whether one is preferable over the other one?
@pvk-developer any thoughts?

csala (changes requested):
I added a few comments about things that I believe could be improved, but the overall implementation and approach seems to be very good @amontanez24 !

sdv/constraints/tabular.py
amontanez24: looks like `drop` is faster https://stackoverflow.com/questions/47426089/python-del-vs-pandas-drop/51194789

csala (approved):
LGTM!

PR #484, opened 2021-06-25
Add rounding constraint

### LINKED ISSUE #482: Add Rounding constraint
### Problem Description

To make the data look more realistic, we should round floats based on the user's specification. Sometimes, there are a set # of digits after the decimal. Other times, numbers might be to the nearest 100 or 1000.

### Expected behavior

Add a Rounding constraint class that rounds floats based on the user's specification. Users should be able to specify the list of columns it needs to round, and how much to round each column.

katxiao:
Add rounding constraint

Resolves #482

### DIFF
--- a/sdv/constraints/tabular.py
+++ b/sdv/constraints/tabular.py
@@ -569,6 +569,7 @@ class Rounding(Constraint):
             self._columns = columns
 
         self._digits = digits
+        self._round_config = {column: self._digits for column in self._columns}
         self._tolerance = tolerance if tolerance else 10**(-1 * (digits + 1))
         super().__init__(handling_strategy=handling_strategy, fit_columns_model=False)
 
@@ -583,8 +584,9 @@ class Rounding(Constraint):
             pandas.Series:
                 Whether each row is valid.
         """
-        rounded = table_data.round({column: self._digits for column in self._columns})
-        valid = (table_data - rounded).abs() <= self._tolerance
+        columns = table_data[self._columns]
+        rounded = columns.round(self._digits)
+        valid = (columns - rounded).abs() <= self._tolerance
 
         return valid.all(1)
 
@@ -601,4 +603,4 @@ class Rounding(Constraint):
             pandas.DataFrame:
                 Transformed data.
         """
-        return table_data.round({column: self._digits for column in self._columns})
+        return table_data.round(self._round_config)
--- a/sdv/constraints/tabular.py
+++ b/sdv/constraints/tabular.py
@@ -17,7 +17,6 @@ Currently implemented constraints are:
       on the other columns of the table.
 """
 
-import decimal
 import uuid
 from datetime import datetime
 
@@ -546,43 +545,32 @@ class Rounding(Constraint):
     Args:
         columns (str or list[str]):
             Name of the column(s) to round.
-        digits (int or dict[str->int]):
-            How much to round each column. If an `int` is provided, all columns
-            will be rounded to that number of digits. If a `dict` that maps column
-            to digits is provided, each column will be represented to the specified
+        digits (int):
+            How much to round each column. All columns will be rounded to this
             number of digits.
         handling_strategy (str):
             How this Constraint should be handled, which can be ``transform``
             or ``reject_sampling``. Defaults to ``transform``.
         tolerance (int):
-            How many differences in decimal places we will tolerante when
-            reject sampling.
+            When reject sampling, the sample data must be within this distance
+            of the desired rounded values.
     """
 
-    def __init__(self, columns, digits, handling_strategy='transform', tolerance=1):
+    def __init__(self, columns, digits, handling_strategy='transform', tolerance=None):
+        if digits > 15:
+            raise ValueError('The value of digits cannot exceed 15')
+
+        if tolerance is not None and tolerance >= 10**(-1 * digits):
+            raise ValueError('Tolerance must be less than the rounding level')
+
         if isinstance(columns, str):
             self._columns = [columns]
         else:
             self._columns = columns
 
-        if isinstance(digits, dict):
-            if set(digits.keys()) != set(columns):
-                raise ValueError('The key values of `digits` do not match the `columns` input')
-
-            self._digits = digits
-        else:
-            self._digits = {column: digits for column in columns}
-
-        self._tolerance = tolerance
+        self._digits = digits
+        self._tolerance = tolerance if tolerance else 10**(-1 * (digits + 1))
         super().__init__(handling_strategy=handling_strategy, fit_columns_model=False)
-
-    @staticmethod
-    def _get_decimal_places(number):
-        if number % 1 == 0:
-            # No decimals, so find number of trailing zeros.
-            return -1 * (len(str(number)) - len(str(number).rstrip('0')))
-
-        return -1 * decimal.Decimal(str(number)).as_tuple().exponent
 
     def is_valid(self, table_data):
         """Determine if the data satisfies the rounding constraint.
@@ -595,17 +583,10 @@ class Rounding(Constraint):
             pandas.Series:
                 Whether each row is valid.
         """
-        table_data = table_data.copy()
-        d_places = table_data[self._columns].applymap(self._get_decimal_places)
+        rounded = table_data.round({column: self._digits for column in self._columns})
+        valid = (table_data - rounded).abs() <= self._tolerance
 
-        for column in self._columns:
-            digits = self._digits[column]
-            low = digits - self._tolerance
-            high = digits + self._tolerance
-            table_data[column] = \
-                table_data[(d_places[column] >= low) & (d_places[column] <= high)][column]
-
-        return table_data.notna().all(1)
+        return valid.all(1)
 
     def reverse_transform(self, table_data):
         """Reverse transform the table data.
@@ -620,4 +601,4 @@ class Rounding(Constraint):
             pandas.DataFrame:
                 Transformed data.
         """
-        return table_data.round(self._digits)
+        return table_data.round({column: self._digits for column in self._columns})
--- a/sdv/constraints/tabular.py
+++ b/sdv/constraints/tabular.py
@@ -548,21 +548,20 @@ class Rounding(Constraint):
             Name of the column(s) to round.
         digits (int or dict[str->int]):
             How much to round each column. If an `int` is provided, all columns
-            will be rounded to that number of digits. If a `dict` mapping column
-            to digit is provided, each column will be represented to the specified
-            digit.
+            will be rounded to that number of digits. If a `dict` that maps column
+            to digits is provided, each column will be represented to the specified
+            number of digits.
         handling_strategy (str):
             How this Constraint should be handled, which can be ``transform``
             or ``reject_sampling``. Defaults to ``transform``.
         tolerance (int):
-            How many differences in decimal places we will tolerant when
+            How many differences in decimal places we will tolerante when
             reject sampling.
     """
 
     def __init__(self, columns, digits, handling_strategy='transform', tolerance=1):
         self._columns = columns
         self._digits = digits
-
         self._tolerance = tolerance
         super().__init__(handling_strategy=handling_strategy, fit_columns_model=False)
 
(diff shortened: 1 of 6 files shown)


sdv/constraints/tabular.py
```diff
@@ -379,3 +381,73 @@ def reverse_transform(self, table_data):
...
+            table_data (pandas.DataFrame):
+                Table data.
+
+        Returns:
+            pandas.Series:
+                Whether each row is valid.
+        """
+        table_data = table_data.copy()
+        d_places = table_data[self._columns].applymap(
+            lambda d: -1 * decimal.Decimal(str(d)).as_tuple().exponent)
```
csala, line 424: I would avoid using lambdas and rather use a `staticmethod`

sdv/constraints/tabular.py
```diff
@@ -379,3 +381,73 @@ def reverse_transform(self, table_data):
...
         return table_data
+
+
+class Rounding(Constraint):
+    """Round a column based on the specified number of digits.
+
+    Args:
+        column (list[str]):
+            Name of the column(s) to round.
+        digits (int or dict[str->int]):
```
csala, line 392: What happens if the dict and the `column` list differ? I would validate that and raise a `ValueError` if they do

sdv/constraints/tabular.py
```diff
@@ -379,3 +381,73 @@ def reverse_transform(self, table_data):
         table_data[self._column] = self._formula(table_data)
 
         return table_data
+
+
+class Rounding(Constraint):
+    """Round a column based on the specified number of digits.
+
+    Args:
+        column (list[str]):
```
csala, line 390: column`s`

csala: If it is a single column, should we support being past a single string too, instead of a list?

sdv/constraints/tabular.py
```diff
@@ -379,3 +381,73 @@ def reverse_transform(self, table_data):
...
+
+        Returns:
+            pandas.Series:
+                Whether each row is valid.
+        """
+        table_data = table_data.copy()
+        d_places = table_data[self._columns].applymap(
+            lambda d: -1 * decimal.Decimal(str(d)).as_tuple().exponent)
+
+        cols = [self._columns] if isinstance(self._digits, int) else [col for col in self._columns]
```
csala, line 426: I do not understand this. It seems like the `[col for col in self._columns]` part is redundant. In any case, I would avoid doing this here: let's do it inside the `__init__` instead, and always store `self._columns` as a list from the begining.

sdv/constraints/tabular.py
```diff
@@ -379,3 +381,73 @@ def reverse_transform(self, table_data):
...
+        for col in cols:
```
csala, line 427: Let's use verbose names: `col` -> `column`, `cols` -> `columns`

sdv/constraints/tabular.py
```diff
@@ -379,3 +381,73 @@ def reverse_transform(self, table_data):
...
+            digits = self._digits if isinstance(self._digits, int) else self._digits[col]
```
csala, line 428: If I understand the proposal correctly, the `columns` argument must always be given.
If that is the case, I would just build a dict inside the `__init__` to avoid having to check for types every time:

```python3
class Rounding(Constraint):
    def __init__(self, digits, columns):
        if isinstance(column, str):
            columns = [columns]    # see comment above about support a string

        if isinstance(digits, dict):
            if set(digits.keys()) != set(columns):
                raise ValueError()   # see comment above about validating inputs
                
            self._digits = digits
        else:
            self._digits = {column: digits for column in columns}
```

After this, `self._digits[column]` will always return an integer: either the constant or the one that had been specifically given for the column.

csala (changes requested):
Added a few comments.
The most important one is about using vector operations instead of an apply (so, iterating=

sdv/constraints/tabular.py
katxiao: That makes sense. I was trying to account for both the case where `_digits` is an `int` and the case where `_digits` is a `dict`. That way if `_digits` is an `int` and applies to all the columns, we can just filter it once instead of iterating through each column and applying the filter separately.

I think the approach of making a dict in `__init__` is more readable though.

sdv/constraints/tabular.py
```diff
@@ -379,3 +381,73 @@ def reverse_transform(self, table_data):
...
+    """
+
+    def __init__(self, columns, digits, handling_strategy='transform', tolerance=1):
+        self._columns = columns
+        self._digits = digits
+        self._tolerance = tolerance
+        super().__init__(handling_strategy=handling_strategy, fit_columns_model=False)
+
+    def is_valid(self, table_data):
+        """Determine if the data satisfies the rounding constranit.
```
amontanez24, line 412: minor: constraint*

sdv/constraints/tabular.py
amontanez24, line 424: what happens if the rounding is a negative number? If a user creates the constraint with the `digits` value set to -2, I don't think this strategy works because the `Decimal` exponent property has a minimum of 0 I think

amontanez24 (dismissed):
I think the `is_valid` method may need to change to handle rounding with a negative value for `digits` (ie. rounding to the hundreds place or something like that)

sdv/constraints/tabular.py
katxiao: Ah good point! I updated this map method to find the number of trailing zeros if there are no decimals.

csala (changes requested):
Thinking a bit more about this, maybe things could be make much simpler by removing the possibility to pass a `dict` in as the `digits` option: We assume that a single `Rounding` constraint will always round to a fixed amount of digits, and if multiple rounding levels are needed multiple constraints can be created.

On the other side, I would reinterpret the `tolerance` as "the maximum distance at which a value can be form its rounded version to consider it valid".

Having done this, the `is_valid` and `reverse_transform` become just this:

```python3
def is_valid(self, data):
    return (data - data.round(self._digits)) <= self._tolerance

def reverse_transform(self, data):
    return data.round(self._digits)
```

On another side of things, I would add a validation in the `__init__` to ensure:
- That the `digits` value is not bigger than 15 (python cannot distinguish floats beyond the 15th digit)
- That the `tolerance` value is smaller than the rounding level

sdv/constraints/tabular.py
```diff
@@ -401,3 +401,68 @@ def reverse_transform(self, table_data):
...
+
+        Args:
+            table_data (pandas.DataFrame):
+                Table data.
+
+        Returns:
+            pandas.Series:
+                Whether each row is valid.
+        """
+        rounded = table_data.round({column: self._digits for column in self._columns})
```
csala, line 450: We will need to restrict the validation to the constraint`[self._columns]`, since the other columns should not be compared.
Then, you can skip the dict and just use digits:

```python3
columns = table_data[self._columns]
rounded = columns.round(self._digits)
valid = (columns - rounded).abs() <= self._tolerance
```

sdv/constraints/tabular.py
```diff
@@ -401,3 +401,68 @@ def reverse_transform(self, table_data):
...
+
+        Args:
+            table_data (pandas.DataFrame):
+                Table data.
+
+        Returns:
+            pandas.DataFrame:
+                Transformed data.
+        """
+        return table_data.round({column: self._digits for column in self._columns})
```
csala, line 468: I would capture the `dict` in a variable inside the `__init__` to avoid having to do the loop at every call.

tests/unit/constraints/test_tabular.py
```diff
@@ -1274,3 +1274,174 @@ def test_reverse_transform(self):
             'c': [5, 7, 9]
         })
         pd.testing.assert_frame_equal(expected_out, out)
+
+
+class TestRounding():
+
+    def test___init__(self):
```
csala, line 1281: We should add tests cases that cover the validations: digits > 15 and tolerance < digits

tests/unit/constraints/test_tabular.py
```diff
@@ -1274,3 +1274,174 @@ def test_reverse_transform(self):
...
+            'b': [4.12, 5.51, 6.901],
+            'c': [5.310, 7.12, 9.120]
+        })
+        out = instance.is_valid(table_data)
+
+        # Assert
+        expected_out = pd.Series([True, True, True])
+        pd.testing.assert_series_equal(expected_out, out)
+
+    def test_is_valid_negative_digit_valid(self):
```
csala, line 1331: I think we could merge some of the `is_valid` tests to cover multiple cases at once.
I would create 3 tests: digits = 0, digits > 0, digits < 0

And each one of the can have:
- 1 Invalid row
- 1 row that is exactly valid
- 1 row that is slightly over the rounded level but still valid (within tolerance)
- 1 row that is slightly below the rounded level but still valid (within tolerance)

Also, at the column level, each test should have:
- Multiple number of rounded columns (make 1 test with 1 column, the others with 2 or more, to cover all cases)
- At least 1 numerical column that is not being rounded, which should have a deeper rounding level to make sure that it is not being rounded
- At least 1 non-numerical column

Finally, all the test cases should have random Null values here and there to ensure that they do not break things

tests/unit/constraints/test_tabular.py
```diff
@@ -1274,3 +1274,174 @@ def test_reverse_transform(self):
...
+            'b': [411, -551, 690050],
+            'c': [5.313, 7.12, 9.120]
+        })
+        out = instance.is_valid(table_data)
+
+        # Assert
+        expected_out = pd.Series([False, False, False])
+        pd.testing.assert_series_equal(expected_out, out)
+
+    def test_reverse_transform(self):
```
csala, line 1418: We should probably test this multiple times two, with different digit levels (0, >0, <0), random Null values, and unaffected columns.

csala (changes requested):

csala (approved):
Looks good to go now! Thanks @katxiao !

sdv/constraints/tabular.py
```diff
@@ -401,3 +401,70 @@ def reverse_transform(self, table_data):
...
+        if tolerance is not None and tolerance >= 10**(-1 * digits):
+            raise ValueError('Tolerance must be less than the rounding level')
+
+        if isinstance(columns, str):
+            self._columns = [columns]
+        else:
+            self._columns = columns
+
+        self._digits = digits
+        self._round_config = {column: self._digits for column in self._columns}
```
amontanez24, line 436: I probably missed something, but is there a reason we don't let users provide the config themselves? It seems like it wouldn't be a big stretch since we already create a map of column names to a digits value

amontanez24 (commented):
This looks good, but I am just curious as to why we don't let users specify the rounding config so that they could make one constraint to handle all the columns they want

sdv/constraints/tabular.py
katxiao: I believe it seemed overly complicated, because users could just specify multiple Rounding constraints if they wanted different columns to be rounded to different digits. (see https://github.com/sdv-dev/SDV/pull/484#pullrequestreview-695337713)

amontanez24: Yeah but I believe that was under the assumption each column would have it's own constraint. If the constraint is implemented in such a way that it already checks all columns for the amount to round them then I think it would be nice for users to be able to specify that instead of having to make however many constraints

katxiao: I think the comment was referring to - one constraint specifies one rounding level, so if you want to implement multiple rounding levels (aka digits) you can specify multiple Rounding constraints. I think I like that abstraction better, rather than extending it to all possible rounding levels in the dataset.

csala: > I think the comment was referring to - one constraint specifies one rounding level, so if you want to implement multiple rounding levels (aka digits) you can specify multiple Rounding constraints. I think I like that abstraction better, rather than extending it to all possible rounding levels in the dataset.

That is right. The goal was to not overcomplicate both the implementation and the user interface.

amontanez24: I see, but I don't think it would change the implementation or UI at all. You would just change fit to not create the dict if it's provided and everything else stays the same

csala (approved):

csala (approved):

PR #485, opened 2021-06-25
Enhance GreaterThan constraint to allow comparing to scalar values

### LINKED ISSUE #368: Improve GreaterThan constraint
## Problem Description

Allow the user to pass a certain value to the GreaterThan constraint instead of just columns. Also allow the possibility to choose when using transform strategy, which value to keep (the lower or the higher).

## Expected behavior
```python
from sdv.constraints import GreaterThan

gt_int_column = GreaterThan(low=1, high='column', handling_strategy='transform', transform_value='high')
gt_column_int = GreaterThan(low='column_a', high=100, handling_strategy='transform', transform_value='low')
gt_columns = GreaterThan(low='column_a', high='column_b', handling_strategy='transform', transform_value='low')
```

amontanez24:
Resolve #368


sdv/constraints/tabular.py
```diff
@@ -211,13 +226,31 @@ def _fit(self, table_data):
...
+            self.constraint_columns = tuple(self._low)
+            self._dtype = table_data[self._low].dtype
+
+        if not self._low_is_scalar and not self._high_is_scalar:
+            self._dtype = table_data[self._high].dtype
+
         separator = '#'
-        while not self._valid_separator(table_data, separator, self.constraint_columns):
-            separator += '#'
+        self._diff_column = separator + separator.join(self.constraint_columns)
```
amontanez24, line 246: @csala The reason that this doesn't use the `_valid_separator` method is that the `self.column_constraints` tuple might now only have one element. In this case, the `_valid_separator` method gets stuck in an infinite loop, because there is nothing to join the only column to, so it finds the column name inside the table. Not sure what you think is a good approach. I didn't use the `uuid` either because it will be good to know what the column name ends up being for integration tests. Also not sure how to keep the `uuid` from being something in the column names of the table

sdv/constraints/tabular.py
```diff
@@ -253,10 +290,11 @@ def _transform(self, table_data):
                 Transformed data.
         """
         table_data = table_data.copy()
-        low_column = table_data[self._low]
-        diff = table_data[self._high] - low_column
+        low = self._low if self._low_is_scalar else table_data[self._low]
```
katxiao, line 293: nit: Could pull this into a class methods `get_low_value`, `get_high_value` since this logic is repeated so many times

sdv/constraints/tabular.py
```diff
@@ -292,22 +325,31 @@ def reverse_transform(self, table_data):
         """
         table_data = table_data.copy()
         diff = (np.exp(table_data[self._diff_column]).round() - 1).clip(0)
-        if self._diff_is_datetime(table_data):
+        if self._is_datetime:
```
katxiao, line 328: Could `_is_datetime` ever be `None` here? Since it depends on calling `fit` first.

sdv/constraints/base.py
```diff
@@ -161,6 +161,8 @@ def fit(self, table_data):
             table_data (pandas.DataFrame):
                 Table data.
         """
+        self._fit(table_data)
```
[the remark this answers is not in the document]
amontanez24: In `_fit` the `_constraint_columns` might be changed. Id the part below runs before this, it will try to access the wrong `_constraint_columns`. Before there were no constraints that changed `_constraint_columns` in `_fit` so this wasn't a problem

sdv/constraints/tabular.py
amontanez24: it could. In a similar way though, `self._diff_column` could also be `None`. So I am assuming we want users to fit the constraint before reverse transforming. Not sure what you think

katxiao: I see. If it becomes a usability problem, we could add an explicit check and throw an informative error. Seems fine for now since most constraints probably assume this.

katxiao (approved):

sdv/constraints/tabular.py
```diff
@@ -292,22 +325,31 @@ def reverse_transform(self, table_data):
...
-            table_data[self._low] = (high_column - diff).astype(self._dtype)
+            high = self._high if self._high_is_scalar else table_data[self._high]
+            table_data[self._low] = (high - diff).astype(self._dtype)
 
         else:
-            low_column = table_data[self._low]
             invalid = ~self.is_valid(table_data)
-            new_high_values = low_column.loc[invalid] + diff.loc[invalid]
-            table_data[self._high].loc[invalid] = new_high_values.astype(self._dtype)
+            if self._high_is_scalar and not self._low_is_scalar:
```
csala, line 341: I think this if/else block could be simplified by only assigning `new_values` and then doing:

```python3
if ...
    new_values = ...
    column = ...
elif ...
    ...
else:
    ...

table_data[column].loc[invalid] = new_values.astype(self._dtype)
```

`column` should also be computed beforehand, in the `__init__`, since it will never change.

csala (changes requested):

sdv/constraints/tabular.py
amontanez24: I don't understand this. If we only validate against the columns not in the `self.constraint_columns`, it is possible to end up with a column name that is already present. If the `self.constraint_columns` contains `('a',)`, then the diff column should be called `'a#'`. But if we just join the columns it will return `'a'`

sdv/constraints/tabular.py
amontanez24: I think if we did that, we should probably just have on in the base class that complains if you try and reverse transform or transform before having fit anything

sdv/constraints/tabular.py
amontanez24: I kind of don't like computing the column beforehand because we still have to do all the if/else statements here to calculate `new_values`, so we'd be doing that again earlier on. Also, it's decoupling the calculation of which column to use from the calculation of the values to replace. I think that could get confusing

sdv/demo.py
```diff
@@ -325,6 +325,11 @@ def _load_tabular_dummy():
     faker = Faker()
     names = [faker.name() for _ in range(12)]
     adresses = [faker.address() for _ in range(12)]
+    start_date = datetime(1980, 1, 1)
+    start_dates = [
+        start_date + timedelta(days=np.random.randint(0, 14600))
```
csala, line 330: I think this field choice is delicate, because we will then need to make the `years_in_the_company` match the `start_date`. Maybe we could use a date which is not related to for how long the employee has been in the company?

csala (changes requested):
This looks great @amontanez24 ! I love how thorough the tests are!

I added a couple of minor comments about the code itself, and one about the demo and the `start_date` field that was added, that may end up being confusing.

sdv/demo.py
amontanez24: good point

sdv/constraints/tabular.py
```diff
@@ -202,35 +207,100 @@ class GreaterThan(Constraint):
...
+        self._low_is_scalar = low_is_scalar
         super().__init__(handling_strategy=handling_strategy,
                          fit_columns_model=fit_columns_model)
 
+    def _get_low_value(self, table_data):
+        if self._low_is_scalar:
+            return self._low
+        elif self._low in table_data.columns:
+            return table_data[self._low]
+        return None
```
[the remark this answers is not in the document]
amontanez24: No because if `self._low_is_scalar` is false, the column could still have been dropped. That value can be false, but `self._drop` can be set to `low`

csala (approved):
This looks good @amontanez24 !

PR #492, opened 2021-06-30
Add rounding and min/max arguments that are passed down to the NumericalTransformer

amontanez24:


sdv/metadata/table.py
```diff
@@ -84,6 +84,12 @@ class Table:
             The columns in the dataframe which are constant within each
             group/entity. These columns will be provided at sampling time
             (i.e. the samples will be conditioned on the context variables).
+        rounding (int, str or None):
+            Define rounding scheme for ``NumericalTransformer``.
```
csala, line 88: I would add an explanation of the possible values here (and on the other models)

csala (approved):
It looks good so far (and after RDT bug has been fixed and adapted)!
I would just add some more detail on the docstrings to explain the possible inputs.

PR #493, opened 2021-06-30
Implements the Between constraint

fealho:


docs/user_guides/single_table/constraints.rst
```diff
@@ -200,7 +228,8 @@ constraints that we just defined as a ``list``:
     constraints = [
         unique_company_department_constraint,
         age_gt_age_when_joined_constraint,
-        years_in_the_company_constraint
+        years_in_the_company_constraint,
+        reasonable_age_constraint
```
amontanez24, line 232: Are there any problems that arise from having to constraints on the age column?

sdv/constraints/tabular.py
```diff
@@ -388,3 +390,155 @@ def reverse_transform(self, table_data):
...
+        """
+        table_data = table_data.copy()
+        low = self._get_low(table_data)
+        high = self._get_high(table_data)
+        data = table_data[self._transformed_column]
+
+        data = 1 / (1 + np.exp(-data))
+        data = (data - 0.025) / 0.95
+        data = data * (high - low) + low
+        data = data.clip(low, high)
```
amontanez24, line 539: We talked about this with @csala but I'm curious how much this transformation reduces the need to clip from say just modeling the column without a transformation and clipping the sampled value to the bounds directly

amontanez24 (commented):
This looks really good but I had a couple of questions

docs/user_guides/single_table/constraints.rst
fealho: I didn't notice any.

sdv/constraints/tabular.py
fealho: It really depends on the constants you use when scaling: if you increase the multiplier from 0.95 to 0.99, it removes virtually all out of bounds values, but then the model just doesn't produce any values close to the tails. If you move the constants the other way, the opposite happens. I didn't play around with it too much so there may be a better set of constants that balance the number of out of bound values/performance ratio.

csala:
@fealho it seems like there are some conflicts with `master` branch. Would you mind merging `master` into this branch and fixing them?

tests/unit/constraints/test_tabular.py
```diff
@@ -875,7 +877,7 @@ def test_is_valid_strict_true(self):
           for the other two.
         """
         # Setup
-        instance = GreaterThan(low='a', high='b', strict=True)
+        instance = GreaterThan(low='a', high='b', strict=False)
```
csala, line 880: I think this change and the ones below sapping True/False should not be here, since now they do not match the test_case names

sdv/constraints/tabular.py
```diff
@@ -539,6 +541,158 @@ def reverse_transform(self, table_data):
...
+            or ``reject_sampling``. Defaults to ``transform``.
+    """
+
+    def __init__(self, column, low, high, strict=False, handling_strategy='transform',
+                 fit_columns_model=True):
+        self.constraint_column = column
+        self._low = low
+        self._high = high
+        self._strict = strict
+        self._transformed_column = f'#{self.constraint_column}#{self._low}#{self._high}'
```
csala, line 575: The column name should be chosen during `fit` like it is being done in the `GreaterThan` constraint, making sure that the separator selected does not produce a column name that already exists in the table.

sdv/constraints/tabular.py
```diff
@@ -539,6 +541,158 @@ def reverse_transform(self, table_data):
...
+        strict (bool):
+            Whether the comparison of the values should be strict ``>=`` or
+            not ``>`` when comparing them. Currently, this is only respected
+            if ``reject_sampling`` or ``all`` handling strategies are used.
+        handling_strategy (str):
+            How this Constraint should be handled, which can be ``transform``
+            or ``reject_sampling``. Defaults to ``transform``.
+    """
+
+    def __init__(self, column, low, high, strict=False, handling_strategy='transform',
```
csala, line 569: We should also add `low_is_scalar` and `high_is_scalar` arguments like in the `GreaterThan` constraint.

sdv/constraints/tabular.py
```diff
@@ -539,6 +541,158 @@ def reverse_transform(self, table_data):
...
+
+        Args:
+            table_data (pandas.DataFrame):
+                The Table data.
+
+        Returns:
+            pandas.DataFrame or float:
+                The lower bound.
+        """
+        if isinstance(self._low, str):
```
csala, line 593: The decision should not be made over whether the value is a string or not, since a column name could be an integer value.
This should rather use the `low_is_scalar` like in the `GreaterThan` constraint.

sdv/constraints/tabular.py
```diff
@@ -539,6 +541,158 @@ def reverse_transform(self, table_data):
...
+
+        Args:
+            table_data (pandas.DataFrame):
+                Table data.
+
+        Returns:
+            pandas.Series:
+                Whether each row is valid.
+        """
+        if self._strict:
```
csala, line 628: Maywe we could use `operator.le` and `operator.lt`?

```python
import operator

def __init__(self, ...strict):
   self._lt = operator.le if strict else operator.le

def is_valid(self, table_data):
    ...
    satisfy_low_bound = self._lt(self._get_low(table_data), table_data[self.constraint_column])
    ...
```

csala (changes requested):
Everything seems ok so far, but there are a few things that maybe should be changed to make this have a behavior closer to the `GreaterThan` constraint, including the `high_is_scalar` and `low_is_scalar` arguments, and figuring out the column name sensibly.

tests/unit/constraints/test_tabular.py
fealho: I forgot to swap the test names. I changed `strict` to `False` to match the documentation.

csala: Oh! I did not realize that the docs did not match the method name. Good catch!

sdv/constraints/tabular.py
fealho: That's neat, I didn't know that 👍

csala (approved):
This looks ready to go now!

PR #500, opened 2021-07-06
NumericalTransformer arguments fix

csala:
Proposal to fix the problem derived from using by default and RDT that learns rounding and min/max arguments.

### DIFF
--- a/sdv/metadata/table.py
+++ b/sdv/metadata/table.py
@@ -407,15 +407,23 @@ class Table:
         """
         meta_dtypes = self.get_dtypes(ids=False)
         dtypes = {}
+        numerical_extras = []
         for column in data.columns:
             if column in meta_dtypes:
                 dtypes[column] = meta_dtypes[column]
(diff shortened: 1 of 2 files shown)


sdv/metadata/table.py
```diff
@@ -407,15 +407,23 @@ def _fit_hyper_transformer(self, data, extra_columns):
         """
         meta_dtypes = self.get_dtypes(ids=False)
         dtypes = {}
+        numerical_extras = []
         for column in data.columns:
             if column in meta_dtypes:
                 dtypes[column] = meta_dtypes[column]
             elif column in extra_columns:
-                dtypes[column] = data[column].dtype.kind
+                dtype_kind = data[column].dtype.kind
```
amontanez24, line 415: is `extra_columns` already defined to be any columns we add?

amontanez24 (approved):
I'm not sure what the purpose of using the real model is, but the fix LGTM!

PR #524, opened 2021-07-26
Integration test that covers a single child row per parent row

### LINKED ISSUE #497: Unit test for HMA1 when working with a single child row per parent row
### Problem Description

Within the code of `HMA1` , https://github.com/sdv-dev/SDV/blob/54fb573156c442171213c9c912903222296beb0e/sdv/relational/hma.py#L106 , is working with an edge case where the length of the child rows is equal to 1. This is because we don't want to learn the distribution for those values, since they will be `0` and will alter the `std` by a lot.

Do we have a test that goes over this edge case?

pvk-developer:
Implemented integration tests to cover the edge case of single child row per parent row.

Resolves #497

### DIFF
--- a/tests/integration/relational/test_hma.py
+++ b/tests/integration/relational/test_hma.py
@@ -124,29 +124,20 @@ def test_hma1_single_child_row_single_parent_row():
         'value': range(5)
     })
 
-    parent_b = pd.DataFrame({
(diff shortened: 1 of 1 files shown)


csala (approved):
LGTM! Thanks @pvk-developer !

PR #527, opened 2021-07-28
Change rtol comparison from < to <=

### LINKED ISSUE #525: Sampling with conditions={column: 0.0} for float columns doesn't work
### Environment Details

* SDV version: 0.11.0
* Python version: 3.8.5
* Operating System: Ubuntu 20.04.1 LTS 

### Error Description

Trying to apply a condition on sampling with the value `0.` for a float column leads to an exception.

### Steps to reproduce

```

data = load_tabular_demo('student_placements')
data['experience_years'] = data['experience_years'].astype(float)  # for the demonstration
model = GaussianCopula()
model.fit(data)
model.sample(1, conditions={'experience_years': 0.})

<ipython-input-147-1a9ae5e719f8> in <module>
      6 model = GaussianCopula()
      7 model.fit(data)
----> 8 model.sample(1, conditions={'experience_years': 0.})

/opt/conda/lib/python3.8/site-packages/sdv/tabular/base.py in sample(self, num_rows, max_retries, max_rows_multiplier, conditions, float_rtol, graceful_reject_sampling)

/opt/conda/lib/python3.8/site-packages/sdv/tabular/base.py in _conditionally_sample_rows(self, dataframe, max_retries, max_rows_multiplier, condition, transformed_condition, float_rtol, graceful_reject_sampling)

ValueError: No valid rows could be generated with the given conditions.
```

### Ability to contribute

I think that I can fix the issue and I'll create a pull request.
I suspect that it happens because of `rtol` with zero. The method `_filter_conditions` checked for `<` instead of `<=` (In numpy, `<=` is used [[source](https://github.com/numpy/numpy/blob/b235f9e701e14ed6f6f6dcba885f7986a833743f/numpy/core/numeric.py#L2337)]):
```
if column_values.dtype.kind == 'f':
distance = value * float_rtol
sampled = sampled[np.abs(column_values - value) < distance]
sampled[column] = value
```

shlomihod:
Resolve #525: trying to apply a condition on sampling with the value 0. for a float column leads to an exception.

### DIFF
--- a/sdv/tabular/base.py
+++ b/sdv/tabular/base.py
@@ -181,7 +181,7 @@ class BaseTabularModel:
             column_values = sampled[column]
             if column_values.dtype.kind == 'f':
                 distance = value * float_rtol
-                sampled = sampled[np.abs(column_values - value) < distance]
+                sampled = sampled[np.abs(column_values - value) <= distance]
                 sampled[column] = value
             else:
                 sampled = sampled[column_values == value]


csala (changes requested):
I think looks good @shlomihod 
What do you think about also adding an integration test to validate that it works as expected (and continues to do so in the future)?

I would add it in `tests/integration/tabular/test_base.py`, something similar to the 
`test_conditional_sampling_graceful_reject_sampling_True_dict` test: https://github.com/sdv-dev/SDV/blob/08af5aa6c8ae4c6e1246c425b735a581cc52fd9b/tests/integration/tabular/test_base.py#L22
But setting dtype float, conditioning on `{'column1': 0.0}`, and then, instead of expecting an Exception, capturing the output and validating that column1 is actually 0.0 in the returned DataFrame.

tssbas:
Dear @csala I am experiencing the same problem as described in #527 . After debugging I have come to the same conclusion that the "less than" should actually be "less or equal to". I wanted to submit a pull request but now I see that this fix already exists, but for some reason still hasn't been merged. Is there any chance this PR gets merged and included in the next release of SDV?

Thanks!

katxiao:
Hi @shlomihod, could you add the unit test that csala requested?

PR #528, opened 2021-07-28
Add datetime functionality to Between constraint

### LINKED ISSUE #504: Add datetime functionality to Between constraint
### Problem Description

The `Between` constraint should be able to handle datetime columns as well.

### Expected behavior
The following should be a valid constraint in SDV
```
birthday_constraint = Between(
    column='birthday',
    low=pd.to_datetime('1900-01-01'),
    high=pd.to_datetime('2021-01-01'),
    handling_strategy='transform'
)
```

katxiao:
Resolves #504

### DIFF
--- a/sdv/constraints/tabular.py
+++ b/sdv/constraints/tabular.py
@@ -591,7 +591,7 @@ class Between(Constraint):
     def __init__(self, column, low, high, strict=False, handling_strategy='transform',
                  fit_columns_model=True, high_is_scalar=None, low_is_scalar=None):
         self.constraint_column = column
-        self.constraint_columns = tuple(column)
+        self.constraint_columns = (column,)
         self._low = low
         self._high = high
         self._strict = strict
--- a/sdv/constraints/tabular.py
+++ b/sdv/constraints/tabular.py
@@ -22,12 +22,12 @@ Currently implemented constraints are:
 
 import operator
 import uuid
-from datetime import datetime
 
 import numpy as np
 import pandas as pd
 
 from sdv.constraints.base import Constraint, import_object
+from sdv.constraints.utils import is_datetime_type
 
 
 class CustomConstraint(Constraint):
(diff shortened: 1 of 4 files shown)


sdv/constraints/tabular.py
```diff
@@ -642,13 +642,36 @@ def _get_diff_column_name(self, table_data):
...
+        high = self._get_high_value(table_data)
+        column = table_data[self.constraint_column]
+
+        is_low_datetime = (pd.api.types.is_datetime64_any_dtype(low)
+                           or isinstance(low, pd.Timestamp)
+                           or isinstance(low, datetime))
+        is_high_datetime = (pd.api.types.is_datetime64_any_dtype(high)
+                            or isinstance(high, pd.Timestamp)
+                            or isinstance(high, datetime))
+        is_column_datetime = (pd.api.types.is_datetime64_any_dtype(high)
```
katxiao, line 656: @csala `is_datetime64_any_dtype` should work for any datetime dtype. I preferred this over the `dtype.kind` check, because doing `low.dtype.kind` will error out if `low` is a scalar.

sdv/constraints/tabular.py
```diff
@@ -723,7 +746,11 @@ def reverse_transform(self, table_data):
         data = data * (high - low) + low
         data = data.clip(low, high)
 
-        table_data[self.constraint_column] = data
+        if self._is_datetime:
+            table_data[self.constraint_column] = pd.to_datetime(data)
```
katxiao, line 750: @csala I know we talked about storing the original column's dtype to use in the `to_datetime` conversion here, but I don't think that's necessary since pandas seems to only have `datetime[ns]`(see [doc](https://pandas.pydata.org/pandas-docs/stable/user_guide/basics.html?highlight=basics#dtypes)). If you do something like `pd.Series([pd.to_datetime(1627497875, unit='s')])` it will give you back a data series of `dtype: datetime64[ns]`

pvk-developer (approved):
LGTM!

csala (approved):
This is excellent! Thanks @katxiao

PR #529, opened 2021-07-28
No longer cast categorical types as strings

### LINKED ISSUE #503: Update metadata.json for duration of student_placements dataset
### Environment Details

* SDV version: Any
* Python version: 3.x
* Operating System: Any

### Error Description

When using the `load_tabular_demo` method to load the `student_placements` dataset, the types for the `duration` column are incorrect. Upon further investigation, I noticed that the `metadata.json` file for that dataset incorrectly annotated the `duration` column as a `Categorical` type. The snippet for that is below
```  
{
    "tables": {
        "student_placements": {
            "fields": {
                "start_date": {
                    "type": "datetime",
                    "format": "%Y-%m-%d"
                },
                "end_date": {
                    "type": "datetime",
                    "format": "%Y-%m-%d"
                },
                "salary": {
                    "type": "numerical",
                    "subtype": "integer"
                },
                "duration": {
                    "type": "categorical"
                },
```

This causes the types for that column to be loaded as strings as seen in the "Steps to Reproduce" section. This causes certain operations to not correctly work during the demo.
### Solution
 The s3 metadata file should be updated to label the correct type for the column.

### Steps to reproduce

```
student_placements = load_tabular_demo('student_placements')
x = student_placements['duration'].apply(type)
>>> x
0        <class 'str'>
1        <class 'str'>
2        <class 'str'>
3      <class 'float'>
4        <class 'str'>
            ...
210      <class 'str'>
211      <class 'str'>
212      <class 'str'>
213      <class 'str'>
214    <class 'float'>
Name: duration, Length: 215, dtype: object
```

amontanez24:
resolves #503 

#503 was filed because a categorical column that was made up of float values was being loaded as a string type. When certain constraints were applied to it, it crashed since the column was expected to have numerical values but actually had string. Since we allow for categorical data to be of any type, we no longer want to cast a column as `str` if it is categorical.

### DIFF
--- a/sdv/metadata/dataset.py
+++ b/sdv/metadata/dataset.py
@@ -21,9 +21,7 @@ def _read_csv_dtypes(table_meta):
     dtypes = dict()
     for name, field in table_meta['fields'].items():
         field_type = field['type']
-        if field_type == 'categorical':
-            dtypes[name] = str
-        elif field_type == 'id' and field.get('subtype', 'integer') == 'string':
+        if field_type == 'id' and field.get('subtype', 'integer') == 'string':
             dtypes[name] = str
 
     return dtypes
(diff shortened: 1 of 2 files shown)


csala (approved):
LGTM!

PR #530, opened 2021-07-28
Fix ColumnFormula conditioning bug

### LINKED ISSUE #507: Conditioning on a non-constraint column using the `ColumnFormula` constraint
Conditioning on any column except the constraint column causes the `ColumnFormula` constraint to produce the following error:
```
KeyError                                  Traceback (most recent call last)
~/opt/anaconda3/envs/test/lib/python3.8/site-packages/pandas/core/indexes/base.py in get_loc(self, key, method, tolerance)
   2894             try:
-> 2895                 return self._engine.get_loc(casted_key)
   2896             except KeyError as err:

pandas/_libs/index.pyx in pandas._libs.index.IndexEngine.get_loc()

pandas/_libs/index.pyx in pandas._libs.index.IndexEngine.get_loc()

pandas/_libs/hashtable_class_helper.pxi in pandas._libs.hashtable.PyObjectHashTable.get_item()

pandas/_libs/hashtable_class_helper.pxi in pandas._libs.hashtable.PyObjectHashTable.get_item()

KeyError: 'b'

The above exception was the direct cause of the following exception:

KeyError                                  Traceback (most recent call last)
/var/folders/3p/f4b70gn5165464cqyvg157qw0000gn/T/ipykernel_42215/1774337226.py in <module>
----> 1 samples = model.sample(10, conditions={'c': 1})
      2 samples

~/Desktop/SDV/sdv/tabular/base.py in sample(self, num_rows, max_retries, max_rows_multiplier, conditions, float_rtol, graceful_reject_sampling)
    440                 raise ValueError(f'Invalid column name `{column}`')
    441 
--> 442         transformed_conditions = self._metadata.transform(conditions, on_missing_column='drop')
    443         condition_columns = list(conditions.columns)
    444         transformed_columns = list(transformed_conditions.columns)

~/Desktop/SDV/sdv/metadata/table.py in transform(self, data, on_missing_column)
    577 
    578         LOGGER.debug('Transforming constraints for table %s', self.name)
--> 579         data = self._transform_constraints(data, on_missing_column)
    580 
    581         LOGGER.debug('Transforming table %s', self.name)

~/Desktop/SDV/sdv/metadata/table.py in _transform_constraints(self, data, on_missing_column)
    539         for constraint in self._constraints:
    540             try:
--> 541                 data = constraint.transform(data)
    542             except MissingConstraintColumnError:
    543                 if on_missing_column == 'error':

~/Desktop/SDV/sdv/constraints/tabular.py in transform(self, table_data)
    520         """
    521         table_data = table_data.copy()
--> 522         del table_data[self._column]
    523 
    524         return table_data

~/opt/anaconda3/envs/test/lib/python3.8/site-packages/pandas/core/generic.py in __delitem__(self, key)
   3709             # there was no match, this call should raise the appropriate
   3710             # exception:
-> 3711             loc = self.axes[-1].get_loc(key)
   3712             self._mgr.idelete(loc)
   3713 

~/opt/anaconda3/envs/test/lib/python3.8/site-packages/pandas/core/indexes/base.py in get_loc(self, key, method, tolerance)
   2895                 return self._engine.get_loc(casted_key)
   2896             except KeyError as err:
-> 2897                 raise KeyError(key) from err
   2898 
   2899         if tolerance is not None:

KeyError: 'b' 
```

Code to reproduce it:

```python3
import pandas as pd
from sdv.constraints import *
from sdv.tabular import *

ones = [1 for i in range(10)]
data = pd.DataFrame({
    'a': ones,
    'b': ones,
    'c': ones
})

def formula(data):
    return data['a']

constraint = ColumnFormula(
    column='b',
    formula=formula
)

model.fit(data)
samples = model.sample(10, conditions={'c': 1})
samples
```

katxiao:
Resolves #507 

If the column we're conditioning on is not the `ColumnFormula` constraint column, then the code errors out [here](https://github.com/sdv-dev/SDV/blob/master/sdv/tabular/base.py#L442) when trying to apply `ColumnFormula.transform` to the `conditions` dataframe, which only contains the conditioned-on column(s). In this scenario, `ColumnFormula.transform` should just return the original data.

### DIFF
--- a/sdv/constraints/tabular.py
+++ b/sdv/constraints/tabular.py
@@ -522,6 +522,7 @@ class ColumnFormula(Constraint):
                 Transformed data.
         """
         table_data = table_data.copy()
+
         if self._drop_column:
             del table_data[self._column]
 
--- a/sdv/constraints/tabular.py
+++ b/sdv/constraints/tabular.py
@@ -483,12 +483,13 @@ class ColumnFormula(Constraint):
         handling_strategy (str):
             How this Constraint should be handled, which can be ``transform``
             or ``reject_sampling``. Defaults to ``transform``.
-        drop_column(str):
+        drop_column (bool):
             Whether or not to drop the constraint column.
     """
 
     def __init__(self, column, formula, handling_strategy='transform', drop_column=True):
         self._column = column
+        self.constraint_columns = (column,)
         self._formula = import_object(formula)
         self._drop_column = drop_column
         super().__init__(handling_strategy, fit_columns_model=False)
@@ -507,7 +508,7 @@ class ColumnFormula(Constraint):
         computed = self._formula(table_data)
         return table_data[self._column] == computed
 
-    def transform(self, table_data):
+    def _transform(self, table_data):
         """Transform the table data.
 
         The transformation consist on simply dropping the indicated column from the
@@ -523,7 +524,7 @@ class ColumnFormula(Constraint):
         """
         table_data = table_data.copy()
 
-        if self._drop_column:
+        if self._drop_column and self._column in table_data:
             del table_data[self._column]
 
         return table_data
--- a/tests/unit/constraints/test_tabular.py
+++ b/tests/unit/constraints/test_tabular.py
@@ -1773,8 +1773,8 @@ class TestColumnFormula():
         expected_out = pd.Series([False, False, False])
         pd.testing.assert_series_equal(expected_out, out)
 
-    def test_transform(self):
-        """Test the ``ColumnFormula.transform`` method.
+    def test__transform(self):
+        """Test the ``ColumnFormula._transform`` method.
 
         It is expected to drop the indicated column from the table.
 
@@ -1793,7 +1793,7 @@ class TestColumnFormula():
             'b': [4, 5, 6],
             'c': [5, 7, 9]
         })
-        out = instance.transform(table_data)
+        out = instance._transform(table_data)
 
         # Assert
         expected_out = pd.DataFrame({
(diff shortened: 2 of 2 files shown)


sdv/constraints/tabular.py
```diff
@@ -519,7 +519,9 @@ def transform(self, table_data):
                 Transformed data.
         """
         table_data = table_data.copy()
-        del table_data[self._column]
+
+        if self._column in table_data:
```
amontanez24, line 523: I'm wondering if we should (or can) do something more generalized that makes sure any constraint doesn't attempt to access columns not present in the `table_data`. We might not be able to though

katxiao: Maybe we could pass an argument through when know we are transforming on partial table view (e.g. conditions), and only ignore invalid columns in that case. Otherwise we can error out, so it's obvious if the constraint is on a non-present column.

csala: > I'm wondering if we should (or can) do something more generalized that makes sure any constraint doesn't attempt to access columns not present in the table_data. We might not be able to though

Well, we actually have that, @amontanez24 ! Those are the `self._constraint_columns` :-)
So in this case, the proper fix would be to add `self._column` to `self._constraint_columns` inside `__init__`, and then change rename `transform` to `_transform`, so the parent `transform` validates that.

On the other side, I would still keep this `if` here, because having the `_transform` be robust enough to not crash if the column is there does not hurt either.

csala (changes requested):
Following up on the in-line comments, I would suggest to do the following changes:
1. Add `self._column` to `self._constraint_columns` in the `__init__` method
2. Rename `transform` to `_transform`
3. Adapt the unit tests accordingly, so `_transform` is tested instead of `transform` (so we hit `ColumnFormula` directly without passing through the `BaseConstraint.transform` implementation.

csala (changes requested):

csala (approved):
LGTM! Let's merge this once the CI gives the OK.

PR #531, opened 2021-07-28
Add argument to drop column in ColumnFormula

### LINKED ISSUE #508: Improve error message for `ColumnFormula` constraint when constraint column used in formula
Using the `ColumnFormula` constraint where the constraint column is used in the constraint formula crashes with a `KeyError`. While this is expected, the error message could be improved to provide a more meaningful explanation of what's causing the issue. 

Below is some code to produce this error:
```python3
import pandas as pd
from sdv.constraints import *
from sdv.tabular import *

ones = [1 for i in range(10)]
data = pd.DataFrame({
    'a': ones,
    'b': ones,
    'c': ones
})

def formula(data):
    return data['a']

constraint = ColumnFormula(
    column='a',
    formula=formula
)

model.fit(data)
samples = model.sample(10)
```

katxiao:
Resolves #508

### DIFF
--- a/tests/unit/constraints/test_tabular.py
+++ b/tests/unit/constraints/test_tabular.py
@@ -1802,6 +1802,36 @@ class TestColumnFormula():
         })
         pd.testing.assert_frame_equal(expected_out, out)
 
+    def test_transform_without_dropping_column(self):
+        """Test the ``ColumnFormula.transform`` method without dropping the column.
+
+        If `drop_column` is false, expect to not drop the constraint column.
+
+        Input:
+        - Table data (pandas.DataFrame)
+        Output:
+        - Table data with the indicated column (pandas.DataFrame)
+        """
+        # Setup
+        column = 'c'
+        instance = ColumnFormula(column=column, formula=new_column, drop_column=False)
+
+        # Run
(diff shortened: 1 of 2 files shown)


tests/unit/constraints/test_tabular.py
```diff
@@ -1802,6 +1802,36 @@ def test_transform(self):
...
 
+    def test_transform_without_dropping_column(self):
+        """Test the ``ColumnFormula.transform`` method without dropping the column.
+
+        If `drop_column` is false, expect to not drop the constraint column.
+
+        Input:
+        - Table data (pandas.DataFrame)
+        Output:
+        - Table data without the indicated column (pandas.DataFrame)
```
amontanez24, line 1813: This should say "with the indicated column" right? Since it isn't being dropped

amontanez24 (approved):
One small comment but besides that LGTM!

tests/unit/constraints/test_tabular.py
katxiao: Yes, thanks!

csala (approved):
LGTM!

PR #533, opened 2021-07-29
Support multi-column specification for positive and negative constraint

sarahmish:
- Expand greaterthan constraint to work with multi-column in the case of scalar values.
- Extension for Positive and Negative classes to support multi-column.

Resolve #533.

### DIFF
--- a/docs/user_guides/single_table/constraints.rst
+++ b/docs/user_guides/single_table/constraints.rst
@@ -249,8 +249,7 @@ constraint by passing it:
     years_in_the_company_constraint = ColumnFormula(
         column='years_in_the_company',
         formula=years_in_the_company,
-        handling_strategy='transform',
-        drop_column=False
+        handling_strategy='transform'
     )
 
 Rounding Constraint
@@ -341,9 +340,9 @@ constraints that we just defined as a ``list``:
     constraints = [
         unique_company_department_constraint,
         age_gt_age_when_joined_constraint,
+        experience_years_gt_one_constraint,
         years_in_the_company_constraint,
         salary_gt_30000_constraint,
-        experience_years_gt_one_constraint,
         positive_age_constraint,
         salary_rounding_constraint,
         reasonable_age_constraint,
@@ -368,3 +367,4 @@ we defined:
     :okwarning:
 
     sampled
+
--- a/docs/user_guides/single_table/constraints.rst
+++ b/docs/user_guides/single_table/constraints.rst
@@ -340,10 +340,10 @@ constraints that we just defined as a ``list``:
     constraints = [
         unique_company_department_constraint,
         age_gt_age_when_joined_constraint,
-        experience_years_gt_one_constraint,
-        years_in_the_company_constraint,
         salary_gt_30000_constraint,
+        experience_years_gt_one_constraint,
         positive_age_constraint,
+        years_in_the_company_constraint,
         salary_rounding_constraint,
         reasonable_age_constraint,
         one_hot_constraint
--- a/docs/user_guides/single_table/constraints.rst
+++ b/docs/user_guides/single_table/constraints.rst
@@ -195,11 +195,14 @@ than one year.
         handling_strategy='reject_sampling'
     )
 
-.. note::
+.. warning::
 
-    To specify more than one column, either ``high`` or ``low`` must
-    be a scalar value, otherwise the constraint cannot be correctly
-    evaluated.
+    Warning! Passing a list of columns to the `high` or `low` arguments is only possible 
+    when the other one has been passed as a single column name or scalar value! If you need 
+    to compare multiple ``high`` columns against multiple ``low`` columns (or vice versa), 
+    you need to decompose one of the ends, ``high`` or ``low``, into multiple single column
+    names and define one ``GreaterThan`` constraint for each one of them.
+
 
 Positive and Negative Constraints
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/user_guides/single_table/constraints.rst
+++ b/docs/user_guides/single_table/constraints.rst
@@ -158,8 +158,8 @@ passing:
     )
 
 The ``GreaterThan`` constraint can also be used to guarantee a column is greater
-than a scalar value or specific datetime value instead of another column. To use
-this functionality, we can pass:
+or lower than a scalar value or specific datetime value instead of another column. 
+To use this functionality, we can pass:
 
 -  the scalar value for ``low`` or the scalar value for ``high``
 -  a flag indicating whether ``low`` or ``high`` is a scalar
@@ -173,6 +173,11 @@ this functionality, we can pass:
         scalar='low',
         handling_strategy='reject_sampling'
     )
+
+.. note::
+    If you want to indicate that the column must be *lower than* a scalar value, 
+    all you need to do is invert the arguments, pass the scalar value as the ``high`` 
+    argument, the column name as the ``low`` argument, and set the `scalar` flag to ``"high"``.
 
 Optionally, when constructing ``GreaterThan`` constraint we can specify 
 more than a single column in either the ``high`` or ``low`` arguments. 
--- a/docs/user_guides/single_table/constraints.rst
+++ b/docs/user_guides/single_table/constraints.rst
@@ -174,25 +174,48 @@ this functionality, we can pass:
         handling_strategy='reject_sampling'
     )
 
+ScalarInequality Constraint
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Similar to the GreaterThan constraint, we can we need to constraint our 
+data to be particularly larger than (or smaller than) a constant value. 
+This constraint guarantees that the given column(s) are always greater 
+(or less than) than a given scalar value. In order to use it, we need 
+to create an instance passing:
+
+-  the name of the column(s) under this constraint
+-  the ``scalar`` value of which we hold the inequality
+-  whether we are user the greater than inequality of less than
+-  the handling strategy that we want to use
+
+.. ipython:: python
+    :okwarning:
+
+    salary_gt_30000_constraint = ScalarInequality(
+        columns='salary',
+        scalar=30000,
+        greater=True,
+        handling_strategy='reject_sampling'
+    )
+
 
 Positive and Negative Constraints
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Similar to the ``GreaterThan`` constraint, we can use the ``Positive``
+Similar to the ``ScalarInequality`` constraint, we can use the ``Positive``
 or ``Negative`` constraints. These constraints enforce that a specified
 column is always positive or negative. We can create an instance passing:
 
-- the name of the ``low`` column for ``Negative`` or the name of the ``high`` column for ``Positive``
+- the name of the column(s) for ``Negative`` or ``Positive`` constraints
 - the handling strategy that we want to use
-- a boolean specifying whether to make the data strictly above or below 0, or include 0 as a possible value
 
 .. ipython:: python
     :okwarning:
 
     from sdv.constraints import Positive
 
-    positive_prior_exp_constraint = Positive(
-        high='prior_years_experience',
+    positive_prior_exp_and_age_constraint = Positive(
+        columns=['prior_years_experience', 'age'],
         strict=False,
         handling_strategy='reject_sampling'
     )
@@ -321,7 +344,7 @@ constraints that we just defined as a ``list``:
         age_gt_age_when_joined_constraint,
         years_in_the_company_constraint,
         salary_gt_30000_constraint,
-        positive_prior_exp_constraint,
+        positive_prior_exp_and_age_constraint,
         salary_rounding_constraint,
         reasonable_age_constraint,
         one_hot_constraint
--- a/docs/user_guides/single_table/constraints.rst
+++ b/docs/user_guides/single_table/constraints.rst
@@ -198,9 +198,9 @@ is more than one year.
 Positive and Negative Constraints
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Similar to the ``ScalarInequality`` constraint, we can use the ``Positive``
-or ``Negative`` constraints. These constraints enforce that a specified
-column is always positive or negative. We can create an instance passing:
+Similar to the ``GreaterThan`` constraint, we can use the ``Positive``
+or ``Negative`` constraints. These constraints enforce that specified
+column(s) are always positive or negative. We can create an instance passing:
 
 - the name of the column(s) for ``Negative`` or ``Positive`` constraints
 - the handling strategy that we want to use
--- a/docs/user_guides/single_table/constraints.rst
+++ b/docs/user_guides/single_table/constraints.rst
@@ -161,9 +161,8 @@ The ``GreaterThan`` constraint can also be used to guarantee a column is greater
 than a scalar value or specific datetime value instead of another column. To use
 this functionality, we can pass:
 
--  the scalar value for ``low``
--  the scalar value for ``high``
--  a boolean indicating ``low`` or ``high`` is a scalar
+-  the scalar value for ``low`` or the scalar value for ``high``
+-  a flag indicating whether ``low`` or ``high`` is a scalar
 
 .. ipython:: python
     :okwarning:
@@ -171,14 +170,15 @@ this functionality, we can pass:
     salary_gt_30000_constraint = GreaterThan(
         low=30000,
         high='salary',
+        scalar='low',
         handling_strategy='reject_sampling'
     )
 
-Optionally, when constructing ``GreaterThan`` constraint for scalar
-comparisons, we can specify more than a single column in either 
-the ``high`` or ``low`` arguments. For example, we can create a 
-``GreaterThan`` constraint that ensures that the years of experience
-is more than one year.
+Optionally, when constructing ``GreaterThan`` constraint we can specify 
+more than a single column in either the ``high`` or ``low`` arguments. 
+For example, we can create a ``GreaterThan`` constraint that that ensures 
+that both the years in the company and prior years of experience is more 
+than one year.
 
 .. ipython:: python
     :okwarning:
@@ -186,6 +186,7 @@ is more than one year.
     experience_years_gt_one_constraint = GreaterThan(
         low=1,
         high=['years_in_the_company', 'prior_years_experience'],
+        scalar='low',
         handling_strategy='reject_sampling'
     )
 
@@ -199,7 +200,7 @@ Positive and Negative Constraints
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Similar to the ``GreaterThan`` constraint, we can use the ``Positive``
-or ``Negative`` constraints. These constraints enforce that specified
+or ``Negative`` constraints. These constraints enforce that the specified
 column(s) are always positive or negative. We can create an instance passing:
 
 - the name of the column(s) for ``Negative`` or ``Positive`` constraints
--- a/docs/user_guides/single_table/constraints.rst
+++ b/docs/user_guides/single_table/constraints.rst
@@ -174,30 +174,26 @@ this functionality, we can pass:
         handling_strategy='reject_sampling'
     )
 
-ScalarInequality Constraint
-~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-Similar to the GreaterThan constraint, we can we need to constraint our 
-data to be particularly larger than (or smaller than) a constant value. 
-This constraint guarantees that the given column(s) are always greater 
-(or less than) than a given scalar value. In order to use it, we need 
-to create an instance passing:
-
--  the name of the column(s) under this constraint
--  the ``scalar`` value of which we hold the inequality
--  whether we are user the greater than inequality of less than
--  the handling strategy that we want to use
+Optionally, when constructing ``GreaterThan`` constraint for scalar
+comparisons, we can specify more than a single column in either 
+the ``high`` or ``low`` arguments. For example, we can create a 
+``GreaterThan`` constraint that ensures that the years of experience
+is more than one year.
 
 .. ipython:: python
     :okwarning:
 
-    salary_gt_30000_constraint = ScalarInequality(
-        columns='salary',
-        scalar=30000,
-        greater=True,
+    experience_years_gt_one_constraint = GreaterThan(
+        low=1,
+        high=['years_in_the_company', 'prior_years_experience'],
         handling_strategy='reject_sampling'
     )
 
+.. note::
+
+    To specify more than one column, either ``high`` or ``low`` must
+    be a scalar value, otherwise the constraint cannot be correctly
+    evaluated.
 
 Positive and Negative Constraints
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -214,8 +210,8 @@ column is always positive or negative. We can create an instance passing:
 
     from sdv.constraints import Positive
 
-    positive_prior_exp_and_age_constraint = Positive(
-        columns=['prior_years_experience', 'age'],
+    positive_age_constraint = Positive(
+        columns='age',
         strict=False,
         handling_strategy='reject_sampling'
     )
@@ -251,7 +247,8 @@ constraint by passing it:
     years_in_the_company_constraint = ColumnFormula(
         column='years_in_the_company',
         formula=years_in_the_company,
-        handling_strategy='transform'
+        handling_strategy='transform',
+        drop_column=False
     )
 
 Rounding Constraint
@@ -344,7 +341,8 @@ constraints that we just defined as a ``list``:
         age_gt_age_when_joined_constraint,
         years_in_the_company_constraint,
         salary_gt_30000_constraint,
-        positive_prior_exp_and_age_constraint,
+        experience_years_gt_one_constraint,
+        positive_age_constraint,
         salary_rounding_constraint,
         reasonable_age_constraint,
         one_hot_constraint
--- a/docs/user_guides/single_table/constraints.rst
+++ b/docs/user_guides/single_table/constraints.rst
@@ -203,6 +203,8 @@ or ``Negative`` constraints. These constraints enforce that specified
 column(s) are always positive or negative. We can create an instance passing:
 
 - the name of the column(s) for ``Negative`` or ``Positive`` constraints
+- a boolean specifying whether to make the data strictly above or below 0, 
+  or include 0 as a possible value
 - the handling strategy that we want to use
 
 .. ipython:: python
--- a/sdv/constraints/tabular.py
+++ b/sdv/constraints/tabular.py
@@ -224,7 +224,7 @@ class GreaterThan(Constraint):
     """
 
     _column_list = False
-    _diff_column = None
+    _diff_columns = None
     _is_datetime = None
     _column_to_reconstruct = None
 
@@ -241,6 +241,12 @@ class GreaterThan(Constraint):
         self._drop = drop
         self._high_is_scalar = high_is_scalar
         self._low_is_scalar = low_is_scalar
+
+        if strict:
+            self.operator = np.greater
+        else:
+            self.operator = np.greater_equal
+
         super().__init__(handling_strategy=handling_strategy,
                          fit_columns_model=fit_columns_model)
 
@@ -260,7 +266,7 @@ class GreaterThan(Constraint):
 
         return None
 
-    def _get_column_to_reconstruct(self):
+    def _get_columns_to_reconstruct(self):
         if self._drop == 'high':
             column = self._high
         elif self._drop == 'low':
@@ -272,7 +278,7 @@ class GreaterThan(Constraint):
 
         return column
 
-    def _get_diff_column_name(self, table_data):
+    def _get_diff_columns_name(self, table_data):
         token = '#'
         names = []
         for column in self.constraint_columns:
@@ -283,7 +289,6 @@ class GreaterThan(Constraint):
             names.append(name)
 
         if not self._high_is_scalar and not self._low_is_scalar:
-            assert len(self.constraint_columns) == 2
             while token.join(self.constraint_columns) in table_data.columns:
                 token += '#'
 
@@ -327,9 +332,9 @@ class GreaterThan(Constraint):
             if len(self.constraint_columns) > 2:
                 raise ValueError('`low` and `high` cannot be more than one column.')
 
-        self._column_to_reconstruct = self._get_column_to_reconstruct()
-        self._dtype = [table_data[column].dtype for column in self._column_to_reconstruct]
-        self._diff_column = self._get_diff_column_name(table_data)
+        self._columns_to_reconstruct = self._get_columns_to_reconstruct()
+        self._dtype = [table_data[column].dtype for column in self._columns_to_reconstruct]
+        self._diff_columns = self._get_diff_columns_name(table_data)
         self._is_datetime = self._get_is_datetime(table_data)
 
     def is_valid(self, table_data):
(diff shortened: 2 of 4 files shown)


sdv/constraints/tabular.py
```diff
@@ -195,11 +197,11 @@ class GreaterThan(Constraint):
     will be added to the diff to reconstruct the ``high`` column.
 
     Args:
-        low (str or int):
-            Either the name of the column that contains the low value,
+        low (str or list[str] or int):
```
csala, line 200: I think we should clean this up. We should either:
- Be consistent and mention `int` and `str` everywhere (so, make it `list[str or int]`, and see if there are other places where we should add `int` too)
- Drop `int` and mention `str` only (since `int` is just one of the many other types supported)
- Change everything to `object`, since anything that is hashable can be a column name.

I personally would go for the second option for reading simplicity, since we mention `str` only in many other palaces. Advanced users will know that other things also work anyway.

sdv/constraints/tabular.py
```diff
@@ -301,22 +312,23 @@ def _fit(self, table_data):
...
-            self._dtype = table_data[self._low].dtype
+            self.constraint_columns = tuple(self._low)
         else:
-            self._dtype = table_data[self._high].dtype
+            self.constraint_columns = tuple(self._low + self._high)
+            if len(self.constraint_columns) > 2:
+                raise ValueError('`low` and `high` cannot be more than one column.')
 
         self._column_to_reconstruct = self._get_column_to_reconstruct()
+        self._dtype = [table_data[column].dtype for column in self._column_to_reconstruct]
```
csala, line 331: Should we rename to `self._columns_to_reconstruct` and `self._get_columns_to_reconstruct`?

csala (changes requested):

docs/user_guides/single_table/constraints.rst
```diff
@@ -228,7 +249,8 @@ constraint by passing it:
     years_in_the_company_constraint = ColumnFormula(
         column='years_in_the_company',
         formula=years_in_the_company,
-        handling_strategy='transform'
+        handling_strategy='transform',
+        drop_column=False
```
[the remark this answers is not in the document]
sarahmish: Since `CustomFormula` drops the column, we encounter a `KeyError: 'years_in_the_company'` when running the `GreaterThan` constraint; specifically when we look up the `dtype` of the column in the fit stage
https://github.com/sdv-dev/SDV/blob/505eac9f8c9bc6ce583163b201f9f08400f1b600/sdv/constraints/tabular.py#L331

We can swap the order of the constraints such that we apply `GreaterThan` then `ColumnFormula` in order to evade this situation.

sdv/constraints/tabular.py
sarahmish: Sounds good, I believe that the same change should also apply to `diff_column` to `diff_columns`.

sdv/constraints/tabular.py
sarahmish: Wouldn't that be a bit awkward for the user to understand?
- option 1 is confusing in a sense that `list[int]` is problematic and would not work
- option 2 is the least confusing; but it also hides that `int` is supported
- option 3 is very general imo

I guess then make the change to `str or list[str]`? or would it just be `str`?

sdv/constraints/tabular.py
```diff
@@ -333,10 +345,11 @@ def is_valid(self, table_data):
         """
         low = self._get_low_value(table_data)
         high = self._get_high_value(table_data)
+
         if self._strict:
-            return high > low
+            return np.greater(high, low).all(axis=1)
```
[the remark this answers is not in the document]
sarahmish: yeah I agree

sdv/constraints/tabular.py
```diff
@@ -268,17 +274,22 @@ def _get_column_to_reconstruct(self):
...
             while name in table_data.columns:
                 name += '#'
 
-            return name
+            names.append(name)
 
-        while token.join(self.constraint_columns) in table_data.columns:
-            token += '#'
+        if not self._high_is_scalar and not self._low_is_scalar:
+            assert len(self.constraint_columns) == 2
```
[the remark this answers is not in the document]
sarahmish: I forgot to remove it 😅

docs/user_guides/single_table/constraints.rst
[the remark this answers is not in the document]
csala: Yes, let's swap the order then

sdv/constraints/tabular.py
csala: > * option 1 is confusing in a sense that `list[int]` is problematic and would not work

Why would it not work? Here the `int` is just a `column_name`, so if you have multiple column names that are integer values, you can pass a list of integers.

> * option 2 is the least confusing; but it also hides that `int` is supported

Right, but `int` is not the only type that is being silently supported, so I think it is fine. 

> I guess then make the change to `str or list[str]`? or would it just be `str`?

I would say `str or list[str]`.

sarahmish: > Why would it not work? Here the `int` is just a `column_name`, so if you have multiple column names that are integer values, you can pass a list of integers.

Right, I got what you mean now!

sdv/constraints/tabular.py
```diff
@@ -221,40 +223,50 @@ class GreaterThan(Constraint):
             by checking if the value provided is a column name.
     """
 
-    _diff_column = None
+    _column_list = False
```
katxiao, line 226: is `_column_list` used? If not we can remove this.

sdv/constraints/tabular.py
```diff
@@ -301,23 +317,24 @@ def _fit(self, table_data):
...
+            self.constraint_columns = tuple(self._high)
         elif self._high_is_scalar:
-            self.constraint_columns = (self._low,)
-            self._dtype = table_data[self._low].dtype
+            self.constraint_columns = tuple(self._low)
         else:
-            self._dtype = table_data[self._high].dtype
+            self.constraint_columns = tuple(self._low + self._high)
+            if len(self.constraint_columns) > 2:
+                raise ValueError('`low` and `high` cannot be more than one column.')
```
katxiao, line 333: If we no longer need to check if `self._high` and `self._low` are in `table_data.columns`, we can move these checks to the `__init__` method. That way we can also make sure `self.constraint_columns` is correctly set after `__init__` (right now it could be a nested tuple, which seems invalid).

katxiao: Also for the last error message, it would be more readable as '`low` and `high` cannot both be more than one column.`

sdv/constraints/tabular.py
```diff
@@ -388,92 +403,91 @@ def reverse_transform(self, table_data):
...
 
-        table_data = table_data.drop(self._diff_column, axis=1)
+        table_data = table_data.drop(self._diff_columns, axis=1)
 
         return table_data
 
 
 class Positive(GreaterThan):
-    """Ensure that the ``high`` column is always positive.
+    """Ensure that the given column is always positive.
```
katxiao, line 437: nit: "... the given column(s) are always positive"

sdv/constraints/tabular.py
```diff
@@ -388,92 +403,91 @@ def reverse_transform(self, table_data):
...
+            Whether the comparison of the values should be strict; disclude
+            zero ``>`` or include it ``>=``. Currently, this is only respected
             if ``reject_sampling`` or ``all`` handling strategies are used.
         handling_strategy (str):
             How this Constraint should be handled, which can be ``transform``
             or ``reject_sampling``. Defaults to ``transform``.
-        drop (str):
-            Which column to drop during transformation. Can be ``'high'``
-            or ``None``.
+        drop (bool):
```
katxiao, line 453: `drop` is a `bool` here but a `str` in the `GreaterThan` constraint. I see that below we're just passing the `drop` argument straight through to `super().__init__`. Do we need some kind of translation? e.g.
```
if drop:
  drop = 'high'
```

Same question for the `Negative` constraint below.

katxiao: We could add unit tests for `Positive` and `Negative` where `drop` is set to a non-None value.

katxiao: I also find it a little confusing that `drop` can be `bool` in some constraints, but `str` in others. We can discuss this as a larger topic though.

katxiao (changes requested):

sdv/constraints/tabular.py
csala: WRT the message, I would change it further to something like: `low and high can only be multiple columns if the other one is scalar`

csala: > If we no longer need to check if `self._high` and `self._low` are in `table_data.columns`, we can move these checks to the `__init__` method. That way we can also make sure `self.constraint_columns` is correctly set after `__init__` (right now it could be a nested tuple, which seems invalid).

Good catch @katxiao I think there is an additional problem here: It seems like the current implementation is assuming that if the value is not a list then it is a scalar, but that should not be the case: A scalar value could also be a column name, so I would rather say that the check this block of code needs to recover the `in table.columns` check somehow.

sdv/constraints/tabular.py
```diff
@@ -221,40 +223,50 @@ class GreaterThan(Constraint):
...
+    _column_list = False
+    _diff_columns = None
     _is_datetime = None
     _column_to_reconstruct = None
 
     def __init__(self, low, high, strict=False, handling_strategy='transform',
                  fit_columns_model=True, drop=None, high_is_scalar=None,
                  low_is_scalar=None):
+        self.constraint_columns = (low, high)
+        low = [low] if isinstance(low, str) else low
```
csala, line 235: Following up on the comment below: This does not look right.
We should take `high` and `low` as they come and not put them inside a `list` until we have already checked that they are actual column names.

sdv/constraints/tabular.py
```diff
@@ -221,40 +223,50 @@ class GreaterThan(Constraint):
...
-    _diff_column = None
+    _column_list = False
+    _diff_columns = None
     _is_datetime = None
     _column_to_reconstruct = None
 
     def __init__(self, low, high, strict=False, handling_strategy='transform',
                  fit_columns_model=True, drop=None, high_is_scalar=None,
                  low_is_scalar=None):
+        self.constraint_columns = (low, high)
```
csala, line 234: I think this needs to be revisited, since `low` and `high` may also be lists.
We should also add all the possible validations such as the inability to pass both `high` and `low` as `lists` here, in the `__init__`.

sdv/constraints/tabular.py
sarahmish: good catch!

sdv/constraints/tabular.py
sarahmish: I'm not sure, if the input is string, then they should be column names, and `_validate_constraint_columns` would handle the case if they did not exist.

sdv/constraints/tabular.py
```diff
@@ -211,74 +214,96 @@ class GreaterThan(Constraint):
...
     _is_datetime = None
-    _column_to_reconstruct = None
+    _columns_to_reconstruct = None
 
     def __init__(self, low, high, strict=False, handling_strategy='transform',
-                 fit_columns_model=True, drop=None, high_is_scalar=None,
-                 low_is_scalar=None):
+                 fit_columns_model=True, drop=None, scalar=None):
+
+        if isinstance(low, list) and isinstance(high, list):
```
csala, line 229: I think this validations block is still too complex.

I would suggest these changes:

1. Move all the validations to a dedicated classmethod method.

    ```python3
    self.low, self.high, self.constraint_columns = self._validate_inputs(low, high, scalar)
    ```

    Another option is to make it an instance method that updates the arguments internally:

    ```python3
    self._validate_inputs(low, high, scalar)   # this sets `self.low` and the others internally
    ```

2. I would validate the `scalar=None` explicitly, and raise an exception if `scalar` is none of the expected values.

3. I would implement the asymmetric validations as another, generic static method.

    This method would expect a scalar value and column names, but without knowing which
    one is high and which one is low, and just validate their types, and would be called with
    `high` and `low` put in the right place depending on the direction. The `scalar` value can
    still be passed as a string to write the right error messages.

Altogether it would be something like:

```python3
@staticmethod
def _validate_scalar(scalar_column, column_names, scalar):
    """Validate scalar comparison inputs.

    - Make sure that the scalar column is not a list and raise the proper error if it is.
    - If the `column_names` is not a list it would make it a list.
    - Return both the scalar column and column names with the right format
    """
    ...

@classmethod
def _validate_inputs(cls, low, high, scalar):
    if scalar is None:
        # validate that neither high nor low are lists and 
    if scalar == 'low':
        low, high = cls._validate_scalar(scalar_column=low, column_names=high, scalar=scalar)
        constraint_columns = tuple(high)
    elif scalar == 'high':
        high, low = cls._validate_scalar(scalar_column=high, column_names=low, scalar=scalar)
        constraint_columns = tuple(low)
    else:
        raise ValueError(...)

    return low, high, constraint_columns

def __init__(self, ...):
    self.low, self.high, self.constraint_columns = self._validate_inputs(low, high, scalar)
```

sdv/constraints/tabular.py
```diff
@@ -300,24 +325,30 @@ def _fit(self, table_data):
...
-        if self._high_is_scalar and self._low_is_scalar:
-            raise TypeError('`low` and `high` cannot be both scalars at the same time')
-        elif self._low_is_scalar:
-            self.constraint_columns = (self._high,)
-            self._dtype = table_data[self._high].dtype
-        elif self._high_is_scalar:
-            self.constraint_columns = (self._low,)
-            self._dtype = table_data[self._low].dtype
+        if self._scalar == 'high':
+            exists = [column in table_data.columns for column in self._low]
```
csala, line 329: This block and the one below could be moved to a method which is called one way or another depending on `_scalar`

sdv/constraints/tabular.py
```diff
@@ -300,24 +325,30 @@ def _fit(self, table_data):
...
+            exists = [column in table_data.columns for column in self._high]
+            if not all(exists):
+                raise KeyError(f'The `high` columns {list(compress(self._high, exists))} '
+                               'were not found in table_data.')
         else:
-            self._dtype = table_data[self._high].dtype
-
-        self._column_to_reconstruct = self._get_column_to_reconstruct()
-        self._diff_column = self._get_diff_column_name(table_data)
+            for column in self._low:
```
csala, line 339: This block and the one below could be moved to a method that is called twice, passing `_low` and `_high` (and maybe the argument name to add to the error message)

sdv/constraints/tabular.py
```diff
@@ -300,24 +325,30 @@ def _fit(self, table_data):
...
+                if column not in table_data.columns:
+                    raise KeyError(f'The column {column} specified in `low` was not found in '
+                                   f'table_data. If {column} is a scalar, set `scalar="low"`.')
+
+            for column in self._high:
+                if column not in table_data.columns:
+                    raise KeyError(f'The column {column} specified in `high` was not found in '
+                                   f'table_data. If {column} is a scalar, set `scalar="high"`.')
+
+        self._columns_to_reconstruct = self._get_columns_to_reconstruct()
```
csala, line 349: This could be moved to the `__init__` method, since it does not use the data at all.

sdv/constraints/tabular.py
```diff
@@ -211,74 +214,96 @@ class GreaterThan(Constraint):
...
         super().__init__(handling_strategy=handling_strategy,
                          fit_columns_model=fit_columns_model)
 
     def _get_low_value(self, table_data):
-        if self._low_is_scalar:
+        if self._scalar == 'low':
             return self._low
-        elif self._low in table_data.columns:
-            return table_data[self._low]
+        elif any([low in table_data.columns for low in self._low]):
```
csala, line 265: Is this check needed? I suspect that at this point, if `scalar` is not `low`, it means that we already validated that all the `low` names are in `table_data`, so this could probably just be:

```python3
if self._scalar == 'low':
    return self._low

return table_data[self._low].values
```

sdv/constraints/tabular.py
```diff
@@ -211,74 +214,96 @@ class GreaterThan(Constraint):
...
+            return table_data[self._low].values
 
         return None
 
     def _get_high_value(self, table_data):
-        if self._high_is_scalar:
+        if self._scalar == 'high':
```
csala, line 271: Same as in other places: Maybe it would be simpler to make a single, generic method instead of two?

```python3
def _get_value(self, table_data, field):
    if self._scalar == field:
        return getattr(self, f'_{field}')

    return table_data[getattr(self, f'_{field}')].values

...
low = self._get_value(table_data, 'low')
high = self._get_value(table_data, 'high')
```

csala (changes requested):
I added several comments about things that could be refactored.
I suspect that we are still dragging too much inheritance from the previous implementation, and the new interface allows us to push implementation changes a bit further

sdv/constraints/tabular.py
katxiao: I like the idea of moving the validation to a class method `_validate_inputs`, and I think the class method should return the validated inputs (the first option), as the name suggests. If we want to update the arguments internally in this method, we should make the name something like `_validate_inputs_and_initialize`, but I think the first way is more explicit.

sdv/constraints/tabular.py
katxiao: +1, we can move the duplicated logic in L329-L332 to a class method `_check_columns_exist` and call it with either `self._low` or `self._high`.

sdv/constraints/tabular.py
```diff
@@ -388,93 +417,102 @@ def reverse_transform(self, table_data):
...
-            or ``None``.
+        drop (bool):
+            Whether to drop columns during transformation.
     """
 
-    def __init__(self, high, strict=False, handling_strategy='transform',
-                 fit_columns_model=True, drop=None):
+    def __init__(self, columns, strict=False, handling_strategy='transform',
+                 fit_columns_model=True, drop=False):
+        if drop:
```
katxiao, line 473: nit: writing it inline would be more concise `drop = 'high' if drop else None`

tests/unit/constraints/test_tabular.py
```diff
@@ -885,7 +1053,57 @@ def test_fit_type__low_is_scalar(self):
...
+        """
+        # Setup
+        instance = GreaterThan(low=['a', 'b'], high=0, scalar='high')
+        dtype_int = pd.Series([1]).dtype
+        dtype_float = np.dtype('float')
+        table_data = pd.DataFrame({
+            'a': [1, 2, 4],
+            'b': [4., 5., 6.]
+        })
+        instance.fit(table_data)
```
katxiao, line 1075: we can call `_fit` here so we're not testing the parent method

tests/unit/constraints/test_tabular.py
```diff
@@ -1272,6 +1542,74 @@ def test_transform_low_is_scalar(self):
...
+        instance._diff_columns = ['a#', 'b#']
+        instance.constraint_columns = ['a', 'b']
+
+        # Run
+        table_data = pd.DataFrame({
+            'a': [1, 2, 3],
+            'b': [4, 5, 6],
+            'c': [7, 8, 9],
+        })
+        out = instance.transform(table_data)
```
katxiao, line 1567: We can call `_transform` here so we're not testing the parent method

tests/unit/constraints/test_tabular.py
```diff
@@ -1654,33 +2162,58 @@ def test__init__(self):
...
+        - drop = True
+        Side effects:
+        - instance._low == 'a'
+        - instance._high == 0
+        - instance._strict == True
+        - instance._scalar == 'low'
+        - instance._drop = 'high'
+        """
+        # Run
+        instance = Positive(columns='a', strict=True, drop=True)
```
katxiao, line 2208: nice!

tests/unit/constraints/test_tabular.py
```diff
@@ -431,44 +431,215 @@ def test___init___strict_false(self):
...
+
+        Make sure to raise an error when both ``high`` and ``low`` are lists.
+
+        Input:
+        - low = ['a']
+        - high = ['b']
+        """
+        # Run / Assert
+        with pytest.raises(ValueError):
+            GreaterThan(low=['a'], high=['b'])
```
katxiao, line 512: Hm. So technically this should be valid because we're only passing in one column for high and one column for low right? 

In our validation we could check 
```
if isinstance(low, list):
    if len(low) > 1:
        raise ValueError('`low` cannot be multiple columns when `scalar` is `None`.')
    low = low[0]
elif isinstance(high, list):
    if len(high) > 1:
        raise ValueError('`high` cannot be a list when `scalar` is `None`.')
    high = high[0]
```

docs/user_guides/single_table/constraints.rst
```diff
@@ -161,38 +161,60 @@ The ``GreaterThan`` constraint can also be used to guarantee a column is greater
 than a scalar value or specific datetime value instead of another column. To use
```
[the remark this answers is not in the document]
sarahmish: I think you mean the opposite: pass the scalar value as the ``high`` argument, the column name as the ``low`` argument, and set the ``scalar`` flag to ``"high"``

docs/user_guides/single_table/constraints.rst
```diff
@@ -161,38 +161,60 @@ The ``GreaterThan`` constraint can also be used to guarantee a column is greater
...
     :okwarning:
 
     salary_gt_30000_constraint = GreaterThan(
         low=30000,
         high='salary',
+        scalar='low',
         handling_strategy='reject_sampling'
     )
 
+Optionally, when constructing ``GreaterThan`` constraint we can specify
```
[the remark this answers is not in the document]
sarahmish: I think we opted to retract this statement, since `GreaterThan(low='a', high=['b', 'c'])` is a valid comparison.

docs/user_guides/single_table/constraints.rst
```diff
@@ -161,38 +161,60 @@ The ``GreaterThan`` constraint can also be used to guarantee a column is greater
...
+    :okwarning:
+
+    experience_years_gt_one_constraint = GreaterThan(
+        low=1,
+        high=['years_in_the_company', 'prior_years_experience'],
+        scalar='low',
+        handling_strategy='reject_sampling'
+    )
+
+.. note::
```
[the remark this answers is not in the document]
sarahmish: Minor adjustment 

```
Warning! Passing a list of columns to the ``high`` or ``low`` arguments is only possible when
the other one has been passed as a scalar or one of them is a single column! In other 
words, passing a more that one column in ``high`` AND more than one column in ``low``
is not possible.
```

tests/unit/constraints/test_tabular.py
sarahmish: I think this is now outdated in a sense that we allow `GreaterThan(low=['a'], high=['b'])`. but in general, internally we are dealing with slices of dataframes so all `low` and `high` values need to be stored in a list (except if they are scalar).

docs/user_guides/single_table/constraints.rst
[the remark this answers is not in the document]
csala: Yeah, it needs to be adjusted wrt the possibility of passing a single column. I would try to phrase it using positive statements rather than negative, though. Here's another proposal, adding the single column to the previous one:

```
Warning! Passing a list of columns to the `high` or `low` arguments is only possible when
the other one has been passed as a single column name or scalar value! If you need to
compare multiple ` high`` columns against multiple ``low`` columns (or vice versa), you
need to decompose one of the ends, ``high`` or ``low``, into multiple single column
names and define one ``GreaterThan`` constraint for each one of them.
```

sdv/constraints/tabular.py
```diff
@@ -211,78 +214,109 @@ class GreaterThan(Constraint):
...
-        if self._high_is_scalar:
-            return self._high
-        elif self._high in table_data.columns:
-            return table_data[self._high]
+        - Make sure that the scalar column is not a list and raise the proper error if it is.
+        - If the `column_names` is not a list it would make it a list.
+        - Return both the scalar column and column names with the right format
+        """
+        if isinstance(scalar_column, list):
+            raise TypeError(f'{scalar} cannot be a list when scalar="{scalar}".')
```
csala, line 235: I would add ticks around `{scalar}`: ```TypeError(f'`{scalar}` cannot...```

sdv/constraints/tabular.py
```diff
@@ -211,78 +214,109 @@ class GreaterThan(Constraint):
...
+
+        column_names = [column_names] if not isinstance(column_names, list) else column_names
```
csala, line 237: I would rather do this in a 2 lines if. Otherwise we are just reassigning unnecessarily:

```
if not isinstance(column_names, list):
     column_names = [column_names]
```

sdv/constraints/tabular.py
```diff
@@ -211,78 +214,109 @@ class GreaterThan(Constraint):
...
+        return column_names
+
+    @classmethod
+    def _validate_inputs(cls, low, high, scalar):
+        if scalar is None:
+            low = [low] if not isinstance(low, list) else low
```
csala, line 243: Same as above. We could actually make this a function: `as_list(value)`

sdv/constraints/tabular.py
```diff
@@ -211,78 +214,109 @@ class GreaterThan(Constraint):
...
+            constraint_columns = tuple(low + high)
+
+        elif scalar == 'low':
+            high = cls._validate_scalar(scalar_column=low, column_names=high, scalar=scalar)
+            constraint_columns = tuple(high)
+        elif scalar == 'high':
+            low = cls._validate_scalar(scalar_column=high, column_names=low, scalar=scalar)
+            constraint_columns = tuple(low)
+        else:
+            raise ValueError(f"scalar {scalar} is unknown. Use either: 'high', 'low', or None.")
```
csala, line 257: ```Invalid `scalar` value: `{scalar}`. Use...```

sdv/constraints/tabular.py
```diff
@@ -211,78 +214,109 @@ class GreaterThan(Constraint):
...
-        if len(self.constraint_columns) == 1:
-            name = self.constraint_columns[0] + token
-            while name in table_data.columns:
-                name += '#'
+    def __init__(self, low, high, strict=False, handling_strategy='transform',
+                 fit_columns_model=True, drop=None, scalar=None):
+        self._strict = strict
+        self._drop = drop
+        self._scalar = scalar
+        self._low, self._high, self.constraint_columns = self._validate_inputs(low, high, scalar)
```
csala, line 278: I just realized about one thing: We should be validating `drop` too, since `scalar == drop` is no possible.

sdv/constraints/tabular.py
```diff
@@ -211,78 +214,109 @@ class GreaterThan(Constraint):
...
+                         fit_columns_model=fit_columns_model)
+
+    def _get_value(self, table_data, field):
+        variable = getattr(self, f'_{field}')
+        if self._scalar == field:
+            return variable
+        elif all([column in table_data.columns for column in variable]):
+            return table_data[getattr(self, f'_{field}')].values
+
+        return None
```
csala, line 296: I think this return None is not necessary, and neither is the `elif` above.

This can be:
```
if self._scalar == field:
    return variable

return table_data[variable].values
```

sdv/constraints/tabular.py
```diff
@@ -211,78 +214,109 @@ class GreaterThan(Constraint):
...
-            token += '#'
+        super().__init__(handling_strategy=handling_strategy,
+                         fit_columns_model=fit_columns_model)
+
+    def _get_value(self, table_data, field):
+        variable = getattr(self, f'_{field}')
+        if self._scalar == field:
+            return variable
+        elif all([column in table_data.columns for column in variable]):
+            return table_data[getattr(self, f'_{field}')].values
```
csala, line 294: `variable`

sdv/constraints/tabular.py
```diff
@@ -293,31 +327,31 @@ def _get_is_datetime(self, table_data):
 
         return is_datetime
 
+    def _check_columns_exist(self, table_data, field):
+        values = getattr(self, f'_{field}')
+        exists = [column not in table_data.columns for column in values]
```
csala, line 332: I would rename `exists` to `missing` (otherwise the `if any(exists)` is very confusing)
This can also be rewritten using sets:

```python3
if set(values) - set(table_data.columns):
    raise ...
```

sdv/constraints/tabular.py
```diff
@@ -293,31 +327,31 @@ def _get_is_datetime(self, table_data):
...
-            self._dtype = table_data[self._high].dtype
-        elif self._high_is_scalar:
-            self.constraint_columns = (self._low,)
-            self._dtype = table_data[self._low].dtype
-        else:
-            self._dtype = table_data[self._high].dtype
-
-        self._column_to_reconstruct = self._get_column_to_reconstruct()
-        self._diff_column = self._get_diff_column_name(table_data)
+        if self._scalar == 'high':
```
csala, line 345: this can be made simpler by negating:

```python3
if self._scalar != 'high':    # <- So high must be columns...
    self._check_columns_exists(table_data, 'high')
if self._scalar != 'low':
    self._check_columns_exists(table_data, 'low')
```

sdv/constraints/tabular.py
```diff
@@ -293,31 +327,31 @@ def _get_is_datetime(self, table_data):
...
+            self._check_columns_exist(table_data, 'low')
+        elif self._scalar == 'low':
+            self._check_columns_exist(table_data, 'high')
+        elif self._scalar is None:
+            self._check_columns_exist(table_data, 'low')
+            self._check_columns_exist(table_data, 'high')
+
+        self._dtype = [table_data[column].dtype for column in self._columns_to_reconstruct]
```
csala, line 353: `self._dtype = table_data[self._columns_to_reconstruct].dtypes`

csala (changes requested):
One final round of tweaks and improvements.

sdv/constraints/tabular.py
sarahmish: Yes, I believe `low = low` and `high = high` is not possible; but `None` and `None` works.. I will add one more check to `_validate_inputs`.

sdv/constraints/tabular.py
sarahmish: In reverse transform L427-428 we try to get the value of high and low:
```
high = self._get_value(table_data, 'high')
low = self._get_value(table_data, 'low')
```

but because of `drop` there is no guarantee these columns exist, raising a `KeyError`.

csala: Ok, there's the problem :-)
Adding a comment on L427

sdv/constraints/tabular.py
```diff
@@ -388,93 +420,94 @@ def reverse_transform(self, table_data):
...
         table_data = table_data.copy()
-        diff = (np.exp(table_data[self._diff_column]).round() - 1).clip(0)
+        diff = (np.exp(table_data[self._diff_columns].values).round() - 1).clip(0)
         if self._is_datetime:
-            diff = pd.to_timedelta(diff)
+            diff = diff.astype('timedelta64[ns]')
 
-        high = self._get_high_value(table_data)
-        low = self._get_low_value(table_data)
+        high = self._get_value(table_data, 'high')
```
csala, line 427: It seems we are always trying to get `high` and `low`, even though we will always need only one of them, and they will need to be always accessed only once.
What about moving these further down, into the `if/else` block?

```python3
if self._drop == 'high':
    low = self._get_value(table_data, 'low')
    new_values = pd.DataFrame(diff + low, columns=self._high)
    table_data[self._high] = new_values.astype(dict(zip(self._high, self._dtype)))
...
```

sarahmish: I thought about it, I feel like it is a bit redundant, I also though about:
```python3
if self._drop != 'high':
    high = self._get_value(table_data, 'high')
if self._drop != 'low':
    low = self._get_value(table_data, 'low')
```

but that also feels redundant because of the succeeding `if` statements. I'm leaning towards your suggestion or keeping it as is with `None` returns.

katxiao (approved):
lgtm after the last round of changes!

csala (approved):
I think this is finally ready to go! :rocket: 
Thanks for the patience @sarahmish !

PR #538, opened 2021-07-30
Removed unused label encoder.

### LINKED ISSUE #494: UniqueCombinations constraint for more than 2 columns
### Environment Details

* SDV version: 0.10
* Python version: 3.9
* Operating System: Windows

### Error Description

I was under the impression that adding a list of columns (more than 2) would constrain the model in respect of the combinations of categories represented by those columns but it does not seem to work. I see nulls generated where there should be none

### Steps to reproduce

I can’t share my example due to privacy etc. but imagine that a toy example where a list of three categories passed to uniquecombinations would probably generate the same result.

### LINKED ISSUE #510: `UniqueCombinations` constraint outputting wrong data type
The `UniqueCombinations` constraint can sometimes produce the wrong output data type. For example, if the passed input only contains integers, the output is expected to be made of only integers as well. Instead, sometimes it can produce floats. Below is some code that runs this example.

```python3
import pandas as pd
from sdv.constraints import *
from sdv.tabular import *

data = pd.DataFrame({
    'a': zero_to_nine,
    'b': zero_to_nine,
    'c': zero_to_nine
})

    columns=['a', 'b'],
)

model.fit(data)
samples = model.sample(10, conditions={'a': 1})
samples
```

The code above returns the following: 
```
	a	b	c
0	1	1.0	5
1	1	1.0	6
2	1	1.0	5
3	1	1.0	2
4	1	1.0	3
5	1	1.0	3
6	1	1.0	6
7	1	1.0	4
8	1	1.0	2
9	1	1.0	4
```

### LINKED ISSUE #521: UniqueCombinations constraint with handling_strategy = 'transform' yields synthetic data with nan values
### Environment Details

SDV version: 0.11.0
Python version: 3.6.8
Operating System: Windows 10

### Error Description

When applying a `UniqueCombinations` constraint with `handling_strategy = 'transform'`, the CTGAN model produces some synthetic observations that exhibit `nan` values for those columns specified within the constraint (_despite training the `CTGAN` model on **nan-free data**_).

This screen capture illustrates the bug:

image

### Steps to reproduce

The full Jupyter Notebook that reproduces this bug is attached (the notebook pulls it's own data from the web): 
[generate_adversarial_data-ctgan.zip](https://github.com/sdv-dev/SDV/files/6856991/generate_adversarial_data-ctgan.zip)

katxiao:
Resolve #510
Resolve #521 
Resolve #494

### DIFF
--- a/sdv/tabular/copulagan.py
+++ b/sdv/tabular/copulagan.py
@@ -211,7 +211,7 @@ class CopulaGAN(CTGAN):
                 distribution=distributions.get(field, default)
             )
             for field in table_data.columns
-            if fields.get(field, dict()).get('type') != 'categorical'
+            if field in fields and fields.get(field, dict()).get('type') != 'categorical'
         }
         self._ht = HyperTransformer(transformers=transformers)
         table_data = self._ht.fit_transform(table_data)
--- a/sdv/tabular/ctgan.py
+++ b/sdv/tabular/ctgan.py
@@ -16,7 +16,7 @@ class CTGANModel(BaseTabularModel):
     _model_kwargs = None
 
     _DTYPE_TRANSFORMERS = {
-        'O': 'label_encoding'
+        'O': None
     }
 
     def _build_model(self):
--- a/tests/integration/tabular/test_copulagan.py
+++ b/tests/integration/tabular/test_copulagan.py
@@ -36,11 +36,11 @@ def test_copulagan():
             },
             'country': {
                 'type': 'categorical',
-                'transformer': 'label_encoding',
+                'transformer': None,
             },
             'gender': {
                 'type': 'categorical',
-                'transformer': 'label_encoding',
+                'transformer': None,
             },
             'age': {
                 'type': 'numerical',
--- a/tests/integration/tabular/test_ctgan.py
+++ b/tests/integration/tabular/test_ctgan.py
@@ -32,11 +32,11 @@ def test_ctgan():
             },
             'country': {
                 'type': 'categorical',
-                'transformer': 'label_encoding',
+                'transformer': None,
             },
             'gender': {
                 'type': 'categorical',
-                'transformer': 'label_encoding',
+                'transformer': None,
             },
             'age': {
                 'type': 'numerical',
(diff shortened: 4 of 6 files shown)


csala (approved):
LGTM!

csala:
Good job catching this @katxiao 

Adding here some more verbose description about what is going on, for the record.

Apparently issue #494, #510 and #521 are a combination of several things:

1. The `UniqueCombinations` constraint replaces the affected columns with a _new_ column that contains `str` values.
2. `ctgan` based models use `label_encoding` internally to transform categorical values into `int` labels. This was done in the past because CTGAN could not process strings in its own, but it is not necessary anymore after PR https://github.com/sdv-dev/CTGAN/pull/77
3. When the `ctgan` models are fitted SDV passes the list of categorical (or discrete) variables, which include all the columns that were originally labelled as `categorical` in the original data, as well as all the new columns generated by the constraints which are of type `str`.
4. Because of point 2. the strings generated on point 1. are not caught by process described in point 3. resulting in the `ctgan` models not knowing that the observed integer values are actually categories. As a result, `ctgan` processes them as numerical variables of type `int`.
5. Because of point 4. when `ctgan` samples it may generate some `int` values out of range, which the `LabelEncodingTransformer` used on point 2. will fail to `reverse_transform`, producing `np.nan` values.
6. When reverse transforming, the `UniqueCombionations` constraint simply forwards the `np.nan` for the rows where `nan` was the input, producing `nan` values for all the constrainted columns.

Once this point is reached, 2 can happen:
a. If no `conditions` had been given, the `np.nan` values are kept in the output data, as reported in issues #494 and #521 
b. If any condition had been given for the constrained columns, the `reject_sampling` loop will discard the rows that contain `nan` values, so they will not be seen. However, if the input columns were of type `int`, the existence of `np.nan` values at some point during the process will have converted them to `float`, as reported in issue #510 

Now, by simply removing the (now unneeded) `label_encoding` transformer and passing the strings down to CTGAN we have 2 advantages:
1. We can now skip an unnecessary `transform/reverse_transform` step
2. The string columns like the one generated by the `UniqueCombinations` are properly labeled as `categorical` when `ctgan` is called, so the output will only contain valid values that will be able to be reverse transformed but the constraint.

PR #539, opened 2021-08-02
Validate UniqueCombinations columns input

### LINKED ISSUE #509: `UniqueCombinations` constraint on only one column gets stuck in an infinite loop
Using the `UniqueCombinations` constraint on only one column gets stuck in an infinite loop. This can be fixed by validating the constraint inputs, since the reason for the crash is that `UniqueCombinations` requires at least two column as input.

Code to reproduce the error:

```python3
import pandas as pd
from sdv.constraints import *
from sdv.tabular import *

data = pd.DataFrame({
    'a': zero_to_nine,
    'b': zero_to_nine,
    'c': zero_to_nine
})

    columns=['a'],
)

model.fit(data)
```

katxiao:
Validate the `UniqueCombinations` `columns` input to make sure `columns` contains at least 2 values, since we expect this constraint to be over multiple columns.

Resolves #509

### DIFF
--- a/sdv/constraints/tabular.py
+++ b/sdv/constraints/tabular.py
@@ -73,7 +73,8 @@ class UniqueCombinations(Constraint):
 
     Args:
         columns (list[str]):
-            Names of the columns that need to produce unique combinations.
+            Names of the columns that need to produce unique combinations. Must
+            contain at least two columns.
         handling_strategy (str):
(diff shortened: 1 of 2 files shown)


csala (approved):
LGTM!

amontanez24 (approved):
LGTM!

PR #540, opened 2021-08-02
Create Unique Constraint

### LINKED ISSUE #532: Create Unique Constraint
### Problem Description

Sometimes, even though a column isn't a key, the values in it need to be unique in order for the data to be valid. Consider the following 2 use cases:
1. There is a sensitive field in my data (such as SSN) that cannot be repeated. Even though I have another primary key (`user_id`), I need to makes sure that `SSN` is also unique
2. I have a table that lists all the shoe types I sell. I have `models` (eg. Nike, Reebok, etc.) and `sizes` (6, 6.5, 7, etc.). I to make sure that each size is unique within a given model. Eg. I can't have 2 rows that are size 7.5 Reebok

In order to support this, write a `Unique` constraint where the user can supply both a `column` (whose values must be unique) and an optional list of `group_by` columns that control the partition that determine uniqueness. In our cases

1. `column='SSN'` with `group_by=None` -- which means it's unique throughout the table
2. `column='size'` while `group_by=['model']` -- which means that within a given model, size is unique

amontanez24:
resolve #532

This PR adds a new constraint class called `Unique`. The class takes in one parameter for its `__init__`: `columns` which is either a string defining the one column name that needs to be unique or a list of column names.

The `Unique` constraint only works with the `reject_sampling` `handling_strategy`. For this reason it only implements a `is_valid` method. 

The `is_valid` method returns a `pd.Series` where the index corresponding to the first occurrence of a unique value for the specified `columns` is set to True, and the rest are set to False.

### DIFF
--- a/sdv/constraints/tabular.py
+++ b/sdv/constraints/tabular.py
@@ -956,3 +956,40 @@ class OneHotEncoding(Constraint):
         table_data[self._columns] = transformed_data
 
         return table_data
+
+
+class Unique(Constraint):
+    """Ensure that each value for a specified column/group of columns is unique.
+
+    This constraint is provided a list of columns, and guarantees that every
+    unique combination of those columns appears at most once in the sampled
+    data.
+
+    Args:
+        columns (str or list[str]):
+            Name of the column(s) to keep unique.
+    """
+
+    def __init__(self, columns):
+        self.columns = [columns] if isinstance(columns, str) else columns 
+        self.handling_strategy = 'reject_sampling'
+
+    def is_valid(self, table_data):
+        """Get indices of first instance of unique rows.
+
+        If a row is the first instance of that combination of column
+        values, it is valid. Otherwise it is false.
+
+        Args:
+            table_data (pandas.DataFrame):
+                Table data.
+
+        Returns:
+            pandas.Series:
+                Whether each row is valid.
+        """
+        data = table_data.reset_index()
+        groups = data.groupby(self.columns)
+        valid = pd.Series([False]*data.shape[0])
+        valid.iloc[groups.first()['index'].values] = True
+        return valid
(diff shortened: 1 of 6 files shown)


sdv/constraints/tabular.py
```diff
@@ -956,3 +956,40 @@ def reverse_transform(self, table_data):
...
+    unique combination of those columns appears at most once in the sampled
+    data.
+
+    Args:
+        columns (str or list[str]):
+            Name of the column(s) to keep unique.
+    """
+
+    def __init__(self, columns):
+        self.columns = [columns] if isinstance(columns, str) else columns
```
csala, line 974: I would do it the other way around: If `columns` is not a `list`, make it a list of a single element.
Otherwise different column name types, such as `int`s will not work.

tests/unit/tabular/test_base.py
```diff
@@ -82,6 +82,51 @@ def test_conditional_sampling_graceful_reject_sampling(model):
         model.sample(5, conditions=conditions, graceful_reject_sampling=False)
 
 
+def test__sample_rows_previous_rows_appended_correctly():
+    """Test the ``BaseTabularModel._sample_rows`` method.
+
+    If ``_sample_rows`` is passed ``previous_rows``, then it
+    shuld reset the index when appending them to the new
```
csala, line 89: should

sdv/tabular/base.py
```diff
@@ -234,7 +234,7 @@ def _sample_rows(self, num_rows, conditions=None, transformed_conditions=None,
             sampled = self._metadata.reverse_transform(sampled)
 
             if previous_rows is not None:
-                sampled = previous_rows.append(sampled)
+                sampled = previous_rows.append(sampled, ignore_index=True)
```
csala, line 237: good one

csala (changes requested):

csala (approved):
I think this is good to go!

sarahmish (approved):
LGTM

PR #542, opened 2021-08-03
Allow conditioning on constraint column in `ColumnFormula`

### LINKED ISSUE #506: Conditioning on the constraint column of the `ColumnFormula` constraint
Conditioning on the constraint column when using the `ColumnFormula` constraint causes the following error:

```
/var/folders/3p/f4b70gn5165464cqyvg157qw0000gn/T/ipykernel_42215/2787579389.py in <module>
----> 1 samples = model.sample(10, conditions={'b': 1})
      2 samples

~/Desktop/SDV/sdv/tabular/base.py in sample(self, num_rows, max_retries, max_rows_multiplier, conditions, float_rtol, graceful_reject_sampling)
    471             else:
    472                 transformed_conditions_in_group = transformed_conditions.loc[condition_indices]
--> 473                 transformed_groups = transformed_conditions_in_group.groupby(transformed_columns)
    474                 for transformed_group, transformed_dataframe in transformed_groups:
    475                     if not isinstance(transformed_group, tuple):

~/opt/anaconda3/envs/test/lib/python3.8/site-packages/pandas/core/frame.py in groupby(self, by, axis, level, as_index, sort, group_keys, squeeze, observed, dropna)
   6513         axis = self._get_axis_number(axis)
   6514 
-> 6515         return DataFrameGroupBy(
   6516             obj=self,
   6517             keys=by,

~/opt/anaconda3/envs/test/lib/python3.8/site-packages/pandas/core/groupby/groupby.py in __init__(self, obj, keys, axis, level, grouper, exclusions, selection, as_index, sort, group_keys, squeeze, observed, mutated, dropna)
    523             from pandas.core.groupby.grouper import get_grouper
    524 
--> 525             grouper, exclusions, obj = get_grouper(
    526                 obj,
    527                 keys,

~/opt/anaconda3/envs/test/lib/python3.8/site-packages/pandas/core/groupby/grouper.py in get_grouper(obj, key, axis, level, sort, observed, mutated, validate, dropna)
    819 
    820     if len(groupings) == 0 and len(obj):
--> 821         raise ValueError("No group keys passed!")
    822     elif len(groupings) == 0:
    823         groupings.append(Grouping(Index([], dtype="int"), np.array([], dtype=np.intp)))

ValueError: No group keys passed!
```

Code to reproduce it:

```python3
import pandas as pd
from sdv.constraints import *
from sdv.tabular import *

ones = [1 for i in range(10)]
data = pd.DataFrame({
    'a': ones,
    'b': ones,
    'c': ones
})

def formula(data):
    return data['a']

constraint = ColumnFormula(
    column='b',
    formula=formula
)

model.fit(data)
samples = model.sample(10, conditions={'b': 1})
samples
```

katxiao:
If the transformed conditions doesn't have any columns, don't pass through the conditions when sampling.

Resolve #506

### DIFF
--- a/sdv/tabular/base.py
+++ b/sdv/tabular/base.py
@@ -457,7 +457,7 @@ class BaseTabularModel:
 
             condition_indices = dataframe[COND_IDX]
             condition = dict(zip(condition_columns, group))
-            if transformed_conditions.empty:
+            if len(transformed_columns) == 0:
                 sampled_rows = self._conditionally_sample_rows(
                     dataframe,
                     max_retries,
--- a/tests/unit/tabular/test_base.py
+++ b/tests/unit/tabular/test_base.py
@@ -16,6 +16,68 @@ MODELS = [
 ]
 
 
+class DataFrameMatcher:
+    def __init__(self, df):
+        self.df = df
+
+    def __eq__(self, other):
+        return self.df.equals(other)
+
+
+class TestBaseTabularModel:
+
+    @patch("sdv.tabular.base.COND_IDX", 'test-cond-idx')
+    def test_sample_no_transformed_columns(self):
+        """Test the ``BaseTabularModel.sample`` method with no transformed columns.
+
+        When the transformed conditions DataFrame has no columns, expect that sample
+        does not pass through any conditions when conditionally sampling.
+
+        Setup:
+            - Mock the ``_make_conditions_df`` method to return a dataframe representing
+              the expected conditions, and the ``get_fields`` method to return metadata
+              fields containing the expected conditioned column.
+            - Mock the ``_metadata.transform`` method to return an empty transformed
+              conditions dataframe.
+            - Mock the ``_conditionally_sample_rows`` method to return the expected
+              sampled rows.
+        Input:
+            - number of rows
+            - one set of conditions
+        Output:
+            - the expected sampled rows
+        Side Effects:
+            - Expect ``_conditionally_sample_rows`` to be called with the given condition
+              and a transformed_condition of None.
+        """
+        # Setup
+        gaussian_copula = Mock(spec_set=GaussianCopula)
+        expected = pd.DataFrame(['a', 'a', 'a'])
+
+        gaussian_copula._make_conditions_df.return_value = pd.DataFrame({'a': ['a', 'a', 'a']})
+        gaussian_copula._metadata.get_fields.return_value = ['a']
+        gaussian_copula._metadata.transform.return_value = pd.DataFrame({}, index=[0, 1, 2])
+        gaussian_copula._conditionally_sample_rows.return_value = pd.DataFrame({
+            'a': ['a', 'a', 'a'],
+            'test-cond-idx': [0, 1, 2]})
+        gaussian_copula._metadata.make_ids_unique.return_value = expected
+
+        # Run
+        out = GaussianCopula.sample(gaussian_copula, num_rows=3, conditions={'a': 'a'})
+
+        # Asserts
+        gaussian_copula._conditionally_sample_rows.assert_called_once_with(
+            DataFrameMatcher(pd.DataFrame({'test-cond-idx': [0, 1, 2], 'a': ['a', 'a', 'a']})),
+            100,
+            10,
+            {'a': 'a'},
+            None,
+            0.01,
+            False,
+        )
+        assert out.equals(expected)
+
+
 @patch('sdv.tabular.base.Table', spec_set=Table)
 def test__init__passes_correct_parameters(metadata_mock):
     """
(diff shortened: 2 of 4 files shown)


sdv/tabular/base.py
```diff
@@ -457,7 +457,7 @@ def sample(self, num_rows=None, max_retries=100, max_rows_multiplier=10,
 
             condition_indices = dataframe[COND_IDX]
             condition = dict(zip(condition_columns, group))
-            if transformed_conditions.empty:
+            if len(transformed_columns) == 0:
```
csala, line 460: Interesting point here. The `transformed_conditions.empty` was actually covering the case where there are no rows in the `DataFrame`, rather than the case where there are no columns (since there always was at least the COND_IDX column).
But since we are inside the `groupby` loop, we know that there is at least 1 `conditions` row, which means that the `transformed_conditions` cannot be empty!
So, if I'm reading it right, this block of code was unreachable in the previous version!

csala: So, to add to the previous comment: Can we delete this test?
https://github.com/sdv-dev/SDV/blob/0d4163ca537f22bb47385badb9e1399fc1f72928/tests/unit/tabular/test_base.py#L85

To be fair, the current unit tests for the `BaseTabularModel` are not properly implemented, since they should be done inside a `class` called `TestBaseTabularModel` which would implement at least 1 test for each one of the methods inside the `BaseTabularModel`.

Can you get started on creating the test class and adding at least the `no transformed_columns` unit test for the `sample` method? The thing should be similar to what can be found in https://github.com/sdv-dev/SDV/blob/master/tests/unit/tabular/test_copulas.py

csala (changes requested):
The PR looks good so far, but I added a comment about expanding a bit the unit tests.

sdv/tabular/base.py
csala: I just raised issue #543 to cover the proper testing of the `BaseTabularModel`. That will be covered in a separated PR, but let's get started on adding at least the test suggested above to the new test class.

katxiao: Yes, I believe it was previously unreachable because we create the `COND_IDX` column in `transformed_conditions` beforehand.

I removed the unneeded test, created the class `TestBaseTabularModel`, and added a `no_transformed_columns` unit test. Moving the existing test cases in a separate issue sounds good to me.

csala (changes requested):
I think we are almost there. I added a couple of comments of things that could be refined in the tests.

sdv/tabular/base.py
amontanez24: This is strange. I remember adding the test for this because of an issue I ran into while testing it, but I can't remember what it was

tests/unit/tabular/test_base.py
```diff
@@ -16,6 +16,68 @@
 ]
 
 
+class DataFrameMatcher:
```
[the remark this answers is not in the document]
katxiao: I moved it to `tests/utils/utils.py`

tests/unit/tabular/test_base.py
```diff
@@ -16,6 +18,60 @@
...
+            'a': ['a', 'a', 'a'],
+            COND_IDX: [0, 1, 2]})
+        gaussian_copula._metadata.make_ids_unique.return_value = expected
+
+        # Run
+        out = GaussianCopula.sample(gaussian_copula, num_rows=3, conditions={'a': 'a'})
+
+        # Asserts
+        gaussian_copula._conditionally_sample_rows.assert_called_once_with(
+            DataFrameMatcher(pd.DataFrame({COND_IDX: [0, 1, 2], 'a': ['a', 'a', 'a']})),
```
amontanez24, line 64: this is great!

amontanez24 (approved):
This looks good! If you end up changing the tests to have a class, then it will likely create merge conflicts with https://github.com/sdv-dev/SDV/pull/540, so let's just be careful about how we merge these

sdv/tabular/base.py
amontanez24: When looking at this again, I found that if you run `df.empty` on a `DataFrame` that has no columns/rows but does have an index, it still returns `True`. So I'm confused how this case wasn't being reached

amontanez24: Sorry I was confused because I forgot that `transformed_conditions.reset_index(inplace=True)`, will move the previous index column into the `DataFrame`

tests/unit/tabular/test_base.py
```diff
@@ -4,9 +4,11 @@
 import pytest
 
 from sdv.metadata.table import Table
+from sdv.tabular.base import COND_IDX
 from sdv.tabular.copulagan import CopulaGAN
 from sdv.tabular.copulas import GaussianCopula
 from sdv.tabular.ctgan import CTGAN, TVAE
+from tests.utils.utils import DataFrameMatcher
```
csala, line 11: Why is this `utils.utils`?
I think it could be directly put under `tests/utils.py`

csala (approved):
I think this is good. I just added a comment about a module that could be moved, the coding is all ready to go,

PR #544, opened 2021-08-04
Multi-column `CustomConstraint`

sarahmish:
``CustomConstraint`` with multi-column + documentation


docs/user_guides/single_table/custom_constraints.rst
```diff
@@ -0,0 +1,236 @@
...
+    def reverse_transform(table_data):
+        table_data['age'] = table_data['age'] / 2
+        return table_data
+
+Lastly, we write the ``is_valid`` function to assess whether the column is even or not.
+
+.. ipython:: python
+    :okwarning:
+
+    def is_valid(table_data):
```
csala, line 61: Doubting about whether to include the `is_valid` in the main example, or just move it to the `note` below as an alternative possibility.

docs/user_guides/single_table/custom_constraints.rst
```diff
@@ -0,0 +1,236 @@
...
+
+    sampled
+
+
+.. note::
+    It is sufficient to define ``is_valid`` function alone. In this case, the constraint will
+    use ``reject_sampling`` strategy. For example, ``CustomConstraint(is_valid=is_valid)``.
+
+
+Applying Functions to Multiple Columns
```
csala, line 99: I like the question asked below. Why not make it the title?

```
Can I apply the same function to multiple columns?
--------------------------------------------------
```

And then start on the `Say we want...`

docs/user_guides/single_table/custom_constraints.rst
```diff
@@ -0,0 +1,236 @@
...
+--------------------------------------
+
+What if I want to apply the even constraint to multiple columns? Say we want ``age``
+and ``age_when_joined`` to be both even. Rather than defining two constraints,
+we provide another style of writing functions such that the function should accept
```
csala, line 104: `...two constraints, or editing the code of our functions for each new column that we want to constraint, we provide...`

docs/user_guides/single_table/custom_constraints.rst
```diff
@@ -0,0 +1,236 @@
...
+
+Viewing ``sampled`` we now see two columns that are always even.
+
+.. ipython:: python
+    :okwarning:
+
+    sampled
+
+
+Applying Functions to Multiple Columns with Access to Table
```
csala, line 161: Same as above, let's make the title a question:

```
Can I access the rest of the table from my column functions?
------------------------------------------------------------
```

docs/user_guides/single_table/custom_constraints.rst
```diff
@@ -0,0 +1,236 @@
...
+
+    sampled = gc.sample(10)
+
+    sampled
+
+This style gives flexibility to access any column in the table while still operating on 
+a column basis.
+
+.. note::
+	The ``transform`` and ``reverse_transform`` methods return a table. Except
```
csala, line 233: I think this note is outdated.

csala (changes requested):
I think the coding is ready. Added a few comments to the docs.

docs/user_guides/single_table/custom_constraints.rst
sarahmish: I have this doubt too. I think there is a branch for ``transform`` and ``reverse_transform``; another branch for ``is_valid`` but I think this distinction should be on a higher level. In other words, we should describe the essence of these functions in the `Constraints` page, specifically in the ``handling_strategy``. And here in ``CustomConstraint`` we link to the place.

Then I would keep all three functions here defined such that the signature of the function is very clear.

docs/user_guides/single_table/custom_constraints.rst
```diff
@@ -0,0 +1,234 @@
...
+
+.. ipython:: python
+    :okwarning:
+
+    sampled
+
+
+.. note::
+    It is sufficient to define ``is_valid`` function alone. In this case, the constraint will
+    use ``reject_sampling`` strategy. For example, ``CustomConstraint(is_valid=is_valid)``.
```
amontanez24, line 99: minor: "the ``reject_sampling`` strategy."

tests/unit/constraints/test_tabular.py
```diff
@@ -50,9 +63,239 @@ def test___init__(self):
...
+        assert instance._is_valid == dummy_is_valid
+
+    def test__run_transform_table(self):
+        """Test the ``CustomConstraint._run`` method.
+
+        The ``_run`` method excutes ``transform`` and ``reverse_transform``
+        based on the signature of the functions.
+
+        Setup:
+        - Create dummy functions of general type.
```
amontanez24, line 77: I see this on most of these. What does create functions of a certain type mean? Can this be slightly more descriptive?

sdv/constraints/tabular.py
```diff
@@ -46,16 +46,60 @@ class CustomConstraint(Constraint):
...
+        return self._run(self._reverse_transform, table_data, reverse=True)
+
+    def _run_is_valid(self, table_data):
+        if self._columns:
+            try:
+                valid = [self._is_valid(table_data, column) for column in self._columns]
+            except TypeError:
+                valid = [self._is_valid(table_data[column]) for column in self._columns]
+
+            return np.logical_and.reduce(valid)
```
amontanez24, line 81: Is this faster than `numpy.all`? I think that can also be used with multi-dimensional arrays

docs/user_guides/single_table/custom_constraints.rst
csala: I'm a bit concerned about this being confusing, so one may thing that it is necessary to implement the three functions every time. What do you think about creating a new section at the end for this instead?

We could cover the three styles using `transform` and `reverse_transform` only, and then add a new section at the end (e.g. titled "Can I write a CustomConstraint based on reject sampling?") that explains only that.
This section could just mention that if there is no transform/reverse transform implementation possible, the `is_valid` functions with the same signature can be passed instead, and then show the three implementations at once at the end.

docs/user_guides/single_table/custom_constraints.rst
```diff
@@ -0,0 +1,263 @@
...
+   result will be a multiple of 500.
+
+.. ipython:: python
+    :okwarning:
+
+    def reverse_transform(table_data, column):
+        base = 500.
+        is_not_contractor = table_data.contractor == 0.
+        table_data[column] = table_data[column].round(4)
+        table_data.at[is_not_contractor, column] = table_data[column].loc[is_not_contractor].round()
```
csala, line 193: I think that `at` would not work here, since `at` [works for single values only](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.at.html#pandas.DataFrame.at). We need to use `loc` instead. Possibly irrelevant if we change to the proposal above.

docs/user_guides/single_table/custom_constraints.rst
```diff
@@ -0,0 +1,263 @@
...
+2. round values to zero for employees that are not contractors such that the end
+   result will be a multiple of 500.
+
+.. ipython:: python
+    :okwarning:
+
+    def reverse_transform(table_data, column):
+        base = 500.
+        is_not_contractor = table_data.contractor == 0.
+        table_data[column] = table_data[column].round(4)
```
csala, line 192: Implementation note: You could take advantage of the possibility of rounding to multiple levels at once:

```python3
round_levels = (table_data.contractor == 0) * 4    # sets 4 if contractor, 0 otherwise
table_column[column] = table_data[column].round(round_levels)
```

docs/user_guides/single_table/custom_constraints.rst
```diff
@@ -0,0 +1,263 @@
...
+
+.. ipython:: python
+    :okwarning:
+
+    constraint = CustomConstraint(
+        transform=transform, 
+        reverse_transform=reverse_transform
+    )
+
+    gc = GaussianCopula(constraints=[constraint])
```
csala, line 76: I think we could skip the modeling and sampling until the last style, so the story telling is continuous and it does not really run the model until all the needed pieces are in place.

docs/user_guides/single_table/custom_constraints.rst
```diff
@@ -0,0 +1,263 @@
...
+    employees = load_tabular_demo()
+    employees
+
+The dataset defined in :ref:`_single_table_constraints` contains basic details about employees.
+We will use this dataset to demonstrate how you can create your own constraint. 
+
+
+Using the ``CustomConstraint``
+------------------------------
+
```
csala, line 31: Storytelling: Rather than talking about "an example", I would just point at how the data is actually crafted.

"If you look at the data above, you will notice that..."

docs/user_guides/single_table/custom_constraints.rst
```diff
@@ -0,0 +1,263 @@
...
+.. ipython:: python
+    :okwarning:
+
+    sampled
+
+
+Can I apply the same function to multiple columns?
+--------------------------------------------------
+
+Say we want ``salary`` and ``annual_bonus`` to be both composed of the base unit. Rather than
```
csala, line 94: Following the same style above, I would not phrase this as a hypothesis, but rather as an observation.

"In the example above we fixed the ``salary`` format, but if we continue observing the data we will see that ``annual_bonus`` is also constrained by the same logic".

docs/user_guides/single_table/custom_constraints.rst
```diff
@@ -0,0 +1,263 @@
...
+------------------------------
+
+Let's consider the following example where we wish to generate synthetic data and we want
+a particular column, ``salary`` for example, to be a multiple of a *base* value, e.g. 500. 
+In other words, the ``salary`` increments by 500. We will define ``transform`` and 
+``reverse_transform`` methods to make sure our data satisfy our constraints.
+
+We can achieve our goal by performing transformations in a 2 step process:
+
+- Divide ``salary`` by the base unit (500). This transformation makes it easier to model the data
```
csala, line 39: wording sounds strange: "...transformation makes the job easier for the model, since..." ?

csala (approved):
I made a couple comments about the docs, but all of them are minor things.
Once these are fixed, I think it can go in

docs/user_guides/single_table/custom_constraints.rst
sarahmish: Is this warning ok to ignore?
```
SettingWithCopyWarning: 
A value is trying to be set on a copy of a slice from a DataFrame
```

I changed the above line to use `loc` and it seems to operate on the actual dataframe. So I am not sure why the warning is raised.

I also want to mention that `at` works for multiple values, I guess it's not the correct way though. We should revisit the change I made for GT to use `at` as well
https://github.com/sdv-dev/SDV/blob/cc6a9e0f4f8038f9a587a20c37f47e6f69eae07d/sdv/constraints/tabular.py#L464

docs/user_guides/single_table/custom_constraints.rst
sarahmish: I'm not sure this would work, since `round` expects the `decimal` argument to be an integer.

csala: Oh, you are right! I got confused with the `DataFrame.round`, where you can pass one level for each column.

docs/user_guides/single_table/custom_constraints.rst
csala: Yes, I think we should change it. Technically, the docs say that `at` is for single values, so we could expect this to change at any point.

csala (approved):
This looks good to go now!

docs/user_guides/single_table/custom_constraints.rst
```diff
@@ -0,0 +1,239 @@
+.. _custom_constraints:
+
+Defining Custom Constraints
+===========================
+
+In some cases, the predefined constraints do not cover all your needs. 
+In such scenarios, you can use ``CustomConstraint`` to define your own 
+logic on how you would like to apply it to your data. There are three
```
amontanez24, line 8: More of a writing comment, but I'm not sure what "it" refers to. Maybe say "define your own logic on how to constrain your data"

docs/user_guides/single_table/custom_constraints.rst
```diff
@@ -0,0 +1,239 @@
...
+
+Defining Custom Constraints
+===========================
+
+In some cases, the predefined constraints do not cover all your needs. 
+In such scenarios, you can use ``CustomConstraint`` to define your own 
+logic on how you would like to apply it to your data. There are three 
+main functions that you can create:
+
+- ``transform`` which is responsible for the forward pass when using ``transform`` strategy.
```
amontanez24, line 11: Maybe also add what this method is meant to do. Like change the data in whatever way necessary to enforce constraint

docs/user_guides/single_table/custom_constraints.rst
```diff
@@ -0,0 +1,239 @@
...
+This style gives flexibility to access any column in the table while still operating on 
+a column basis.
+
+
+Can I write a ``CustomConstraint`` based on reject sampling?
+------------------------------------------------------------
+
+In the previous section, we defined our ``CustomConstraint`` using ``transform`` and 
+``reverse_transform`` functions. Sometimes, our constraints are not possible to implement 
+using these methods, that is when we rely on ``reject_sampling`` strategy. In ``reject_sampling``
```
amontanez24, line 201: minor: "the reject_sampling strategy"

docs/user_guides/single_table/custom_constraints.rst
```diff
@@ -0,0 +1,239 @@
...
+we need to implement an ``is_valid`` function that identifies which rows do not follow the 
+said constraint, in our case, which rows are not a multiple of the *base* unit.
+
+We can define ``is_valid`` according to the three styles mentioned in the previous section:
+
+1. function with ``table_data`` argument.
+2. function with ``column_data`` argument.
+3. function with ``table_data`` and ``column`` argument.
+
+Here is an example of how you would define ``is_valid`` for each one of the mentioned styles:
```
amontanez24, line 211: I would say somewhere explicitly what the expected output is. (A pd.Series where every valid row is 1 and the rest are 0).

amontanez24 (approved):
Some comments regarding the wording but besides that this looks good to go!

docs/user_guides/single_table/custom_constraints.rst
sarahmish: I made a minor adjustment and made it *True* and *False*, I think that's easier to understand.

PR #546, opened 2021-08-04
Check for invalid conditions when sampling

### LINKED ISSUE #511: Passing invalid conditions when using constraints produces unreadable errors
Currently, using invalid conditions when using constraints produces unreadable errors. Instead, it would make more sense to validate the inputs and produce more meaningful error messages.

For example, if we use the `UniqueCombination` constraint and condition on the two constraint columns being different (which is an invalid condition) it produces a `ValueError: No objects to concatenate`, which doesn't provide much feedback about where the error is.

```python3
import pandas as pd
from sdv.constraints import *
from sdv.tabular import *

data = pd.DataFrame({
    'a': zero_to_nine,
    'b': zero_to_nine,
    'c': zero_to_nine
})

    columns=['a', 'b'],
)

model.fit(data)
samples = model.sample(10, conditions={'a': 0, 'b': 1})
```

katxiao:
If we don't have any non-null transformed conditions when sampling, return a more informative error (`InvalidConditionsError`), which tells the user that there are no valid conditions to sample with.

Resolves #511

### DIFF
--- a/sdv/metadata/table.py
+++ b/sdv/metadata/table.py
@@ -578,8 +578,7 @@ class Table:
 
         Args:
             data (pandas.DataFrame):
-                Table data. If the table data is not valid for the provided constraints,
-                a `ValueError` is raised.
+                Table data.
             on_missing_column (str):
                 If the value is error, then a `MissingConstraintColumnError` is raised.
                 If the value is drop, then the columns involved in the constraint that
@@ -588,6 +587,10 @@ class Table:
         Returns:
             pandas.DataFrame:
                 Transformed data.
+
+        Raises:
+            ConstraintsNotMetError:
+                If the table data is not valid for the provided constraints.
         """
         if not self.fitted:
             raise MetadataNotFittedError()
--- a/sdv/metadata/table.py
+++ b/sdv/metadata/table.py
@@ -11,6 +11,7 @@ from faker import Faker
 
 from sdv.constraints.base import Constraint
 from sdv.constraints.errors import MissingConstraintColumnError
+from sdv.errors import ConstraintsNotMetError
 from sdv.metadata.errors import MetadataError, MetadataNotFittedError
 from sdv.metadata.utils import strings_from_regex
 
@@ -553,12 +554,32 @@ class Table:
 
         return data
 
+    def _validate_data_on_constraints(self, data):
+        """Make sure the given data is valid for the given constraints.
+
+        Args:
+            data (pandas.DataFrame):
+                Table data.
+
+        Returns:
+            bool:
+                Whether or not the table data is valid for the constraints.
+        """
+        is_valid = True
+
+        for constraint in self._constraints:
+            if set(constraint.constraint_columns).issubset(data.columns.values):
+                is_valid = is_valid and constraint.is_valid(data).all()
+
+        return is_valid
+
     def transform(self, data, on_missing_column='error'):
         """Transform the given data.
 
         Args:
             data (pandas.DataFrame):
-                Table data.
+                Table data. If the table data is not valid for the provided constraints,
+                a `ValueError` is raised.
             on_missing_column (str):
                 If the value is error, then a `MissingConstraintColumnError` is raised.
                 If the value is drop, then the columns involved in the constraint that
@@ -574,6 +595,9 @@ class Table:
         fields = [field for field in self.get_dtypes(ids=False) if field in data.columns]
         LOGGER.debug('Anonymizing table %s', self.name)
         data = self._anonymize(data[fields])
+
+        if not self._validate_data_on_constraints(data):
+            raise ConstraintsNotMetError('Data is not valid for the given constraints')
 
         LOGGER.debug('Transforming constraints for table %s', self.name)
         data = self._transform_constraints(data, on_missing_column)
--- a/sdv/metadata/table.py
+++ b/sdv/metadata/table.py
@@ -562,16 +562,16 @@ class Table:
                 Table data.
 
         Returns:
-            bool:
-                Whether or not the table data is valid for the constraints.
-        """
-        is_valid = True
+            None
 
+        Raises:
+            ConstraintsNotMetError:
+                If the table data is not valid for the provided constraints.
+        """
         for constraint in self._constraints:
             if set(constraint.constraint_columns).issubset(data.columns.values):
-                is_valid = is_valid and constraint.is_valid(data).all()
-
-        return is_valid
+                if not constraint.is_valid(data).all():
+                    raise ConstraintsNotMetError('Data is not valid for the given constraints')
 
     def transform(self, data, on_missing_column='error'):
         """Transform the given data.
@@ -599,8 +599,7 @@ class Table:
         LOGGER.debug('Anonymizing table %s', self.name)
         data = self._anonymize(data[fields])
 
-        if not self._validate_data_on_constraints(data):
-            raise ConstraintsNotMetError('Data is not valid for the given constraints')
+        self._validate_data_on_constraints(data)
 
         LOGGER.debug('Transforming constraints for table %s', self.name)
         data = self._transform_constraints(data, on_missing_column)
--- a/sdv/tabular/base.py
+++ b/sdv/tabular/base.py
@@ -8,6 +8,7 @@ from warnings import warn
 import numpy as np
 import pandas as pd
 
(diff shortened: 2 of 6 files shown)


tests/unit/tabular/test_base.py
```diff
@@ -71,6 +72,39 @@ def test_sample_no_transformed_columns(self):
...
+            - invalid conditions
+        Side Effects:
+            - Expect an ``InvalidConditionsError``.
+        """
+        # Setup
+        gaussian_copula = Mock(spec_set=GaussianCopula)
+
+        gaussian_copula._make_conditions_df.return_value = pd.DataFrame({
+            'a': [0, 0, 0],
+            'b': [1, 1, 1]})
```
pvk-developer, line 99: I think it's better if we add a break between the bracket:
```python
        gaussian_copula._make_conditions_df.return_value = pd.DataFrame({
            'a': [0, 0, 0],
            'b': [1, 1, 1]
        })
```

tests/unit/tabular/test_base.py
```diff
@@ -71,6 +72,39 @@ def test_sample_no_transformed_columns(self):
...
+        gaussian_copula._metadata.get_fields.return_value = ['a', 'b']
+        gaussian_copula._metadata.transform.return_value = pd.DataFrame({
+            'a#b': [None, None, None]})
```
pvk-developer, line 102: Style
```python
        gaussian_copula._metadata.transform.return_value = pd.DataFrame({
            'a#b': [None, None, None]
        })
```

sdv/tabular/base.py
```diff
@@ -471,6 +472,8 @@ def sample(self, num_rows=None, max_retries=100, max_rows_multiplier=10,
             else:
                 transformed_conditions_in_group = transformed_conditions.loc[condition_indices]
                 transformed_groups = transformed_conditions_in_group.groupby(transformed_columns)
+                if len(transformed_groups.groups) == 0:
+                    raise InvalidConditionsError
```
pvk-developer, line 476: Add a break line for consistency

pvk-developer (changes requested):
Other than styling looks good to me! :tada:

sdv/tabular/base.py
csala: I'm not sure about defining a new type of constraint for this. I would rather just use a `ValueError` and put the proper message on it, since this is just an invalid user input scenario.

sdv/tabular/base.py
```diff
@@ -471,6 +472,8 @@ def sample(self, num_rows=None, max_retries=100, max_rows_multiplier=10,
             else:
                 transformed_conditions_in_group = transformed_conditions.loc[condition_indices]
                 transformed_groups = transformed_conditions_in_group.groupby(transformed_columns)
+                if len(transformed_groups.groups) == 0:
```
csala, line 475: I think this implementation proposal breaks the possibility to fall back to conditioning based on reject-sampling if the constraints cannot transform the conditions data.

Instead of this, I suggest adding a new method that validates the input by iterating over the constraints and running the input data through their `is_valid` method if all the `constraint_columns` are contained in the `conditions`

csala (changes requested):

sdv/tabular/base.py
katxiao: Just to clarify..  if there are no transformed groups, we should fall into the case on L461?

sdv/tabular/base.py
katxiao: Follow up question - what is the disadvantage to defining a custom error type that extends a standard one (e.g. `ValueError`). Is it confusing for users?

sdv/metadata/table.py
```diff
@@ -553,12 +554,32 @@ def _transform_constraints(self, data, on_missing_column='error'):
...
+        return is_valid
+
     def transform(self, data, on_missing_column='error'):
         """Transform the given data.
 
         Args:
             data (pandas.DataFrame):
-                Table data.
+                Table data. If the table data is not valid for the provided constraints,
+                a `ValueError` is raised.
```
csala, line 582: This could be moved to a `Raises` section below (inside the docstring).
I think the exception type should also change:

```
Raises:
    ConstraintsNotMet:
        If the table data is not valid for the provided constraints.
```

sdv/tabular/base.py
```diff
@@ -411,7 +412,8 @@ def sample(self, num_rows=None, max_retries=100, max_rows_multiplier=10,
                 value, then this method generates `num_rows` samples, all of
                 which are conditioned on the given variables. If this is a DataFrame,
                 then it generates an output DataFrame such that each row in the output
-                is sampled conditional on the corresponding row in the input.
+                is sampled conditional on the corresponding row in the input. If the
+                given conditions are not able to be met, raises a `ValueError`.
```
csala, line 416: Same as above, this could be moved to the `Raises` section

sdv/tabular/base.py
```diff
@@ -439,7 +441,11 @@ def sample(self, num_rows=None, max_retries=100, max_rows_multiplier=10,
             if column not in self._metadata.get_fields():
                 raise ValueError(f'Invalid column name `{column}`')
 
-        transformed_conditions = self._metadata.transform(conditions, on_missing_column='drop')
+        try:
+            transformed_conditions = self._metadata.transform(conditions, on_missing_column='drop')
+        except ConstraintsNotMetError:
+            raise ValueError("The given conditions are not able to be met.")
```
csala, line 447: I would not raise a new exception, since this will add a new layer to the traceback.
I would just updated the message and re-raise:
```
except ConstraintsNotMetError as cnme:
    cnme.message = 'Passed conditions are not valid for the given constraints'
    raise
```

csala (changes requested):
A couple of minor things and we should be ready to go

tests/integration/test_constraints.py
```diff
@@ -42,29 +40,3 @@ def test_constraints(tmpdir):
...
-
-def _is_valid(rows):
-    if not _IS_VALID_CALLED:
-        _IS_VALID_CALLED.append(True)
-        return pd.Series([False] * len(rows), index=rows.index)
-
-    return pd.Series([True] * len(rows), index=rows.index)
-
-
-def test_constraints_reject_sampling_zero_valid():
```
katxiao, line 58: @csala This test started failing because the first time `constraint.is_valid` is called is now within the `BaseTabularMethod.transform` method instead of when we are actually sampling data. 

I was thinking about updating the mocked `is_valid` function to return False on the second time it's called, but that seems very fragile. Instead, I moved the test for this scenario (returning zero valid rows on the first sample iteration) to a unit test in `tests/unit/tabular/test_base.py`. Let me know what you think

sdv/tabular/base.py
katxiao: In the end, we decided that the code should not reach the case of no transformed groups -- if the conditions are invalid, the transformed_conditions should have zero columns (a case that is already handled).

sdv/metadata/table.py
```diff
@@ -553,6 +554,25 @@ def _transform_constraints(self, data, on_missing_column='error'):
...
+
+        Returns:
+            bool:
+                Whether or not the table data is valid for the constraints.
+        """
+        is_valid = True
+
+        for constraint in self._constraints:
+            if set(constraint.constraint_columns).issubset(data.columns.values):
+                is_valid = is_valid and constraint.is_valid(data).all()
```
csala, line 572: What about raising the exception directly from here, so we do not need to continue validating?

tests/integration/test_constraints.py
csala: Yes, sounds good!

csala (commented):
Added one comment about raising an exception sooner. I think doing the change will make little difference so @katxiao your call on whether to take the suggestion or not.

sdv/metadata/table.py
katxiao: So I personally prefer raising it in `transform`. I feel like it scopes this method to answering the question, "is it valid or not?", and then the caller can determine what to do if not valid.

I can make it return early if any of the `constraint.is_valid` is False though, since it doesn't need to keep iterating in that case.

katxiao: Update: I moved the error to the helper method, so that it is thrown immediately

csala (approved):

pvk-developer (approved):
LGTM!

PR #547, opened 2021-08-04
Multi-row constraints documentation

amontanez24:


sdv/demo.py
```diff
@@ -322,17 +322,14 @@ def _load_tabular_dummy():
     """Load a dummy tabular demo dataframe."""
     age = np.random.randint(30, 50, 12)
     age_when_joined = age - np.random.randint(0, 10, 12)
-    faker = Faker()
-    names = [faker.name() for _ in range(12)]
```
amontanez24, line 326: @csala I was actually thinking that the `Unique` constraint might make more sense with the names and addresses since they would not be expected to be primary keys. Not sure what you think

sdv/demo.py
```diff
@@ -322,17 +322,14 @@ def _load_tabular_dummy():
...
-    adresses = [faker.address() for _ in range(12)]
+    employee_ids = np.random.randint(100000, 999999, 12)
```
amontanez24, line 325: I was going for a 6 digit id but this range is so large I don't think the constraint really will have to do much

csala (approved):
Added a couple of comments. Other than that, it looks good to go to me.

docs/user_guides/single_table/predefined_constraints.rst
```diff
@@ -0,0 +1,410 @@
...
+.. ipython:: python
+    :okwarning:
+
+    from sdv.constraints import OneHotEncoding
+
+    one_hot_constraint = OneHotEncoding(
+        columns=['full_time', 'part_time', 'contractor']
+    )
+
+Multi-row Constraints
```
sarahmish, line 336: Suggestions: I find it difficult to follow how is "multi-row" different from the previous constraints because in the previous constraints, we do not explicitly mention they are "column-based" or whatever. I would either remove it or include "Multi-column Constraints" or simply "Row Constraints" and "Column Constraints" headers instead.

docs/user_guides/single_table/predefined_constraints.rst
```diff
@@ -0,0 +1,410 @@
...
+
+.. ipython:: python
+    :okwarning:
+
+    experience_years_gt_one_constraint = GreaterThan(
+        low=1,
+        high=['years_in_the_company', 'prior_years_experience'],
+        scalar='low',
+        handling_strategy='reject_sampling'
+    )
```
sarahmish, line 196: I just realized that the training data contradicts this constraints because we have "years_in_the_company" equal to zero which mainly falls on my part. Maybe we can change this to `low=0`.

docs/user_guides/single_table/predefined_constraints.rst
```diff
@@ -0,0 +1,410 @@
...
+In the cases where applying a **Transform** strategy is not possible or
+may affect the quality of the learning process, **SDV** can apply a
+**Reject Sampling** strategy.
+
+When using this strategy, **SDV** validates the sampled rows, discards
+the ones that do not adjust to the constraint, and re-samples them. This
+process is repeated until enough rows have been sampled.
+
+Predefined Constraints
+-------------------------------
```
sarahmish, line 107: Suggestion: there is another documentation page for "CustomConstraints" coming where it's on the same level as "Predefined Constraints". Therefore, I would make this a higher header
```rst
Predefined Constraints
======================
```
and I would actually split this into two pages, such that it will be:
```
├── single_table/
    ├── constraints.rst
    ├── predefined_constraints.rst
    └── custom_constraints.rst
```

Otherwise just truncate the number of '--' to be of equal length to the header.

sarahmish (commented):
It looks ready, I just have a few suggestions if you think they make sense.

docs/user_guides/single_table/predefined_constraints.rst
```diff
@@ -0,0 +1,410 @@
...
+example, let us create a ``GaussianCopula`` model passing it the
+constraints that we just defined as a ``list``:
+
+.. ipython:: python
+    :okwarning:
+
+    from sdv.tabular import GaussianCopula
+
+    constraints = [
+        unique_employee_id_company_constraint,
```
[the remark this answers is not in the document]
amontanez24: this won't work because the next constraint modifies a column it uses

docs/user_guides/single_table/predefined_constraints.rst
```diff
@@ -0,0 +1,410 @@
...
+of columns for which each unique combination of their values can only show up once
+in the table. The ``Unique`` constraint enforces that the provided column(s) at
+most have one instance of each possible combination of values in the synthetic data.
+
+As an example, let us apply this constraint to the ``employee_id`` and ``company``
+columns, since the ``employee_id`` should be unique for each ``company``.
+
+To use this constraint, we must make an instance and provide:
+
+- A list of column names or string representing the one column to keep unique
```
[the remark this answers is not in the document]
amontanez24: I want to specify that a single column name can be passed as a string still. How about "A list of the names of the columns that's value combination must be unique, or a string if there is only one column"?

docs/user_guides/single_table/predefined_constraints.rst
amontanez24: Wouldn't you just use the `Positive` constraint then?

amontanez24: Changed demo data to not generate 0s

sarahmish (approved):
LGTM

amontanez24:
> Added a couple of comments. Other than that, it looks good to go to me.

Made some changes, so you might want to take a quick look again in the morning. If it's still all good I can merge or you can feel free to

csala (approved):
Ready to go!

PR #548, opened 2021-08-05
Raise error when multiple constraints can't be enforced

### LINKED ISSUE #541: Raise error when multiple constraints can't be enforced
### Problem Description

Certain constraints change or drop the input columns when using the `transform` `handling_strategy`. This means that if multiple constraints are applied to the same column, and one of them does this, then the constraints will not work properly since one of them will attempt to make changes to a column that was either altered or dropped. 

This problem does not occur if constraints use the `reject_sampling` `handling_strategy`. 

### Expected behavior

The goal of this issue is to sort constraints so that any using the `reject_sampling` strategy occur first. On top of that, if multiple constraints are using `transform` and they touch the same column, we should raise an error saying that those constraints might not be enforceable.

amontanez24:
Resolve #541

This PR:
- adds a `rebuild_columns` attribute to the `Constraint` classes. This list will contain any column names that the constraint needs to rebuild during `reverse_transform`
- Adds functionality to sort the constraints during the `metadata.Table.__init__`. It will put any constraints that have `reject_sampling` as the `handling_strategy` before the rest
- Adds a validation method to the `__init__` that checks to make sure the `rebuild_columns` in a constraint do not show up in the `constraint_columns` of a later constraint.

### DIFF
--- a/sdv/metadata/table.py
+++ b/sdv/metadata/table.py
@@ -207,6 +207,7 @@ class Table:
 
     @staticmethod
     def _prepare_constraints(constraints):
+        constraints = constraints or []
         rebuild_columns = set()
         transform_constraints = []
         reject_sampling_constraints = []
@@ -248,7 +249,6 @@ class Table:
         self._sequence_index = sequence_index
         self._entity_columns = entity_columns or []
         self._context_columns = context_columns or []
-        self._constraints = constraints or []
         self._constraints = self._prepare_constraints(self._constraints)
         self._dtype_transformers = self._DTYPE_TRANSFORMERS.copy()
         self._transformer_templates = self._TRANSFORMER_TEMPLATES.copy()
--- a/sdv/metadata/table.py
+++ b/sdv/metadata/table.py
@@ -249,7 +249,7 @@ class Table:
         self._sequence_index = sequence_index
         self._entity_columns = entity_columns or []
         self._context_columns = context_columns or []
-        self._constraints = self._prepare_constraints(self._constraints)
+        self._constraints = self._prepare_constraints(constraints)
         self._dtype_transformers = self._DTYPE_TRANSFORMERS.copy()
         self._transformer_templates = self._TRANSFORMER_TEMPLATES.copy()
         self._update_transformer_templates(rounding, min_value, max_value)
--- a/sdv/metadata/table.py
+++ b/sdv/metadata/table.py
@@ -205,6 +205,16 @@ class Table:
                 'float': custom_float
             })
 
+    @staticmethod
+    def _get_constraint_sort_key(constraint):
+        try:
+            if constraint.handling_strategy == 'reject_sampling':
+                return 0
+            else:
+                return 1
+        except AttributeError:
+            return 1
+
     def __init__(self, name=None, field_names=None, field_types=None, field_transformers=None,
                  anonymize_fields=None, primary_key=None, constraints=None,
                  dtype_transformers=None, model_kwargs=None, sequence_index=None,
@@ -222,6 +232,7 @@ class Table:
         self._entity_columns = entity_columns or []
         self._context_columns = context_columns or []
         self._constraints = constraints or []
+        self._constraints.sort(key=self._get_constraint_sort_key)
         self._dtype_transformers = self._DTYPE_TRANSFORMERS.copy()
         self._transformer_templates = self._TRANSFORMER_TEMPLATES.copy()
         self._update_transformer_templates(rounding, min_value, max_value)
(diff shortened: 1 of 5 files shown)


sdv/metadata/table.py
```diff
@@ -205,6 +205,51 @@ def _update_transformer_templates(self, rounding, min_value, max_value):
...
+                pass
+
+        for i in range(len(transform_constraints) - 1):
+            cur_constraint = transform_constraints[i]
+            for j in range(i + 1, len(transform_constraints)):
+                next_constraint = transform_constraints[j]
+                try:
+                    for column in cur_constraint.rebuild_columns:
+                        if column in next_constraint.constraint_columns:
+                            raise Exception('Multiple constraints will modify the same column: '
```
fealho, line 246: I think this message could be more helpful if it told the user which constraints caused the issue. Also, I think there's a more efficient way to implement this without using nested for loops O(n^2) by using a counter?

fealho (commented):
Besides that one comment, looks good.

sdv/metadata/table.py
amontanez24: What do you mean by using a counter?

fealho (approved):

sdv/metadata/table.py
```diff
@@ -205,6 +206,57 @@ def _update_transformer_templates(self, rounding, min_value, max_value):
...
+                constraint = constraint().to_dict()
+            elif isinstance(constraint, Constraint):
+                constraint = constraint.to_dict()
+
+            constraint = Constraint.from_dict(constraint)
+            self._constraints[idx] = constraint
+
+        self._constraints.sort(key=self._get_constraint_sort_key)
+
+    def _validate_constraint_order(self):
```
csala, line 231: It seems like we are doing the sorting and the validation in two different steps, which all iterate over the constraints multiple times.
Could everything be put in a single loop instead?
Another thing is that the current loop seems to keep track of the individual constraints rather than the columns that they modify. We could probably just keep track of the columns and not the constraints.

So, the logic could be something like:

```python3
@staticmethod
def _prepare_constraints(constraints):
    rebuilt_columns = set()
    transform = []
    reject_sampling = []
    for constraint in constraints:
        if not constraint.rebuild_columns:
            reject_sampling.append(constraint)
        elif rebuild_columns & set(constraint.constraint_columns):
            # previous rebuild columns overlap with current constraint_columns
            # we can recompute the intersection to know which columns overlap
            # and put them on the error message
            raise error
        else:
            transform.append(constraint)
            rebuild_columns.update(constraint.rebuild_columns)

    return reject_sampling + transform
```

tests/unit/metadata/test_table.py
```diff
@@ -36,6 +36,121 @@ def test__init__(self, transformer_mock):
         transformer_mock.assert_any_call(
             dtype=float, rounding=-1, max_value=100, min_value=-50)
 
+    @patch.object(Constraint, 'from_dict')
+    def test__init__sorts_constraints(self, from_dict_mock):
```
csala, line 40: I would move part of this logic to specific `unit` tests for the exact method instead of proxying the test through the `__init__` call.
At the `__init__` level I would just validate that we are properly calling the methods that do the work, but the internal methods should be tested on their own.

csala (changes requested):

sdv/metadata/table.py
```diff
@@ -222,6 +249,7 @@ def __init__(self, name=None, field_names=None, field_types=None, field_transfor
         self._entity_columns = entity_columns or []
         self._context_columns = context_columns or []
         self._constraints = constraints or []
```
csala, line 251: I suspect this line could be dropped now, and rely on the next on only.
I would also take care of the `None` possibility (so, the `constraints or []` part) inside the `_prepare` method, so it is hidden away.

csala (approved):
Added one comment about a line that may not be necessary anymore, but other than that, this looks ready to go!
Thanks @amontanez24 !

sdv/metadata/table.py
```diff
@@ -221,7 +249,7 @@ def __init__(self, name=None, field_names=None, field_types=None, field_transfor
         self._sequence_index = sequence_index
         self._entity_columns = entity_columns or []
         self._context_columns = context_columns or []
-        self._constraints = constraints or []
+        self._constraints = self._prepare_constraints(self._constraints)
```
csala, line 252: Hey! This should the `(constraints)` now, since `self._constraints` is still unassigned.

amontanez24: yeah, tests caught it thankfully

Issue #522, opened 2021-07-22
Contrary to documentation, the transform strategy with UniqueCombinations does slow down model fitting


tjhallum:
### Environment Details

SDV version: 0.11.0
Python version: 3.6.8
Operating System: Windows 10

### Error Description

When applying a `UniqueCombinations` constraint with `handling_strategy = 'transform'` to TVAE/CTGAN models being trained on > 2,000 observations, the time required for model fitting increases considerably (even worse, the increased time requirement does not appear linearly correlated to the size of the training set).

As an example, my computer can fit non-constrained TVAE/CTGAN models on 10,000 observations of airline data within just a few minutes. By contrast, I added a `UniqueCombinations` constraint with `handling_strategy = 'transform'` and had to stop the model fitting process at 1 hour...because I wasn't sure exactly how long the model fitting was going to take.

This behavior is in direct conflict with the Handling Constraints User Guide, which states that the transform strategy for constraints "[is very efficient and does not affect the speed of the modeling and sampling process, but in some cases might affect the quality of the learning process or simply not be possible.](https://sdv.dev/SDV/user_guides/single_table/constraints.html#:~:text=The%20Transform%20strategy%20is%20very%20efficient%20and%20does%20not%20affect%20the%20speed%20of%20the%20modeling%20and%20sampling%20process%2C%20but%20in%20some%20cases%20might%20affect%20the%20quality%20of%20the%20learning%20process%20or%20simply%20not%20be%20possible.)" 

### Steps to reproduce

The full Jupyter Notebook that reproduces this behavior is attached (the notebook pulls it's own data from the web): [generate_adversarial_data-tvae.zip](https://github.com/sdv-dev/SDV/files/6863552/generate_adversarial_data-tvae.zip)

katxiao:
Thanks for the issue @tjhallum! 

We believe that this is because the current default behavior of UniqueCombinations sets `fit_column_model=True`, which fits a GaussianCopula model to the constraint columns. We are working on updating it to `fit_column_model=False` by default. In the meantime, you could try manually setting the parameter to `False` and seeing if that solves the performance issue.

PR #549, opened 2021-08-06
Add documentation for fit_column_model and update default

### LINKED ISSUE #517: Add explanation of `fit_columns_model` to API docs
SDV Version: 0.11.0

A few of the tabular constraints now have a new kwarg called `fit_columns_model`. While it shows up in the API, there is no explanation of if it the parameters section so I'm not sure how to use it.

Affected constraints:
* [UniqueCombinations](https://sdv.dev/SDV/api_reference/constraints/api/sdv.constraints.UniqueCombinations.html)
* [GreaterThan](https://sdv.dev/SDV/api_reference/constraints/api/sdv.constraints.GreaterThan.html)
* [Positive](https://sdv.dev/SDV/api_reference/constraints/api/sdv.constraints.Positive.html)
* [Negative](https://sdv.dev/SDV/api_reference/constraints/api/sdv.constraints.Negative.html)
* [Between](https://sdv.dev/SDV/api_reference/constraints/api/sdv.constraints.Between.html)

katxiao:
Add documentation for fit_column_model and change the default value to `False`. Also add a warning for when reject sampling may take a long time

Resolves #517, #522, and #550

### DIFF
--- a/sdv/constraints/base.py
+++ b/sdv/constraints/base.py
@@ -4,6 +4,7 @@ import copy
 import importlib
 import inspect
 import logging
+import warnings
 
 import pandas as pd
 from copulas.multivariate.gaussian import GaussianMultivariate
@@ -111,7 +112,7 @@ class Constraint(metaclass=ConstraintMeta):
     def _identity(self, table_data):
         return table_data
 
-    def __init__(self, handling_strategy, fit_columns_model=True):
+    def __init__(self, handling_strategy, fit_columns_model=False):
         self.fit_columns_model = fit_columns_model
         if handling_strategy == 'transform':
             self.filter_valid = self._identity
@@ -226,6 +227,14 @@ class Constraint(metaclass=ConstraintMeta):
         """
         missing_columns = [col for col in self.constraint_columns if col not in table_data.columns]
         if missing_columns:
+            if not self._columns_model:
+                warning_message = (
+                    'The following constraint columns are missing from the table data '
+                    f'- `{missing_columns}`. When `fit_columns_model` is `False` and one or more '
+                    'constraint columns are missing, reject sampling can become slow.'
+                )
+                warnings.warn(warning_message, UserWarning)
+
             all_columns_missing = len(missing_columns) == len(self.constraint_columns)
             if self._columns_model is None or all_columns_missing:
                 raise MissingConstraintColumnError()
--- a/sdv/constraints/base.py
+++ b/sdv/constraints/base.py
@@ -229,9 +229,10 @@ class Constraint(metaclass=ConstraintMeta):
         if missing_columns:
             if not self._columns_model:
                 warning_message = (
-                    'The following constraint columns are missing from the table data '
-                    f'- `{missing_columns}`. When `fit_columns_model` is `False` and one or more '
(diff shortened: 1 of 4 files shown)


sdv/constraints/base.py
```diff
@@ -226,6 +227,14 @@ def _validate_constraint_columns(self, table_data):
         """
         missing_columns = [col for col in self.constraint_columns if col not in table_data.columns]
         if missing_columns:
+            if not self._columns_model:
```
katxiao, line 230: @csala I wasn't sure if this is the correct scenario to warn for. Could you elaborate on why missing columns might cause the reject sampling to become slow?

sdv/constraints/base.py
```diff
@@ -226,6 +227,14 @@ def _validate_constraint_columns(self, table_data):
...
+                warning_message = (
```
amontanez24, line 231: Could we add a small unit test for this?

sdv/constraints/base.py
csala: @katxiao I submitted issue #550 for completion, so the change and its motivation is actually registered on the issues side. I included there a small description of why conditional sampling will become slower if the columns model is not fitted.

sdv/constraints/base.py
katxiao: Yes, I just added it!

amontanez24: Awesome. Now that I look at this, I wonder if it is confusing to a user. @csala The case where this happens is usually because of conditioning right? Would they be confused to see a warning that says "constraints are missing from table data". Should we say that it's because of conditioning to provide more context?

tests/unit/constraints/test_base.py
```diff
@@ -756,3 +757,35 @@ def func(x):
...
+        constraint = Mock()
+        constraint.constraint_columns = ['a', 'b']
+        constraint._columns_model = False
+        constraint._sample_constraint_columns.return_value = pd.DataFrame({'a': [0, 1, 2]})
+
+        table_data = pd.DataFrame({'a': [0, 1, 2]})
+
+        # Run and assert
+        with warnings.catch_warnings(record=True) as w:
+            warnings.simplefilter("always")
```
amontanez24, line 788: minor: should we use single quotes here or is this an exception?

amontanez24 (commented):
This looks good but I have a question about the warning message that I'd like to clarify before merging

tests/unit/constraints/test_base.py
katxiao: Oh yes, we should use single quotes. I'll update it

sdv/constraints/base.py
katxiao: Hm that's a good point. I didn't completely understand the use case when I wrote this message.

I can update the comment to say, "When `fit_columns_model` is `False` and we are conditioning on a subset of the constraint columns, reject sampling can become slow."

What do you think?

amontanez24: I think that makes sense

csala (approved):

amontanez24 (approved):
Thanks for addressing! LGTM!

PR #561, opened 2021-08-12
Add codeowners config

katxiao:
Add codeowners config to add `sdv-dev/core-contributors` as owner


csala (approved):
It seems it should be fine. Go ahead!

PR #564, opened 2021-08-18
Fix `CustomConstraint` docs

fealho:


docs/user_guides/single_table/custom_constraints.rst
```diff
@@ -69,6 +69,8 @@ Then, we pack every thing together in ``CustomConstraint``.
 .. ipython:: python
     :okwarning:
 
+    from sdv.constraints import CustomConstraint
```
sarahmish, line 72: good catch! In line 172, we are also missing `from sdv.tabular import GaussianCopula`

sarahmish (approved):

PR #587, opened 2021-09-13
Fix repository language

### LINKED ISSUE #464: Fix repository language
The SDV repository currently shows up as a `Jupyter notebook` repository, which is obviously not the expected.

We should fix this by adding a [`.gitattributes` file that sets `*.ipynb` as `linguist-vendored`](https://github.com/github/linguist/blob/master/docs/overrides.md).

fealho:
Resolves #464.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,1 @@
-*.ipynb -linguist-vendored
+*.ipynb linguist-vendored
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.ipynb -linguist-vendored


pvk-developer (changes requested):
I think that this file has to go inside the `github` folder, as I don't see the changes reflected on the branch.

pvk-developer:
> @pvk-developer I'm not sure if the branch is supposed to show the changes? I created a fork and it worked there.

Yes you are right! There is a delay when that is happening and when I first tried it didn't work because there is a few minutes gap until it reads it properly.

pvk-developer (approved):
LGTM!

Issue #576, opened 2021-08-30
Gaussian Copula is generating different data with metadata and without metadata.


kvrameshreddy:
When we create a Gaussian copula model without defining metadata explicitly, the generated synthetic data holds the properties of the sample data( like min and max values). When we define the same by explicitly passing meta data the generated data is loosing the properties. 
```
from sdv.tabular import GaussianCopula

df1=pd.read_csv("rounding_data_test.csv")
meta_default={'fields': {'COL1': {'type': 'numerical',
                                  'subtype': 'float'} ,
                         'COL2': {'type': 'numerical',
                                   'subtype': 'float'},
                          'COL3': {'type': 'numerical',
                                    'subtype': 'float'},
                          'COL4': {'type': 'numerical',
                                    'subtype': 'float'}}}
model1 = GaussianCopula(table_metadata=meta_default)
model1.fit(df1)
new_data_with_meta = model1.sample(100)

''' without meta data'''

model2 = GaussianCopula()
model2.fit(df1)

new_data_without_meta = model2.sample(100)
```
[rounding_data_test.csv](https://github.com/sdv-dev/SDV/files/7078377/rounding_data_test.csv)

![gen_data_issue](https://user-images.githubusercontent.com/49742709/131344566-c9208a38-dc93-4cb7-ab51-41b7885a8ebd.PNG)
From the above screenshot consider the 'col2' column in the original data(df1) min value is 8.37 and max is 196.8 .
now the generated new_data_without_meta holds the min and max values of the sample but the new_data_with_meta generates negative values. I even verified for the distributions of the data for each column, both the models have same distributions.
![data_dist](https://user-images.githubusercontent.com/49742709/131345234-799c9812-20b7-4ff9-a65b-a2e631a79724.PNG)

The same behaviour is seen with constraints also, when passed directly gives similar data, when passed with explicit metadata gives out of bound values.
can you please look into this issue. 
Thankyou.

PR #599, opened 2021-09-21
Automatically detect min and max when not set in metadata

katxiao:
When we are creating the metadata from a user-provided dict, we should still automatically detect the min and max values.

Resolve issue #576


sdv/metadata/table.py
```diff
@@ -769,6 +769,8 @@ def from_dict(cls, metadata_dict, dtype_transformers=None):
             entity_columns=metadata_dict.get('entity_columns') or [],
             context_columns=metadata_dict.get('context_columns') or [],
             dtype_transformers=dtype_transformers,
+            min_value=metadata_dict.get('min_value') or 'auto',
```
fealho, line 772: Any reason not to do

> metadata_dict.get('min_value', 'auto')

instead of this `or` syntax?

fealho (approved):
👍

sdv/metadata/table.py
```diff
@@ -769,6 +769,8 @@ def from_dict(cls, metadata_dict, dtype_transformers=None):
...
+            max_value=metadata_dict.get('max_value') or 'auto',
```
csala, line 773: Maybe we should also add the `rounding` argument?

katxiao: I did try adding the rounding argument at first, but it seems to crash because of the nans issue (https://github.com/sdv-dev/RDT/pull/244).

We could wait for that PR to be merged first

sdv/metadata/table.py
csala: I agree with @fealho: The `or` syntax makes sense in the cases above because it avoids having to create new `list` or `dict` objects that will end up not being used, but with the `string` this is not a problem

katxiao: Sure, will update

PR #601, opened 2021-09-22
Update column formula is valid to handle nan equality

### LINKED ISSUE #593: ColumnFormula raises ConstraintsNotMetError when checking NaN values
### Environment Details
* SDV version: 0.12.0

### Error Description
I have written a `ColumnFormula` constraint that I know is correct. (I have used it on the real data and confirmed it outputs the same values.) 

However, whenever I apply it to my model, I get an error: `ConstraintsNotMetError: Data is not valid for the given constraints`. Upon closer inspection, I think it's because:

- My `ColumnFormula` will sometimes output `NaN` values (for eg. `[1, 2, NaN, 2, 3, NaN, NaN]`) -- same values exist in my real data
- The SDV is using a simple `==` to check equivalence ([ref](https://github.com/sdv-dev/SDV/blob/469486cbb5b892779707d6c012a3da603ebf9e0d/sdv/constraints/tabular.py#L633)). Since `NaN != NaN`, it thinks my formula's output is invalid.

### Expected Outcome 

I expect the `ColumnFormula` should consider `NaN == NaN` for the purposes of validity.
Can we use some sort of other check here to ensure that the formula passes when it includes `NaNs`?

katxiao:
Currently, if a column formula is expected to produce NaNs, the is_valid method will still flag it as false because nan == nan evaluates to false. We can account for this case by explicitly checking for rows where both the table data and the expected column are NaN.

Resolves #593

### DIFF
--- a/sdv/constraints/tabular.py
+++ b/sdv/constraints/tabular.py
@@ -630,7 +630,9 @@ class ColumnFormula(Constraint):
                 Whether each row is valid.
         """
         computed = self._formula(table_data)
-        return table_data[self._column] == computed
+        isnan = table_data[self._column].isna() & computed.isna()
+
+        return table_data[self._column].eq(computed) | isnan
 
     def _transform(self, table_data):
         """Transform the table data.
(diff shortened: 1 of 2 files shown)


pvk-developer (approved):
LGTM!

csala (approved):
Good to go!

PR #619, opened 2021-10-29
Fixes #616 and #617 enables Unique constraint on custom index columns

### LINKED ISSUE #616: The `Unique` constraint crashes when the data contains a column called `index`
### Environment Details

* SDV version: v0.12.1

### Error Description

The internal implementation of the `Unique` constraint wrongfully creates a and relies on a dummy column called `index`.
As a consequence, if the input data already contains a column called `index`, the `Unique` constraint cannot handle it properly.

### Steps to reproduce

Fit a tabular model which has been configured with a `Unique` constraint using data that contains a column called `index`.

<Replace this text with a description of the steps that anyone can follow to reproduce the error. If the error happens only on a specific dataset, please consider attaching some example data to the issue so that others can use it to reproduce the error.>

```python3
In [1]: import sdv
   ...: 
   ...: data = sdv.demo.load_tabular_demo('student_placements')
   ...: 
   ...: unique = sdv.constraints.Unique(columns=['student_id'])
   ...: 
   ...: data['index'] = [1] * len(data)
   ...: 
   ...: gc = sdv.tabular.GaussianCopula(constraints=[unique])
   ...: gc.fit(data.head(10))
ConstraintsNotMetError                    Traceback (most recent call last)
<ipython-input-1-79c7bc7c5ae7> in <module>
      8 
      9 gc = sdv.tabular.GaussianCopula(constraints=[unique])
---> 10 gc.fit(data.head(10))

/home/ubuntu/.virtualenvs/sdv-test/lib/python3.7/site-packages/sdv/tabular/base.py in fit(self, data)
    136 
    137         LOGGER.debug('Transforming table %s; shape: %s', self._metadata.name, data.shape)
--> 138         transformed = self._metadata.transform(data)
    139 
    140         if self._metadata.get_dtypes(ids=False):

/home/ubuntu/.virtualenvs/sdv-test/lib/python3.7/site-packages/sdv/metadata/table.py in transform(self, data, on_missing_column)
    620         data = self._anonymize(data[fields])
    621 
--> 622         self._validate_data_on_constraints(data)
    623 
    624         LOGGER.debug('Transforming constraints for table %s', self.name)

/home/ubuntu/.virtualenvs/sdv-test/lib/python3.7/site-packages/sdv/metadata/table.py in _validate_data_on_constraints(self, data)
    592             if set(constraint.constraint_columns).issubset(data.columns.values):
    593                 if not constraint.is_valid(data).all():
--> 594                     raise ConstraintsNotMetError('Data is not valid for the given constraints')
    595 
    596     def transform(self, data, on_missing_column='error'):

ConstraintsNotMetError: Data is not valid for the given constraints
```

### LINKED ISSUE #617: The `Unique` constraint cannot handle non-default index
### Environment Details

* SDV version: v0.12.0

### Error Description

The internal implementation of the `Unique` constraint relies on the input data `index`.
Because of this, errors such as #616 and #610  happen, but other weird behaviors can be observed, such as the constraint consider that valid data is invalid:

```python3
In [30]: data = pd.DataFrame({'unique': [1, 2, 3]}, index=[0, 0, 0])

In [31]: unique.is_valid(data)
Out[31]: 
0     True
1    False
2    False
dtype: bool
```

Or producing inconsistent results:

```python3
In [38]: data = pd.DataFrame({'unique': [1, 2, 1]}, index=[0, 2, 1])

In [39]: unique.is_valid(data)
Out[39]: 
0     True
1    False
2     True
dtype: bool
```

We should review its implementation to prevent all these scenarios.

xamm:
# This PR suggests a possible solution to the problems described in #616 and #617.

## Problem description
As described in #616, the "Unique" constraint cannot handle DataFrames when there is a column named "index" in the DataFrame.

This is because the `Unique` constraint relies on the `index` to check if a particular row is unique.

In #617, the problem of inconsistencies is described when a custom `index` is added to the DataFrame and this custom `index` is not sorted or contains unexpected values, such as the same index for all rows.

## PR description
The PR adds tests for these behaviors and changes the implementation of the `unique` constraint.

To check if a row is unique the following is done:
1. The DataFrame is grouped based on the user-defined unique columns.
2. The elements in each group are then numbered.
3. The decision whether a row is unique or the first occurrence of a column combination is based on checking whether the numbered element is 0, i.e. the first occurrence of the group.

Resolves #616
Resolves #617


tests/unit/constraints/test_tabular.py
```diff
@@ -5188,6 +5188,128 @@ def test_is_valid(self):
...
+        """
+        # Setup
+        instance = Unique(columns=['a', 'b', 'c'])
+
+        # Run
+        data = pd.DataFrame({
+            'a': [1, 1, 2, 2, 3],
+            'b': [5, 5, 6, 6, 7],
+            'c': [8, 8, 9, 9, 10]
+        }, index=[2, 1, 3, 5, 4])
```
katxiao, line 5244: It seems like this test and the one above both test the same combination of columns, and also that the original index is preserved. Is it enough just to have one of these tests? Is having the index set to 0 a special case?

tests/integration/tabular/test_base.py
```diff
@@ -94,6 +94,68 @@ def test_conditional_sampling_graceful_reject_sampling_False_dataframe(model):
...
+            "B",
+            "C",
+            "D",
+            "E",
+        ]
+    })
+    unique = Unique(
+        columns=["index"]
+    )
+    err_metadata = {
```
katxiao, line 134: nit: can we rename the variables `error` and `err_metadata` to something like `table` and `table_metadata` since it isn't supposed to error out anymore?

tests/integration/tabular/test_base.py
```diff
@@ -94,6 +94,68 @@ def test_conditional_sampling_graceful_reject_sampling_False_dataframe(model):
...
+        "index": [
+            "A",
+            "B",
+            "C",
+            "D",
+            "E",
+        ]
+    })
+    unique = Unique(
+        columns=["index"]
```
katxiao, line 132: We can also add a test where the table has 3 columns - `key`, `index`, and `test_col`, and the Unique constraint is on `test_col`. This way we can check that it works when the unique constraint is not on the column called "index", but the table contains another column called "index".

tests/integration/tabular/test_base.py
```diff
@@ -94,6 +94,68 @@ def test_conditional_sampling_graceful_reject_sampling_False_dataframe(model):
...
+            "key": {
+                "type": "id", "subtype": "integer"
+            },
+            "index": {
+                "type": "categorical",
+            }
+        }
+    }
+
+    model = GaussianCopula(table_metadata=err_metadata)
```
katxiao, line 148: We can pass in `constraints=[unique]` directly, instead of constructing the metadata.

katxiao (changes requested):

tests/integration/tabular/test_base.py
xamm: Good Idea!

I kept the `primary_key` as a parameter as well and created the unique constraint in a variable.
I think this looks nicer than moving everything to the Model creation line.

```
unique = Unique(columns=["index"])
model = GaussianCopula(primary_key="key", constraints=[unique])
```

tests/integration/tabular/test_base.py
xamm: when removing the metadata specification and specifying the constraint directly during model creation as suggested in https://github.com/sdv-dev/SDV/pull/619#discussion_r748369723, this will also be resolved.

tests/unit/constraints/test_tabular.py
xamm: I think it is a very similar test, since the indices of `pandas` do not have to be unique.
I created the test from the Issues to confirm that this problem is reproducible.

I would probably leave it as is because I like to see problems specified in Issues created as tests.

tests/integration/tabular/test_base.py
xamm: I added a new test for this.

xamm:
The error in the checks occurs due to the new graphviz Version 0.18.1 from the 14th of November.

katxiao:
Thanks for the graphviz fix! It looks good to go, once that's merged into this branch.

katxiao (approved):
lgtm!

csala (approved):
LGTM!

PR #627, opened 2021-11-03
Update metadata tests

### LINKED ISSUE #624: Add scaffolding for Metadata integration tests
### Problem Description

Add scaffolding for `Metadata` integration tests. We should add scaffolding to write integration test that accounts for different `Metadata` use cases (e.g. Faker).

katxiao:
Resolves #624


tests/integration/metadata/test_table.py
```diff
@@ -0,0 +1,67 @@
...
+                        'pii_category': 'company',
+                        'pii_locales': ['en_US'],
+                    },
+                },
+            },
+        }
+    }
+    data = {table_name: pd.DataFrame({'company': [str(i) for i in range(num_values)]})}
+
+    sdv = SDV()
```
csala, line 27: I think that testing through the `SDV` class is not necessary.
For these integration tests I would suggest to work with the `Table` object only, and test that all its functionalities are correct by configuring an instance using a `dict`, calling `fit` on some data, and then calling `transform` and `reverse_transform` on that data, making sure that the output values are correct in each step

tests/integration/metadata/test_table.py
```diff
@@ -0,0 +1,67 @@
+import pandas as pd
+
+from sdv import SDV
+
+
+def test_pii_with_single_localization():
```
csala, line 6: Instead of having multiple test cases, I suggest creating a single test cases that covers multiple things at once by creating a metadata that has:
- Fields that are not `pii`
- Fields that are `pii` without locales
- Fields that are `pii` with one locale
- Fields that are `pii` with multiple locales

And then we validate that each field has been manipulated as expected.

tests/unit/metadata/test_table.py
```diff
@@ -136,124 +136,92 @@ def test__get_faker_method_pass_args(self):
...
-            'fields': {
-                'foo': {
-                    'type': 'categorical',
-                    'pii': True,
-                    'pii_category': 'company',
-                    'pii_locales': 'sv_SE'
-                }
-            }
+        metadata._fields_metadata = {
+            'foo': foo_metadata
```
csala, line 211: I suggest creating the other possible types of fields too, so we validate that the `if` logical checks are correct.
We should at least have:
- A field of type `id`, which should be skipped
- A field which is neither an `id` nor `pii`, which should also be skipped
- A field which is not `id` but is `pii`, for which mappings should be created

tests/unit/metadata/test_table.py
```diff
@@ -136,124 +136,92 @@ def test__get_faker_method_pass_args(self):
...
-        `pii_locales` is specified in the metadata description.
+        The ``_make_anonymization_mappings`` method should map equal values in the original data
+        to the same faked value.
 
         Input:
         - DataFrame with a field that should be anonymized based on the metadata description.
-        Metadata description specifies multiple localizations using `pii_locales`.
         Output:
-        - The mappings created from the original values to localized faked values.
+        - The mappings created from the original values to faked values.
```
csala, line 200: I think this should be marked as a `Side Effect` rather than `Output`, since the method does not really return anything

tests/unit/metadata/test_table.py
```diff
@@ -136,124 +136,92 @@ def test__get_faker_method_pass_args(self):
...
-        data = pd.DataFrame({'foo': list(range(N_VALUES))})
-        metadata = Table.from_dict(metadata_dict)
+        data = pd.DataFrame({
+            'foo': ['test1@example.com', 'test2@example.com', 'test1@example.com']
+        })
 
         # Run
-        metadata._make_anonymization_mappings(data)
+        Table._make_anonymization_mappings(metadata, data)
         foo_mappings = metadata._ANONYMIZATION_MAPPINGS[id(metadata)]['foo']
```
csala, line 219: We should validate that no other mappings have been created:

```python3
mappings = metadata._ANONYMIZATION_MAPPINGS[id(metadata)] 
assert len(mappings) == 1
```

csala (changes requested):

tests/unit/metadata/test_table.py
katxiao: I added checks for other types of fields in the `test__make_anonymization_mappings` test. This `test__make_anonymization_mappings_unique_faked_value_in_field ` just tests if there are repeats of a value in the id field.

csala: Oh! My bad, I missed this. In any case, I like how you laid it out now!

csala (approved):
This looks good to me! I think it's ready to go.

pvk-developer (approved):
LGTM!

PR #629, opened 2021-11-08
Add pip check to workflows

### LINKED ISSUE #626: Add `pip check` to CI workflows
The current CI workflows and local test and lint commands do not catch dependency incompatibilities.

For example, installing the repository for development on the `master` branch results in these errors:

```
ERROR: packaging 21.2 has requirement pyparsing<3,>=2.0.2, but you'll have pyparsing 3.0.4 which is incompatible.
ERROR: deepecho 0.2.1 has requirement torch<1.8,>=1.4, but you'll have torch 1.10.0 which is incompatible.
ERROR: numba 0.54.1 has requirement numpy<1.21,>=1.17, but you'll have numpy 1.21.3 which is incompatible.
ERROR: autopep8 1.6.0 has requirement pycodestyle>=2.8.0, but you'll have pycodestyle 2.7.0 which is incompatible.
```
A pip check command should be made part of the local and CI tests to make sure that our dependency tree is always clean.

pvk-developer:
Resolves #626


Makefile
```diff
@@ -119,7 +119,11 @@ fix-lint: ## fix lint issues using autoflake, autopep8, and isort
 
 .PHONY: test-unit
 test-unit: ## run tests quickly with the default Python
-	invoke pytest
+	invoke unit
+
+.PHONY: test-integration
+test-unit: ## run tests quickly with the default Python
```
csala, line 125: I bet this was intended to be called `test-integration` :-P

csala (approved):
I left a comment about a typo, but other than that this seems good to go

setup.py
```diff
@@ -14,17 +14,18 @@
 install_requires = [
     'Faker>=3.0.0,<10',
     'graphviz>=0.13.2,<1',
-    'numpy>=1.18.0,<2',
+    'numpy>=1.19.3,<2',
```
katxiao, line 17: What is the reason for `numpy>=1.19.3`? So far I have that `sktime` requires `numpy>=1.19.0`

pvk-developer: This is because I removed the `sktime` dependency from SDV: `sktime 0.8.1 has requirement numpy>=1.19.3, but you have numpy 1.19.0.`
If I add ``sktime>=0.6,<1` then yes, I will roll back the change (I removed it because we didn't use it here but I think that we should attempt to keep wider range on numpy.

katxiao (approved):
lgtm after the tests pass

PR #639, opened 2021-11-16
Fix between constraint is valid for nan columns

### LINKED ISSUE #632: ConstraintsNotMetError when applying Between constraint on datetime columns containing null values
### Environment Details

* SDV version: 0.12.1
* Python version: 3.6
* Operating System: ubuntu

### Error Description
I am modeling a relational dataset where some tables contain datetime columns that should be subject to certain range constraints, e.g. start_date should be between two datetimes and end_date should be greater than start_date. I've noticed that the Between constraint produces ConstraintsNotMetError when there are missing values in the datetime columns.

I've checked out-of-bound dates and made sure e.g., start_date.min() > low (scalar) and start_date.max() < high (scalar). I think the problem might be in the is_valid method, which uses operator.lt to check the condition and returns False when either value is NaT.  

### Steps to reproduce

```

```

katxiao:
Resolves #632 

Handle NaNs in Between constraint's `is_valid` method.
- If the constraint column is `NaN`, return True
- If both `low` and `high` are `NaN`, return True
- If constraint column is not `NaN` and only one of `low` or `high` is `NaN`, ignore the `NaN` value when making comparisons.

### DIFF
--- a/sdv/constraints/tabular.py
+++ b/sdv/constraints/tabular.py
@@ -824,14 +824,18 @@ class Between(Constraint):
             pandas.Series:
                 Whether each row is valid.
         """
-        satisfy_low_bound = self._lt(
-            self._get_low_value(table_data), table_data[self.constraint_column]
-        )
-        satisfy_high_bound = self._lt(
-            table_data[self.constraint_column], self._get_high_value(table_data)
+        low = self._get_low_value(table_data)
+        high = self._get_high_value(table_data)
+
+        satisfy_low_bound = self._lt(low, table_data[self.constraint_column])
+        satisfy_high_bound = self._lt(table_data[self.constraint_column], high)
+
+        isnull = np.logical_or(
+            np.isnan(table_data[self.constraint_column]),
+            np.logical_or(np.isnan(low), np.isnan(high))
         )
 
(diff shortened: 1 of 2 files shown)


amontanez24 (approved):
LGTM!

sdv/constraints/tabular.py
```diff
@@ -824,14 +831,22 @@ def is_valid(self, table_data):
...
         )
-        satisfy_high_bound = self._lt(
-            table_data[self.constraint_column], self._get_high_value(table_data)
+        satisfy_high_bound = np.logical_or(
+            self._lt(table_data[self.constraint_column], high),
+            np.isnan(high),
         )
 
-        return satisfy_low_bound & satisfy_high_bound
+        return np.logical_or(
```
csala, line 846: Pure curiosity: It seems like we are mixing the `np.logical_or` with `&`. Maybe we could use `np.logical_and` too, to make it uniform?

csala (approved):
LGTM!
I just added a very minor question, but I think this can go in.

sdv/constraints/tabular.py
katxiao: Yes, that's a good point, I can change it to be consistent.

PR #641, opened 2021-11-18
Adding support for Python 3.9

### LINKED ISSUE #631: Add support for Python 3.9
### Problem Description

SDV should support Python 3.9

amontanez24:
Resolves #631


sdv/tabular/base.py
```diff
@@ -234,7 +234,6 @@ def _sample_rows(self, num_rows, conditions=None, transformed_conditions=None,
                     sampled = self._sample(num_rows, transformed_conditions)
                 except NotImplementedError:
                     sampled = self._sample(num_rows)
-
```
pvk-developer, line 237: Maybe this was deleted by mistake, but I think it should stay there.

pvk-developer (changes requested):
Looks good and it works as expected.

sdv/metadata/table.py
```diff
@@ -661,7 +661,10 @@ def transform(self, data, on_missing_column='error'):
         data = self._transform_constraints(data, on_missing_column)
 
         LOGGER.debug('Transforming table %s', self.name)
-        return self._hyper_transformer.transform(data)
+        try:
+            return self._hyper_transformer.transform(data)
+        except rdt.errors.NotFittedError:
+            return data
```
csala, line 667: Why do we return `data` if `NotFitted`? Should we not let the error be raised?

sdv/metadata/table.py
```diff
@@ -693,7 +696,10 @@ def reverse_transform(self, data):
         if not self.fitted:
             raise MetadataNotFittedError()
 
-        reversed_data = self._hyper_transformer.reverse_transform(data)
+        try:
+            reversed_data = self._hyper_transformer.reverse_transform(data)
+        except rdt.errors.NotFittedError:
+            reversed_data = data
```
csala, line 702: Same comment as above

sdv/constraints/tabular.py
```diff
@@ -1028,11 +1029,18 @@ def _sample_constraint_columns(self, table_data):
                         proposed_table_data[column] = 1.0
 
             else:
+                should_transform = True
+                conditions = sub_table_data[condition_columns]
+                transformed_conditions = self._hyper_transformer.transform(conditions)
                 proposed_table_data = self._columns_model.sample(
                     num_rows=sub_table_data[condition_columns].shape[0],
```
csala, line 1036: Isn't this just `num_rows=len(sub_table_data)`?

sdv/constraints/tabular.py
```diff
@@ -1020,6 +1020,7 @@ def _sample_constraint_columns(self, table_data):
         has_one = table_data[condition_columns].sum(axis=1) == 1.0
```
csala, line 1020: Not really part of this PR, but we could think about doing the fix here:
It seems like we are accessing `table_data[condition_columns]` several times, and also computing `sum(axis=1)` on them at least twice. Could we capture these in variables?

sdv/constraints/tabular.py
```diff
@@ -1028,11 +1029,18 @@ def _sample_constraint_columns(self, table_data):
                         proposed_table_data[column] = 1.0
 
             else:
+                should_transform = True
+                conditions = sub_table_data[condition_columns]
+                transformed_conditions = self._hyper_transformer.transform(conditions)
```
csala, line 1034: It seems like this introduces a `transform` that was not originally done, so the behavior is being changed. Why is this done?

sdv/metadata/table.py
amontanez24: This happens when data that is all primary keys gets passed. The `HyperTransformer` raises a not fitted error even though it was fit because the `_transformers_sequence` is empty

sdv/constraints/tabular.py
amontanez24: It should have been transformed before. It wasn't because the values are numeric and the column names were the same before so it didn't matter. Now that the column names are different, the model doesn't know the conditions' original names, and they should be transformed.

sdv/metadata/table.py
csala: I see. So it seems to rather be a bug on RDT, right?
I think the change is fine, though, because this code is protected by the `if not self.fitted` above, meaning that this `rdt.errors.NotFitted` will only be raised in that scenario that you described.

amontanez24: Yeah, I think we will have to fix this in RDT

csala (approved):
After resolving some doubts, this looks good to go.
@amontanez24 it would be great if we can add the small fixes I commented that are unrelated to the PR, but it's also OK to merge as it is and open an issue to come back and optimize this some other time.

sdv/constraints/tabular.py
amontanez24: Hmm I think so, Idk why this is like that

pvk-developer (approved):
:shipit:

PR #649, opened 2021-11-23
Add py39 to github workflows

katxiao:
issue #631


amontanez24:
While you're doing this, can we add `3.9` here as well? https://github.com/sdv-dev/SDV/blob/08bed6d8fde6d5348162ba8e29bf27d3e2927d37/setup.py#L89-L93

I missed this in the original PR

amontanez24 (approved):
LGTM! Thanks for fixing

PR #671, opened 2021-12-17
Update metadata doc

### LINKED ISSUE #664: Make true/false values lowercase in Metadata Schema specification
Boolean values in JSON are lowercase (`true`/`false`) even though they are uppercase in Python (`True`/`False`).

The [User Guide](https://sdv.dev/SDV/developer_guides/sdv/metadata.html#categorical-fields-data-anonymization) for the Metadata Schema says it's describing JSON files at the top. However, it's including an uppercase boolean value as an example.

```
"social_security_number": {
    "type": "categorical",
    "pii": True,
    "pii_category": "ssn"
}
```

This might confuse users if they're following the guide to write their `metadata.json`.

katxiao:
Resolves #664 

Additionally, update hma1's `_load_table` method to try reading a table from metadata if the `table_name` does not exist in the `tables` map.

### DIFF
--- a/docs/developer_guides/sdv/metadata.rst
+++ b/docs/developer_guides/sdv/metadata.rst
@@ -130,7 +130,7 @@ the following keys.
             "fields": {
                 "social_security_number": {
                     "type": "categorical",
-                    "pii": True,
+                    "pii": true,
                     "pii_category": "ssn"
                 },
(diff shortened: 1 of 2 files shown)


amontanez24 (approved):
LGTM!

PR #698, opened 2022-01-27
Fix conditional sampling with custom constraint

### LINKED ISSUE #696: Conditional sampling doesn't work if the model has a `CustomConstraint`
### Environment Details
* SDV version: `0.13.1`
* Python version: `3.7`

### Error Description

Any time you create a tabular model using a [`CustomConstraint`](https://sdv.dev/SDV/user_guides/single_table/custom_constraints.html), you cannot sample it with conditions. (Sampling entire rows -- without conditions -- works.)

Currently, we only using reject sampling, so it should always be possible to sample an entire row and then filter out invalid conditions.

### Steps to reproduce

Follow the code in issue #560 .

### Current Workaround

It's possible to sample entire rows without constraints. So the workaround is to call `sample()`, and then manually delete any invalid rows that are returned. _Note: It may take more tries to get the desired output, so you may want to try sampling a greater number of rows that you want._

### Other notes

- Until we figure out how to fix this, it would at least be nice to give a better error message to users (filed in #694 )
- Related user issues: #560 , #687

katxiao:
If we use conditional sampling and a custom constraint together, we currently get a KeyError from the constraint's transform method. This is expected when we transform the conditions, as it is not the complete table data.

Update the custom constraint to return a `MissingConstraintColumnError` in this case, so that the error can be handled properly.  

Resolves #696

### DIFF
--- a/sdv/constraints/tabular.py
+++ b/sdv/constraints/tabular.py
@@ -29,6 +29,7 @@ import numpy as np
 import pandas as pd
 
 from sdv.constraints.base import Constraint, import_object
+from sdv.constraints.errors import MissingConstraintColumnError
 from sdv.constraints.utils import is_datetime_type
 
 
@@ -51,6 +52,10 @@ class CustomConstraint(Constraint):
     def _run(self, function, table_data, reverse=False):
         table_data = table_data.copy()
         if self._columns:
+            missing_columns = [col for col in self._columns if col not in table_data.columns]
+            if missing_columns:
+                raise MissingConstraintColumnError()
+
             if reverse:
                 columns = reversed(self._columns)
             else:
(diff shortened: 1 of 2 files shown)


sdv/constraints/tabular.py
```diff
@@ -68,6 +69,11 @@ def _run(self, function, table_data, reverse=False):
         return table_data
 
     def _run_transform(self, table_data):
+        if self._columns:
+            missing_columns = [col for col in self._columns if col not in table_data.columns]
```
amontanez24, line 73: minor: since we're not using any of the columns we could do 
```
if any(col not in table_data.columns for col in self._columns)
```
so that it terminates early.

amontanez24 (approved):
LGTM!

sdv/constraints/tabular.py
katxiao: Sounds good, made the change

csala (approved):

PR #705, opened 2022-02-09
Refactor sample API (1/3)

### LINKED ISSUE #690: Improve `sample()` method
### Problem Description

Let's make sampling more user friendly. We can create multiple methods for different user needs.

The `sample()` method should only be used to sample entire rows (no conditions).

### Expected behavior

Parameters:
- `num_rows` is a required param
- `randomize_samples` will determine whether or not there should be a fixed seed (default: True)
- Remove all existing params

```python
# works with any tabular model
from sdv.tabular import CTGAN
model = CTGAN()
model.fit(data)

# creates 1000 rows. Different rows created every time
model.sample(num_rows=1000)

# Fix a seed so that this call always creates rows in the same order
model.sample(num_rows=1000, randomize_samples=False)

```

## Error Handling

Guidance if the user does not specify `num_rows`.
```python
>>> model.sample()
Error: You must specify the number of rows to sample (eg. num_rows=100).
```

Guidance if the user passes in the `conditions` argument.

```python
>>> model.sample(num_rows=100, conditions={'age': 45})
Error: This method does not support the conditions parameter. Please create Condition objects
and pass them into the sample_conditions method. See User Guide or API for more details.
```

katxiao:
Resolves #690

### DIFF
--- a/sdv/tabular/base.py
+++ b/sdv/tabular/base.py
@@ -3,6 +3,7 @@
 import logging
 import pickle
 import uuid
+from collections import defaultdict
 from warnings import warn
 
 import numpy as np
@@ -318,8 +319,8 @@ class BaseTabularModel:
 
         return sampled.head(min(len(sampled), num_rows))
 
-    def _make_conditions_df(self, conditions):
-        """Transform `conditions` into a dataframe.
+    def _make_condition_dfs(self, conditions):
+        """Transform `conditions` into a list of dataframes.
 
         Args:
             conditions (list[sdv.sampling.Condition]):
@@ -328,18 +329,25 @@ class BaseTabularModel:
                 to generate for that condition.
 
         Returns:
-            pandas.DataFrame:
-                `conditions` as a dataframe.
+            list[pandas.DataFrame]:
+                A list of `conditions` as dataframes.
         """
-        condition_dataframes = []
+        condition_dataframes = defaultdict(list)
         for condition in conditions:
-            condition_dataframes.append(
+            column_values = condition.get_column_values()
+            condition_dataframes[tuple(column_values.keys())].append(
                 pd.DataFrame(
                     condition.get_column_values(),
                     index=range(condition.get_num_rows())
                 )
             )
-        return pd.concat(condition_dataframes, ignore_index=True)
+
+        return [
+            pd.concat(
+                condition_list,
+                ignore_index=True,
+            ) for condition_list in condition_dataframes.values()
+        ]
 
     def _conditionally_sample_rows(self, dataframe, condition, transformed_condition,
                                    max_tries=None, batch_size_per_try=None, float_rtol=None,
@@ -439,7 +447,7 @@ class BaseTabularModel:
         try:
             transformed_conditions = self._metadata.transform(conditions, on_missing_column='drop')
         except ConstraintsNotMetError as cnme:
-            cnme.message = 'Passed conditions are not valid for the given constraints'
+            cnme.message = 'Provided conditions are not valid for the given constraints'
             raise
 
         condition_columns = list(conditions.columns)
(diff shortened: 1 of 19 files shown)


sdv/tabular/base.py
```diff
@@ -390,126 +411,33 @@ def _conditionally_sample_rows(self, dataframe, max_retries, max_rows_multiplier
 
         return sampled_rows
 
-    def sample(self, num_rows=None, max_retries=100, max_rows_multiplier=10,
-               conditions=None, float_rtol=0.01, graceful_reject_sampling=False):
+    def _set_fixed_seed(self, randomize_samples):
+        if randomize_samples:
+            # TODO: set random state on copulas.
```
fealho, line 416: Todo

tests/unit/tabular/test_base.py
```diff
@@ -108,6 +56,88 @@ def test__sample_batch_zero_valid(self):
...
+    def test_sample_no_num_rows(self):
+        """Test the `BaseTabularModel.sample` method with no `num_rows` input.
+
+        Expect that an error is thrown.
+        """
+        # Setup
+        model = BaseTabularModel()
+
+        # Run and assert
+        with pytest.raises(TypeError):
```
fealho, line 96: check that the error messages match.

sdv/tabular/base.py
```diff
@@ -390,126 +411,33 @@ def _conditionally_sample_rows(self, dataframe, max_retries, max_rows_multiplier
...
-        all_sampled_rows = pd.concat(all_sampled_rows)
-        all_sampled_rows = all_sampled_rows.set_index(COND_IDX)
-        all_sampled_rows.index.name = conditions.index.name
-        all_sampled_rows = all_sampled_rows.sort_index()
-        all_sampled_rows = self._metadata.make_ids_unique(all_sampled_rows)
-
-        return all_sampled_rows
+        if num_rows is None:
+            raise ValueError(
+                'Error: You must specify the number of rows to sample (eg. num_rows=100).')
```
fealho, line 436: Remove the `Error`, yes? Otherwise it becomes `ValueError: Error: You must...`

sdv/tabular/base.py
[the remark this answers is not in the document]
katxiao: I'll remove all the `randomize_samples` implementation and tests. It was meant to go into this PR, but it ended up being more work than expected, so I'll add it in a follow-up PR.

tests/integration/timeseries/test_par.py
```diff
@@ -62,7 +62,7 @@ def test_column_after_date_simple():
 
     model = PAR(entity_columns=['col'], sequence_index='date', epochs=1)
     model.fit(data)
-    sampled = model.sample()
+    sampled = model.sample(1)
```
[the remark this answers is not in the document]
katxiao: I believe it has something to do with the number of sequences, and how PAR generates sequences after the usual sample process.

fealho (approved):

sdv/tabular/base.py
```diff
@@ -8,11 +8,32 @@
 import numpy as np
 import pandas as pd
 
-from sdv.errors import ConstraintsNotMetError
 from sdv.metadata import Table
 
 LOGGER = logging.getLogger(__name__)
 COND_IDX = str(uuid.uuid4())
+FIXED_SEED = 21321
```
csala, line 15: This does not seem to be used anywhere

sdv/tabular/base.py
```diff
@@ -8,11 +8,32 @@
...
+
+
+def validate_sample_args(function):
```
csala, line 18: Using a decorator for this seems overkill, since this is being used only in one place. On the other hand, using a decorator as it is done here would shadow the builtin `help(model.sample)` (or `model.sample?` in ipython) functionality by showing the `wrapper` docstring (so, nothing) help instead of the actual `sample` docstring.

I think the right way to do this is to keep the argument in the `sample` method for some time, stating that the argument is deprecated in its description, and validate the value right there, inside the `sample` method.

tests/integration/tabular/test_copulagan.py
```diff
@@ -82,82 +82,82 @@ def test_recreate():
...
-    model.fit(data)
-    conditions = {
-        "column1": 1.0,
-    }
-    sampled = model.sample(5, conditions=conditions)
-
-    assert list(sampled.column1) == [1.0] * 5
-
-
+#def test_conditional_sampling_dict():
```
csala, line 92: Why are all these tests commented out?

csala (changes requested):

sdv/tabular/base.py
katxiao: I see. Sure I can change it to keeping the argument in sample.

csala (approved):

PR #713, opened 2022-02-18
Fix seed to generate deterministic samples

### LINKED ISSUE #690: Improve `sample()` method

katxiao:
Resolves #690


setup.py
```diff
@@ -18,8 +18,8 @@
     "numpy>=1.20.0,<2;python_version>='3.7'",
     'pandas>=1.1.3,<2',
     'tqdm>=4.15,<5',
-    'copulas>=0.6.0,<0.7',
-    'ctgan>=0.5.0,<0.6',
+    'copulas>=0.6.1.dev0,<0.7',
```
katxiao, line 21: I'll update these version numbers once `copulas` and `ctgan` are released.

sdv/tabular/base.py
```diff
@@ -390,6 +391,20 @@ def _conditionally_sample_rows(self, dataframe, max_retries, max_rows_multiplier
 
         return sampled_rows
 
+    def _randomize_samples(self, randomize_samples):
```
fealho, line 394: If it's easy, I'd add a test case for this as well.

sdv/tabular/copulas.py
```diff
@@ -318,6 +318,15 @@ def _sample(self, num_rows, conditions=None):
         """
         return self._model.sample(num_rows, conditions=conditions)
 
+    def _set_random_state(self, random_state):
+        """Set the random state of the model's random number generator.
+
+        Args:
+            random_state (int, np.random.RandomState, or None):
```
fealho, line 325: What's the expected behavior if it's passed None? Maybe add a test for it, if the behavior is any different than passing an integer.

fealho (approved):
Looks good to me, just some small details.

sdv/tabular/copulas.py
katxiao: It should set the underlying model's random state to None. Not sure if we should test it here since it's the internal model behavior (e.g. within CTGAN and copulas)

sdv/tabular/ctgan.py
```diff
@@ -79,6 +79,15 @@ def _sample(self, num_rows, conditions=None):
 
         raise NotImplementedError(f"{self._MODEL_CLASS} doesn't support conditional sampling.")
 
+    def _set_random_state(self, random_state):
+        """Set the random state of the model's random number generator.
+
+        Args:
+            random_state (int, np.random.RandomState, or None):
```
csala, line 86: Wouldn't this be `tuple[np.random.RandomState, torch.Generator]` instead of just `np.random.RandomState`?

sdv/tabular/base.py
```diff
@@ -390,6 +391,19 @@ def _conditionally_sample_rows(self, dataframe, max_retries, max_rows_multiplier
...
+        """Randomize the samples according to user input.
+
+        If ``randomize_samples`` is false, fix the seed that the random number generator
+        uses in the underlying models.
+
+        Args:
+            randomize_samples (bool):
+                Whether or not to randomize the generated samples.
+        """
+        if randomize_samples:
```
csala, line 404: Shouldn't this be `if not`?

sdv/tabular/base.py
```diff
@@ -390,6 +391,19 @@ def _conditionally_sample_rows(self, dataframe, max_retries, max_rows_multiplier
 
         return sampled_rows
 
+    def _randomize_samples(self, randomize_samples):
```
csala, line 394: I don't see this used anywhere apart from the unit tests.
On the other hand, what happens if `randomize_samples` is `True`?
Should we set the seed to None?

katxiao: @csala Yes, if `randomize_samples` is true, we should set the seed to `None`. I updated the code to call the `_randomize_samples` method, and I also added an integration test to test that setting the seed to None works as expected.

PR #720, opened 2022-03-03
Create sdv presets

### LINKED ISSUE #715: Create SDV presets
### Problem Description

With so many configuration options available in the SDV, let's create a module where we can deliver specific models based on the user goal.

### Expected behavior

- Create an sdv lite module
- In that module, create a TabularPreset object to be used for single table models

Users should be able to import:

```python
from sdv.lite import TabularPreset
```

### LINKED ISSUE #716: Create a speed optimized Preset
### Problem Description

After adding the framework in #715, we should add a configuration that allows the user to optimize a tabular model for speed.

### Expected behavior

A speed optimized tabular model will:
- Use the `GaussianCopula`
- Not round numerical data
- Enforce min/max values on numerical data
- Use the `CategoricalTransformer` (to be renamed: `FrequencyEncoder`) on any categorical columns
- Do not model null values. Instead, randomly generate null values based on observed proportions
- Default all univariate distributions to `gaussian` instead of performing a search

### API
The `TabularPreset` will accept 2 parameters:
- (required) `optimize_for`, the name of the preset. We'll add `"SPEED"`
- `metadata` a dictionary or `TableMetadata` object, similar to tabular models today

It will print out hard-coded info about the configuration & benchmarks.

```python
>>> from sdv.lite import TabularPreset

# the PresetModel will not allow too many extra settings, making it simple to use
>>> model = TabularPreset(optimize_for='SPEED', metadata=my_metadata)
Info: This config optimizes the modeling speed above all else.

Your exact runtime is dependent on the data. Benchmarks:
100K rows and 100 columns may take around 1 minute.
1M rows and 250 columns may take around 30 minutes.

# these work like any other tabular model
>>> model.fit(my_data)
>>> model.sample(my_data)
```

### Edge Cases
If the user does not pass in metadata, we should still run the model, but we should provide a warning:

```python
>>> from sdv.lite import TabularPreset
>>> model = TabularPreset(optimize_for='SPEED')
Warning: No metadata provided. Metadata will be automatically detected from your data. This process may
not be accurate. We recommend writing metadata to ensure correct data handling.

Info: This config optimizes the modeling speed above all else.

Your exact runtime is dependent on the data. Benchmarks:
100K rows and 100 columns may take around 1 minute.
1M rows and 250 columns may take around 30 minutes.
```

If the user does not pass in the required optimization parameter, throw an error.
```python
>>> from sdv.lite import TabularPreset
>>> model = TabularPreset()
Error: You must provide the name of a preset using the 'optimize_for' parameter. Use 
'TabularPreset.list_available_presets()' to browse through the options.
```

### LINKED ISSUE #717: Create a method to browse presets
### Problem Description

When we add the `"SPEED"` preset in #716, it will be the first of multiple. We should allow users a way to browse presets.

### Expected behavior

Create a class function `list_available_presets` that prints out the available presets. See #715.

```python
>>> from sdv.lite import TabularPreset
>>> TabularPreset.list_available_presets()
Available presets:
    "SPEED": Use this preset to minimize the time needed to create a synthetic data model

Supply the preset using the 'opimize_for' parameter.

Have any requests for custom presets? Contact the SDV team to learn more an SDV Premium license.
```

katxiao:
Resolves #715 
Resolves #716 
Resolves #717

### DIFF
--- a/sdv/lite/tabular.py
+++ b/sdv/lite/tabular.py
@@ -1,6 +1,7 @@
 """Base class for tabular model presets."""
 
 import logging
+import warnings
 
 import copulas
 import rdt
@@ -34,21 +35,21 @@ class TabularPreset:
         if optimize_for not in PRESETS:
             raise ValueError(f'`optimize_for` must be one of {PRESETS}.')
         if metadata is None:
-            LOGGER.info('Warning: No metadata provided. Metadata will be automatically '
-                        'detected from your data. This process may not be accurate. '
-                        'We recommend writing metadata to ensure correct data handling.')
+            warnings.warn('No metadata provided. Metadata will be automatically '
+                          'detected from your data. This process may not be accurate. '
+                          'We recommend writing metadata to ensure correct data handling.')
 
         if optimize_for == SPEED_PRESET:
             self._model = GaussianCopula(
                 table_metadata=metadata,
                 field_transformers={'categorical': rdt.transformers.CategoricalTransformer},
-                model_kwargs={'default_distribution': copulas.univariate.GaussianUnivariate},
+                default_distribution=copulas.univariate.GaussianUnivariate,
                 rounding=None,
             )
-            LOGGER.info('This config optimizes the modeling speed above all else.\n\n'
-                        'Your exact runtime is dependent on the data. Benchmarks:\n'
-                        '100K rows and 100 columns may take around 1 minute.\n'
-                        '1M rows and 250 columns may take around 30 minutes.')
+            print('This config optimizes the modeling speed above all else.\n\n'
+                  'Your exact runtime is dependent on the data. Benchmarks:\n'
+                  '100K rows and 100 columns may take around 1 minute.\n'
+                  '1M rows and 250 columns may take around 30 minutes.')
 
     def fit(self, data):
         """Fit this model to the data."""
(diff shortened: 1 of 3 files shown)


sdv/lite/tabular.py
```diff
@@ -0,0 +1,69 @@
...
+    def fit(self, data):
+        """Fit this model to the data."""
+        self._model.fit(data)
+
+    def sample(self, num_rows):
+        """Sample rows from this table."""
+        return self._model.sample(num_rows)
+
+    @classmethod
+    def list_available_presets(cls):
```
amontanez24, line 64: Can we add unit tests for this and make sure it prints properly in the scenario where there are multiple presets?

amontanez24 (changes requested):
This looks good. One question I have is why not make it a child class of the tabular model that sets these parameters based on the preset selected during the `__init__`. If this were the case, it would also be able to use functions like `sample_conditions` or anything else the tabular model offers

katxiao:
> This looks good. One question I have is why not make it a child class of the tabular model that sets these parameters based on the preset selected during the `__init__`? If this were the case, it would also be able to use functions like `sample_conditions` or anything else the tabular model offers

Yeah that's a good question. We also want to add logic that wraps around the tabular model, such as some data pre- and post- processing. I thought it would be more straightforward to wrap around the Tabular model itself.

amontanez24 (approved):
:shipit:

PR #721, opened 2022-03-03
Fix dependencies for windows

katxiao:
Fix `pywinpty` installation for windows with python 3.6.

Also fix flaky tutorial.


pvk-developer (approved):
LGTM :shipit:

PR #724, opened 2022-03-04
Save dependency versions in model.pkl

### LINKED ISSUE #722: Check for library versions when loading a previously saved model
### Problem Description

While we're still in pre-alpha, parts of our API may change. If a user previously saved a model and tries to load it with a newer version, it may be incompatible.

### Expected behavior

1. When saving a model, we should store the relevant SDV libraries & their versions (eg. in a dictionary)
_Note: We only need to save the relevant libraries. Eg. for the `GaussianCopula` model, we don't have to save the version of `DeepEcho`, `SDGym`, etc._

```python
# this should work for any model: GaussianCopula, CopulaGAN, CTGAN, TVAE, HMA1, PAR, etc.
# though the exact set libraries we save will be different per model
>>> model.save('model.pkl')
```

2.  When loading the previously-saved model, we should check against this dictionary and warn the user if there are mismatches. The warning should include info about previous vs. current versions so that they can appropriately downgrade (or upgrade).

```python
# should work for any model, though the exact libraries checked will be different
>>> from sdv.tabular import GaussianCopula
>>> GaussianCopula.load('model.pkl')
Warning: The libraries used to create the model have different versions than your
current setup. This may cause errors when sampling.

copulas used version 0.6.0; current version is 0.6.1
rdt used version 0.6.2; current version is 0.6.3

```

### Edge Cases

The user may have created the model so long ago that it _doesn't_ have the dictionary of libraries & versions saved. In this case, print a generic warning.

```python
# should work for any model, though the exact libraries checked will be different
>>> from sdv.tabular import CopulaGAN
>>> CopulaGAN.load('old_model.pkl')
Warning: The libraries used to create the model have older versions than your
current setup. This may cause errors when sampling.
```

katxiao:
Resolves #722

### DIFF
--- a/sdv/sdv.py
+++ b/sdv/sdv.py
@@ -8,6 +8,7 @@ import warnings
 from sdv.errors import NotFittedError
 from sdv.relational.hma import HMA1
 from sdv.tabular.copulas import GaussianCopula
+from sdv.utils import generate_version_mismatch_warning, get_package_versions
 
 
 class SDV:
@@ -147,6 +148,8 @@ class SDV:
             path (str):
                 Path where the SDV instance will be serialized.
         """
+        self._package_versions = get_package_versions(getattr(self, '_model', None))
+
         with open(path, 'wb') as output:
             pickle.dump(self, output)
 
@@ -159,4 +162,10 @@ class SDV:
                 Path from which to load the SDV instance.
         """
         with open(path, 'rb') as f:
-            return pickle.load(f)
+            model = pickle.load(f)
+            warning_str = generate_version_mismatch_warning(
+                getattr(model, '_package_versions', None))
+            if warning_str:
+                warnings.warn(warning_str)
+
+            return model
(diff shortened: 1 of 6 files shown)


sdv/sdv.py
```diff
@@ -159,4 +162,10 @@ def load(cls, path):
                 Path from which to load the SDV instance.
         """
         with open(path, 'rb') as f:
-            return pickle.load(f)
+            model = pickle.load(f)
+            warning_str = generate_version_mismatch_warning(
+                getattr(model, '_package_versions', None))
+            if warning_str:
+                warnings.warn(warning_str)
```
amontanez24, line 169: Is there any reason not to move this whole chunk to utils and just call a method that raises the warning if necessary?

katxiao: I like to have the warning originating from the actual method, not a utils method. That way it's more obvious if a method raises a warning / error.

amontanez24 (approved):
LGTM :shipit:

PR #728, opened 2022-03-07
Fixes for sampling update

katxiao:
Fix the following sampling issues
- Create the output file at the beginning of sampling run, regardless of if there are successfully sampled rows
- Add a way to disable writing to output file for internal use
- Only display the progress bar if there are multiple batches or multiple conditions & retries
- Update error messages and error handling


sdv/tabular/utils.py
```diff
@@ -133,3 +134,47 @@ def unflatten_dict(flat):
...
+        output_file_path (str):
+            The output file path.
+        sampling_error:
+            The error to raise.
+
+    Side Effects:
+        The error will be raised.
+    """
+    if is_tmp_file:
+        print('Error: Sampling terminated. Partial results are stored in a temporary file: '
```
amontanez24, line 173: Why do we print instead of making it part of the error message?

sdv/tabular/base.py
```diff
@@ -712,16 +739,21 @@ def sample_remaining_columns(self, known_columns, max_tries=100, batch_size_per_
...
-        except (Exception, KeyboardInterrupt) as e:
-            print('Error: Sampling terminated. Partial results are stored in a temporary file: '
-                  f'{output_file_path}. This file will be overridden the next time you sample. '
-                  'Please rename the file if you wish to save these results.')
-            raise e
+            if len(known_columns) == 1 and max_tries == 1:
+                sampled = _sample_function()
+            else:
+                sampled = progress_bar_wrapper(
+                    _sample_function, len(known_columns), 'Sampling remaining columns')
```
amontanez24, line 753: Can we add some unit tests to make sure this logic of only making the progress bar under certain conditions is working?

sdv/tabular/utils.py
```diff
@@ -133,3 +134,47 @@ def unflatten_dict(flat):
             unflattened[key] = value
 
     return unflattened
+
+
+def progress_bar_wrapper(function, pb_total, pb_description):
```
amontanez24, line 139: Can we add some very basic unit tests for these util functions?

amontanez24 (commented):
Looks good! Just a question and some comments about testing

sdv/tabular/utils.py
katxiao: We wanted to make it print before the actual error is thrown, versus it being a stacked error.

sdv/tabular/base.py
katxiao: Added asserts to the existing unit tests!

amontanez24 (approved):

PR #730, opened 2022-03-08
Update conditional sampling errors

katxiao:
Throw warnings when we cannot create any rows in a condition instead of crashing with an error.
Throw a ValueError only when we cannot sample ANY rows from any of the conditions

### DIFF
--- a/sdv/tabular/base.py
+++ b/sdv/tabular/base.py
@@ -396,29 +396,28 @@ class BaseTabularModel:
         if num_sampled_rows < num_rows:
             # Didn't get enough rows.
             if len(sampled_rows) == 0:
-                error = ('Unable to sample any rows for the given conditions: '
-                         f'`{transformed_condition}`. '
-                         f'Try increasing `max_tries` (currently: {max_tries}) or increasing '
-                         f'`batch_size_per_try` (currently: {batch_size_per_try}). Note that '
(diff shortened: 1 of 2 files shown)


sdv/tabular/base.py
```diff
@@ -396,29 +396,28 @@ def _conditionally_sample_rows(self, dataframe, condition, transformed_condition
...
-                         f'`{transformed_condition}`. '
-                         f'Try increasing `max_tries` (currently: {max_tries}) or increasing '
-                         f'`batch_size_per_try` (currently: {batch_size_per_try}). Note that '
-                         'increasing these values will also increase the sampling time.')
-                raise ValueError(error)
-
-            elif not graceful_reject_sampling:
-                error = f'Could not get enough valid rows within {max_tries} trials.'
-                raise ValueError(error)
+                user_msg = ('Unable to sample any rows for the given conditions: '
```
amontanez24, line 399: can we move this into the next if since it seems like it's only used if that condition is met

amontanez24 (approved):
Just one comment then :shipit:

sdv/tabular/base.py
katxiao: It's used as a warning if `graceful_reject_sampling=True`

PR #731, opened 2022-03-08
Update Gaussian copula conditional sampling API

### LINKED ISSUE #729: Updates to `GaussianCopula` conditional sampling methods
### Problem Description

When conditional sampling, the `GaussianCopula` model directly synthesizes the values without reject sampling. We should update the API so that it does not make reference to reject sampling for this model only. _Note: All other tabular models use reject sampling, so will have a slightly different API._ 

### Expected behavior

The following applies to both `sample_conditions` and `sample_remaining_columns`.
(See #691 and #692 )

- Remove the `batch_size_per_try` and `max_tries` arguments
- Add `batch_size` argument. This will work exactly like in `sample`
- Update the error and warning messages

```python
model = GaussianCopula()
model.fit(data)

# see Issue #689 for Condition object details
from sdv.tabular.sampling import Conditions
female_users = Condition(column_values={'sex': 'F', 'active_user': True}, num_rows=50)
inactive_users = Condition(column_values={'active_user': False}, num_rows=100)
conditions = [female_users, inactive_users]

# pass in list of conditions
model.sample_conditions(conditions, batch_size=50, randomize_sample=False)

# or pass in a partial dataframe
model.sample_remaining_conditions(known_columns=partial_dataframe, batch_size=100)
``` 

### Edge Cases

Scenario 1: Unable to create any rows: Either bc the input is out-of-bounds or due to mathematical reasons (eg. matrix inversion). **Reject sampling is not a reason why this method can fail.**

```python
>>> synthetic_data = model.sample_remaining_columns(known_columns=difficult_dataframe)
Error: Unable to sample any rows. This may be because the provided values are out-of-bounds in the current model.
Please try again with a different set of values.
```

Scenario 2: Invalid input -- Same as #691 and #692 
```python
# Unexpected column name
>>> c = Condition(column_values={'New_Column': 42})
>>> synthetic_data = model.sample_conditions([c])
Error: Unexpected column name 'New_Column'. Use a column name that was present in the original data.
```

katxiao:
Resolves #729 

Since GaussianCopula can do conditional sampling without reject sampling, we can update its conditional sampling methods to remove the reject sampling arguments (`max_tries`, and `batch_size_per_try`). Also update the warning message for GaussianCopula

### DIFF
--- a/sdv/tabular/base.py
+++ b/sdv/tabular/base.py
@@ -9,6 +9,7 @@ import uuid
 import warnings
 from collections import defaultdict
 
+import copulas
 import numpy as np
 import pandas as pd
 import tqdm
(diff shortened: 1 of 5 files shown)


pvk-developer (approved):
LGTM!

PR #732, opened 2022-03-09
Fix write to file in sampling

katxiao:
Fix write to file in sampling - we should append if the file size is greater than 0.


amontanez24 (approved):
:shipit:

pvk-developer (approved):

PR #737, opened 2022-03-17
Handle null values in speed preset

### LINKED ISSUE #716: Create a speed optimized Preset

katxiao:
In the speed preset, we want to avoid modeling nulls in the GaussianCopula model. Instead, we capture the null value percentage of each column with null values before fitting, and add nulls back with the same likelihood after sampling. 

Resolves #716

### DIFF
--- a/sdv/lite/tabular.py
+++ b/sdv/lite/tabular.py
@@ -47,20 +47,19 @@ class TabularPreset(BaseTabularModel):
         if optimize_for == SPEED_PRESET:
             self._model = GaussianCopula(
                 table_metadata=metadata,
-                categorical_transformer='categorical',
+                categorical_transformer='label_encoding',
                 default_distribution='gaussian',
                 rounding=None,
             )
 
             dtype_transformers = {
-                'numerical': rdt.transformers.NumericalTransformer(null_column=False),
-                'integer': rdt.transformers.NumericalTransformer(
+                'i': rdt.transformers.NumericalTransformer(
                     dtype=np.int64, null_column=False),
-                'float': rdt.transformers.NumericalTransformer(
+                'f': rdt.transformers.NumericalTransformer(
                     dtype=np.float64, null_column=False),
-                'categorical': rdt.transformers.CategoricalTransformer(fuzzy=True),
-                'boolean': rdt.transformers.BooleanTransformer(null_column=False),
-                'datetime': rdt.transformers.DatetimeTransformer(null_column=False),
+                'O': rdt.transformers.CategoricalTransformer(fuzzy=True),
+                'b': rdt.transformers.BooleanTransformer(null_column=False),
+                'M': rdt.transformers.DatetimeTransformer(null_column=False),
             }
             self._model._metadata._dtype_transformers.update(dtype_transformers)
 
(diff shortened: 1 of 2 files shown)


sdv/lite/tabular.py
```diff
@@ -42,22 +47,50 @@ def __init__(self, optimize_for=None, metadata=None):
...
 
     def sample(self, num_rows):
         """Sample rows from this table."""
-        return self._model.sample(num_rows)
+        sampled = self._model.sample(num_rows)
+
+        if self._null_percentages:
+            for field, percentage in self._null_percentages.items():
+                sampled[field] = sampled[field].apply(
+                    lambda x: np.nan if random.random() < percentage else x)
```
amontanez24, line 91: I think `apply` can sometimes be slow. Can you try making a mask from the columns based on the null percentage for each one and applying the mask to the `DataFrame` instead? I think that should be faster. Here is an example https://cmdlinetips.com/2019/05/how-to-randomly-add-nan-to-pandas-dataframe/, although in this case the percentage of nulls is different for each field

amontanez24 (commented):
This is looking really good! I just left one comment about potentially making it faster

sdv/lite/tabular.py
katxiao: Cool, sounds good! I'll update it

amontanez24 (approved):
LGTM! Thanks for addressing the changes

PR #743, opened 2022-03-23
Fix field distribution arg in GaussianCopula

katxiao:
In GaussianCopula _fit, we received transformed columns which have columns `a.value` and `a.is_null` if the original dataset has a column 'a'. When this happens, the field distribution does not get properly passed through to GaussianCopula since it has been specified under the key 'a'. We can fix this by matching transformed columns to the original column names.

### DIFF
--- a/sdv/tabular/copulas.py
+++ b/sdv/tabular/copulas.py
@@ -292,7 +292,7 @@ class GaussianCopula(BaseTabularModel):
             if column not in self._field_distributions:
                 # Check if the column is a derived column.
                 column_name = column.replace('.value', '').replace('.is_null', '')
-                self._field_distributions[column] = self._field_distribution.get(
+                self._field_distributions[column] = self._field_distributions.get(
                     column_name, self._default_distribution)
 
         self._model = copulas.multivariate.GaussianMultivariate(
--- a/sdv/tabular/copulas.py
+++ b/sdv/tabular/copulas.py
@@ -289,9 +289,11 @@ class GaussianCopula(BaseTabularModel):
                 Data to be fitted.
         """
         for column in table_data.columns:
-            distribution = self._field_distributions.get(column)
-            if not distribution:
-                self._field_distributions[column] = self._default_distribution
+            if column not in self._field_distributions:
+                # Check if the column is a derived column.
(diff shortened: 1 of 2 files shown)


sdv/tabular/copulas.py
```diff
@@ -289,9 +289,11 @@ def _fit(self, table_data):
                 Data to be fitted.
         """
         for column in table_data.columns:
-            distribution = self._field_distributions.get(column)
-            if not distribution:
-                self._field_distributions[column] = self._default_distribution
+            if column not in self._field_distributions:
+                # Check if the column is a derived column.
+                column_name = column.replace('.value', '').replace('.is_null', '')
```
amontanez24, line 294: The `is_null` columns are just boolean columns that say whether or not their row should be null. I would apply a fixed distribution to them.

I think it only makes sense to apply the specified distribution to the one ending in `.value`

amontanez24 (changes requested):

sdv/tabular/copulas.py
```diff
@@ -289,9 +289,11 @@ def _fit(self, table_data):
                 Data to be fitted.
         """
         for column in table_data.columns:
-            distribution = self._field_distributions.get(column)
-            if not distribution:
-                self._field_distributions[column] = self._default_distribution
+            if column not in self._field_distributions:
+                # Check if the column is a derived column.
+                column_name = column.replace('.value', '')
+                self._field_distributions[column] = self._field_distributions.get(
```
amontanez24, line 295: One thing to note is that the `_field_distributions` dictionary will have some field names that match the `RDT` output, and some that match the input. For example, if null columns are created and there is a column 'a', then the dictionary will have 
```
{
    'a': dist,
    'a.is_null': default_dist
}
````
Idk if it makes sense to have them all match the `HyperTransformer` output names or not (ie. keep the `.value` extension)

amontanez24 (approved):
I think this looks good. I left a comment but it's up to you on how to implement

sdv/tabular/copulas.py
katxiao: I created #744 to track this question, not completely sure what we should do right now.

PR #750, opened 2022-03-29
Fix field distributions bug for copulagan

### LINKED ISSUE #747: Field distributions bug in `CopulaGAN`
### Environment Details

* SDV version: 0.14.0

### Error Description

The same bug described in #746 is also observed on the `CopulaGAN` model: The `field_distributions` aren't being honored because the RDT is changing `column_name` to `column_name.value`, so we cannot find the right column to apply the distribution to.

katxiao:
Resolves #747

### DIFF
--- a/sdv/tabular/copulagan.py
+++ b/sdv/tabular/copulagan.py
@@ -146,6 +146,9 @@ class CopulaGAN(CTGAN):
     """
 
     DEFAULT_DISTRIBUTION = 'parametric'
+    _field_distributions = None
+    _default_distribution = None
+    _ht = None
 
     def __init__(self, field_names=None, field_types=None, field_transformers=None,
                  anonymize_fields=None, primary_key=None, constraints=None, table_metadata=None,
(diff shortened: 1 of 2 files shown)


sdv/tabular/copulagan.py
```diff
@@ -209,11 +212,13 @@ def _fit(self, table_data):
...
-                distribution=distributions.get(field, default)
+                distribution=distributions.get(field.replace('.value', ''), default)
             )
             for field in table_data
-            if field.replace('.value', '')
-            in fields and fields.get(field, dict()).get('type') != 'categorical'
+            if field.replace('.value', '') in fields and fields.get(
+                field.replace('.value', ''),
+                dict(),
+            ).get('type') != 'categorical'
```
amontanez24, line 221: minor: but can we rewrite this using a normal for loop and storing `field.replace('.value', '')` as a variable. I think it would be easier to read

amontanez24 (approved):
minor comment but besides that :shipit:

pvk-developer (approved):

PR #755, opened 2022-03-31
Add constraints to preset

### LINKED ISSUE #753: Presets should be able to handle constraints
### Problem Description

In #716 we defined the preset functionality. Let's also add in the ability to input `constraints` in this preset. 

- Right now, both metadata & constraints cannot be input into a tabular model. The user can supply either-or
- If the user wishes to add constraints, they have to put it into the metadata, which is not straighforward

For for now, we should allow users to input their constraints as a separate parameter for presets only.

katxiao:
Resolves #753 

Allow users to pass in constraints to tabular preset. If both constraints and metadata are passed in, first add the conditions to the metadata, since the tabular model will error out if both are provided.

Additional changes:
- update the tabular preset to include sampling, save, and load methods
- update preset categorical transformer to one that produces continuous values
- remove print on initialization
- add custom representation that includes preset name

### DIFF
--- a/sdv/lite/tabular.py
+++ b/sdv/lite/tabular.py
@@ -41,6 +41,9 @@ class TabularPreset():
                              'through the options.')
         if name not in PRESETS:
             raise ValueError(f'`name` must be one of {PRESETS}.')
+
+        self.name = name
+
         if metadata is None:
             warnings.warn('No metadata provided. Metadata will be automatically '
                           'detected from your data. This process may not be accurate. '
@@ -230,3 +233,11 @@ class TabularPreset():
                   'Supply the desired preset using the `name` parameter.\n\n'
                   'Have any requests for custom presets? Contact the SDV team to learn '
                   'more an SDV Premium license.\n')
+
+    def __repr__(self):
+        """Represent tabular preset instance as text.
+
+        Returns:
+            str
+        """
+        return f'TabularPreset(name={self.name})'
--- a/sdv/lite/tabular.py
+++ b/sdv/lite/tabular.py
@@ -7,8 +7,8 @@ import warnings
 import numpy as np
 import rdt
 
+from sdv.metadata import Table
 from sdv.tabular import GaussianCopula
-from sdv.tabular.base import BaseTabularModel
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


sdv/lite/tabular.py
```diff
@@ -43,10 +46,21 @@ def __init__(self, name=None, metadata=None):
...
+                metadata['constraints'].append(constraint.to_dict())
+
+            constraints = None
+
         if name == SPEED_PRESET:
             self._model = GaussianCopula(
                 table_metadata=metadata,
-                categorical_transformer='label_encoding',
+                constraints=constraints,
+                categorical_transformer='categorical_fuzzy',
```
amontanez24, line 63: why are we changing this?

sdv/lite/tabular.py
```diff
@@ -62,13 +76,13 @@ def __init__(self, name=None, metadata=None):
             }
             self._model._metadata._dtype_transformers.update(dtype_transformers)
 
-            print('This config optimizes the modeling speed above all else.\n\n'
```
amontanez24, line 65: Why is this being removed?

tests/unit/lite/test_tabular.py
```diff
@@ -182,11 +293,55 @@ def test_list_available_presets(self):
...
+        preset = Mock()
+        preset._model = model
+
+        # Run
+        TabularPreset.save(preset, 'test-path')
+
+        # Assert
+        model.save.assert_called_once_with('test-path')
+
+    @patch('sdv.lite.tabular.GaussianCopula')
```
amontanez24, line 327: Do we need this mock?

amontanez24 (commented):
This looks good, I just have some small questions

sdv/lite/tabular.py
katxiao: @npatki mentioned that for GaussianCopula, it expects continuous values, and that label encoding and categorical are not continuous.

tests/unit/lite/test_tabular.py
katxiao: Ah we don't, thanks.

amontanez24 (approved):
Thanks for addressing! :shipit:

PR #757, opened 2022-04-01
Don't modify my metadata object

### LINKED ISSUE #754: Don't modify my metadata object
### Environment Details

* SDV version: 0.14.0 (and previous)

### Description

Right now, if I pass in a [`metadata.Table` object](https://sdv.dev/SDV/api_reference/metadata/table.html) into a model and fit it, the process will modify my metadata (add extra fields to it).

This can lead to unexpected behavior due to the extra fields that were added, especially if I want to re-use the metadata object in a different model.

Can we make a deep copy of the `metadata.Table` object instead?

### Steps to reproduce

See the code below for unexpected usage

```python

metadata, data = load_tabular_demo('student_placements', metadata=True)

# create a model that estimates everything as a gamma distribution
model = GaussianCopula(table_metadata=metadata,
                       categorical_transformer='label_encoding',
                       default_distribution='gamma')
model.fit(data)
print(model.get_distributions()) # each distribution correctly is a Gamma

# create a new model that estimates everything as a beta
model2 = GaussianCopula(table_metadata=metadata,
                        categorical_transformer='label_encoding',
                        default_distribution='beta')

model2.fit(data)
print(model2.get_distributions()) # Everything is still gamma. This is unexpected!
```

This is happening because, unbeknownst to me, my metadata is modified to include a 'gamma' distribution field for each and every column. I am unknowingly passing it into `model2`, which is favoring the metadata over the `default_distribution` parameter.

amontanez24:
resolves #754

### DIFF
--- a/sdv/lite/tabular.py
+++ b/sdv/lite/tabular.py
@@ -49,16 +49,15 @@ class TabularPreset():
                           'detected from your data. This process may not be accurate. '
                           'We recommend writing metadata to ensure correct data handling.')
 
-        else:
+        if metadata is not None and constraints is not None:
             if isinstance(metadata, Table):
                 metadata = metadata.to_dict()
 
-            if constraints is not None:
-                metadata['constraints'] = []
-                for constraint in constraints:
-                    metadata['constraints'].append(constraint.to_dict())
+            metadata['constraints'] = []
+            for constraint in constraints:
+                metadata['constraints'].append(constraint.to_dict())
 
-                constraints = None
+            constraints = None
 
         if name == SPEED_PRESET:
             self._model = GaussianCopula(
--- a/sdv/lite/tabular.py
+++ b/sdv/lite/tabular.py
@@ -49,15 +49,16 @@ class TabularPreset():
                           'detected from your data. This process may not be accurate. '
                           'We recommend writing metadata to ensure correct data handling.')
 
-        if metadata is not None and constraints is not None:
+        else:
             if isinstance(metadata, Table):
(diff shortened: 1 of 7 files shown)


katxiao (approved):
lgtm!

sdv/lite/tabular.py
```diff
@@ -49,15 +49,16 @@ def __init__(self, name=None, metadata=None, constraints=None):
                           'detected from your data. This process may not be accurate. '
                           'We recommend writing metadata to ensure correct data handling.')
 
-        if metadata is not None and constraints is not None:
+        else:
```
csala, line 52: Wouldn't this now fail if `metadata` is `None` and `constraints` are not `None`?

tests/integration/tabular/test_base.py
```diff
@@ -26,6 +27,37 @@ def _isinstance_side_effect(*args, **kwargs):
...
+                           categorical_transformer='label_encoding',
+                           default_distribution='gamma')
+    model.fit(data)
+    model2 = GaussianCopula(table_metadata=metadata,
+                            categorical_transformer='label_encoding',
+                            default_distribution='beta')
+    model2.fit(data)
+
+    # Assert
+    assert model._metadata != metadata
```
katxiao, line 56: Could we also assert that the different metadata have the expected distributions? Just to be sure.

katxiao (approved):
LGTM with a comment

katxiao (approved):

PR #759, opened 2022-04-11
resolved FutureWarning with Pandas replaced append by concat

Deathn0t:
Hello, 

I am just providing a simple fix to replace the `append` function which is now triggering a `FutureWarning` for deprecation from Pandas.

Best regards,
Romain.


katxiao (approved):
Thanks! LGTM :shipit:

PR #764, opened 2022-04-14
Model nulls normally when tabular preset has constraints

### LINKED ISSUE #763: Tabular Preset with constraints should handle nulls correctly
### Problem Description

SDV Tabular preset should handle constraints correctly when the constraints add null values. Currently, if I use a CustomConstraint that specifies a certain column should have null values based on a custom formula, those null values get overwritten.

katxiao:
Resolves #763 

If tabular preset has constraints, model nulls normally instead of pre- and post-processing the data. Also, update the preset name.

### DIFF
--- a/sdv/lite/tabular.py
+++ b/sdv/lite/tabular.py
@@ -12,9 +12,9 @@ from sdv.tabular import GaussianCopula
 
 LOGGER = logging.getLogger(__name__)
 
-SPEED_PRESET = 'SPEED'
+FAST_ML_PRESET = 'FAST_ML'
 PRESETS = {
-    SPEED_PRESET: 'Use this preset to minimize the time needed to create a synthetic data model.',
+    FAST_ML_PRESET: 'Use this preset to minimize the time needed to create a synthetic data model.'
 }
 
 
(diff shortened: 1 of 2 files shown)


sdv/lite/tabular.py
```diff
@@ -86,13 +91,14 @@ def fit(self, data):
...
                 Data to fit the model to.
         """
-        self._null_percentages = {}
-
-        for column, column_data in data.iteritems():
-            num_nulls = column_data.isna().sum()
-            if num_nulls > 0:
-                # Store null percentage for future reference.
-                self._null_percentages[column] = num_nulls / len(column_data)
+        if not self._null_column:
```
amontanez24, line 94: can we add the test case for when this should not learn the null percentages?

amontanez24 (approved):
Just one comment about adding test cases. Otherwise LGTM!

pvk-developer (approved):

sdv/lite/tabular.py
katxiao: @amontanez24 That should be covered in the `test_fit` unit test, and I also assert that null percentages is null at the end.

katxiao: Added!

PR #776, opened 2022-04-27
Update preset min and max params

katxiao:
Update min/max params to 'auto' in tabular preset.

When we update the dtype transformers, we override the min/max params that are set in the TabularBase class.


amontanez24 (approved):
:shipit:

PR #788, opened 2022-05-10
Fix missing case for None in _validate_file_path

davebulaval:
As by the doc of `sample` if `output_file_path` is set to `None`, there should be no temporary file writing.
However, [`_validate_file_path`](https://github.com/sdv-dev/SDV/blob/8d3d847505805801c0bb5662838b11ccefa3f934/sdv/tabular/base.py#L422) does not implement a None case, thus it always falls in the [else](https://github.com/sdv-dev/SDV/blob/8d3d847505805801c0bb5662838b11ccefa3f934/sdv/tabular/base.py#L434) option and uses `TMP_FILE_NAME`.

This PR adds a None case and adds a unit test to test this case.

### DIFF
--- a/sdv/tabular/base.py
+++ b/sdv/tabular/base.py
@@ -422,7 +422,7 @@ class BaseTabularModel:
     def _validate_file_path(self, output_file_path):
         """Validate the user-passed output file arg, and create the file."""
         output_path = None
-        if output_file_path == DISABLE_TMP_FILE:
+        if output_file_path == DISABLE_TMP_FILE or output_file_path is None:
             # Temporary way of disabling the output file feature, used by HMA1.
             return output_path
 
--- a/tests/unit/tabular/test_base.py
+++ b/tests/unit/tabular/test_base.py
@@ -635,6 +635,13 @@ class TestBaseTabularModel:
         with pytest.raises(AssertionError, match='path/to/file already exists'):
             BaseTabularModel._validate_file_path(model, 'file_path')
 
+    def test_given_none_output_file_path_when_validate_file_path_then_return_none(self):
+        model = Mock(spec_set=CTGAN)
+        expected = None
+
+        actual = BaseTabularModel._validate_file_path(model, output_file_path=None)
+        assert actual is expected
+
     @patch('sdv.tabular.base.os')
     def test_sample_with_default_file_path(self, os_mock):
         """Test the `BaseTabularModel.sample` method with the default file path.


katxiao:
@davebulaval Actually this is the desired behavior - we always default to writing to a temp file. If the `output_file_path` is specified, then we use that file path instead of the default file path.

davebulaval:
> @davebulaval Actually this is the desired behavior - we always default to writing to a temp file. If the `output_file_path` is specified, then we use that file path instead of the default file path.

Thus, the doc is not written accordingly. Also, this behaviour creates a strange error when you sample a singleton sample multiple times (around 15K). At one point, it cannot find the temps file that was supposed to have been created.

I am not sure what causes this error after a certain amount of singleton sampling.

katxiao:
> > @davebulaval Actually this is the desired behavior - we always default to writing to a temp file. If the `output_file_path` is specified, then we use that file path instead of the default file path.
> 
> Thus, the doc is not written accordingly. Also, this behaviour creates a strange error when you sample a singleton sample multiple times (around 15K). At one point, it cannot find the temps file that was supposed to have been created.
> 
> I am not sure what causes this error after a certain amount of singleton sampling.

To clarify, the expected behavior is as follows: 
- When sampling runs successfully and `output_file_path=None`, there is no output file. 
- If `output_file_path=None` and there is an error with sampling, all the sampling progress up until the error will be saved in the default temp file.

If you have a question about a specific use case, please file an issue. We will not be moving forward with this PR.

davebulaval:
> > > @davebulaval Actually this is the desired behavior - we always default to writing to a temp file. If the `output_file_path` is specified, then we use that file path instead of the default file path.
> > 
> > 
> > Thus, the doc is not written accordingly. Also, this behaviour creates a strange error when you sample a singleton sample multiple times (around 15K). At one point, it cannot find the temps file that was supposed to have been created.
> > I am not sure what causes this error after a certain amount of singleton sampling.
> 
> To clarify, the expected behavior is as follows:
> 
> * When sampling runs successfully and `output_file_path=None`, there is no output file.
> * If `output_file_path=None` and there is an error with sampling, all the sampling progress up until the error will be saved in the default temp file.
> 
> If you have a question about a specific use case, please file an issue. We will not be moving forward with this PR.

That is not what you just said. If the desired output is as so, the code does not do that.

RN the code can be read as if `output_file_path=None` it will create a dummy file since if `output_file_path` is not it goes into the else, thus it creates a file with the `TMP_FILE_NAME`. Then, the `_validate_file_path`  will return this path and later on this path will be used by the [_sample_function](https://github.com/sdv-dev/SDV/blob/84dd2ad374abdfd3290842435aaa140c4fe4eb36/sdv/tabular/base.py#L507). Thus, it will try to write into a file.

My specific use case is that the codes fail due to writing attempts on this temp file.

CroCodeDial:
I apologize for bumping this closed issue. I had this problem too.

I was running a while-loop around the sampling method of a CTGAN model followed by a different method that checks the Levenshtein distance between the attributes of the generated data. It would then remove non-optimum rows and sample again. I got the same problem as mentioned by OP, where the SSH I am connected to concludes that it cannot find the temporary file set by the "_validate_file_path()" method.

Sometimes it would bug out on the first iteration, sometimes later iterations. 

I understand SDV already considers these features with constraints and other functionalities, so my setup might be to blame. 

An easy "tongue-in-cheek" fix was to put the "output_file_path" argument as 'disable.' However, I'm unsure if this has consequences on the sampled data as it's meant initially for HMA1...

davebulaval:
And I think the doc is misleading. Yes, in the end, no file is kept, but it is false to advertise that **no file is created** since there is one created in the process. It is just that you deleted it before the end.

The impact is that if I launch the process from a computer without writing permission, it would automatically fail, and it could be unnecessarily difficult to debug the reason for such failure.

PR #799, opened 2022-05-17
Fix `Unique` constraint for nans

### LINKED ISSUE #797: Make Unique constraint works with nans
### Problem Description

As a user, it is confusing and inconvenient when my `Unique` constraint fails validation because of missing values. 

Currently
- missing values are treated as unique values in the single column case
- missing values cause failure 100% of the time in the multi column case

We want to make the behavior for the `Unique` constraint consistent in both cases when handling missing values.

### Expected behavior

We should treat missing values (`np.nan`, `None`, etc.) as a unique category/value. Consider the following cases:

Should PASS → YES (right now this fails)
```
Col1		Col2
A		NaN
NaN		B
C		D
NaN		NaN
```

Should this pass? → NO
```
Col1	
A		
NaN	
C	
NaN
```

Should this pass? – NO
```
Col1		Col2
A		NaN
A		NaN
NaN		B
C		D
NaN		NaN
```

Should this pass? – NO
```
Col1		Col2
A		NaN
NaN		B
C		D
NaN		NaN
NaN		NaN
```

Currently all these cases will fail validation, but the first one should pass

fealho:
Resolve #797.

### DIFF
--- a/tests/unit/constraints/test_tabular.py
+++ b/tests/unit/constraints/test_tabular.py
@@ -5560,3 +5560,57 @@ class TestUnique():
         # Assert
         expected = pd.Series([True, False, False, True, True, False])
         pd.testing.assert_series_equal(valid, expected)
+
+    def test_is_valid_one_column_nans(self):
+        """Test the ``Unique.is_valid`` method for one column with nans.
+
+        This method should return a pd.Series where the index
+        of the first occurence of a unique value of ``instance.columns``
+        is set to ``True``, and every other occurence is set to ``False``.
+        ``None``, ``np.nan`` and ``float('nan')`` should be treated as the same category.
+
+        Input:
+        - DataFrame with some repeated values, some of which are nan's.
+        Output:
+        - Series with the index of the first occurences set to ``True``.
+        """
+        # Setup
+        instance = Unique(columns=['a'])
+
+        # Run
+        data = pd.DataFrame({
+            'a': [1, None, 2, np.nan, float('nan'), 1],
+            'b': [np.nan, 1, None, float('nan'), float('nan'), 1],
+        })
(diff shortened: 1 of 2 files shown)


tests/unit/constraints/test_tabular.py
```diff
@@ -5560,3 +5560,57 @@ def test_is_valid_one_column(self):
...
+    def test_is_valid_multiple_columns_nans(self):
+        """Test the ``Unique.is_valid`` method for multiple columns with nans.
+
+        This method should return a pd.Series where the index
+        of the first occurence of a unique combination of ``instance.columns``
+        is set to ``True``, and every other occurence is set to ``False``.
+        ``None``, ``np.nan`` and ``float('nan')`` should be treated as the same category.
+
+        Input:
+        - DataFrame with multiple of the same combinations of columns, some of which are nan's.
```
amontanez24, line 5600: minor: I don't think `nans` needs the apostrophe

tests/unit/constraints/test_tabular.py
```diff
@@ -5560,3 +5560,57 @@ def test_is_valid_one_column(self):
...
+    def test_is_valid_one_column_nans(self):
+        """Test the ``Unique.is_valid`` method for one column with nans.
+
+        This method should return a pd.Series where the index
+        of the first occurence of a unique value of ``instance.columns``
+        is set to ``True``, and every other occurence is set to ``False``.
+        ``None``, ``np.nan`` and ``float('nan')`` should be treated as the same category.
+
+        Input:
+        - DataFrame with some repeated values, some of which are nan's.
```
amontanez24, line 5573: minor: same

amontanez24 (approved):
LGTM!

katxiao (approved):

PR #807, opened 2022-05-19
Raise constraint errors together + misc.

### LINKED ISSUE #801: Improve error message for invalid constraints
### Problem Description

As a user, it would be helpful to know which rows failed the constraint if my data does not meet a constraint I specified.

### Expected behavior
- Augment the current error message that a user gets when their data fails constraint validation.
    - Print the first five rows that fail
    - If more than 5 rows fail print the number of extra rows as follows:: `+X more`
    - Only print the columns needed by the constraint
- Change the way we validate constraints to catch all of the errors and surface them at once. Ie. instead of failing on the first invalid constraint, validate all of them and then surface those errors together

Here are some examples:

```
ConstraintsNotMetError: Data is not valid for the 'GreaterThan' constraint:
     parent_age       child_age
10   13               32
20   34               50
23   10               37
80   0                1
81   -1               50
+50 more
```
```
Data is not valid for the 'Unique' constraint:
     first_name       last_name
45   'Annie'          'Hall'
14   'Harry'          'Potter'
98   'Harry'          'Potter'
```

fealho:
Resolve #801.

Note: this PR also fixes several typos and minor fixes.


sdv/constraints/base.py
```diff
@@ -236,6 +262,8 @@ def _validate_constraint_columns(self, table_data):
             table_data (pandas.DataFrame):
                 Table data.
         """
+        self._validate_data_on_constraint(table_data)
```
fealho, line 265: These two methods: `_validate_data_on_constraint` and `_validate_constraint_columns` should be merged in the future, when the `fit_columns_model` logic gets dropped.

tests/unit/constraints/test_base.py
```diff
@@ -301,22 +301,22 @@ def test_fit_trains_column_model(self, ht_mock, gm_mock):
     def test_transform(self):
         """Test the ``Constraint.transform`` method.
 
-        It is an identity method for completion, to be optionally
-        overwritten by subclasses.
+        When no constraints are passed, it behaves like an identity method,
```
fealho, line 304: Unrelated to my changes, it was just unclear documentation.

npatki:
> @amontanez24 @npatki What is the expected behavior of a constraint when `handling_strategy=reject_sampling`? Right now it doesn't go through validation, so if no rows can be sampled it simply returns and empty dataframe.

The expected behavior is to handle this the same as conditional sampling with reject sampling. See https://github.com/sdv-dev/SDV/issues/809. It would be nice to make these changes in tandem with the constraint changes.

sdv/constraints/base.py
```diff
@@ -120,13 +121,17 @@ class Constraint(metaclass=ConstraintMeta):
     def _identity(self, table_data):
         return table_data
 
+    def _identity_transform(self, table_data):
```
amontanez24, line 124: Why not just put in `identity`?

sdv/constraints/base.py
amontanez24: Let's keep these separate for a couple of reasons:
1. We may end up keeping the columns model
2. These are actually two different types of validation. One is checking that the data being transformed actually adheres to the constraint. The other is checking if any columns in the constraint are missing and need to be sampled. Let's keep the logic separate. Maybe changing the name of `_validate_constraint_columns` to `_generate_missing_columns` would help or something like that

sdv/constraints/base.py
```diff
@@ -220,6 +225,31 @@ def _sample_constraint_columns(self, table_data):
         sampled_data = pd.concat(all_sampled_rows, ignore_index=True)
         return sampled_data
 
+    def _validate_data_on_constraint(self, table_data):
```
amontanez24, line 228: We also need to check this during `fit`

sdv/constraints/tabular.py
```diff
@@ -1135,6 +1137,7 @@ class Unique(Constraint):
 
     def __init__(self, columns):
         self.columns = columns if isinstance(columns, list) else [columns]
+        self.constraint_columns = tuple(self.columns)
```
amontanez24, line 1140: Why do we need this?

sdv/metadata/table.py
```diff
@@ -643,10 +629,6 @@ def transform(self, data, on_missing_column='error'):
         Returns:
             pandas.DataFrame:
                 Transformed data.
-
-        Raises:
-            ConstraintsNotMetError:
-                If the table data is not valid for the provided constraints.
```
amontanez24, line 649: Should this be replaced with the new error that gets raised?

amontanez24 (changes requested):

sdv/metadata/table.py
fealho: The `self._validate_data_on_constraints(data)` was moved from here to the `transform` function, so the error is not really from here anymore.

sdv/constraints/tabular.py
fealho: Because of this line in the validation: `if set(self.constraint_columns).issubset(table_data.columns.values):`.

If a constraint doesn't have the `constraint_columns` attribute, it won't get validated

sdv/constraints/base.py
fealho: This is a little complicated because (1) `fit` is only really used when `fit_column_model == True`, otherwise all `fit` does is call `_fit`, (2) `_fit` needs to run before we can call `is_valid`, since `_fit` creates some of the attributes used in `is_valid` for some constraints, (3) `_validate_data_on_constraint` runs `is_valid`.

I could (1) move `_validate_data_on_constraint` to run in `fit` right after the `_fit` runs, (2) remove it from the `transform` part, otherwise we would just be running the validation on the same data twice and (3) add it back to the `Table.transform` method.

The reason I didn't do that is because we discussed removing the `fit_columns_model` logic, in which case the logic of both cases is the same, but cleaner the way I did.

sdv/constraints/tabular.py
amontanez24: I see. I wonder why we even need that condition tbh

sdv/constraints/base.py
fealho: `identity` gets used for `self.filter_valid = self._identity` and `self.reverse_transform = self._identity`, cases in which we don't want to run the validation.

sdv/constraints/base.py
amontanez24: ok, this is fine for now. The goal eventually is to separate out all of the transformations into their own component. If this happens, it may be possible for users to fit on one dataset and end up using a different one to be transformed, in which case we would have to validate both times. For now though it's fine to just do it in one place

sdv/metadata/table.py
amontanez24: Does it still surface here though?

sdv/constraints/tabular.py
fealho: It was there originally. I didn't remove it because the code would crash if a `constraint_column` is not in the `table_data.columns`, and it seems we want to do stuff with it in the function right after, `_validate_constraint_columns`.

sdv/constraints/base.py
amontanez24: ok, let's name this more clearly then. Like `_identity_with_validation`

sdv/metadata/table.py
fealho: It's possible, I never did test whether errors that didn't happen in the first transform could be happen in the second. Regardless, I don't think we usually write the errors of children methods in the parent one, right? If so, then we need to add the error from `_validate_constraint_columns` as well.

sdv/constraints/tabular.py
amontanez24: Ok. That's fine but in general we should try to decouple the validation from the stuff related to the columns model. Originally they were more tied together because we were only validating constraints with a transform, but since that is no longer the case, they don't need to be tied together. Every constraint needs to be validated now, whether or not it will even need the columns model.

tests/unit/metadata/test_table.py
```diff
@@ -582,93 +581,6 @@ def test__transform_constraints_drops_columns(self):
         }, index=[0, 1, 2])
         assert result.equals(expected_result)
 
-    def test__validate_data_on_constraints(self):
```
amontanez24, line 585: Can we keep the tests for this method but just move them to the appropriate place

amontanez24 (commented):
I think this is almost ready, but we should keep the tests for the `_validate_on_constraint` method

fealho:
@amontanez24 The old tests didn't really match the new code, so I rewrote them a bit.

pvk-developer (approved):
LGTM. Just minor question about the `Errors` or `Error`

sdv/constraints/errors.py
```diff
@@ -3,3 +3,7 @@
 
 class MissingConstraintColumnError(Exception):
     """Error to use when constraint is provided a table with missing columns."""
+
+
+class MultipleConstraintsErrors(Exception):
```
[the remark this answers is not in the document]
fealho: I think `Errors` is clearer, since it is a list of multiple errors

amontanez24 (approved):
LGTM!

PR #823, opened 2022-05-26
Split `GreaterThan` into `Inequality` and `ScalarInequality`

### LINKED ISSUE #814: Split GreaterThan constraint into Inequality and ScalarInequality
### Problem Description

_As a user, it is sometimes confusing to use the `GreaterThan` constraint because the parameters can take many different types of values._

For example, `low` and `high` can be either a column name, a scalar, or a list of column names. If it is a list, then it isn't clear what the behavior is. If both `high` and `low` are lists, then the values are compared by their indices (ie. the first value of low must be less than the first value of high, the second value of low must be less than the second value of high, etc.). If only one is a list, then each value in the list is compared to the other value. If they are both lists and the lengths don't match, there will be an error.

To remove all of this confusion, we propose splitting the `GreaterThan` constraint into two different constraints: `Inequality` and `ScalarInequality`. The first will compare two columns and the second will compare a column to a scalar. There will no longer be support for lists of columns.

### Expected behavior

- Remove `GreaterThan` constraint class
- Add `Inequality` class
    - Init params should be:
        - low_column_name: String that is the name of the lower column
        - high_column_name: String that is the name of the higher column
        - strict_boundaries: Boolean on whether or not to be inclusive with the inequality
    - Logical changes
        - Since there is no longer a `drop` parameter, we should just pick one of the columns to drop and one to use
- Add `ScalarInequality` class
    - Init params should be:
        - column_name: String that is the name of the column to compare
        - value: Scalar value to compare to
        - relation: String that is either `'>'`, `'>='`, `'<'`, `'<='`

fealho:
Resolve #814.

### DIFF
--- a/tests/unit/constraints/test_tabular.py
+++ b/tests/unit/constraints/test_tabular.py
@@ -1634,121 +1634,59 @@ class TestInequality():
 class TestPositive():
 
     def test__init__(self):
-        """
-        Test the ``Positive.__init__`` method.
+        """Test the ``Positive.__init__`` method.
 
-        The method is expected to set the ``_low`` instance variable
-        to 0, the ``_scalar`` variable to ``'low'``. The rest of the
-        parameters should be passed. Check that ``_drop`` is set to
-        ``None`` when ``drop`` is ``False``.
-
-        Input:
-        - strict = True
-        - low = 'a'
-        - drop = False
-        Side effects:
-        - instance._low == 'a'
-        - instance._high == 0
-        - instance._strict == True
-        - instance._scalar == 'low'
-        - instance._drop = None
+        Ensure the attributes are correctly set.
         """
         # Run
-        instance = Positive(columns='a', strict=True, drop=False)
+        instance = Positive(column_name='abc')
 
         # Asserts
-        assert instance._low == 0
-        assert instance._high == ['a']
-        assert instance._strict is True
-        assert instance._scalar == 'low'
-        assert instance._drop is None
+        assert instance._value == 0
+        assert instance._column_name == 'abc'
+        assert instance._operator == np.greater
 
-    def test__init__drop_true(self):
-        """
-        Test the ``Positive.__init__`` method with drop is ``True``.
+    def test__init__strict_True(self):
+        """Test the ``Positive.__init__`` method.
 
-        Check that ``_drop`` is set to 'high' when ``drop`` is ``True``.
-
-        Input:
-        - strict = True
-        - low = 'a'
-        - drop = True
-        Side effects:
-        - instance._low == 'a'
-        - instance._high == 0
-        - instance._strict == True
-        - instance._scalar == 'low'
-        - instance._drop = 'high'
+        Ensure the attributes are correctly set when ``strict`` is True.
         """
         # Run
-        instance = Positive(columns='a', strict=True, drop=True)
+        instance = Positive(column_name='abc', strict=True)
 
         # Asserts
-        assert instance._low == 0
-        assert instance._high == ['a']
-        assert instance._strict is True
-        assert instance._scalar == 'low'
-        assert instance._drop == 'high'
+        assert instance._value == 0
+        assert instance._column_name == 'abc'
+        assert instance._operator == np.greater_equal
 
 
 class TestNegative():
 
     def test__init__(self):
-        """
-        Test the ``Negative.__init__`` method.
+        """Test the ``Negative.__init__`` method.
 
-        The method is expected to set the ``_high`` instance variable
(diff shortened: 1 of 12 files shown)


tests/unit/constraints/test_tabular.py
```diff
@@ -787,1789 +787,283 @@ def test_reverse_transform_non_string(self):
...
         Input:
-        - Table with values above and below low.
-        Output:
-        - True should be returned for the rows where the high
-        column is above low.
+        - a string
+        - a string
+        - a non-bool
+        Side effect:
+        - Raise ``ValueError`` because column names must be strings
```
amontanez24, line 820: This should say: `because strict_boundaries must be a boolean`

tests/unit/constraints/test_tabular.py
```diff
@@ -787,1789 +787,283 @@ def test_reverse_transform_non_string(self):
...
-        column is below `high`.
+        - low_column_name = 'a'
+        - high_column_name = 'b'
+        Side effects:
+        - instance._low_column_name = 'a'
+        - instance._high_column_name = 'b'
+        - instance._diff_column_name = 'a#b'
+        - instance._operator = np.greater_equal
+        - instance.rebuild_columns = 'b'
+        - instance._dtype = None
```
amontanez24, line 843: This is basically just copying the code and doesn't really add anything. Instead can we describe why certain parameters are being set as certain values?

tests/unit/constraints/test_tabular.py
```diff
@@ -787,1789 +787,283 @@ def test_reverse_transform_non_string(self):
...
 
         Input:
-        - Table with a row that contains only one NaN value.
-        Output:
-        - True should be returned for the row with the NaN value.
+        - Table data with integers.
+        Side Effect:
+        - _validate_columns_exist should be called once
+        - _get_is_datetime should be called once
+        - _is_datetime should receive the output of _get_is_datetime
```
amontanez24, line 906: Does `is_datetime` have the preceding `_` or not, or are there two different variables?

tests/unit/constraints/test_tabular.py
```diff
@@ -787,1789 +787,283 @@ def test_reverse_transform_non_string(self):
...
-        - ``_drop`` is set to ``high``.
         Input:
-        - Table with two columns two constrained columns at a constant distance of
-          exactly 3 and one additional dummy column.
-        Output:
-        - Same table with a diff column of the logarithms of the distances + 1,
-        which is np.log(4) and the high column dropped.
+        - Table data with floats.
+        Side Effect:
+        - _dtype should be a list of float dtypes.
```
amontanez24, line 935: Is it a list or just a single dtype?

tests/unit/constraints/test_tabular.py
```diff
@@ -787,1789 +787,283 @@ def test_reverse_transform_non_string(self):
...
-        - ``_drop`` is set to ``None``, so all original columns will be in output.
         Input:
-        - Table with two constrained columns at a constant distance of
-          exactly 3 and one additional dummy column.
-        Output:
-        - Same table with a diff column of the logarithms of the distances + 1,
-        which is np.log(4).
+        - Table data with datetimes.
+        Side Effect:
+        - _dtype should be a list of datetime dtypes.
```
amontanez24, line 958: same

tests/unit/constraints/test_tabular.py
```diff
@@ -2744,594 +1233,483 @@ def test_reverse_transform_datetime_drop_high(self):
...
+        - low_column_name = 'a'
+        - high_column_name = 'b'
+        - relation = '>'
+        Side effects:
+        - instance._column_name = 'a'
+        - instance._value = 1
+        - instance._diff_column_name = 'a#'
+        - instance._operator = np.greater
+        - instance._dtype = None
+        - instance._is_datetime = None
```
amontanez24, line 1303: Same thing here. Let's describe why the parameters are being set to these values rather than listing them

amontanez24 (changes requested):
I checked the coverage report and for `sdv/contraints/tabular.py`, the following lines need to be tested:
https://github.com/sdv-dev/SDV/blob/issue-814-split-greaterthan/sdv/constraints/tabular.py#L304

https://github.com/sdv-dev/SDV/blob/issue-814-split-greaterthan/sdv/constraints/tabular.py#L420

https://github.com/sdv-dev/SDV/blob/issue-814-split-greaterthan/sdv/constraints/tabular.py#L441

All of them seem related to raising errors. Can we add these unit tests in?

amontanez24 (approved):
Assuming the test coverage is now 100% this looks good to go!

pvk-developer (approved):

PR #828, opened 2022-06-01
Create FixedIncrements constraint

### LINKED ISSUE #818: Create FixedIncrements constraint
### Problem Description

_As a user, it would be useful to have a constraint that specifies the lowest common increment the data can appear in. For example, if my data must be a multiple of 5 or 7, it would be useful to be able to specify this._

### Expected behavior

- Remove the `Rounding` constraint. This functionality should be handled by the `HyperTransformer`
- Add a `FixedIncrements` constraint
    - Init params should be:
        - `column_name`: String that is the name of the column to be constrained
        - `increment_value`: Int that is the value to make sure all data should be divisible by. Must be greater than 0.
    - Validation
        - If the `increment_value` is less than 0 or 0, we should raise an error during initialization.
        - If the data is not evenly divisible by the `increment_value` then that row is invalid
        - Nans should be ignored during validation
     - `reverse_transform`
        - Round synthesized values to nearest multiple of `increment_value`

### Additional context

- one option for transform is to divide all values by `increment_value`. Then on reverse you can round to the nearest int and just multiply.

amontanez24:
resolves #818


pvk-developer (approved):
LGTM :shipit:

PR #831, opened 2022-06-01
Change columns to column_names in OneHotEncoding and Unique constraints

### LINKED ISSUE #816: Change columns to column_names in OneHotEncoding and Unique constraints
### Problem Description

_As a user, it is confusing when the parameter names for different constraints differ slightly since it implies that there might be a difference in usage even when there isn't._

### Expected behavior

- Change the `columns` parameter in the `Unique` constraint to be `column_names`
- Change the `columns` parameter in the `OneHotEncoding` constraint to be `column_names`

amontanez24:
Resolves #816


pvk-developer (approved):
LGTM 🚢 !

PR #832, opened 2022-06-03
Split Between into Range and ScalarRange

### LINKED ISSUE #815: Split Between constraint into Range and ScalarRange
### Problem Description

_As a user, it is sometimes unclear how to define a `Between` constraint since some parameters can take multiple values, and some parameters are required depending on the values for other parameters._

For example, `low` can be a column name or a scalar, but if it is a scalar than I must provide the `high_is_scalar` option. To simplify, we propose splitting this constraint into two different constraints: `Range` and `ScalarRange`. The first constrains a column between two other columns and the second constrains a column between two scalar values.

### Expected behavior

- Remove `Between` constraint
- Add `Range` constraint
    - Init parameters should be:
        - `low_column_name`: String that is the column name for the lower bound
        - `middle_column_name`: String that is the column name for the column to be constrained
        - `high_column_name`: String that is the column name for the upper bound
        - `strict_boundaries`: Boolean for whether or not to make the range inclusive. Defaults to `True`
- Add `ScalarRange` constraint:
    - Init parameters should be:
        - `column_name`: String that is the name of the column to constrain
        - `low_value`: Int or float that is the lower bound
        - `high_value`: Int or float that is the upper bound
        - `strict_boundaries`: Boolean for whether or not to make the range inclusive. Defaults to `True`
- Logical changes
    - Currently the data sometimes doesn't round back to the appropriate number of decimals. For example, if the data passed in to `reverse_transform` is rounded to the nearest hundredth, then the returned data should be as well.

### Additional context

- The issue with rounding might require some investigation

pvk-developer:
Resolves #815


sdv/constraints/tabular.py
```diff
@@ -17,8 +17,10 @@
     * Negative: Ensure that the values in given columns are always negative.
     * ColumnFormula: Compute the value of a column based on applying a formula
       on the other columns of the table.
-    * Between: Ensure that the value in one column is always between the values
-      of two other columns/scalars.
+    * Range: Ensure that the value in one column is always between the values
+      of two other columns.
+    * ScalarRange: Ensure that the value in one column is always between the values
+      of two other scalar values.
```
amontanez24, line 23: nitpick: you can just says scalars since you already said values before

sdv/constraints/tabular.py
```diff
@@ -899,21 +970,16 @@ def reverse_transform(self, table_data):
...
-        high = self._get_high_value(table_data)
-        data = table_data[self._transformed_column]
+        data = table_data[self.column_name]
 
         data = 1 / (1 + np.exp(-data))
-        data = (data - 0.025) / 0.95
-        data = data * (high - low) + low
-        data = data.clip(low, high)
+        data = data * (self.high_value - self.low_value) + self.low_value
+        data = data.clip(self.low_value, self.high_value)
```
amontanez24, line 977: Perhaps we should make util functions called `sigmoid` and `logit` that do these operations. That way the code doesn't get reused in two places and we provide some context to the operations. We can also handle the precision problem in those functions

sdv/constraints/tabular.py
```diff
@@ -899,21 +970,16 @@ def reverse_transform(self, table_data):
...
 
         if self._is_datetime:
-            table_data[self.constraint_column] = pd.to_datetime(data)
+            table_data[self.column_name] = pd.to_datetime(data.round('1000ms'))
```
amontanez24, line 980: if the precision problem is fixed, lets get rid of this rounding

amontanez24 (changes requested):
A few nitpicks and a suggestion to move the sigmoid/logit functions into a utils module

tests/unit/constraints/test_tabular.py
```diff
@@ -3606,1085 +3607,815 @@ def test_reverse_transform(self):
...
+            'join_date': pd.to_datetime(['2021-02-10', '2021-05-10', '2021-08-11']),
+            'retirement_date': pd.to_datetime(['2050-10-11', '2058-10-04', '2075-11-14']),
+            'promotion_date#join_date#retirement_date': expected_transform
+        })
         pd.testing.assert_frame_equal(expected_out, out)
 
-    def test_reverse_transform_scalar_column(self):
-        """Test ``Between.reverse_transform`` with ``low`` as scalar and ``high`` as a column.
+    def test_reverse_transform_numerical_data(self):
+        """"Test the ``reverse_transform`` method for ``Range``.
```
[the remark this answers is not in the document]
pvk-developer: Nice catch 👍🏻

tests/unit/constraints/test_tabular.py
```diff
@@ -3998,7 +3993,7 @@ def test_reverse_transform_numerical_data(self):
             'retirement_age': [65, 68, 75],
             'current_age': [21, 22, 25],
         })
-        expected_transform = transform(
+        expected_transform = logit(
```
amontanez24, line 3996: For the unit tests, let's just make sure `logit` and `sigmoid` are called with the correct parameters and we use whatever they return correctly. Then we can add unit tests for those functions specifically

amontanez24 (commented):
Almost ready! One last comment about unit tests

fealho (approved):
Looks good! Left a couple of comments, after that should be good to go.

amontanez24 (approved):
Thanks for addressing the comments :shipit:

fealho (commented):
👍

PR #841, opened 2022-06-13
Remove `fit_columns_model` parameter

### LINKED ISSUE #834: Remove fit_columns_model parameter
### Problem Description

As a user, it is hard to understand what the `fit_columns_model` parameter does because it forces me to learn about the specific method of how conditional sampling for some constraints is handled.

### Expected behavior

- Remove the `fit_columns_model` parameter
- Move any code related to the columns model into its own columns model class

### Additional context

- In `constraints.base.py` the [_validate_constraint_columns](https://github.com/sdv-dev/SDV/blob/master/sdv/constraints/base.py#L253-L290) method should be changed to just raise the `MissingConstraintColumnsError` if any columns are missing.
- The `_reject_sample` and `sample_constraint_columns` methods can be moved into the `columns_model` class which can be renamed to something more appropriate.

pvk-developer:
Resolves #834


amontanez24 (changes requested):
This is looking really good! Just a few comments and suggestions, but after that and the test coverage, it should be ready

tests/unit/constraints/test_base.py
```diff
@@ -878,6 +897,48 @@ def test__reject_sampling_duplicates_valid_rows(self):
         instance._model.sample.assert_any_call(num_rows=5, conditions={'b': 1})
         pd.testing.assert_frame_equal(transformed_data, expected_result)
 
+    def test__reject_sampling_no_valid_rows(self):
+        """Test the ``Constraint.transform`` method's reject sampling fall back.
```
amontanez24, line 914: This docstring should say testing `ColumnsModel._reject_sampling` right?

amontanez24 (approved):
One small comment but besides that I think this is ready!

sdv/constraints/base.py
```diff
@@ -250,44 +169,16 @@ def _validate_data_on_constraint(self, table_data):
...
-            if not self._columns_model:
-                warning_message = (
-                    'When `fit_columns_model` is False and we are conditioning on a subset '
-                    'of the constraint columns, conditional sampling uses reject sampling '
-                    'which can be slow. Changing `fit_columns_model` to True can improve '
-                    'the performance.'
-                )
-                warnings.warn(warning_message, UserWarning)
-
-            all_columns_missing = len(missing_columns) == len(self.constraint_columns)
```
fealho, line 280: Why did you remove this line?

fealho (approved):
Good job! I only had one question, but it looks good to go.

sdv/constraints/base.py
pvk-developer: Since we don't have the `columns_model` in here anymore, and we raise the error whenever a column is missing (doesn't really matter if it's one or all).

PR #842, opened 2022-06-13
Clip corrrectly when using strict boundaries for `Range` and `Inequality` constraints

### LINKED ISSUE #820: Support strict boundaries even when transform strategy is used
### Problem Description

_As a user, it would be useful if I could define range or inequality constraints without having to depend on reject sampling._

### Expected behavior

Currently, the `strict` parameter in the range and inequality constraints can only be enforced using reject sampling. It is not guaranteed to work when set to `True` if the `handling_strategy` is set to `transform`.

- Improve both constraints to support the `strict` parameter being set to `True` when using the `transform` strategy.

### Additional context

- Currently the `strict` parameter is only used to set the `operator` parameter. This is only used in the `is_valid` method which is only used to validate and during `reject_sampling`: 
https://github.com/sdv-dev/SDV/blob/master/sdv/constraints/tabular.py#L368-L371
https://github.com/sdv-dev/SDV/blob/master/sdv/constraints/tabular.py#L454
https://github.com/sdv-dev/SDV/blob/master/sdv/constraints/tabular.py#L741
https://github.com/sdv-dev/SDV/blob/master/sdv/constraints/tabular.py#L845-L849
- We may want to use it to make sure that any values that are equal to the boundary are shifted up to the next acceptable value if `strict` is `True`
    - This is tricky because we will need to know if the next acceptable value is an int or float, and how many places it can be rounded to.

fealho:
Resolve #820.

### DIFF
--- a/sdv/constraints/tabular.py
+++ b/sdv/constraints/tabular.py
@@ -389,6 +389,15 @@ class Inequality(Constraint):
         """
         table_data = table_data.copy()
         diff_column = np.exp(table_data[self._diff_column_name]) - 1
+
+        # Clip the column to ensure the reverse transformed values satisfy the constraint
+        if self._operator is np.greater_equal:
+            diff_column = diff_column.clip(0)
+        elif self._dtype == np.dtype('int'):
+            diff_column = diff_column.clip(1)
+        else:
+            diff_column = diff_column.clip(1e-6)
+
         if self._dtype != np.dtype('float'):
             diff_column = diff_column.round()
 
@@ -523,6 +532,15 @@ class ScalarInequality(Constraint):
         """
         table_data = table_data.copy()
         diff_column = np.exp(table_data[self._diff_column_name]) - 1
+
+        # Clip the column to ensure the reverse transformed values satisfy the constraint
+        if self._operator in [np.less_equal, np.greater_equal]:
+            diff_column = diff_column.clip(0)
+        elif self._dtype == np.dtype('int'):
+            diff_column = diff_column.clip(1)
+        else:
+            diff_column = diff_column.clip(1e-6)
+
         if self._dtype != np.dtype('float'):
             diff_column = diff_column.round()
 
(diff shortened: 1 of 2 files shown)


sdv/constraints/tabular.py
```diff
@@ -389,6 +387,17 @@ def reverse_transform(self, table_data):
...
+
+        # Clip the column to ensure the reverse transformed values satisfy the constraint
+        if self._operator is np.greater_equal:
+            diff_column = diff_column.clip(0)
+        elif self._dtype == np.dtype('int'):
+            diff_column = diff_column.clip(1)
+        elif self._dtype == np.dtype('float'):
+            diff_column = diff_column.clip(1e-6)
+        elif self._is_datetime:
+            diff_column = diff_column.clip(1_000_000_000)
```
fealho, line 399: Datetimes have the same problem as the floats: what's the minimal unit we want to increase it by? Here I'm increasing by 1sec, but maybe the user wants to increase by 1 day? 1 year? 1 nanosecond? 

Again, I don't think this matters too much, as this very rarely (if ever) happens, so I just clip to 1sec.

sdv/constraints/tabular.py
```diff
@@ -285,7 +283,7 @@ def _validate_inputs(low_column_name, high_column_name, strict_boundaries):
         if not isinstance(strict_boundaries, bool):
             raise ValueError('`strict_boundaries` must be a boolean.')
 
-    def __init__(self, low_column_name, high_column_name, strict_boundaries=False,
+    def __init__(self, low_column_name, high_column_name, strict_boundaries=True,
```
fealho, line 286: Changed to match the Range defaults.

tests/unit/constraints/test_tabular.py
```diff
@@ -1201,24 +1246,25 @@ def test_reverse_transform_floats(self):
...
 
         # Assert
         expected_out = pd.DataFrame({
             'a': [1.1, 2.2, 3.3],
             'c': [7, 8, 9],
-            'b': [4.1, 5.2, 6.3],
+            'b': [4.1, 2.200001, 3.300001],
         })
-        pd.testing.assert_frame_equal(out, expected_out)
+        pd.testing.assert_frame_equal(out, expected_out, rtol=1e-8)
```
fealho, line 1259: the default rtol=1e-6 is not precise enough to test the numbers (the test always passes, even when the data is incorrect.)

fealho:
We decided not to clip at all, and instead simply reject sample any invalid values, so this PR was scraped.

PR #843, opened 2022-06-14
Remove handling_strategy parameter

### LINKED ISSUE #833: Remove handling_strategy parameter
### Problem Description

As a user, it is sometimes confusing to know which `handling_strategy` should be used for a constraint. On top of that, if the transform strategy fails, `reject_sampling` could still work so it should be an automatic fall back.

### Expected behavior

- Remove the `handling_strategy` parameter from all constraints
- All constraints should attempt to do the transform strategy, and if that fails because of a `MissingConstraintColumnError`, then it should do nothing to the data and fallback on reject sampling.
    - Raise the following warning if transforming fails:
    ```
    Warning: <constraint name> cannot be transformed because columns [<names>] are not found. Using the reject sampling approach instead.
    ```

### Additional context

- This change will need to be addressed in a couple of other places besides the constraints themselves. 
    - In `metadata.table.py`, there is a method called [_prepare_constraints](https://github.com/sdv-dev/SDV/blob/master/sdv/metadata/table.py#L248-L273) that orders the constraints and raises an error if any constraints touch the same columns. Instead, those constraints should just be set to use `reject_sampling` by simply skipping their transformations. These constraints should still run their `is_valid` check on the data being transformed. This is tricky to handle because if two constraints touch the same columns, then the one that will transform needs to go last meaning we will need a way to know how to skip transformations for certain constraints or to set them to use the `identity` method.
    - In `mtadata.table.py` there is a method called [_transform_constraints](https://github.com/sdv-dev/SDV/blob/master/sdv/metadata/table.py#L600-L616). This method should remove the `on_missing_column` parameter and just always drop (ie. default to reject sampling).

amontanez24:
resolves #833 

TODO:

- Add unit tests
- Decide if constraints need to be ordered


sdv/constraints/base.py
```diff
@@ -91,61 +92,22 @@ class Constraint(metaclass=ConstraintMeta):
...
-        handling_strategy (str):
-            How this Constraint should be handled, which can be ``transform``,
-            ``reject_sampling`` or ``all``.
     """
 
     constraint_columns = ()
     rebuild_columns = ()
     _hyper_transformer = None
+    _use_reject_sampling = False
+    is_custom = False
```
fealho, line 107: Since this is a constant, use capital letters.

sdv/constraints/base.py
```diff
@@ -198,9 +182,22 @@ def transform(self, table_data):
...
+        self._use_reject_sampling = False
+        self._validate_all_columns_present(table_data)
+
+        try:
+            transformed = self._transform(table_data)
+            if self.is_custom:
+                self.reverse_transform(transformed)
+            return transformed
+
+        except Exception:
```
fealho, line 194: Can you use the specific errors?

sdv/metadata/utils.py
```diff
@@ -1,11 +1,12 @@
 """Tools to generate strings from regular expressions."""
 
 import re
-import sre_parse
 import string
 
 import numpy as np
 
+import sre_parse  # isort:skip
```
fealho, line 8: lol

sdv/constraints/base.py
```diff
@@ -198,9 +182,22 @@ def transform(self, table_data):
...
-        self._validate_data_on_constraint(table_data)
-        self.check_missing_columns(table_data)
-        return self._transform(table_data)
+        self._use_reject_sampling = False
+        self._validate_all_columns_present(table_data)
+
+        try:
+            transformed = self._transform(table_data)
+            if self.is_custom:
+                self.reverse_transform(transformed)
```
fealho, line 191: Why is this needed for custom transform?

fealho (approved):
👍 just left some minor comments.

sdv/constraints/base.py
amontanez24: custom constraints can have a noop `transform` but a `reverse_transform` that requires specific columns, so the `transform` won't raise any errors but the `reverse_transform` will.

sdv/constraints/base.py
amontanez24: we want to catch all possible exceptions here

fealho (approved):
Looks good, 2 details.

sdv/constraints/base.py
```diff
@@ -169,16 +127,21 @@ def _validate_data_on_constraint(self, table_data):
 
                 raise ConstraintsNotMetError(err_msg)
 
-    def check_missing_columns(self, table_data):
-        """Check ``table_data`` for missing columns.
+    def _fit(self, table_data):
+        del table_data
```
pvk-developer, line 131: I'm wondering if we need this. Couldn't we just use `pass` or I'm missing something ?

pvk-developer (changes requested):
LGTM! I left a minor change to the `sre_parse` I believe it was fine before:

```
[SDV] ***@lgn:~/Projects/SDV$ isort -df sdv/metadata/utils.py 
--- /home/***/Projects/SDV/sdv/metadata/utils.py:before	2022-06-28 14:17:23.330414
+++ /home/***/Projects/SDV/sdv/metadata/utils.py:after	2022-06-28 14:17:24.234020
@@ -1,11 +1,10 @@
 """Tools to generate strings from regular expressions."""
 
 import re
+import sre_parse
 import string
 
 import numpy as np
-
-import sre_parse
 
 
```

sdv/metadata/utils.py
```diff
@@ -1,11 +1,12 @@
 """Tools to generate strings from regular expressions."""
 
 import re
-import sre_parse
```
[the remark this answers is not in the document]
amontanez24: It fails on windows and mac

sdv/constraints/base.py
amontanez24: we could, this is just from before

fealho (approved):
Still looks good 👍

PR #848, opened 2022-06-17
Add `create_custom_constraint` method

### LINKED ISSUE #836: Add create_custom_constraint factory method
### Problem Description

As a user, it would be easier to create my own custom constraints if the only thing I had to supply was the functions for transforming and validating.

### Expected behavior

- Add a factory class method called `create_custom_constraint` to the `constraints` module.
- This method should take in the following parameters:
    - `is_valid_fn`: required. A function with the following signature `is_valid_fn(column_names, data, **kwargs)`. Must return a series with the rows that are valid for the constraint set to `True`.
    - `transform_fn`: optional. A function with the following signature - `transform_fn(column_names, data, **kwargs)`
    - `reverse_transform_fn`: optional. A function with the following signature - `reverse_transform_fn(column_names, data, **kwargs)`
- The method should return a new constraint class object (not instance) with the `is_valid`, `transform` and `reverse_transform` methods set to what the user provides. The `is_valid` must be provided, and if either of the two tranform methods are provided, then they both must be.
- Delete the current `CustomConstraint` class

### Additional context

- Use the following prototype as inspiration:
```python
def create_custom_constraint(transform_fn=None, reverse_transform_fn=None,
                             is_valid_fn=None):
  
  # User must provide a validity function
  if is_valid_fn is None:
    raise Exception("Missing required parameter 'is_valid'.")
  
  # Transform & reverse are optional but should be provided together or not at all
  if transform_fn is None and reverse_transform_fn is not None:
    raise Exception('No transform is given for the reverse transform.')
  
  if transform_fn is not None and reverse_transform_fn is None:
    raise Exception('No reverse transform is given for the transform.')
  
  # Create a Class that binds to the provided functions
  class CustomConstraint:

    # can keep track of class-level attributes
    TYPE='CUSTOM_CONSTRAINT' # this lets us know how to handle it

    def __init__(self, column_names, **kwargs):
      self.column_names = column_names
      self.kwargs = kwargs
    
    def transform(self, data):
      if transform_fn is None:
        raise Exception('Transform is not defined for this custom constraint.')

      result = transform_fn(self.column_names, data, **self.kwargs)

      if data.shape[0] != result.shape[0]:
        raise Exception('Transformation did not produce the same number of rows as the original')
      return result
    
    def reverse_transform(self, data):
      if reverse_transform_fn is None:
        raise Exception('Reverse transform is not defined for this custom constraint.')

      result = reverse_transform_fn(self.column_names, data, **self.kwargs)
      
      if data.shape[0] != result.shape[0]:
        raise Exception('Reverse transformation did not produce the same number of rows as the original')
      return result
    
    def is_valid(self, data):
      result = is_valid_fn(self.column_names, data, **self.kwargs)
      
      if len(result) != data.shape[0]:
        raise Exception('The validity check should produce exactly 1 True/False value for every row in the data')
      return result

  # return the class to the user 
  return CustomConstraint
```

fealho:
Resolve #836.


sdv/constraints/tabular.py
```diff
@@ -59,69 +73,99 @@ class CustomConstraint(Constraint):
...
+            transform (callable):
+                Function to replace the ``transform`` method.
+            reverse_transform (callable):
+                Function to replace the ``reverse_transform`` method.
+            is_valid (callable):
+                Function to replace the ``is_valid`` method.
+        """
 
-        return table_data
+        TYPE = 'CUSTOM_CONSTRAINT'  # this lets us know how to handle it
```
amontanez24, line 95: Can we either make this a boolean flag like `is_custom` or create an enum class for this? The reason would be that whenever someone is doing logic around this attribute, they would have to know the string value for custom constraints and any other type. It would be easier if they could see all the possible values. Here is an example of enums: https://docs.python.org/3/library/enum.html

Personally though, I think this can just be a boolean

amontanez24 (changes requested):
Could we also add an integration test with a `CustomConstraint`?

sdv/constraints/base.py
```diff
@@ -245,30 +245,6 @@ def filter_valid(self, table_data):
 
         return table_data[valid]
 
-    @classmethod
-    def from_dict(cls, constraint_dict):
```
fealho, line 249: This is not used anywhere else right now (and it causes problems for this PR), which is why it's been deleted.

sdv/constraints/errors.py
```diff
@@ -2,11 +2,18 @@
 
 
 class MissingConstraintColumnError(Exception):
-    """Error to use when constraint is provided a table with missing columns."""
+    """Error used when constraint is provided a table with missing columns."""
+
+    def __init__(self, missing_columns):
+        self.missing_columns = missing_columns
 
     def __init__(self, missing_columns):
```
amontanez24, line 10: looks like it accidentally has 2 inits

sdv/constraints/tabular.py
```diff
@@ -61,70 +77,95 @@ class CustomConstraint(Constraint):
...
-    def _run_is_valid(self, table_data):
-        if self._columns:
-            try:
-                valid = [self._is_valid(table_data, column) for column in self._columns]
-            except TypeError:
-                valid = [self._is_valid(table_data[column]) for column in self._columns]
+            return valid
 
-            return np.logical_and.reduce(valid)
+        def _transform(self, data):
```
amontanez24, line 121: I thought we were going to override the `transform` method and do something to catch errors so that it is possible to fall back on reject sampling. Currently, the logic only falls back if the constraint raises a `MissingConstraintColumnError`, but that can't happen with a custom constraint. We either need to catch most/all errors and surface it as a `MissingConstraintColumnError`, or provide some special logic to handle that

tests/integration/test_constraints.py
```diff
@@ -62,6 +64,9 @@ def test_failing_constraints():
         ScalarInequality('j', '>=', 5.5),
         Range('a', 'b', 'c'),
         ScalarRange('a', 0, 0),
+        create_custom_constraint(
+            lambda _, x: pd.Series([True if x_i > 0 else False for x_i in x['k']])
+        )('k')
```
amontanez24, line 69: minor: can we break this into two lines: One to create the class and one to initialize?

amontanez24 (approved):
One small comment but besides that :shipit:

pvk-developer (approved):
LGTM!

PR #849, opened 2022-06-17
Move `data.copy` to base class of constraints

### LINKED ISSUE #845: Move `table_data.copy` to base class of constraints
It seems every constraint copies the table data in both the transform and reverse transform. Instead of doing that for every constraint, maybe we should just do it in the base class instead.
@amontanez24

fealho:
Resolve #845.

### DIFF
--- a/sdv/constraints/base.py
+++ b/sdv/constraints/base.py
@@ -4,6 +4,7 @@ import copy
 import importlib
 import inspect
 import logging
+import warnings
 
 import pandas as pd
 from copulas.multivariate.gaussian import GaussianMultivariate
@@ -116,7 +117,6 @@ class Constraint(metaclass=ConstraintMeta):
         return table_data
 
     def _identity_with_validation(self, table_data):
-        self._validate_data_on_constraint(table_data)
         return table_data
 
     def __init__(self, handling_strategy):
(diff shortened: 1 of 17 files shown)


sdv/constraints/base.py
```diff
@@ -184,6 +184,7 @@ def transform(self, table_data):
         """
         self._use_reject_sampling = False
         self._validate_all_columns_present(table_data)
+        table_data = table_data.copy()
```
amontanez24, line 187: this may have conflicts when you rebase after the handling strategy issue gets merged. When resolving, can you just move this line to the top of the method?

tests/unit/constraints/test_base.py
```diff
@@ -496,10 +496,10 @@ def test_reverse_transform(self):
         """
         # Run
         instance = Constraint()
-        output = instance.reverse_transform('input')
+        output = instance.reverse_transform(pd.DataFrame())
 
         # Assert
-        assert output == 'input'
+        pd.testing.assert_frame_equal(output, pd.DataFrame())
```
amontanez24, line 502: maybe add a small check to see that it is a different object from what was passed in (like compare id or something)

sdv/constraints/base.py
```diff
@@ -200,6 +200,7 @@ def reverse_transform(self, table_data):
             pandas.DataFrame:
                 Input data unmodified.
         """
+        table_data = table_data.copy()
```
amontanez24, line 203: we should do this for `transform` to no?

fealho: Yep, sorry, I pushed that version to check if I rebased correctly. I just pushed the fixed version with the changes.

amontanez24 (approved):
:shipit:

pvk-developer (approved):
LGTM!

PR #850, opened 2022-06-18
Solve strict boundary clipping issue for `Range` and `Inequality`

### LINKED ISSUE #820: Support strict boundaries even when transform strategy is used

fealho:
Resolve #820.

### DIFF
--- a/sdv/constraints/tabular.py
+++ b/sdv/constraints/tabular.py
@@ -4,6 +4,11 @@ This module contains constraints that are evaluated within a single table,
 and which can affect one or more columns at a time, as well as one or more
 rows.
 
+Note: the data produced by the reverse transform of a constraint does not
+necessarily satisfy that constraint. Any invalid rows produced will have to
+be reject sampled at some later stage. For more information,
+check link.
+
 Currently implemented constraints are:
 
     * CustomConstraint: Simple constraint to be set up by passing the python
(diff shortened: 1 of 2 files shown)


amontanez24 (approved):
LGTM!

pvk-developer (approved):

PR #853, opened 2022-06-23
Improve datetime handling in `ScalarInequality` and `ScalarRange`

### LINKED ISSUE #819: Improve datetime handling in ScalarInequality and ScalarRange constraints
### Problem Description

_As a user, it would be useful if I could provide my datetime value as a string for the `ScalarInequality` and `ScalarRange` constraints. The synthetic data in the end should also match the same format as the input data._

### Expected behavior

- Enforce that string values are used to define datetimes for the `value` parameter of the `ScalarInequality` constraint and the `low_value` and `high_value` parameters of the `ScalarRange` constraints.
- Enforce that the data after `reverse_transform` is in the same datetime format as the data that was fitted.
- Investigate the log warning that sometimes occurs when there are null values in the datetime column.

### Additional context

- For the warning, investigate the cause. If it is a simple fix, implement in this issue, otherwise file another issue.

pvk-developer:
Resolves #819


sdv/constraints/tabular.py
```diff
@@ -424,10 +425,17 @@ def _validate_inputs(column_name, value, relation):
         if not (isinstance(value, (int, float)) or is_datetime_type(value)):
             raise ValueError('`value` must be a number or datetime.')
 
+        if is_datetime_type(value):
+            if not isinstance(value, str):
+                raise ValueError('Datetime must be represented as a string.')
+
+            return cast_to_datetime64(value)
+
+        return value
```
amontanez24, line 434: I don't think the `_validate_inputs` method should return a value. If we want to group the logic we can make a `parse_value` method that calls `_validate_inputs` and then parses to datetime if necessary

sdv/constraints/tabular.py
```diff
@@ -460,6 +468,8 @@ def _fit(self, table_data):
         self._validate_columns_exist(table_data)
         self._is_datetime = self._get_is_datetime(table_data)
         self._dtype = table_data[self._column_name].dtypes
+        if self._is_datetime:
```
amontanez24, line 471: how does the `self._get_is_datetime` method differ from the `is_datetime_type` function? Can we get away with using only one of them? Also, maybe we should just do the datetime validation here since we are already getting whether or not it is a datetime type

sdv/constraints/tabular.py
```diff
@@ -531,6 +541,11 @@ def reverse_transform(self, table_data):
             original_column = self._value - diff_column
 
         table_data[self._column_name] = pd.Series(original_column).astype(self._dtype)
+        if self._is_datetime and self._datetime_format:
```
amontanez24, line 544: is there ever a case where `self._datetime_format` is True, but `self._is_datetime` isn't? There shouldn't be right because the former is only set if the latter is True

sdv/constraints/tabular.py
```diff
@@ -746,32 +761,41 @@ class ScalarRange(Constraint):
             not ``>`` when comparing them.
     """
 
+    @staticmethod
+    def _validate_inputs(low_value, high_value):
```
amontanez24, line 765: again minor, but I don't think validate methods should return anything. There purpose should just be to raise errors

amontanez24:
Has the test coverage been updated? If not it seems like some cases are missing

tests/unit/constraints/test_utils.py
```diff
@@ -92,6 +93,46 @@ def test_is_datetime_type_with_int():
     assert is_datetime is False
 
 
+def test_is_datetime_type_with_datetime_str():
+    """Test the ``is_datetime_type`` function when an valid datetime string is passed.
+
+    Expect to return False when an int variable is passed.
```
fealho, line 99: Should be expected to return True.

tests/unit/constraints/test_utils.py
```diff
@@ -92,6 +93,46 @@ def test_is_datetime_type_with_int():
...
+
+def test_is_datetime_type_with_invalid_str():
+    """Test the ``is_datetime_type`` function when an invalid string is passed.
+
+    Expect to return False when an int variable is passed.
+
+    Input:
+    - string
+    Output:
+    - True
```
fealho, line 124: False

tests/unit/constraints/test_utils.py
```diff
@@ -92,6 +93,46 @@ def test_is_datetime_type_with_int():
...
+    is_datetime = is_datetime_type(value)
+
+    # Assert
+    assert is_datetime
+
+
+def test_is_datetime_type_with_invalid_str():
+    """Test the ``is_datetime_type`` function when an invalid string is passed.
+
+    Expect to return False when an int variable is passed.
```
fealho, line 119: Str, not int

tests/unit/constraints/test_tabular.py
```diff
@@ -2517,21 +2568,20 @@ def test_reverse_transform_is_datetime(self, mock_sigmoid, mock_pd):
...
-        mock_pd.to_datetime.side_effect = lambda x: x
+        mock_pd.to_datetime.side_effect = lambda x: pd.to_datetime('2021-02-02 10:10:59')
+        instance._datetime_format = '%Y-%m-%d'
 
         # Run
-        out = instance.reverse_transform(transformed_data)
+        instance.reverse_transform(transformed_data)
 
         # Assert
-        pd.testing.assert_frame_equal(table_data, out)
```
fealho, line 2532: Why did you delete this check?

fealho (changes requested):
Looking good, I just had a few questions + some typos.

sdv/constraints/tabular.py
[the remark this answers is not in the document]
pvk-developer: Sometimes `self._datetime_format` may not be learnt (it can return `None`) since the learning of the format does not support `ms`: 
```
In [32]: detect_datetime_format('2021-02-02T00:00:00.000000')
In [33]: detect_datetime_format('2021-02-02T00:00:00')
Out[33]: '%Y-%m-%dT%H:%M:%S'
```

sdv/constraints/utils.py
```diff
@@ -5,23 +5,70 @@
...
+
+    Args:
+        value (pandas.Series, np.ndarray, list, or str):
+            Input data to convert to ``numpy.datetime64``.
+
+    Return:
+        ``numpy.datetime64`` value or values.
+    """
+    if isinstance(value, str):
+        value = pd.to_datetime(value).to_datetime64()
```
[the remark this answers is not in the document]
pvk-developer: I used it to ensure it ends up being `numpy.datetime64[ns]` that way all the datetimes are in the same magnitude.

sdv/constraints/tabular.py
pvk-developer: If `self._is_datetime` is `True` but we are not able to `detect` the format of the date time, then we can't format with `None`.

sdv/constraints/tabular.py
pvk-developer: `_get_is_datetime` checks both, value and column, to be `datetime`. The reason behind is that we end raising a value error inside that function when one of the values is `datetime` and the other one is not.

amontanez24 (approved):
Just a couple small comments and address Felipe's comment on the test file and then this is good to go!

amontanez24 (approved):

fealho (approved):

PR #862, opened 2022-07-01
Unify sampling params for reject sampling

### LINKED ISSUE #809: Unify sampling params for reject sampling
### Problem Description

1. Our sampling methods assume `sample` (for all models) and `GaussianCopula` would never use reject sampling. This is incorrect. Due to constraints, any sampling method for any model might need reject sampling.
2. We are using `batch_size` and `batch_size_per_try` as interchangeable concepts when they are opposites: `batch_size` is intended to be small (to control output and save progress), but `batch_size_per_try` is intended to be large (for reject sampling)

### Expected behavior

Have the same parameters for all sampling methods (`sample`, `sample_conditions` and `sample_remaining_columns`) for all single table models.

- `batch_size` which will be used in the intended way: Users should be able to split up large sampling tasks into smaller sizes
  -  Requirement: `batch_size <= total num_rows`
  - Default: `batch_size = total num rows`
- `max_tries_per_batch`, which will be used when there is any kind of reject sampling involved
  - Default: `10`

#### Logic

For every batch, for every try: Dynamically update the # of rows we try to sample to achieve the full `batch_size`.

#### TabularPreset
- Now that GaussianCopula has the same sampling params as other models, we need a special case for it. Make sure this works. (ref: #796)

amontanez24:
resolves #809 

This PR:

- Changes the `batch_size_per_try` param to be `batch_size` everywhere
- Changes the `max_tries` param to be `max_tries_per_batch`
- Uses the progress bar all the time since it is relevant for constraints
- Changes the logic of `_sample_batch` to attempt to sample the full batch size. If it fails, it updates the number of rows requested for the next call based on the success rate of the previous call. It caps at 10x the batch size since very large calls can be slow for certain models


sdv/tabular/base.py
```diff
@@ -504,23 +503,16 @@ def sample(self, num_rows, randomize_samples=True, batch_size=None, output_file_
...
-                        batch_size_per_try=batch_size,
+                        batch_size=batch_size,
                         progress_bar=progress_bar,
                         output_file_path=output_file_path,
                     )
                     sampled.append(sampled_rows)
 
-                return sampled
-
-            if batch_size == num_rows:
```
katxiao, line 519: Do we still want to hide the progress bar if batch_size = num_rows? I guess it's fine to show it in that case, but it will just be one 0 -> 100 increment.

katxiao (approved):
lgtm with one small comment

sdv/tabular/base.py
amontanez24: Oh, the reason I put it is because we could get reject sampling with constraints and so the progress bar would update

sdv/tabular/base.py
```diff
@@ -265,13 +266,13 @@ def _sample_rows(self, num_rows, conditions=None, transformed_conditions=None,
...
 
-    def _sample_batch(self, num_rows=None, max_tries=100, batch_size_per_try=None,
+    def _sample_batch(self, batch_size=None, max_tries=100,
                       conditions=None, transformed_conditions=None, float_rtol=0.01,
                       progress_bar=None, output_file_path=None):
         """Sample a batch of rows with the given conditions.
 
         This will enter a reject-sampling loop in which rows will be sampled until
-        all of them are valid and match the requested conditions. If `max_tries`
+        all of them are valid and match the requested conditions. If `max_tries_per_batch`
```
pvk-developer, line 275: here we use `max_tries_per_batch` however we take as input `max_tries`.

sdv/tabular/base.py
```diff
@@ -683,32 +638,25 @@ def _sample_conditions(self, conditions, max_tries, batch_size_per_try, randomiz
...
-                sampled = progress_bar_wrapper(_sample_function, num_rows, 'Sampling conditions')
-
             check_num_rows(
-                len(sampled),
-                num_rows,
-                (hasattr(self, '_model') and not isinstance(
+                num_rows=len(sampled),
+                expected_num_rows=num_rows,
+                is_reject_sampling=(hasattr(self, '_model') and not isinstance(
                     self._model, copulas.multivariate.GaussianMultivariate)),
```
pvk-developer, line 657: Could we store this in a variable and just pass it down here?

pvk-developer (approved):
Just left some minor comments but LGTM!

sdv/tabular/base.py
katxiao: Ah okay, that's fine then.

PR #872, opened 2022-07-07
Numerical Instability in conditioned constraint

### LINKED ISSUE #806: Numerical Instability in Constrained GaussianCopula
### Environment Details

* SDV version: 0.14.2.dev0
* Python version: Python 3.7.13
* Operating System: Linux 5.4.0-88-generic x86_64 GNU/Linux

### Error Description

When using **integer** values in a constrained GaussianCopula, numerical instability can cause valid inputs to fail to produce any rows with the `sample_conditions()` call. This doesn't happen for all inputs, but the bug can be reliably produced for some value(s) in the range of valid inputs for virtually any `Between` constraint instance. I do not believe the problem is strictly isolated to `Between`, but I have not attempted to reproduce the error for other constraint classes.

### What happens
Given a particular `Between` constraint bounded by `low` and `high`, conditional sampling of some values such that `low < x < high` fails to produce any values, but other valid inputs work as expected. The following exception traceback will be generated, erroneously claiming that the input value violated the constraint:
```python
Traceback (most recent call last):
  File "SDV/sdv/tabular/copulas.py", line 285, in sample_conditions
    conditions, 100, batch_size, randomize_samples, output_file_path)
  File "SDV/sdv/tabular/base.py", line 715, in _sample_conditions
    handle_sampling_error(output_file_path == TMP_FILE_NAME, output_file_path, error)
  File "SDV/sdv/tabular/utils.py", line 175, in handle_sampling_error
    raise sampling_error
  File "SDV/sdv/tabular/base.py", line 711, in _sample_conditions
    batch_size_per_try,
  File "SDV/sdv/tabular/utils.py", line 224, in check_num_rows
    raise ValueError(user_msg)
ValueError: Unable to sample any rows for the given conditions. This may be because the provided values are out-of-bounds in the current model.
Please try again with a different set of values.
```

### What should happen
The function call is expected to work without issue for all valid integer inputs when constraints are specified as integer types.

### Steps to reproduce

```python
import sdv 
from sdv.constraints import Between
from sdv.sampling.tabular import Condition
import pandas as pd
import numpy as np

# The high and low values alter which exact integers are unstable
# Another script can be used to search for unstable conversions for arbitrary low/high values
constraint_input = Between(column='input', low=49, high=100)
model = GaussianCopula(
            field_names=['input', 'output'],
            field_transformers = {'input': 'integer', # Problematic conversions may occur
                                  'output': 'float',},
            constraints=[constraint_input],
            min_value = None,
            max_value = None)
# The particular data (and amount used) do not matter, but should be present for the model to have a sampling basis
i, j = 50, 80
arbitrary_data = pd.DataFrame(data={'input': [_ for _ in range(i,j)],
                                    'output': [np.random.rand() for _ in range(j-i)]})
model.fit(arbitrary_data)
# In this case of low=49, high=100; input=88 is the only unstable value
conditions = Condition({'input': 88}, num_rows=3)
output = model.sample_conditions([conditions])
```

Running the above as `mwe.py`:
```python
$ python mwe.py 
Sampling conditions:   0%|         | 0/3 [00:00<?, ?it/s]
Traceback (most recent call last):
  File "mwe.py", line 25, in <module>
    output = model.sample_conditions([conditions])
  File "SDV/sdv/tabular/copulas.py", line 285, in sample_conditions
    conditions, 100, batch_size, randomize_samples, output_file_path)
  File "SDV/sdv/tabular/base.py", line 715, in _sample_conditions
    handle_sampling_error(output_file_path == TMP_FILE_NAME, output_file_path, error)
  File "SDV/sdv/tabular/utils.py", line 175, in handle_sampling_error
    raise sampling_error
  File "SDV/sdv/tabular/base.py", line 711, in _sample_conditions
    batch_size_per_try,
  File "SDV/sdv/tabular/utils.py", line 224, in check_num_rows
    raise ValueError(user_msg)
ValueError: Unable to sample any rows for the given conditions. This may be because the provided values are out-of-bounds in the current model. 
Please try again with a different set of values.
```

In the particular example above, debugging can show that during the reverse transformation, the interval floating point value is 87.9999999 rather than 88.0, so casting to an integer demotes it to 87 and makes all sampled rows fail to pass the constraint.

### Known Workarounds

I believe the issue should be fixed for SDV's benefit, but in the meantime it is possible to get around the issue by avoiding SDV's integer conversions. However, it would be ideal if the numerical stability was reliable on integer values as expected.

* Using the same values in floating point format for each of the constraint, sampling input, and training data avoids the integer conversion/numerical instability as it is smaller than `rtol`'s default value.

* Instead of using integers, using floating point values `low=0`, `high=1` and doing the appropriate data conversions (ie: `x = x*(high-low)+low` and `x = (x-low)/(high-low)`, where `high` and `low` are the desired integer constraint bounds) before modeling/requesting samples and after sampling can get around the issue.

fealho:
Resolve #806.


sdv/metadata/table.py
```diff
@@ -445,7 +445,8 @@ def _transform_constraints(self, data, is_condition=False):
         for constraint in self._constraints:
             try:
                 data = constraint.transform(data)
-                self._constraints_to_reverse.append(constraint)
+                if not is_condition:
+                    self._constraints_to_reverse.append(constraint)
```
amontanez24, line 449: I think if you rebase this will already be done

sdv/metadata/table.py
```diff
@@ -698,6 +699,9 @@ def reverse_transform(self, data):
             else:
                 field_data = reversed_data[name]
 
+            if field_metadata['transformer'] == 'integer':
+                field_data = field_data.round()
```
amontanez24, line 703: can we add a unit test for this line?

tests/integration/constraints/test_tabular.py
```diff
@@ -148,3 +149,28 @@ def test_ScalarRange():
     # Assert
     assert sampled.column.min() >= 5
     assert sampled.column.max() <= 11
+
+
+def test_ScalarRange_conditions():
+    """Test ``ScalarRange`` with conditions."""
```
amontanez24, line 155: minor, but can we add to the description to say why we needed to check this (to make sure the conditions weren't being altered by the constraint transformation)

sdv/metadata/table.py
fealho: Sure. I didn't add any because there were no reverse_transform tests in the first place, presumably because it gets tested by the integration ones.

tests/unit/metadata/test_table.py
```diff
@@ -621,6 +621,40 @@ def test__transform_constraints_is_condition_false_returns_data(self):
...
+        table._constraints_to_reverse = []
+        table._dtypes = {'bar': 'int'}
+        table._field_names = ['bar']
+
+        # Run
+        output = table.reverse_transform(data)
+
+        # Assert
+        expected_data = pd.DataFrame({'bar': [0, 2, 2]})
+        pd.testing.assert_frame_equal(output, expected_data, check_dtype=False)
```
amontanez24, line 656: why was this failing with the dtype? Shouldn't it be converted back into an int?

fealho: Because Windows has `int32` instead of `int64`.

amontanez24 (approved):
:shipit:

pvk-developer (approved):

PR #882, opened 2022-07-08
Fix error message for ScalarInequality and ScalarRange.

### LINKED ISSUE #868: Fix error message in ScalarInequality
### Environment Details

* SDV version: 0.16.0.dev0
* Python version: Any
* Operating System: Any

### Error Description
If I pass in an invalid value as the "value" parameter, the error says:
```ValueError: `value` must be a number or datetime.```

This is not technically correct. The `value` should be a string that represents a datetime.
New Error should say:
```Value Error: `value` must be a number or a string that represents a datetime.```

### Steps to reproduce

```python
constraint = ScalarInequality(column_name='age', relation='>=', value=False)
```

pvk-developer:
Resolves #868


fealho (approved):

amontanez24 (approved):
:shipit:

PR #898, opened 2022-07-11
Update rdt to v1 0 0

### LINKED ISSUE #897: Update the RDT version to 1.0
Update the current `RDT` version to `1.0` and the new `API` that this presents. Also update `CTGAN` and `SDMetrics` to the new release candidates.

pvk-developer:
Resolves #897


sdv/metadata/table.py
```diff
@@ -404,25 +417,24 @@ def _get_transformers(self, dtypes):
...
+                    transformer = transformer_template()
+                else:
+                    transformer = copy.deepcopy(transformer_template)
 
-            LOGGER.debug('Loading transformer %s for field %s',
-                         transformer.__class__.__name__, name)
-            transformers[name] = transformer
+                LOGGER.debug('Loading transformer %s for field %s',
+                             transformer.__class__.__name__, name)
+                transformers[name] = transformer
```
amontanez24, line 437: why are these indented?

sdv/metadata/table.py
```diff
@@ -513,10 +525,17 @@ def _fit_hyper_transformer(self, data, extra_columns):
...
+        self._hyper_transformer = rdt.HyperTransformer()
+        self._hyper_transformer.detect_initial_config(data)
+        if transformers_dict:
+            self._hyper_transformer.update_transformers(transformers_dict)
 
-        self._hyper_transformer = rdt.HyperTransformer(field_transformers=transformers_dict)
-        self._hyper_transformer.fit(data[list(transformers_dict.keys())])
+        try:
+            self._hyper_transformer.fit(data)
+        except rdt.errors.Error:
```
amontanez24, line 537: why do we need to catch errors here?

sdv/metadata/dataset.py
```diff
@@ -419,64 +419,14 @@ def _get_pii_fields(self, table_name):
...
-        transformers_dict = dict()
-        for name, dtype in dtypes.items():
-            dtype = np.dtype(dtype)
-            if dtype.kind == 'i':
-                transformer = transformers.NumericalTransformer(dtype=int)
-            elif dtype.kind == 'f':
-                transformer = transformers.NumericalTransformer(dtype=float)
-            elif dtype.kind == 'O':
-                anonymize = pii_fields.get(name)
-                transformer = transformers.CategoricalTransformer(anonymize=anonymize)
```
amontanez24, line 448: I can't even find this `anonymize` parameter in the currently used versions of rdt

sdv/metadata/table.py
```diff
@@ -682,9 +701,16 @@ def reverse_transform(self, data):
...
+            for column in self._hyper_transformer._output_columns
+            if column in data.columns
+        ]
         try:
-            reversed_data = self._hyper_transformer.reverse_transform(data)
-        except rdt.errors.NotFittedError:
+            reversed_data = self._hyper_transformer.reverse_transform_subset(
+                data[reversable_columns]
+            )
+        except (rdt.errors.NotFittedError, rdt.errors.Error):
```
amontanez24, line 713: what are the scenarios where we need to catch these other errors

sdv/metadata/dataset.py
```diff
@@ -495,14 +445,18 @@ def transform(self, table_name, data):
...
         if hyper_transformer is None:
-            hyper_transformer = self._load_hyper_transformer(table_name)
-            fields = list(hyper_transformer.transformers.keys())
-            hyper_transformer.fit(data[fields])
+            hyper_transformer = HyperTransformer()
+            hyper_transformer.detect_initial_config(data)
+            pii_fields = self._get_pii_fields(table_name)
+            if pii_fields:
+                # TODO: We have to define the pii and how we should pass to ``AnonymizedFaker``.
+                self._update_hyper_transformer(pii_fields, hyper_transformer)
```
amontanez24, line 453: I think this part can probably just be dropped

amontanez24 (commented):
This is looking mostly good! I just have some questions about errors that we are catching. It might be worth making sure that there aren't bugs in the `HyperTransformer` that we need to fix

sdv/constraints/base.py
```diff
@@ -323,9 +324,9 @@ def fit(self, table_data):
                 Table data.
         """
         data_to_model = table_data[self.constraint_columns]
-        self._hyper_transformer = HyperTransformer(
-            default_data_type_transformers={'categorical': 'OneHotEncodingTransformer'}
-        )
+        self._hyper_transformer = HyperTransformer()
+        self._hyper_transformer.detect_initial_config(data_to_model)
```
katxiao, line 328: Is it ok that we are detecting the config here? I know that's probably what we were doing previously, but I'm confused about how that works.

sdv/metadata/table.py
pvk-developer: Those are indented in order to be able to use a ``None`` transformer from the lines above:
```python
            if self._dtype_transformers[np.dtype(dtype).kind] is None:
                transformers[name] = None

            elif transformer_template is None:
```

sdv/metadata/table.py
pvk-developer: Fixed by using `NotFittedError`. Before the `hypertransformer` was always being fitted, and some cases we couldn't reverse transform which raised some `Error` from `RDT`.

sdv/constraints/base.py
pvk-developer: I think that it is ok. The reason is that mainly we get the `data_to_model` which is a subset of the dataframe, we then `fit_transform` only that portion of the data.

sdv/metadata/table.py
pvk-developer: This error was produced because the `data` was empty, we check now if the `data` is empty we just don't fit.

pvk-developer:
Closing this `PR` since we want to merge to `master` with slightly different approach: https://github.com/sdv-dev/SDV/pull/923

PR #908, opened 2022-07-18
Add `load_from_json` and `save_to_json`  to `SingleTableMetadata`

### LINKED ISSUE #874: Add load_from_json and save_to_json methods to SingleTableMetadata
### Problem Description

As a user, it would be useful if I could create an instance of the `SingleTableMetadata` class from a json file. It would also be useful if I could save an existing instance to a json file.

### Expected behavior

- Add a `load_from_json` method
    - Parameters:
        - `filepath`: string that is the path to json file
    - Errors:
        - If the json file doesn't contain the `SCHEMA_VERSION` attribute, then raise the following error
        ```python
        Error: This metadata file is incompatible with the SingleTableMetadata class and version.
        ``` 
```python
>>> metadata = SingleTableMetadata.load_from_json(filepath='project/metadata.json')
```
- Add a `save_to_json` method
    - Add `SCHEMA_VERSION` to dict privately
    - Parameters:
        - `filepath`: string that is the path to save the file to
    - Errors:
        - If there is already a file in the path location, raise the following error
        ```python
        Error: A file named 'metadata.json' already exists in this folder. Please specify a different filename.
        ```
```python
>>> metadata.save_to_json(filepath='project/metadata.json')
```

pvk-developer:
Resolves #874


tests/unit/metadata/test_single_table.py
```diff
@@ -105,6 +109,183 @@ def test__load_from_dict(self):
...
+            - String representing a filepath.
+
+        Side Effects:
+            - A ``ValueError`` is raised pointing that the ``file`` does not exist.
+        """
+        # Setup
+        mock_path.return_value.exists.return_value = False
+        mock_path.return_value.name = 'filepath.json'
+
+        # Run
```
katxiao, line 131: small nit: I have seen the comment `# Run and Assert` in cases like these.

katxiao (approved):
LGTM!

tests/unit/metadata/test_single_table.py
pvk-developer: Nice catch, I usually write `# Run / Assert` but this time I missed it. WIll address it asap.

amontanez24 (changes requested):
One thing concerning constraints, but the rest looks good!

Issue #910, opened 2022-07-18
Conditional sampling using `GaussianCopula` inefficient when categories are noised


npatki:
### Environment Details

Please indicate the following details about the environment in which you found the bug:

* SDV version: 0.16.0.dev2
* Python version: 3.7
* Operating System: Colab Notebook

### Error Description

The `GaussianCopula` is listed as the most efficient way to perform conditional sampling. Yet, there are some configurations of this model that are very inefficient compared to others.

Inefficient configurations: `categorical_transformer='categorical_fuzzy'`
Efficient configurations (**up to 100x faster**): `categorical_transformer='categorical'` or `'label-encoding'` or `'one_hot_encoding'`

### Steps to reproduce

The following is inefficient:

```python
from sdv.demo import load_tabular_demo
from sdv.tabular import GaussianCopula
from sdv.sampling import Condition

data = load_tabular_demo('student_placements')
data.head()

model = GaussianCopula() # no constraints
model.fit(data)

condition1 = Condition({'gender': 'F', 'high_spec': 'Science'}, num_rows=100)
condition2 = Condition({'gender': 'M', 'high_spec': 'Science'}, num_rows=100)

model.sample_conditions(conditions=[condition1, condition2])

Sampling conditions: 93%|█████████▎| 186/200 [02:05<00:09, 1.49it/s]
/usr/local/lib/python3.7/dist-packages/sdv/tabular/utils.py:211: UserWarning: Only able to sample 186 rows for the given conditions. To sample more rows, try increasing `max_tries_per_batch` (currently: 100). Note that increasing this value will also increase the sampling time.
 warnings.warn(user_msg)
```

Meanwhile, changing it to `label_encoding` is 100x faster:
```python
model = GaussianCopula(categorical_transformer='label_encoding')
...
Sampling conditions: 100%|██████████| 200/200 [00:00<00:00, 327.30it/s]
```

PR #911, opened 2022-07-18
Update release notes `v0.16.0`

pvk-developer:


pvk-developer:
> This looks good to me. Will just need to add #910 when that gets merged and update the date. I can do that if it gets finished by today

It's already added 😅

fealho (approved):

PR #913, opened 2022-07-20
only showing progress bar for single table

amontanez24:
This PR solves the problem of tons of progress bars being printed in the multi-table sample case. It also only shows the progress bar for `sample` if there are multiple batches to sample or constraints.

### DIFF
--- a/sdv/tabular/base.py
+++ b/sdv/tabular/base.py
@@ -481,6 +481,57 @@ class BaseTabularModel:
         else:
             self._set_random_state(FIXED_RNG_SEED)
 
+    def _sample_with_progress_bar(self, num_rows, randomize_samples=True, max_tries_per_batch=100,
+                                  batch_size=None, output_file_path=None, conditions=None,
+                                  show_progress_bar=True):
+        if conditions is not None:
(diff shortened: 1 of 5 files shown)


sdv/tabular/base.py
```diff
@@ -525,13 +503,23 @@ def sample(self, num_rows, randomize_samples=True, max_tries_per_batch=100, batc
...
+                        output_file_path=output_file_path
+                    )
+
+            else:
                 sampled = self._sample_in_batches(
                     num_rows=num_rows,
                     batch_size=batch_size,
                     max_tries_per_batch=max_tries_per_batch,
-                    progress_bar=progress_bar,
+                    progress_bar=None,
```
pvk-developer, line 522: we could remove the `if - else` by only adding `tqdm.tqdm(total=num_rows, disable=not show_progress_bar)` (When `True` is not being shown, when `False` is being shown, defaults to `False`).

pvk-developer (changes requested):

pvk-developer (approved):

katxiao (approved):
lgtm!

PR #914, opened 2022-07-20
Add `_validate_inputs` class method to constraints

### LINKED ISSUE #878: Add _validate_inputs class method to each constraint
### Problem Description

As a developer, it would be useful to check that the provided parameters for each constraint are valid since this is used when validating the metadata and also when creating a constraint.

### Expected behavior

- Add `_validate_inputs` class method to each constraint
- The specific functionality differs for each constraint and is described below
- Should catch and raise all errors together

### Additional context
- **Constraint validation**
    - `Unique`
        - "column_names" must be present
        `Error: Missing required "column_names" in the Unique constraint`
        - No additional attributes should be present
        `Error: Invalid values ("pii") are present in a Unique constraint.`
    - `FixedCombinations`
        - "column_names" must be present
        `Error: Missing required values ("column_names") in a FixedCombinations constraint`
        - No additional attributes should be present
        `Error: Invalid values ("pii") are present in a FixedCombinations constraint.`
    - `Inequality`
        - "high_column_name" and "low_column_name" must both be present 
        `Error: Missing required values ("high_column_name", "low_column_name") in an Inequality constraint`
        - No additional attributes should be present
        `Error: Invalid values ("pii") are present in a FixedCombinations constraint.`
    - `ScalarInequality`
        - "relation", "column_name" and "value" must all be present
        `Error: Missing required values ("relation", "column_name", "value") in a ScalarInequality constraint`
        - "relation" must be one of  ">", ">=", "<", "<="
        `Error: Invalid relation value ("==") in a ScalarInequality constraint. The relation must be one of: ">", ">=", "<" or "<=".`
        - No additional attributes should be present
        `Error: Invalid values ("pii") are present in a ScalarInequality constraint.`
    - `Range`
        - "high_column_name", "between_column_name" and "low_column_name" must both all present 
        `Error: Missing required values ("high_column_name", "between_column_name", "low_column_name") in a Range constraint.`
        - No additional attributes should be present
        `Error: Invalid values ("pii") are present in a Range constraint.`
    - `ScalarRange`
        - Required params must be present
        `Error: Missing required values ("column_name", "low_value", "high_value") in a ScalarRange constraint.`
        - No additional attributes should be present
        `Error: Invalid values ("pii") are present in a ScalarRange constraint.`
    - `Positive`
        - Required params must be present
        `Error: Missing required values ("column_name") in a Positive constraint.`
        - No additional attributes should be present
        `Error: Invalid values ("pii") are present in a Positive constraint.`
    - `Negative`
        - Required params must be present
        `Error: Missing required values ("column_name") in a Negative constraint.`
        - No additional attributes should be present
        `Error: Invalid values ("pii") are present in a Negative constraint.`
    - `FixedIncrements`
        - All required values should be present
        `Error: Missing required values ("column_name", "increment") in a FixedIncrements constraint`
        - increment_value should be an integer >0
        `Error: Invalid increment value (0.5) in a FixedIncrements constraint. Increments must be positive integers.`
        - No additional attributes should be present
        `Error: Invalid values ("pii") are present in a FixedIncrements constraint.`
    - OneHotEncoding
        - "column_names" must be present
        `Error: Missing required values ("column_names") in a OneHotEncoding constraint`
        - No additional attributes should be present
        `Error: Invalid values ("pii") are present in a OneHotEncoding constraint.`
    - `CustomConstraint`
        - "column_names" must be present
        `Error: Missing required values ("column_names") in a <name> constraint`

fealho:
Resolve #878.

### DIFF
--- a/sdv/constraints/base.py
+++ b/sdv/constraints/base.py
@@ -117,7 +117,7 @@ class Constraint(metaclass=ConstraintMeta):
                 f'Invalid values {invalid_values} are present in {constraint} constraint.'
             ))
 
-        raise Exception(errors)
+        return errors
 
     def _validate_data_meets_constraint(self, table_data):
         """Make sure the given data is valid for the constraint.
--- a/sdv/constraints/base.py
+++ b/sdv/constraints/base.py
@@ -10,7 +10,7 @@ from copulas.multivariate.gaussian import GaussianMultivariate
 from copulas.univariate import GaussianUnivariate
 from rdt import HyperTransformer
 
-from sdv.constraints.errors import MissingConstraintColumnError
+from sdv.constraints.errors import MissingConstraintColumnError, MultipleConstraintsErrors
 from sdv.errors import ConstraintsNotMetError
 
 LOGGER = logging.getLogger(__name__)
@@ -102,22 +102,41 @@ class Constraint(metaclass=ConstraintMeta):
     constraint_columns = ()
     _hyper_transformer = None
 
-    @staticmethod
-    def _validate_inputs(valid_parameters, constraint, **kwargs):
+    @classmethod
+    def _validate_inputs(cls, **kwargs):
         errors = []
-        missing_values = valid_parameters - set(kwargs)
+        args = []
+        params = inspect.signature(cls).parameters
+        for arg_name, value in params.items():
+            if value.default is inspect._empty:
+                args.append(arg_name)
+
+        missing_values = set(args) - set(kwargs)
+        constraint_name = cls.__name__
         if missing_values:
-            errors.append(ValueError(
-                f'Missing required values {missing_values} in {constraint} constraint.'
-            ))
+            if constraint_name == 'Inequality':
+                errors.append(ValueError(
+                    f'Missing required values {missing_values} in an {constraint_name} constraint.'
+                ))
+            else:
+                errors.append(ValueError(
+                    f'Missing required values {missing_values} in a {constraint_name} constraint.'
+                ))
 
-        invalid_values = set(kwargs.keys()) - valid_parameters
+        invalid_values = set(kwargs) - set(args)
         if invalid_values:
-            errors.append(ValueError(
-                f'Invalid values {invalid_values} are present in {constraint} constraint.'
-            ))
+            if constraint_name == 'Inequality':
+                errors.append(ValueError(
+                    f'Invalid values {invalid_values} are present in an'
+                    f' {constraint_name} constraint.'
+                ))
+            else:
+                errors.append(ValueError(
+                    f'Invalid values {invalid_values} are present in a'
+                    f' {constraint_name} constraint.'
+                ))
 
-        return errors
+        raise MultipleConstraintsErrors('\n' + '\n\n'.join(map(str, errors)))
 
     def _validate_data_meets_constraint(self, table_data):
         """Make sure the given data is valid for the constraint.
--- a/sdv/constraints/base.py
+++ b/sdv/constraints/base.py
@@ -112,29 +112,18 @@ class Constraint(metaclass=ConstraintMeta):
                 args.append(arg_name)
 
         missing_values = set(args) - set(kwargs)
-        constraint_name = cls.__name__
+        constraint = cls.__name__
+        article = 'an' if constraint == 'Inequality' else 'a'
         if missing_values:
-            if constraint_name == 'Inequality':
-                errors.append(ValueError(
-                    f'Missing required values {missing_values} in an {constraint_name} constraint.'
-                ))
-            else:
-                errors.append(ValueError(
-                    f'Missing required values {missing_values} in a {constraint_name} constraint.'
-                ))
+            errors.append(ValueError(
+                f'Missing required values {missing_values} in {article} {constraint} constraint.'
+            ))
 
-        invalid_values = set(kwargs) - set(args)
-        if invalid_values:
-            if constraint_name == 'Inequality':
-                errors.append(ValueError(
-                    f'Invalid values {invalid_values} are present in an'
-                    f' {constraint_name} constraint.'
-                ))
-            else:
-                errors.append(ValueError(
-                    f'Invalid values {invalid_values} are present in a'
-                    f' {constraint_name} constraint.'
-                ))
+        invalid_vals = set(kwargs) - set(args)
+        if invalid_vals:
+            errors.append(ValueError(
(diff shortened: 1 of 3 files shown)


sdv/constraints/tabular.py
```diff
@@ -206,6 +213,15 @@ class FixedCombinations(Constraint):
...
     _uuids_to_combinations = None
 
+    @classmethod
+    def _validate_inputs(cls, **kwargs):
+        errors = super()._validate_inputs(
+            valid_parameters={'column_names'},
+            constraint='a FixedCombinations',
+            **kwargs
+        )
+        raise MultipleConstraintsErrors('\n' + '\n\n'.join(map(str, errors)))
```
amontanez24, line 223: each of these just passes the parameters from the init right? And the constraint parameter is always 'a' + the name of the constraint? If so, I think all of this can just be moved into the base. We can use [introspection ](https://stackoverflow.com/questions/36849837/introspecting-arguments-from-the-constructor-function-init-in-python) to get the names of the required parameters for the class, and just access the name directly (ie. `class.__name__`). I don't think each class needs to override this method

amontanez24 (changes requested):
I think a lot of the classes can just use the base method if we make the base method use introspection to get the parameters names from `__init__`

sdv/constraints/tabular.py
fealho: Sure, I can go for it. I'll have to add an if statement in the base to catch the`**an** Inequality` article, but that's minor. I'll also have to make the base raise the error, instead of returning a list of strings, so the two classes that do overwrite `_validate_inputs` will have to try/catch the error.

amontanez24: Yeah, or the base method can just gather the strings still, and another method can call it and raise the `MultipleConstraintsError`

sdv/constraints/base.py
```diff
@@ -102,6 +102,42 @@ class Constraint(metaclass=ConstraintMeta):
...
+        for arg_name, value in params.items():
+            if value.default is inspect._empty:
+                args.append(arg_name)
+
+        missing_values = set(args) - set(kwargs)
+        constraint_name = cls.__name__
+        if missing_values:
+            if constraint_name == 'Inequality':
+                errors.append(ValueError(
+                    f'Missing required values {missing_values} in an {constraint_name} constraint.'
```
fealho, line 119: I don't really like these if statements. @amontanez24 @pvk-developer what do you guys think of deleting it, so the error would just say `a Inequality constraint` instead of `an Inequality constraint`. It doesn't sound too wrong to me, and I'd rather the code be cleaner than a very minor grammatical thing.

amontanez24: instead of doing the ifs we can just ake a variable ccalled article and use that in the f-string. eg.
```python
article = 'an' if constraint_name == 'Inequality' else 'a'
errors.append(ValueError(f'Missing required values {missing_values} in {article} {constraint_name} constraint.'))
```

amontanez24 (approved):
I left a comment about removing the ifs for the `Inequality` cases. Once that change is done this should be good to go!

sdv/constraints/base.py
pvk-developer, line 119: I think that the message should be with `('param'...)` instead of `{'param'...}` (tuple vs set).  You could convert the `missing_values` to string:
```python
missing_values = list(missing_values)
# This way the list is always in the same order and you can assert easier the tests
missing_values.sort()
missing_values = ', '.join(missing_values)
```

sdv/constraints/tabular.py
```diff
@@ -97,6 +97,13 @@ class CustomConstraint(Constraint):
                 Function to replace the ``is_valid`` method.
         """
 
+        @classmethod
+        def _validate_inputs(cls, **kwargs):
+            if 'column_names' not in set(kwargs):
+                raise MultipleConstraintsErrors(
+                    "Missing required values {'column_names'} in a CustomConstraint constraint."
```
pvk-developer, line 104: `()` instead of `{}`

sdv/constraints/tabular.py
```diff
@@ -459,8 +466,26 @@ class ScalarInequality(Constraint):
...
+        errors = []
+        try:
+            super()._validate_inputs(**kwargs)
+        except Exception as e:
+            errors.append(str(e))
+
+        if 'relation' in kwargs and kwargs['relation'] not in {'>', '>=', '<', '<='}:
+            wrong_relation = {kwargs['relation']}
+            errors.append(
+                f'Invalid relation value {wrong_relation} in a ScalarInequality constraint.'
```
pvk-developer, line 480: `()` instead of `{}`

sdv/constraints/tabular.py
```diff
@@ -954,6 +980,24 @@ class FixedIncrements(Constraint):
...
+        errors = []
+        try:
+            super()._validate_inputs(**kwargs)
+        except Exception as e:
+            errors.append(str(e))
+
+        if 'increment_value' in kwargs and kwargs['increment_value'] <= 0:
+            wrong_increment = {kwargs['increment_value']}
+            errors.append(
+                f'Invalid increment value {wrong_increment} in a FixedIncrements constraint.'
```
pvk-developer, line 994: `()` instead of `{}`

sdv/constraints/base.py
```diff
@@ -102,6 +102,42 @@ class Constraint(metaclass=ConstraintMeta):
...
+            else:
+                errors.append(ValueError(
+                    f'Missing required values {missing_values} in a {constraint_name} constraint.'
+                ))
+
+        invalid_values = set(kwargs) - set(args)
+        if invalid_values:
+            if constraint_name == 'Inequality':
+                errors.append(ValueError(
+                    f'Invalid values {invalid_values} are present in an'
```
pvk-developer, line 130: same as above

pvk-developer (changes requested):

sdv/constraints/base.py
fealho: I kinda really didn't want to do the sorting (which is why I implemented the regex tests instead of using the normal asserts), since I'd be adding multiple sorts in the code (one for each method that overwrites the _validate_inputs). If it was just in one place I wouldn't mind too much, but since it's not I prefer to leave the burden to the tests instead of in the code.

The same thing is why I decided to use `{}` instead of `()`, since I'd have to convert the sets to tuples everywhere, and using brackets in error messages is pretty standard.

pvk-developer (approved):

PR #915, opened 2022-07-20
Add `update_column` and `add_column` to `SingleTableMetadata`

### LINKED ISSUE #877: Add update_column and add_column methods to SingleTableMetadata
### Problem Description

_As a user, it would be helpful to be able to modify or add to the metadata programmatically._

### Expected behavior

- Add `update_column` method
    - Not all fields are required. Just update the ones provided in the `SingleTableMetadata`'s dict
    - Parameters:
        - `column_name`: string that is the name of the column to update
        - `**kwargs`: any key word arguments that describe metadata for a column
    - Errors:
        - If the column doesn't already exist in the metadata raise the following error
        `Error: Column name ('ssnn') does not exist in the table. Use 'add_column' to add new columns.`
        - Run all of the column validations that are done in the `validate` method. eg. #879 
        `Error: Invalid values ("pii") for numerical column "age".`
```python
>>> metadata.update_column(column_name='ssn', pii=True)
>>> metadata.update_column(
         column_name='education',
         order=['hs', 'bachelor', 'masters', 'PhD'])
```

- Add `add_column` method
    - All required parameters for the column metadata must be provided
    - Parameters:
        - `column_name`: string that is the name of the column to update
        - `**kwargs`: any key word arguments that describe metadata for a column
    - Errors:
        - If the column already is in the metadata, raise the following error:
        `Error: Column name 'user_id' already exists. Use 'update_column' to update an existing column.`
        - Run all of the column validations that are done in the `validate` method. eg.
        `Error: Invalid values ("pii") for numerical column "age".`
        `Error: Invalid values ("pii") for numerical column "age".`
```python
>>> metadata.add_column(column_name='cell', sdtype='phone_number', pii=True)
```

pvk-developer:
Resolves #877

### DIFF
--- a/sdv/metadata/single_table.py
+++ b/sdv/metadata/single_table.py
@@ -42,17 +42,19 @@ class SingleTableMetadata:
         representation = kwargs.get('representation')
         if representation and representation not in representations:
             raise ValueError(
-                f"Invalid value for 'representation' {representation} for column '{column_name}.")
+                f"Invalid value for 'representation' {representation} for column '{column_name}'.")
 
     @staticmethod
     def _validate_datetime(column_name, **kwargs):
         datetime_format = kwargs.get('datetime_format')
         if datetime_format:
             formated_date = datetime.now().strftime(datetime_format)
-            match = re.search('%.', formated_date)
-            if match:
+            matches = re.findall('(%.)|(%)', formated_date)
+            if matches:
+                matches = ', '.join(item for match in matches for item in match if item)
+                matches = ''.join(char if not char.isalnum() else '.' for char in matches)
                 raise ValueError(
-                    f"Invalid datetime fromat stringa '{match.group(0)}' "
+                    f"Invalid datetime format string '{matches}' "
                     f"for datetime column '{column_name}'."
                 )
 
@@ -65,12 +67,13 @@ class SingleTableMetadata:
                 f"Categorical column '{column_name}' has both an 'order' and 'order_by' "
                 'attribute. Only 1 is allowed.'
             )
-        elif order_by not in ('numerical_value', 'alphabetical'):
+        if order_by and order_by not in ('numerical_value', 'alphabetical'):
             raise ValueError(
                 f"Unknown ordering method '{order_by}' provided for categorical column "
                 f"'{column_name}'. Ordering method must be 'numerical_value' or 'alphabetical'."
             )
-        elif order and not isinstance(order, list):
+        if (isinstance(order, list) and not len(order)) or\
+           (not isinstance(order, list) and order is not None):
             raise ValueError(
                 f"Invalid order value provided for categorical column '{column_name}'. "
                 "The 'order' must be a list with 1 or more elements."
@@ -104,6 +107,8 @@ class SingleTableMetadata:
         expected_kwargs = self._EXPECTED_KWARGS.get(sdtype, ['pii'])
         unexpected_kwargs = set(list(kwargs)) - set(expected_kwargs)
         if unexpected_kwargs:
+            unexpected_kwargs = list(unexpected_kwargs)
+            unexpected_kwargs.sort()
             unexpected_kwargs = ', '.join(unexpected_kwargs)
             raise ValueError(
                 f"Invalid values '({unexpected_kwargs})' for {sdtype} column '{column_name}'.")
@@ -112,7 +117,7 @@ class SingleTableMetadata:
         if column_name not in self._columns:
             raise ValueError(
                 f"Column name ('{column_name}') does not exist in the table. "
-                "Use 'add_column'  to add new column."
+                "Use 'add_column' to add new column."
             )
 
     def _validate_column(self, column_name, sdtype, **kwargs):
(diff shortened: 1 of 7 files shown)


sdv/metadata/single_table.py
```diff
@@ -22,6 +32,63 @@ class SingleTableMetadata:
...
+
+    @staticmethod
+    def _validate_datetime(column_name, **kwargs):
+        datetime_format = kwargs.get('datetime_format')
+        if datetime_format:
+            formated_date = datetime.now().strftime(datetime_format)
+            matches = re.findall('(%.)|(%)', formated_date)
+            if matches:
+                matches = ', '.join(item for match in matches for item in match if item)
+                matches = ''.join(char if not char.isalnum() else '.' for char in matches)
```
amontanez24, line 55: what is this regex doing? Can we just use the same logic as in the actual transformer? As in try to parse the datetime and if it comes back with an error than we catch that here?

tests/unit/metadata/test_single_table.py
```diff
@@ -34,6 +35,544 @@ def test___init__(self):
...
+        instance._validate_unexpected_kwargs('age', 'numerical', representation='int')
+        instance._validate_unexpected_kwargs('start_date', 'datetime')
+        instance._validate_unexpected_kwargs('start_date', 'datetime', datetime_format='%Y-%d')
+        instance._validate_unexpected_kwargs('name', 'categorical')
+        instance._validate_unexpected_kwargs('name', 'categorical', order_by='alphabetical')
+        instance._validate_unexpected_kwargs('name', 'categorical', order=['a', 'b', 'c'])
+        instance._validate_unexpected_kwargs('synthetic', 'boolean')
+        instance._validate_unexpected_kwargs('phrase', 'text')
+        instance._validate_unexpected_kwargs('phrase', 'text', regex_format='[A-z]')
+        instance._validate_unexpected_kwargs('phone', 'phone_number')
```
amontanez24, line 236: Can we [parameterize ](https://docs.pytest.org/en/7.1.x/example/parametrize.html)this test instead of calling all these different cases like this? We can make the passed kwargs, and the error message parameters

amontanez24 (commented):
A few comments but I think this is looking really good so far!

sdv/metadata/single_table.py
pvk-developer: There is no such way to tell if a `strftime` is valid (or at least I didn't find one). This `regex` detects if there are `%` left on the `formated_date`, which is a way to tell whether or not the user passed an extra `%` which is the character that is usually used to specify any datetime like formating string.
The last line maybe I can remove it since it's just to clear the string a bit. I don't want ot confuse the end user since it can be a bit confusing if we find a `%` next to a `1` that comes from the `datetime.now()`.

tests/unit/metadata/test_single_table.py
pvk-developer: 🤔 yes let me re-do it

sdv/metadata/single_table.py
```diff
@@ -22,6 +32,68 @@ class SingleTableMetadata:
     KEYS = ['columns', 'primary_key', 'alternate_keys', 'constraints', 'SCHEMA_VERSION']
     SCHEMA_VERSION = 'SINGLE_TABLE_V1'
 
+    @staticmethod
+    def _validate_numerical(column_name, **kwargs):
+        representations = [
```
katxiao, line 37: Can we pull this out as a constant?

tests/unit/metadata/test_single_table.py
```diff
@@ -34,6 +66,514 @@ def test___init__(self):
...
+            - sdtype numerical
+            - representation
+
+        Side Effects:
+            - Passes with the correct ``representation``
+            - ``ValueError`` is raised stating that the ``representation`` is wrong.
+        """
+        # Setup
+        instance = SingleTableMetadata()
+        representations = [
```
katxiao, line 87: You could parameterize the unit test with the list of representations. That way if one fails, it's easy to tell which representation failed.

tests/unit/metadata/test_single_table.py
```diff
@@ -47,6 +47,12 @@ class TestSingleTableMetadata:
          re.escape("Invalid values '(anonymization, order_by)' for phone_number column 'phone'."))
     ]
 
+    NUMERICAL_REPRESENTATIONS = [
```
amontanez24, line 50: can you just access this from the class like ```SingleTableMetadata._NUMERICAL_REPRESENTATIONS```

sdv/metadata/single_table.py
```diff
@@ -29,18 +29,18 @@ class SingleTableMetadata:
         'b': 'boolean',
         'M': 'datetime',
     }
+
+    _NUMERICAL_REPRESENTATIONS = [
```
amontanez24, line 33: can we make this a frozenset instead of a list? That way it is immutable

sdv/metadata/single_table.py
```diff
@@ -12,16 +14,86 @@
 class SingleTableMetadata:
     """Single Table Metadata class."""
 
+    _EXPECTED_KWARGS = {
```
amontanez24, line 17: can we make this a frozendict as well?

amontanez24 (approved):
A couple of minor comments but I think this is ready!

katxiao (approved):
lgtm!

PR #923, opened 2022-07-25
Update `RDT` to v1.0.0

### LINKED ISSUE #897: Update the RDT version to 1.0

pvk-developer:
Resolves #897

### DIFF
--- a/sdv/metadata/table.py
+++ b/sdv/metadata/table.py
@@ -44,14 +44,14 @@ class Table:
             Dictinary specifying which transformers to use for each field.
             Available transformers are:
 
-                * ``integer``: Uses a ``FloatFormatter`` of dtype ``int``.
-                * ``float``: Uses a ``FloatFormatter`` of dtype ``float``.
-                * ``categorical``: Uses a ``FrequencyEncoder`` without gaussian noise.
-                * ``categorical_noised``: Uses a ``FrequencyEncoder`` adding gaussian noise.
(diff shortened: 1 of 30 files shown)


amontanez24 (changes requested):
Mostly minor comments and questions

sdv/metadata/table.py
```diff
@@ -404,24 +404,25 @@ def _get_transformers(self, dtypes):
         for name, dtype in dtypes.items():
             field_metadata = self._fields_metadata.get(name, {})
             transformer_template = field_metadata.get('transformer')
-            if self._dtype_transformers[np.dtype(dtype).kind] is None:
-                transformers[name] = None
-
-            elif transformer_template is None:
+            if transformer_template is None:
                 transformer_template = self._dtype_transformers[np.dtype(dtype).kind]
```
amontanez24, line 408: you can probably add this as the default to the get on line 406 so you only have to do this check once. ie
```python
transformer_template = field_metadata.get('transformer', self._dtype_transformers[np.dtype(dtype).kind])
```

amontanez24 (approved):
One more minor comment, but after that looks ready to go to me!

fealho (approved):
Looks good! Just double check why the tutorials are failing.

Issue #888, opened 2022-07-08
Add validate method to MultiTableMetadata


amontanez24:
### Problem Description

_As a user, it would be useful to make sure the multi-table metadata I've specified is formatted correctly._

### Expected behavior

- Add a `validate` method
- The validation should have 2 steps
    1. First, run the `SingleTableMetadata.validate` method for each table and group the error messages.
    2. Validate the relationships (described in detail in the Additional context section)
- If any errors show it should raise a `InvalidMetadataError` with all of the errors grouped by table
```python
>>> metadata.validate()
InvalidMetadataError: The metadata is not valid

Table: 'users'
Error: Invalid values ("pii") for datetime column "start_date".
Error: A Unique constraint is being applied to column "age". This column is already a key for that table.

Table: 'transactions'
Error: Invalid regex format string "[A-{6}" for text column "transaction_id"
Error: Unknown key value 'ttid'. Keys should be columns that exist in the table.
Error: Invalid increment value (0.5) in a FixedIncrements constraint. Increments must be positive integers.

Relationships:
Error: Relationship between tables ('users', 'transactions') contains an unknown primary key 'userr_id'.
Error: The relationships in the dataset are disjointed. Tables ('sessions') are not connected to any of the other tables.
```

### Additional context

**Relationship validation**
- "parent_table_name" and "child_table_name" should refer to tables that are defined in the "tables" portion of the metadata
`Error: Relationship contains an unknown table 'userss'.`

- "parent_primary_key" value should be the same as "primary_key" in the table definition
`Error: Relationship between tables ('users', 'transactions') contains an unknown primary key 'userr_id'.`

- "child_foreign_key" should refer to column(s) that are defined in the "tables" portion of the metadata
`Error: Relationship between tables ('users', 'transactions') contains an unknown foreign key 'transactions_idd'.`

- The length of "child_foreign_key" should match the length of "parent_primary_key" (you cannot have a single column foreign key referencing a multi-column composite primary key)
`Error: Relationship between tables ('users', 'transactions') is invalid. Primary key has length 2 but the foreign key has length 1.`

- The sdtype & attributes of the primary key should exactly match the sdtype & attributes of the foreign key
`Error: Relationship between tables ('users', 'transactions') is invalid. The primary and foreign key columns are not the same type.`
    - If it's a composite key then match by the order in the list; eg. check that the 0-index has the same definitions, then the 1-index

[...]

PR #932, opened 2022-08-02
Add Relationship method for `MultiTableMetadata`

### LINKED ISSUE #890: Add add_relationship method to MultiTableMetadata
### Problem Description

_As a user, it would be helpful to be able to programmatically specify and add relationships between tables in my metadata._

### Expected behavior

- Add a `add_relationship` method
- The relationship should be added to the private dict created in #883 
-  Parameters:
    - `parent_table_name`: A string representing the name of the parent table
    - `child_table_name`: A string representing the name of the child table
    - `parent_primary_key`: A string or tuple of strings representing the primary key of the parent
    - `child_foreign_key`: A string or tuple of strings representing the foreign key of the child
- Errors:
    - Check for the same errors as described in #888 

```python
metadata.add_relationship(
  parent_table_name='users',
  child_table_name='transactions',
  parent_primary_key='user_id', # can be a tuple for composite keys
  child_foreign_key='purchaser_id' # can be a tuple for composite key
)

metadata.add_relationship(
  parent_table_name='users',
  child_table_name='transactions',
  parent_primary_key='user_id', # can be a tuple for composite keys
  child_foreign_key='start_date' # can be a tuple for composite key
)
Error: Relationship between tables ('users', 'transactions') is invalid. The primary and foreign key columns are not the same type.
```

pvk-developer:
Resolves #890 
As part of #888

### DIFF
--- a/sdv/metadata/multi_table.py
+++ b/sdv/metadata/multi_table.py
@@ -7,7 +7,7 @@ from copy import deepcopy
 
 from sdv.metadata.errors import InvalidMetadataError
 from sdv.metadata.single_table import SingleTableMetadata
-from sdv.metadata.utils import read_json, validate_file_does_not_exist
+from sdv.metadata.utils import cast_to_iterable, read_json, validate_file_does_not_exist
 from sdv.metadata.visualization import visualize_graph
 
 
@@ -17,27 +17,27 @@ class MultiTableMetadata:
     def __init__(self):
         self._tables = {}
         self._relationships = []
-        self._child_map = defaultdict(set)
-        self._parent_map = defaultdict(set)
 
     @staticmethod
     def _validate_missing_relationship_keys(parent_table, parent_table_name, parent_primary_key,
                                             child_table_name, child_foreign_key):
-        missing_key = set(parent_table._primary_key) - set(parent_primary_key)
-        if missing_key:
+
+        missing_keys = set()
+        parent_primary_key = cast_to_iterable(parent_primary_key)
+        table_primary_keys = set(cast_to_iterable(parent_table._primary_key))
+        for key in parent_primary_key:
+            if key not in table_primary_keys:
+                missing_keys.add(key)
+
+        if missing_keys:
             raise ValueError(
                 f'Relationship between tables ({parent_table_name}, {child_table_name}) contains '
-                'an unknown foreign key {missing_key}.'
+                f'an unknown primary key {missing_keys}.'
             )
 
-        missing_keys = {}
-        if isinstance(child_foreign_key, str) and child_foreign_key not in parent_table.columns:
-            missing_keys.add(child_foreign_key)
-
-        elif isinstance(child_foreign_key, (list, tuple)):
-            for key in child_foreign_key:
-                if key not in parent_table._columns:
-                    missing_keys.add(key)
+        for key in set(cast_to_iterable(child_foreign_key)):
+            if key not in parent_table._columns:
+                missing_keys.add(key)
 
         if missing_keys:
             raise ValueError(
@@ -47,15 +47,15 @@ class MultiTableMetadata:
 
     @staticmethod
     def _validate_missing_relationship_tables(parent_table_name, child_table_name, tables):
-        missing_table_name = set(parent_table_name, child_table_name) - set(tables)
+        missing_table_name = set([parent_table_name, child_table_name]) - set(tables)
         if missing_table_name:
-            raise ValueError(f"Relationship contains an unknown table '{missing_table_name}'.")
+            raise ValueError(f'Relationship contains an unknown table {missing_table_name}.')
 
     @staticmethod
     def _validate_relationship_key_length(parent_table_name, parent_primary_key,
                                           child_table_name, child_foreign_key):
-        pk_len = len(set(parent_primary_key))
-        fk_len = len(set(child_foreign_key))
+        pk_len = len(set(cast_to_iterable(parent_primary_key)))
+        fk_len = len(set(cast_to_iterable(child_foreign_key)))
         if pk_len != fk_len:
             raise ValueError(
                 f"Relationship between tables ('{parent_table_name}', '{child_table_name}') is "
--- a/sdv/metadata/multi_table.py
+++ b/sdv/metadata/multi_table.py
@@ -2,6 +2,7 @@
 
 import json
 import warnings
+from collections import defaultdict
 from copy import deepcopy
 
 from sdv.metadata.errors import InvalidMetadataError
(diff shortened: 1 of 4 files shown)


sdv/metadata/multi_table.py
```diff
@@ -13,6 +15,165 @@ def __init__(self):
         self._tables = {}
         self._relationships = []
 
+    @staticmethod
+    def _validate_missing_relationship_keys(parent_table, parent_table_name, parent_primary_key,
+                                            child_table_name, child_foreign_key):
+        missing_keys = set()
+        parent_primary_key = cast_to_iterable(parent_primary_key)
+        table_primary_keys = set(cast_to_iterable(parent_table._primary_key))
```
fealho, line 23: This chunk of the code (all the way to the error message) can be moved into its own function.

That way you don't have to repeat the same code twice in lines 34-42

sdv/metadata/multi_table.py
```diff
@@ -13,6 +15,165 @@ def __init__(self):
...
+
+        if missing_keys:
+            raise ValueError(
+                f'Relationship between tables ({parent_table_name}, {child_table_name}) '
+                f'contains an unknown foreign key {missing_keys}.'
+            )
+
+    @staticmethod
+    def _validate_missing_relationship_tables(parent_table_name, child_table_name, tables):
+        missing_table_name = set([parent_table_name, child_table_name]) - set(tables)
```
fealho, line 46: What happens if `parent_table_name` or `child_table_name` are None? If it automatically crashes somewhere this is fine, otherwise you have to remove it:

```python3
set([parent_table_name, child_table_name]) - set(tables) - {None}
```

sdv/metadata/multi_table.py
```diff
@@ -13,6 +15,165 @@ def __init__(self):
...
+
+        if error:
+            raise ValueError(
+                f"Relationship between tables ('{parent_table_name}', '{child_table_name}') is "
+                'invalid. The primary and foreign key columns are not the same type.'
+            )
+
+    def _validate_circular_relationships(self, parent, children=None,
+                                         parents=None, child_map=None, errors=None):
+        """Validate that there is no circular relatioship in the metadata."""
```
fealho, line 87: missing n, relationship

sdv/metadata/multi_table.py
```diff
@@ -13,6 +15,165 @@ def __init__(self):
...
+        )
+
+        child_map = defaultdict(set)
+        for relation in self._relationships:
+            parent_name = relation['parent_table_name']
+            child_name = relation['child_table_name']
+            child_map[parent_name].add(child_name)
+
+        child_map[parent_table_name].add(child_table_name)
+        errors = []
```
fealho, line 160: Can you move this logic (160-168) to its own method? I think it makes this method easier to read, makes this logic easier to test and harder for us to mess this up in the future (since you are mutating errors inplace I'd prefer to be extra careful).

tests/unit/metadata/test_multi_table.py
```diff
@@ -17,6 +20,386 @@ def test___init__(self):
...
+            'sessions': Mock(),
+            'transactions': Mock()
+        }
+        instance._relationships = [
+            {'parent_table_name': 'users', 'child_table_name': 'sessions'},
+            {'parent_table_name': 'sessions', 'child_table_name': 'users'}
+        ]
+
+        # Run / Assert
+        # TODO fix match error message
```
fealho, line 353: ?

fealho (changes requested):

sdv/metadata/multi_table.py
```diff
@@ -13,6 +15,165 @@ def __init__(self):
         self._tables = {}
         self._relationships = []
 
+    @staticmethod
+    def _validate_missing_relationship_keys(parent_table, parent_table_name, parent_primary_key,
+                                            child_table_name, child_foreign_key):
+        missing_keys = set()
+        parent_primary_key = cast_to_iterable(parent_primary_key)
```
amontanez24, line 22: I think there are other places in this code that could now use the `cast_to_iterable` method. There are for sure some in the `SingleTableMetadata`

sdv/metadata/multi_table.py
```diff
@@ -13,6 +15,165 @@ def __init__(self):
...
+                missing_keys.add(key)
+
+        if missing_keys:
+            raise ValueError(
+                f'Relationship between tables ({parent_table_name}, {child_table_name}) '
+                f'contains an unknown foreign key {missing_keys}.'
+            )
+
+    @staticmethod
+    def _validate_missing_relationship_tables(parent_table_name, child_table_name, tables):
```
amontanez24, line 45: very minor: can we name this `_validate_relationship_missing_tables` or `_validate_no_missing_tables_in_relationship`?

sdv/metadata/multi_table.py
```diff
@@ -13,6 +15,165 @@ def __init__(self):
...
+        missing_table_name = set([parent_table_name, child_table_name]) - set(tables)
+        if missing_table_name:
+            raise ValueError(f'Relationship contains an unknown table {missing_table_name}.')
```
amontanez24, line 48: Isn't it possible for multiple tables to be missing? Should it say `Relationship contains an unknown table` or `Relationship contains unknown tables` depending on the set length?

sdv/metadata/multi_table.py
```diff
@@ -13,6 +15,165 @@ def __init__(self):
...
+        error = False
+        if isinstance(parent_primary_key, (list, tuple)):
+            for pk, fk in zip(parent_primary_key, child_foreign_key):
+                if parent_columns[pk]['sdtype'] != parent_columns[fk]['sdtype']:
+                    error = True
+                    break
+
+        else:
+            pk_sdtype = parent_columns[parent_primary_key]['sdtype']
+            fk_sdtype = parent_columns[child_foreign_key]['sdtype']
```
amontanez24, line 75: if we made a set of all columns involved in the primary key and foreign key, then we could just loop through them instead of doing this if right? To me it might make sense to combine `_validate_relationship_key_length` with this method and do the following steps:
1. Convert keys to sets
2. If the lengths are different, raise an error
3. Otherwise, loop through the keys and make sure the sdtypes are the same

amontanez24 (changes requested):

tests/unit/metadata/test_multi_table.py
pvk-developer: This is a `set` and always comes in 2 pairs, so I will fix with the `SingleTableMetadata` improvement that I'm doing

sdv/metadata/multi_table.py
pvk-developer: No need to remove `None` since if they provide `None` it should crash. And if it does here it's fine too.
```
bool(set([None])) -> True
```

sdv/metadata/utils.py
```diff
@@ -10,6 +10,14 @@
 import sre_parse  # isort:skip
 
 
+def cast_to_iterable(value):
+    """Return a ``list`` if the input object is not a ``list`` or ``tuple``."""
+    if isinstance(value, (list, tuple)):
+        return value
+
+    return [value]
```
amontanez24, line 18: should we add a small test or two to cover these cases?

amontanez24 (approved):
Small comment about tests but I think this is good!

fealho (approved):

PR #934, opened 2022-08-02
Add update_column and add_column methods to MultiTableMetadata

### LINKED ISSUE #893: Add update_column and add_column methods to MultiTableMetadata
### Problem Description

_As a user, it would be helpful to be able to modify or add to the columns for a table in my_ `MultiTableMetadata` _programmatically._

### Expected behavior

- Add `update_column` method
    - Should just be calling the `SingleTableMetadata.update_column` method for the appropriate table under the hood
    - Not all fields are required. Just update the ones provided in the `SingleTableMetadata`'s dict
    - Parameters:
        - `table_name`: string that is the name of the table to update the column for
        - `column_name`: string that is the name of the column to update
        - `**kwargs`: any key word arguments that describe metadata for a column
    - Errors:
        - If the column doesn't already exist in the metadata raise the following error
        `Error: Column name ('ssnn') does not exist in the table. Use 'add_column' to add new columns.`
        - Run all of the column validations that are done in the `validate` method. eg.
        `Error: Invalid values ("pii") for numerical column "age".`
        - If the table doesn't exist, raise the following error
        `Error: Unknown table name ('userss')`
```python
>>> metadata.update_column(table_name='userss', ...)
Error: Unknown table name ('userss')

>>> metadata.update_column(table_name='users', column_name='ages', pii=True)
Error: Invalid values ("pii") for numerical column "age".
```

- Add `add_column` method
    - Should just be calling the `SingleTableMetadata.add_column` method for the appropriate table under the hood
    - All required parameters for the column metadata must be provided
    - Parameters:
        - `table_name`: string that is the name of the table to add the column for
        - `column_name`: string that is the name of the column to update
        - `**kwargs`: any key word arguments that describe metadata for a column
    - Errors:
        - If the column already is in the metadata, raise the following error:
        `Error: Column name 'user_id' already exists. Use 'update_column' to update an existing column.`
        - Run all of the column validations that are done in the `validate` method. eg.
        `Error: Invalid values ("pii") for numerical column "age".`
        `Error: Invalid values ("pii") for numerical column "age".`
        - If the table doesn't exist, raise the following error
        `Error: Unknown table name ('userss')`

```python
>>> metadata.add_column(table_name='userss', ...)
Error: Unknown table name ('userss')

>>> metadata.add_column(table_name='users', column_name='user_id', sdtype='text')
Error: Column name 'user_id' already exists. Use 'update_column' to update an existing column name.
```

amontanez24:
resolves #893

### DIFF
--- a/tests/unit/metadata/test_multi_table.py
+++ b/tests/unit/metadata/test_multi_table.py
@@ -1,6 +1,9 @@
 """Test Multi Table Metadata."""
 
+import re
 from unittest.mock import Mock, patch
+
+import pytest
 
 from sdv.metadata.multi_table import MultiTableMetadata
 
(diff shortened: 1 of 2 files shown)


tests/unit/metadata/test_multi_table.py
```diff
@@ -435,3 +438,99 @@ def test_visualize_show_table_details_only(self, visualize_graph_mock):
...
+        # Setup
+        metadata = MultiTableMetadata()
+        table = Mock()
+        metadata._tables = {'table': table}
+
+        # Run
+        metadata.add_column('table', 'column', sdtype='numerical', pii=False)
+
+        # Assert
+        table.add_column.assert_called_once_with('column', sdtype='numerical', pii=False)
```
fealho, line 467: If it's easy, I'd also check that metadata._tables was updated properly (since that's the actually only important part).

tests/unit/metadata/test_multi_table.py
```diff
@@ -435,3 +438,99 @@ def test_visualize_show_table_details_only(self, visualize_graph_mock):
...
+        # Setup
+        metadata = MultiTableMetadata()
+        table = Mock()
+        metadata._tables = {'table': table}
+
+        # Run
+        metadata.update_column('table', 'column', sdtype='numerical', pii=False)
+
+        # Assert
+        table.update_column.assert_called_once_with('column', sdtype='numerical', pii=False)
```
fealho, line 515: Same here, check column actually. updated

fealho (approved):
A few nitpicks, but it looks good 👍

tests/unit/metadata/test_multi_table.py
amontanez24: same

tests/unit/metadata/test_multi_table.py
fealho: Oh, you are patching the whole thing, I hadn't noticed. Then it's fine to leave as is (I just usually like to double check the values themselves just to be sure it actually worked, but only if it's easy.).

amontanez24: yeah it wouldn't be doable here. The `MultiTableMetadata` just stores a dict containing `SingleTableMetadata` objects, so the column gets added at that level

pvk-developer (approved):

PR #938, opened 2022-08-03
Add `load_json` and `save_json` to `MutliTableMetadata`

### LINKED ISSUE #891: Add load_from_json and save_to_json methods to the MultiTableMetadata
### Problem Description

_As a user, it would be useful if I could create an instance of the `MultiTableMetadata` class from a json file. It would also be useful if I could save an existing instance to a json file._

### Expected behavior

- Add a `load_from_json` method
    - Parameters:
        - `filepath`: string that is the path to json file
    - Errors:
        - If the json file doesn't contain the `SCHEMA_VERSION` attribute, then raise the following error
        ```python
        Error: This metadata file is incompatible with the SingleTableMetadata class and version.
        ``` 
    - To create the `SingleTableMetadata` instances you can call `SingleTableMetadata._load_from_dict`.

```python
>>> metadata = SingleTableMetadata.load_from_json(filepath='project/metadata.json')
```
- Add a `save_to_json` method
    - Parameters:
        - `filepath`: string that is the path to save the file to
    - Errors:
        - If there is already a file in the path location, raise the following error
        ```python
        Error: A file named 'metadata.json' already exists in this folder. Please specify a different filename.
        ```
```python
>>> metadata.save_to_json(filepath='project/metadata.json')
```

fealho:
Resolve #891.

### DIFF
--- a/sdv/metadata/single_table.py
+++ b/sdv/metadata/single_table.py
@@ -12,7 +12,7 @@ import pandas as pd
 from sdv.constraints import Constraint
 from sdv.constraints.errors import MultipleConstraintsErrors
 from sdv.metadata.errors import InvalidMetadataError
-from sdv.metadata.utils import read_json, validate_path
+from sdv.metadata.utils import read_json, validate_file_does_not_exist
 
 
 class SingleTableMetadata:
@@ -508,7 +508,7 @@ class SingleTableMetadata:
         Raises:
             Raises an ``Error`` if the path already exists.
         """
-        validate_path(filepath)
+        validate_file_does_not_exist(filepath)
         metadata = self.to_dict()
         metadata['SCHEMA_VERSION'] = self.SCHEMA_VERSION
         with open(filepath, 'w', encoding='utf-8') as metadata_file:
--- a/sdv/metadata/single_table.py
+++ b/sdv/metadata/single_table.py
@@ -529,6 +529,6 @@ class SingleTableMetadata:
             json.dump(metadata, metadata_file, indent=4)
 
     def __repr__(self):
-        """Pretty print the ``SingleTableMetadata```SingleTableMetadata``."""
+        """Pretty print the ``SingleTableMetadata``."""
         printed = json.dumps(self.to_dict(), indent=4)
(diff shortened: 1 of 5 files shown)


sdv/metadata/utils.py
```diff
@@ -150,3 +152,26 @@ def strings_from_regex(regex, max_repeat=16):
             sizes.append(size)
 
     return _from_generators(generators, max_repeat), np.prod(sizes)
+
+
+def open_path(filepath):
```
pvk-developer, line 157: Instead of `open_path` maybe rename to `read_json` or similar.

pvk-developer (approved):
Small nit but LGTM!

sdv/metadata/utils.py
```diff
@@ -150,3 +152,26 @@ def strings_from_regex(regex, max_repeat=16):
...
+            f"A file named '{filepath.name}' does not exist. "
+            'Please specify a different filename.'
+        )
+
+    with open(filepath, 'r', encoding='utf-8') as metadata_file:
+        return json.load(metadata_file)
+
+
+def validate_path(filepath):
+    """Validate a file path."""
```
amontanez24, line 171: Can we rename this and change the description to be more descriptive like `validate_file_does_not_exist`

sdv/metadata/single_table.py
```diff
@@ -516,19 +508,13 @@ def save_to_json(self, filepath):
...
-            )
-
+        validate_path(filepath)
         metadata = self.to_dict()
         metadata['SCHEMA_VERSION'] = self.SCHEMA_VERSION
         with open(filepath, 'w', encoding='utf-8') as metadata_file:
             json.dump(metadata, metadata_file, indent=4)
 
     def __repr__(self):
-        """Pretty print the ``SingleTableMetadata```SingleTableMetadata``."""
```
amontanez24, line 532: good catch

amontanez24 (commented):
Just a small comment. After that this should be good to go!

amontanez24 (approved):
:shipit:

PR #942, opened 2022-08-08
Incorrect rounding in Custom Constraint example

### LINKED ISSUE #941: Incorrect rounding in `Custom Constraint` example
If I follow the exact code in the [Custom Constraint User Guide](https://sdv.dev/SDV/user_guides/single_table/custom_constraints.html#transformations), the resulting constraint doesn't work as intended -- it never produces employees that are not constractors.

The issue is in the `reverse_transform` method. Instead of `round(2)` the code should be `round()`.

i.e. change this line

```
rounded_data = transformed_data[is_included][column_name].round()
```

### Further Context

In previous versions of the docs, we were multiplying by the increment first. Now, we are multiplying by the increment last. Therefore, it is necessary to round to the nearest whole number instead of 2 digits.

amontanez24:
resolves #941

### DIFF
--- a/docs/user_guides/single_table/custom_constraints.rst
+++ b/docs/user_guides/single_table/custom_constraints.rst
@@ -106,7 +106,7 @@ salaries won't necessarily be divisible by 500. Instead we should:
         column_name = column_names[0]
   
         is_included = (transformed_data[exclusion_column] == 0)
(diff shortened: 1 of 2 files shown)


fealho (approved):

katxiao (approved):
lgtm

PR #945, opened 2022-08-09
Cap `nbconvert` version

pvk-developer:
Cap the `nbconvert` version since it's causing tutorials build to fail.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,6 +38,7 @@ tests_require = [
     'pytest>=3.4.2',
     'pytest-cov>=2.6.0',
     'pytest-rerunfailures>10',
+    'nbconvert>5,<6.5.0',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -38,6 +38,7 @@
     'pytest>=3.4.2',
     'pytest-cov>=2.6.0',
     'pytest-rerunfailures>10',
+    'nbconvert>5,<6.5.0',
```
amontanez24, line 41: do we need this?

amontanez24 (approved):
One comment, but after addressing feel free to merge

setup.py
pvk-developer: Not really, since I couldn't reproduce it.

PR #954, opened 2022-08-12
Add validate method for MultiTableMetadata

### LINKED ISSUE #888: Add validate method to MultiTableMetadata

pvk-developer:
Resolves #888

### DIFF
--- a/sdv/metadata/multi_table.py
+++ b/sdv/metadata/multi_table.py
@@ -140,6 +140,24 @@ class MultiTableMetadata:
             child_foreign_key
         )
 
+    def _get_child_map(self):
+        child_map = defaultdict(set)
+        for relation in self._relationships:
+            parent_name = relation['parent_table_name']
+            child_name = relation['child_table_name']
+            child_map[parent_name].add(child_name)
+
+        return child_map
+
+    def _get_parent_map(self):
+        parent_map = defaultdict(set)
+        for relation in self._relationships:
+            parent_name = relation['parent_table_name']
+            child_name = relation['child_table_name']
+            parent_map[child_name].add(parent_name)
+
+        return parent_map
+
     def add_relationship(self, parent_table_name, child_table_name,
                          parent_primary_key, child_foreign_key):
         """Add a relationship between two tables.
@@ -166,12 +184,7 @@ class MultiTableMetadata:
         self._validate_relationship(
             parent_table_name, child_table_name, parent_primary_key, child_foreign_key)
 
-        child_map = defaultdict(set)
-        for relation in self._relationships:
-            parent_name = relation['parent_table_name']
-            child_name = relation['child_table_name']
-            child_map[parent_name].add(child_name)
-
+        child_map = self._get_child_map()
         child_map[parent_table_name].add(child_table_name)
         self._validate_child_map_circular_relationship(child_map)
 
@@ -443,7 +456,7 @@ class MultiTableMetadata:
                     'The following errors were found in the metadata:\n', tittle)
                 errors.append(error)
 
-    def _validate_disjoined_tables(self, parent_map, child_map):
+    def _validate_all_tables_connected(self, parent_map, child_map):
         nodes = list(self._tables.keys())
         queue = [nodes[0]]
         connected = {table_name: False for table_name in nodes}
@@ -472,6 +485,15 @@ class MultiTableMetadata:
 
             raise ValueError(f'The relationships in the dataset are disjointed. {table_msg}')
 
+    def _append_relationships_errors(self, errors, method, *args, **kwargs):
+        try:
+            method(*args, **kwargs)
+        except Exception as error:
+            if '\nRelationships:' not in errors:
+                errors.append('\nRelationships:')
+
+            errors.append(error)
+
     def validate(self):
         """Validate the metadata.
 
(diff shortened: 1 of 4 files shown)


sdv/metadata/multi_table.py
```diff
@@ -427,6 +432,92 @@ def add_constraint(self, table_name, constraint_name, **kwargs):
         table = self._tables.get(table_name)
         table.add_constraint(constraint_name, **kwargs)
 
+    def _validate_single_table(self, errors):
+        for table_name, table in self._tables.items():
+            try:
+                table.validate()
+            except Exception as error:
+                errors.append('\n')
+                tittle = f'Table: {table_name}'
```
fealho, line 441: title

sdv/metadata/single_table.py
```diff
@@ -269,7 +269,8 @@ def set_primary_key(self, id):
             id (str, tuple):
                 Name (or tuple of names) of the primary key column(s).
         """
-        self._validate_key(id, 'primary')
+        if id is not None:
```
fealho, line 272: I think it makes more sense to check if id is None inside the `_validate_key` method. That way you don't need any of the changes in lines 400-409.

sdv/metadata/single_table.py
```diff
@@ -395,11 +397,17 @@ def validate(self):
                 errors.append(reformated_errors)
 
         # Validate keys
-        self._append_error(errors, self._validate_key, self._primary_key, 'primary')
-        self._append_error(errors, self._validate_key, self._sequence_key, 'sequence')
+        if self._primary_key:
```
fealho, line 400: With change above, none of these are needed.

tests/unit/metadata/test_multi_table.py
```diff
@@ -50,25 +111,39 @@ def test__validate_missing_relationship_keys_foreign_key(self):
...
+        child_table = Mock()
+        child_table._primary_key = 'session_id'
+        child_table._columns = {
+            'user_id': {'sdtype': 'numerical'},
+            'session_id': {'sdtype': 'numerical'},
+            'transactions': {'sdtype': 'numerical'},
+        }
         child_table_name = 'sessions'
-        child_foreign_key = 'session_id'
+        child_foreign_key = 'id'
```
fealho, line 131: Is the description of this test case wrong? The child foreign key seems to be in the parent_table primary keys.

sdv/metadata/multi_table.py
```diff
@@ -427,6 +432,92 @@ def add_constraint(self, table_name, constraint_name, **kwargs):
...
+                error = str(error).replace(
+                    'The following errors were found in the metadata:\n', tittle)
+                errors.append(error)
+
+    def _validate_disjoined_tables(self, parent_map, child_map):
```
amontanez24, line 446: can we rename to match the positive case? so `_validate_all_tables_connected` or something like that?

tests/unit/metadata/test_multi_table.py
```diff
@@ -361,11 +394,7 @@ def test_validate_child_map_circular_relationship(self):
         'sdv.metadata.multi_table.MultiTableMetadata._validate_no_missing_tables_in_relationship'
     )
     @patch('sdv.metadata.multi_table.MultiTableMetadata._validate_relationship_key_length')
-    @patch('sdv.metadata.multi_table.MultiTableMetadata._validate_relationship_sdtypes')
-    @patch('sdv.metadata.multi_table.MultiTableMetadata._validate_missing_relationship_keys')
```
amontanez24, line 365: why are we changing these but not the other two?

sdv/metadata/multi_table.py
```diff
@@ -427,6 +432,92 @@ def add_constraint(self, table_name, constraint_name, **kwargs):
...
+
+                errors.append(error)
+
+        parent_map = defaultdict(set)
+        child_map = defaultdict(set)
+        for relation in self._relationships:
+            parent_name = relation['parent_table_name']
+            child_name = relation['child_table_name']
+            parent_map[child_name].add(parent_name)
+            child_map[parent_name].add(child_name)
```
amontanez24, line 498: can we pull this into separate a method? I think it would be useful to have a `_get_child_map` and `_get_parent_map` method. It's also used in `add_relationship`

amontanez24 (commented):
Mostly minor comments and questions

tests/unit/metadata/test_multi_table.py
pvk-developer: Yyes the description maybe wrong I need to revisit them!

tests/unit/metadata/test_multi_table.py
pvk-developer: I changed those to `instance` methods instead since the previous validation that I was performing wasn't as intended.

sdv/metadata/multi_table.py
```diff
@@ -427,6 +445,79 @@ def add_constraint(self, table_name, constraint_name, **kwargs):
         table = self._tables.get(table_name)
         table.add_constraint(constraint_name, **kwargs)
 
+    def _validate_single_table(self, errors):
+        for table_name, table in self._tables.items():
+            try:
+                table.validate()
+            except Exception as error:
+                errors.append('\n')
+                tittle = f'Table: {table_name}'
```
fealho, line 454: Typo.

fealho (approved):
Looks good! Just one typo.

sdv/metadata/multi_table.py
```diff
@@ -427,6 +445,79 @@ def add_constraint(self, table_name, constraint_name, **kwargs):
...
+            connected[node] = True
+            for child in list(child_map[node]) + list(parent_map[node]):
+                if not connected[child] and child not in queue:
+                    queue.append(child)
+
+        if not all(connected.values()):
+            not_connected_tables = set()
+            for table, value in connected.items():
+                if not value:
+                    not_connected_tables.add(table)
```
amontanez24, line 475: very minor but can this be rewritten as:
```python
disconnected_tables = {table for table, value in connected.items if not value)
```

amontanez24 (approved):
One minor comment, but besides that I think this is ready!

PR #957, opened 2022-08-16
Create DataProcessor class

### LINKED ISSUE #946: Create DataProcessor class
### Problem Description

With all of the metadata management moved to `SingleTableMetadata` and `MultiTableMetadata`, the [table.py](https://github.com/sdv-dev/SDV/blob/master/sdv/metadata/table.py) class no longer needs to manage metadata. As engineers, it would be useful to have a class that takes in the `SingleTableMetadata` and uses it to manage the `HyperTransformer`, constraints and anonymization of data.

### Expected behavior

- Add a `data_processing` module
- Create a `DataProcessor` class and put it in that module
- The `__init__` params should be:
    - `metadata` (SingleTableMetadata): The single table metadata instance that will be used to apply constraints and transformations to the data.
    - `learn_rounding_scheme` (bool): Define rounding scheme for ``FloatFormatter``. If ``True``, the data returned by ``reverse_transform`` will be rounded to that place. Defaults to ``True``.
    - `enforce_min_max_values` (bool): Specify whether or not to clip the data returned by ``reverse_transform`` of the numerical transformer, ``FloatFormatter``, to the min and max values seen during ``fit``. Defaults to ``True``.
    - `model_kwargs` (dict): Dictionary specifying the kwargs that need to be used in each tabular model when working on this table. This dictionary contains as keys the name of the TabularModel class and as values a dictionary containing the keyword arguments to use. This argument exists mostly to ensure that the models are fitted using the same arguments when the same DataProcessor is used to fit different model instances on different slices of the same table.
- During the init we should create a `constraints` attribute that loads all of the constraint instance from the `SingleTableMetadata`
    - It can be done similarly to what's done [here](https://github.com/sdv-dev/SDV/blob/master/sdv/metadata/table.py#L247-L256) except the constraints should always be loaded from the json.

amontanez24:
resolves #946

### DIFF
--- a/sdv/data_processing/__init__.py
+++ b/sdv/data_processing/__init__.py
@@ -1,0 +1,7 @@
+"""Data processing subpackage."""
+
+from sdv.data_processing.data_processor import DataProcessor
+
+__all__ = (
+    'DataProcessor',
+)
--- a/sdv/data_processing/data_processor.py
+++ b/sdv/data_processing/data_processor.py
@@ -10,7 +10,7 @@ class DataProcessor:
 
     This class handles all pre and post processing that is done to a single table to get it ready
     for modeling and finalize sampling. These processes include formatting, transformations,
-    anonymization and constraint hanlding.
+    anonymization and constraint handling.
 
     Args:
         metadata (metadata.SingleTableMetadata):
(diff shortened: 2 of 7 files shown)


fealho (approved):
Looks good!

PR #958, opened 2022-08-16
Update default metrics

### LINKED ISSUE #949: Change default metrics for evaluate
### Problem Description

As a user, it is confusing when multi-table metrics are applied to single table data. Also many of the metrics seem to result in errors.

### Expected behavior

- When running evaluate, the following metrics should be applied by default in the single table case
    - CSTest
    - KSTest
- In the multi-tale case, the multi-table version of the following metrics should be used
    - CSTest
    - KSTest

### Additional context

The metrics to use are currently selected in this method: https://github.com/sdv-dev/SDV/blob/65c8b5327190e28ad340bfca89e55171aa675292/sdv/evaluation.py#L74
We should change this to only use the metrics described above. One thing to note is that if a user provides metadata and a single table as input, it will crash because of this line https://github.com/sdv-dev/SDV/blob/65c8b5327190e28ad340bfca89e55171aa675292/sdv/evaluation.py#L130
This is not necessary to fix for this issue, but if it's easy enough then we should make the change

fealho:
Resolve #949.

### DIFF
--- a/sdv/evaluation.py
+++ b/sdv/evaluation.py
@@ -11,6 +11,7 @@ def _validate_arguments(synthetic_data, real_data, metadata, root_path, table_na
 
     If ``metadata`` is an instance of dict create the ``Metadata`` object.
     If ``metadata`` is ``None``, ``real_data`` has to be a ``pandas.DataFrane``.
+    If ``metadata`` is a dict, ``root_path`` must be passed.
 
     If ``real_data`` is ``None`` load all the tables and assert that ``synthetic_data`` is
     a ``dict``. Otherwise, ``real_data`` and ``synthetic_data`` must be of the same type.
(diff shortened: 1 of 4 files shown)


sdv/evaluation.py
```diff
@@ -52,6 +53,10 @@ def _validate_arguments(synthetic_data, real_data, metadata, root_path, table_na
     elif not isinstance(synthetic_data, type(real_data)):
         raise TypeError('`real_data` and `synthetic_data` must be of the same type')
 
+    # Get table name from metadata for single tables when table_name is not passed
+    if table_name is None and not isinstance(synthetic_data, dict):
+        table_name = list(metadata.to_dict()['tables'].keys())[0]
```
amontanez24, line 58: I think it would be better to just raise an error and tell them they have to provide the table name in this case

amontanez24 (approved):
One comment but besides that I think this looks good!

PR #959, opened 2022-08-17
Fixes for rdt 1.2.0

amontanez24:
This PR changes some tests that failed with updates made to the `FloatFormatter` in the new `RDT` release. These were not bugs in `RDT` but rather tests that were misusing the `FloatFormatter`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ install_requires = [
     'copulas>=0.7.0,<0.8',
     'ctgan>=0.5.2.dev0,<0.6',
     'deepecho>=0.3.0.post1,<0.4',
-    'rdt>=1.1.0,<2.0',
+    'rdt==1.2.0.dev0',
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -12,7 +12,7 @@
     history = history_file.read()
 
 install_requires = [
-    'Faker>=3.0.0,<10',
+    'Faker>=3.0.0,<15',
```
fealho, line 15: Can you change this in the conda file as well?

fealho (approved):

setup.py
amontanez24: I'm actually not sure if the conda file will work for this because I'm not sure if we ever released rdt >= 1.0.0 on our conda channel. I think we've just been releasing on conda forge

katxiao (approved):
lgtm!

Issue #948, opened 2022-08-10
Add utility methods to DataProcessor


amontanez24:
### Problem Description

Outside of managing transformations, the `DataProcessor` is responsible for validating constraints, making sure all the ids are unique, and converting back and forth to dicts and json.

### Expected behavior

- Add `filter_valid(self, data)`: Filter the data using the constraints and return only the valid rows.
    - Borrow logic from here https://github.com/sdv-dev/SDV/blob/e5fc3f62b2eb20fa18f284e4c51b26bf80d33f88/sdv/metadata/table.py#L726-L740
- Add `make_ids_unique(self, data)`: Repopulate any id fields in provided data to guarantee uniqueness.
    - Borrow logic from here https://github.com/sdv-dev/SDV/blob/e5fc3f62b2eb20fa18f284e4c51b26bf80d33f88/sdv/metadata/table.py#L742-L759
- Add `to_dict`
    - Borrow logic from here https://github.com/sdv-dev/SDV/blob/e5fc3f62b2eb20fa18f284e4c51b26bf80d33f88/sdv/metadata/table.py#L765-L784
    - fields, constraints, primary_key, entity_columns, sequence_index and context_columns all now live in the `SingleTableMetadata` so that will be handled by calling it's `to_dict` method
- Add `to_json`
    - Borrow logic from here https://github.com/sdv-dev/SDV/blob/e5fc3f62b2eb20fa18f284e4c51b26bf80d33f88/sdv/metadata/table.py#L786-L794
- Add `from_dict`
    - Borrow logic from here https://github.com/sdv-dev/SDV/blob/e5fc3f62b2eb20fa18f284e4c51b26bf80d33f88/sdv/metadata/table.py#L796-L823
    - Again, a lot of the variables now live inside the `SingleTableMetadata`
- Add `From_json`
    - Borrow logic from here https://github.com/sdv-dev/SDV/blob/e5fc3f62b2eb20fa18f284e4c51b26bf80d33f88/sdv/metadata/table.py#L825-L834

PR #965, opened 2022-08-19
Add utility methods

fealho:
Resolve https://github.com/sdv-dev/SDV/issues/948.


sdv/data_processing/data_processor.py
```diff
@@ -69,3 +73,75 @@ def __init__(self, metadata, learn_rounding_scheme=True, enforce_min_max_values=
...
+    def to_dict(self):
+        """Get a dict representation of this DataProcessor.
+
+        Returns:
+            dict:
+                Dict representation of this DataProcessor.
+        """
+        return {
+            'metadata': copy.deepcopy(self.metadata.to_dict()),
+            'constraints_to_reverse': copy.deepcopy(self._constraints_to_reverse),
```
fealho, line 102: This implementation assumes `constraints_to_reverse` is a list of constraint_dicts. If we decide not to do so, this needs to be revised.

sdv/data_processing/data_processor.py
```diff
@@ -69,3 +73,75 @@ def __init__(self, metadata, learn_rounding_scheme=True, enforce_min_max_values=
...
+        return data
+
+    def to_dict(self):
+        """Get a dict representation of this DataProcessor.
+
+        Returns:
+            dict:
+                Dict representation of this DataProcessor.
+        """
+        return {
```
fealho, line 100: I'm only passing `metadata, constraints_to_reverse, model_kwargs` because:

1. `_constraints` are derived from `metadata` during the __init__ (so they are not needed).
2. `_transformers_by_sdtype` cannot be fully transformed into dict, since RDT transformers don't have to_dict/from_dict methods.

sdv/data_processing/data_processor.py
```diff
@@ -69,3 +73,75 @@ def __init__(self, metadata, learn_rounding_scheme=True, enforce_min_max_values=
...
+            'metadata': copy.deepcopy(self.metadata.to_dict()),
+            'constraints_to_reverse': copy.deepcopy(self._constraints_to_reverse),
+            'model_kwargs': copy.deepcopy(self._model_kwargs)
+        }
+
+    @classmethod
+    def from_dict(cls, metadata_dict, learn_rounding_scheme=True, enforce_min_max_values=True):
```
fealho, line 107: Since `to_dict` doesn't provide `_transformers_by_sdtype`, we have to pass `learn_rounding_scheme` and `enforce_min_max_values` to derive it from scratch.

sdv/data_processing/data_processor.py
amontanez24: `_constraints_to_reverse` will actually be a list of constraint instances. When this class gets initialized it creates all the constraint instances from the metadata and then adds specific ones to this list when needed

sdv/data_processing/data_processor.py
```diff
@@ -69,3 +73,75 @@ def __init__(self, metadata, learn_rounding_scheme=True, enforce_min_max_values=
...
+            enforce_min_max_values (bool):
+                If passed, set the ``enforce_min_max_values`` on the new instance.
+        """
+        instance = cls(
+            metadata=SingleTableMetadata.from_dict(metadata_dict['metadata']),
+            learn_rounding_scheme=learn_rounding_scheme,
+            enforce_min_max_values=enforce_min_max_values,
+            model_kwargs=metadata_dict.get('model_kwargs')
+        )
+        instance._constraints_to_reverse = metadata_dict.get('constraints_to_reverse', [])
```
amontanez24, line 124: this will have to change as well

amontanez24 (changes requested):
Just a comment about the `_constraints_to_reverse`. Everything else looks good

sdv/data_processing/data_processor.py
```diff
@@ -97,9 +97,10 @@ def to_dict(self):
             dict:
                 Dict representation of this DataProcessor.
         """
+        constraints_to_reverse = [cnt.to_dict() for cnt in self._constraints_to_reverse]
```
amontanez24, line 100: minor: does it not fit to do `constraint` instead of `cnt`. In general I think we should try to use full descriptive names

amontanez24 (approved):
One small comment, but not a big deal. :shipit:

PR #968, opened 2022-08-19
Adding fit and transform methods

amontanez24:
I still need to add tests. We will also have to figure out how to handle primary and alternate keys in another issue. For this issue, they should just be treated as categorical columns.

### DIFF
--- a/sdv/data_processing/data_processor.py
+++ b/sdv/data_processing/data_processor.py
@@ -10,7 +10,7 @@ import rdt
 from sdv.constraints import Constraint
 from sdv.constraints.errors import (
     FunctionError, MissingConstraintColumnError, MultipleConstraintsErrors)
-from sdv.metadata.errors import MetadataNotFittedError
+from sdv.data_processing.errors import NotFittedError
 from sdv.metadata.single_table import SingleTableMetadata
 
 
@@ -93,6 +93,7 @@ class DataProcessor:
         self._hyper_transformer = None
         self.table_name = table_name
         self._dtypes = None
+        self.fitted = False
 
     def get_model_kwargs(self, model_name):
         """Return the required model kwargs for the indicated model.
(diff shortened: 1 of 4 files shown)


sdv/data_processing/data_processor.py
```diff
@@ -1,8 +1,17 @@
 """Single table data processing."""
 
+import copy
```
pvk-developer, line 3: Since we are using only `deepcopy` I think we can just import `from copy import deepcopy`.

sdv/data_processing/data_processor.py
```diff
@@ -62,10 +80,217 @@ def _update_numerical_transformer(self, learn_rounding_scheme, enforce_min_max_v
...
+        if not is_condition:
+            self._constraints_to_reverse = []
+
+        for constraint in self._constraints:
+            try:
+                data = constraint.transform(data)
+                if not is_condition:
+                    self._constraints_to_reverse.append(constraint)
+
+            except (MissingConstraintColumnError, FunctionError) as e:
```
pvk-developer, line 135: I think it would be good if instead of `e` we used `error` for the future linting plugins.

sdv/data_processing/data_processor.py
```diff
@@ -62,10 +80,217 @@ def _update_numerical_transformer(self, learn_rounding_scheme, enforce_min_max_v
...
+                    LOGGER.info(
+                        f'Error transforming {constraint.__class__.__name__}. '
+                        'Using the reject sampling approach instead.'
+                    )
+                if is_condition:
+                    indices_to_drop = data.columns.isin(constraint.constraint_columns)
+                    columns_to_drop = data.columns.where(indices_to_drop).dropna()
+                    data = data.drop(columns_to_drop, axis=1)
+
+            except Exception as e:
```
pvk-developer, line 152: same as above.

sdv/data_processing/data_processor.py
```diff
@@ -62,10 +80,217 @@ def _update_numerical_transformer(self, learn_rounding_scheme, enforce_min_max_v
...
+                are needed.
+
+        Returns:
+            dict:
+                Keyword arguments to use on the indicated model.
+        """
+        return copy.deepcopy(self._model_kwargs.get(model_name))
+
+    def set_model_kwargs(self, model_name, model_kwargs):
+        """Set the model kwargs used for the indicated model."""
```
pvk-developer, line 110: missing args description.

pvk-developer (approved):
Some minor comments, but overall looks good.

fealho (approved):
Everything looks reasonable to me 🎊

PR #974, opened 2022-08-26
Add flake8 libraries which require no/minimal changes

fealho:
Add flake8's which require minimal code change.

### DIFF
--- a/sdv/tabular/utils.py
+++ b/sdv/tabular/utils.py
@@ -154,18 +154,20 @@ def handle_sampling_error(is_tmp_file, output_file_path, sampling_error):
     if 'Unable to sample any rows for the given conditions' in str(sampling_error):
         raise sampling_error
 
+    error_msg = None
     if is_tmp_file:
-        print(  # noqa: T001
+        error_msg = (
             'Error: Sampling terminated. Partial results are stored in a temporary file: '
             f'{output_file_path}. This file will be overridden the next time you sample. '
             'Please rename the file if you wish to save these results.'
         )
     elif output_file_path is not None:
-        print(  # noqa: T001
-            'Error: Sampling terminated. '
-            f'Partial results are stored in {output_file_path}.'
+        error_msg = (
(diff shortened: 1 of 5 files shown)


tests/unit/metadata/test_multi_table.py
```diff
@@ -385,8 +385,11 @@ def test__validate_child_map_circular_relationship(self):
         }
 
         # Run / Assert
-        # TODO fix match error message
-        with pytest.raises(ValueError):
+        err_msg = (
+            'The relationships in the dataset describe a circular dependency between '
+            "tables {'users', 'sessions'}."
```
fealho, line 390: Did we ever decide how to deal with these?

amontanez24: @pvk-developer was going to write some functionality for this, but honestly I think it would be easier to just do 
```python
f'Example string with set {sorted(set_variable)}'
```
everywhere with sets, granted this would print it with square brackets

sdv/tabular/utils.py
```diff
@@ -155,12 +155,16 @@ def handle_sampling_error(is_tmp_file, output_file_path, sampling_error):
...
+            f'{output_file_path}. This file will be overridden the next time you sample. '
+            'Please rename the file if you wish to save these results.'
+        )
     elif output_file_path is not None:
-        print('Error: Sampling terminated. '
-              f'Partial results are stored in {output_file_path}.')
+        print(  # noqa: T001
+            'Error: Sampling terminated. '
+            f'Partial results are stored in {output_file_path}.'
+        )
```
amontanez24, line 167: It seems like the error gets raised in the end so maybe we should just make this the error message and raise it with the error. Not sure why we print. @katxiao any specific reason for this?

katxiao: My guess is because this part is more informational, and not really related to what the error is. I'm fine with moving it to the error, as long as this statement is always shown at the top of the stack trace, so the user can see it first.

amontanez24: @fealho any update on this?

Issue #620, opened 2021-10-29
Increase code style lint


pvk-developer:
### Problem Description

Currently our code is being validated only by `flake8` 'vanilla' and just a few plugins. We would like to increase the code style checks by adding more add-on's that follow our code style and our standards.

Also we would like to ensure that our docstrings are properly written and follow the rest of our format.

### Additional context

We have performed this task already on [RDT](https://github.com/sdv-dev/RDT/) , more precisely on the following issue:
https://github.com/sdv-dev/RDT/issues/248#issuecomment-932085083

### Docstring plugin

We need to add `pydocstyle` plugin with the following lines on our `setup.cfg` file as we are following the [google convention](https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings).

```
[pydocstyle]
convention = google
add-ignore = D107, D407, D417
```

### Flake8 plugins to be added
Flake8 comes with a lot of different addons that we can use to adapt it to our codestyle and checking, here is a list of plugins that I found to be interesting for us:

* `flake8-builtins` - Check for python builtins being used as variables or parameters.
* `flake8-comprehensions` - Helps you write better list/set/dict comprehensions.
* `flake8-debugger` - Debug statement checker.
* `flake8-variables-names` - Extension that helps to make more readable variables names.
* `Dlint` - Tool for encouraging best coding practices and helping ensure Python code is secure.
* `flake8-mock` - Provides checking mock non-existent methods.
* `flake8-fixme` - Check for FIXME, TODO and other temporary developer notes.
* `flake8-eradicate` - Plugin to find commented out or dead code.
* `flake8-mutable` - Extension for mutable default arguments.
* `flake8-print` - Check for print statements in python files.
* `flake8-pytest-style` - Plugin checking common style issues or inconsistencies.
* `flake8-quotes` - Extension for checking quotes in python.
* `flake8-multiline-containers` - Plugin to ensure a consistent format for multiline containers.
* `pandas-vet` - Plugin that provides opinionated linting for pandas code.
* `pep8-naming` - Check the PEP-8 naming conventions.
* `flake8-expression-complexity` - Plugin to validate expressions complexity.
*  `flake8-sfs` - String formatting.

PR #977, opened 2022-08-26
Add dlint

fealho:
Add dlint as part of https://github.com/sdv-dev/SDV/issues/620.

### DIFF
--- a/sdv/lite/tabular.py
+++ b/sdv/lite/tabular.py
@@ -257,7 +257,7 @@ class TabularPreset():
                 The loaded tabular model.
         """
         with open(path, 'rb') as f:
-            model = pickle.load(f)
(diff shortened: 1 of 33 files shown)


sdv/lite/tabular.py
```diff
@@ -257,7 +257,7 @@ def load(cls, path):
                 The loaded tabular model.
         """
         with open(path, 'rb') as f:
-            model = pickle.load(f)
+            model = pickle.load(f)  # noqa: DUO103
```
amontanez24, line 260: this is something we might want to think about moving forward. Perhaps there is a more secure way to save and load models

amontanez24 (approved):
:shipit:

PR #987, opened 2022-08-31
Add pep8-naming

fealho:
Add pep8-naming as part of #620.

### DIFF
--- a/sdv/constraints/errors.py
+++ b/sdv/constraints/errors.py
@@ -8,7 +8,7 @@ class MissingConstraintColumnError(Exception):
         self.missing_columns = missing_columns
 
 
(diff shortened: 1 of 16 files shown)


sdv/constraints/errors.py
```diff
@@ -8,7 +8,7 @@ def __init__(self, missing_columns):
         self.missing_columns = missing_columns
 
 
-class MultipleConstraintsErrors(Exception):
+class MultipleConstraintsError(Exception):
```
amontanez24, line 11: This name kind of makes it sound like the problem is that there are multiple constraints. Could we call this `AggregateConstraintsError` or something like that

PR #989, opened 2022-08-31
User Guide code fixes

### LINKED ISSUE #983: User Guide code fixes
### Environment Details

* SDV version: 0.17.0.dev1

### Error Description

Some areas of the documentation should be updated to account for the changes in the API. 

1. [Evaluation Framework](https://github.com/sdv-dev/SDV/blob/master/docs/user_guides/evaluation/evaluation_framework.rst): In "Can I control which metrics are applied?" the examples are exactly the same as the new defaults. We can update to different metrics instead: `['KSTest', 'LogisticDetection']`
2. [Custom Constraints](https://github.com/sdv-dev/SDV/blob/master/docs/user_guides/single_table/custom_constraints.rst): The model creation step references `min_value` and `max_value` even though these parameters don't exist anymore. Replace with `enforce_min_max_values=False`.

We should also run the rest of the code to ensure that there are no errors. (The second bug should've been caught if we had run the code.)

amontanez24:
resolves #983

### DIFF
--- a/docs/api_reference/metrics/relational.rst
+++ b/docs/api_reference/metrics/relational.rst
@@ -35,9 +35,9 @@ Multi Table Statistical Metrics
     CSTest
     CSTest.get_subclasses
     CSTest.compute
-    KSTest
-    KSTest.get_subclasses
-    KSTest.compute
+    KSComplement
+    KSComplement.get_subclasses
+    KSComplement.compute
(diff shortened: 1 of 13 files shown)


fealho (approved):
👍

katxiao (approved):
Nice, LGTM!

PR #992, opened 2022-09-01
Add pytest style

fealho:
Add flake8-pytest-style as part of #620.

### DIFF
--- a/tests/integration/tabular/test_base.py
+++ b/tests/integration/tabular/test_base.py
@@ -1,4 +1,3 @@
-import re
 from unittest.mock import patch
 
 import pandas as pd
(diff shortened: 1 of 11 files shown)


tests/integration/tabular/test_base.py
```diff
@@ -78,7 +78,7 @@ def test_conditional_sampling_graceful_reject_sampling_True_dict(model):
         })
     ]
 
-    with pytest.raises(ValueError):
+    with pytest.raises(ValueError):  # noqa: PT011
```
fealho, line 81: This can crash with different errors:

1) For GaussianCopula: `cannot reshape array of size 0 into shape (0)`
2) For other models: `Unable to sample any rows for the given conditions. Try increasing max_tries_per_batch (currently: 100). Note that increasing this value will also increase the sampling time.` 

Not sure if this is expected, I'll investigate. Regardless, the fix shouldn't be part of this PR, I think.

tests/integration/tabular/test_base.py
```diff
@@ -97,7 +97,7 @@ def test_conditional_sampling_graceful_reject_sampling_True_dataframe(model):
         'column3': [93]
     })
 
-    with pytest.raises(ValueError):
+    with pytest.raises(ValueError):  # noqa: PT011
```
fealho, line 100: same

amontanez24 (approved):
:shipit:

PR #995, opened 2022-09-06
Set HyperTransformer configuration manually

### LINKED ISSUE #982: Set HyperTransformer config manually, based on Metadata if given
### Environment Details

* SDV version: 0.17.0.dev1

### Error Description

1. The entire RDT HyperTransformer config is being printed out during modeling. This is unnecessarily verbose because I don't have any way of changing it right now. (Especially the case for HMA1.)
2. The HyperTransformer config does not properly correspond to the metadata that I provide. For example, if I specify that a column is type `categorical`, it is being detected and modeled as numerical.

### Steps to reproduce

```python
metadata, data = load_tabular_demo('student_placements', metadata=True)
model = GaussianCopula(
    table_metadata=metadata
)
model.fit(data)
synthetic_data = model.sample(num_rows=100)
```

Observe that the config is printed out. Also observe that the column `duration` is listed as categorical but is being read in as `numerical`. It's also being modeled as `numerical`, as the synthetic data includes values that are not previously see in the real data.

### Expected Fix

If we use HyperTransformer `set_config`, we can solve both issues.

- The config is printed out during `detect_initial_config`, which we don't need to do
- The config we set should be based on the metadata that the user provides (if applicable). If not provided, it should be based on the pandas dtypes of the passed in data.

pvk-developer:
Resolves #982

### DIFF
--- a/sdv/metadata/table.py
+++ b/sdv/metadata/table.py
@@ -389,7 +389,7 @@ class Table:
 
         return fields_metadata
 
-    def _get_transformers(self, dtypes):
+    def _get_hypertransformer_config(self, dtypes):
         """Create the transformer instances needed to process the given dtypes.
 
         Args:
@@ -398,15 +398,19 @@ class Table:
 
         Returns:
             dict:
-                mapping of field names and transformer instances.
+                A dict containing the ``sdtypes`` and ``transformers`` config for the
+                ``rdt.HyperTransformer``.
         """
         transformers = dict()
+        sdtypes = dict()
         for name, dtype in dtypes.items():
+            dtype = np.dtype(dtype).kind
             field_metadata = self._fields_metadata.get(name, {})
             transformer_template = field_metadata.get(
-                'transformer', self._dtype_transformers[np.dtype(dtype).kind])
+                'transformer', self._dtype_transformers[dtype])
 
             if transformer_template is None:
+                sdtypes[name] = self._DTYPES_TO_TYPES.get(dtype, {}).get('type', 'categorical')
                 transformers[name] = None
                 continue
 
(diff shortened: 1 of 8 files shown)


sdv/metadata/table.py
```diff
@@ -510,20 +515,20 @@ def _fit_hyper_transformer(self, data, extra_columns):
                 else:
                     dtypes[column] = dtype_kind
 
-        transformers_dict = self._get_transformers(dtypes)
+        hp_config = self._get_hypertransformer_config(dtypes)
```
amontanez24, line 518: should this be `ht_config`?

sdv/metadata/table.py
```diff
@@ -398,15 +398,19 @@ def _get_transformers(self, dtypes):
...
-                mapping of field names and transformer instances.
+                A dict containing the ``sdtypes`` and ``transformers`` config for the
+                ``rdt.HyperTransformer``.
         """
         transformers = dict()
+        sdtypes = dict()
         for name, dtype in dtypes.items():
+            dtype = np.dtype(dtype).kind
             field_metadata = self._fields_metadata.get(name, {})
             transformer_template = field_metadata.get(
```
amontanez24, line 409: does this take the metadata into account at all? I'm trying to figure out why it wasn't before and what has changed now. For example in the `student_placements` demo it wasn't making the `duration` column `categorical` even though the metadata said it was

amontanez24 (commented):
This mostly looks good to me. I just have a question about where the metadata is being used and why it didn't work before

sdv/metadata/table.py
pvk-developer: Old code was reading properly the `dtype` but was printing the wrong config which leaded to confusions as the `transformers` were updated after the `detect`. Now both `sdtypes` and `transformers` are being set corresponding to the `dtype` from the `metadata` with the `set_config` method.

amontanez24 (approved):
Looks good! :shipit:

sdv/metadata/table.py
```diff
@@ -398,15 +398,19 @@ def _get_transformers(self, dtypes):
...
-                'transformer', self._dtype_transformers[np.dtype(dtype).kind])
+                'transformer', self._dtype_transformers[dtype])
 
             if transformer_template is None:
+                sdtypes[name] = self._DTYPES_TO_TYPES.get(dtype, {}).get('type', 'categorical')
```
katxiao, line 413: Is there a unit test that checks this case (when `transformer_template` is `None`)?

katxiao (approved):
lgtm, with a question

PR #997, opened 2022-09-07
Reorder methods in single_table.py and multi_table.py files 

fealho:
Reorder methods in the metadata/multi_table.py file to be similar to the metadata/single_table.py ones.

Reorder helper methods in both files, so they are immediately above the methods which use them.


amontanez24 (approved):
:shipit:

PR #1004, opened 2022-09-11
Remove unnecessary `ignores` for lint

fealho:
This PR removes lint rules from the `ignore` list in the `setup.cfg` file, as was done in our other libraries. The removed rules (D203, D213, D413, D406, W503) don't require any code change. The rule (SFS2), however, doesn't allow the usage of `format` strings, so those instances were updated to f-strings throughout the code.

It also removes `pylint`, since it doesn't seem to be used, and reorders the style check libraries in `setup.py`, removing repeated entries.


setup.cfg
```diff
@@ -53,17 +55,4 @@ collect_ignore = ['setup.py']
 
 [pydocstyle]
 convention = google
-add-ignore = D105, D107, D407, D203, D213, D413, D406, D417
-
-[pylint]
```
fealho, line 58: @pvk-developer @amontanez24 Any idea why this is here? It doesn't seem to be used.

pvk-developer: I don't recall having `pylint` on `SDV`.

pvk-developer (approved):

setup.cfg
amontanez24: it comes from this PR https://github.com/sdv-dev/SDV/commit/6ed01d67a5670b511284b21f94f0ccf712af1a0d

amontanez24 (approved):
LGTM :shipit:

PR #1007, opened 2022-09-14
Create `NumericalFormatter`

### LINKED ISSUE #970: Create formatter
### Problem Description

As a user, it would be great if synthesized data matched the same numerical format (decimal places) as the fitted data.

The constraints have their `reverse_transform` after the `HyperTransformer` and sometimes ruin the formatting. As opposed to adding this formatting check in every constraint, we can pull out this logic into formatting classes that the `DataProcessor` can use.

### Expected behavior

- Create a `NumericalFormatter` class
- Add a `learn_format` method that learns the required information from a column.
- Add a `format_data` method that applies the learned format to a passed in column.

### Additional context
- We can decide whether to use this class in the constraints or at a higher level
- The `learn_format` method can basically copy these lines from the `FloatFormatter.fit` 
https://github.com/sdv-dev/RDT/blob/8d2894ce30635902ccbb5bc9b1e8bc60f388db78/rdt/transformers/numerical.py#L156-L163
- The `format_data` method can basically copy these lines from the `FloatFormatter.reverse_transform`
https://github.com/sdv-dev/RDT/blob/master/rdt/transformers/numerical.py#L204-L217

fealho:
Resolve #970.


sdv/data_processing/numerical_formatter.py
```diff
@@ -0,0 +1,101 @@
...
+            ``'UInt32'``, ``'UInt64'``, ``'Float'``.
+            Defaults to ``'Float'``.
+    """
+
+    _dtype = None
+    _min_value = None
+    _max_value = None
+    _rounding_digits = None
+
+    def __init__(self, learn_rounding_scheme=False, enforce_min_max_values=False,
```
fealho, line 43: SDV sets `learn_rounding_scheme` and `enforce_min_max_values` to True by default, while RDT uses False. I'm not sure why the defaults are different, so I used RDT's since then I didn't have to change the tests.

It doesn't matter much either way, since I don't rely on the default values anywhere.

sdv/metadata/single_table.py
```diff
@@ -35,9 +35,8 @@ class SingleTableMetadata:
     }
 
     _NUMERICAL_REPRESENTATIONS = frozenset([
-        'int', 'int64', 'int32', 'int16', 'int8',
-        'uint', 'uint64', 'uint32', 'uint16', 'uint8',
-        'float', 'float64', 'float32', 'float16', 'float8',
+        'Float', 'Int64', 'Int32', 'Int16', 'Int8',
```
fealho, line 38: Changing these to match RDT. Not sure about the naming scheme though, whether to use upper case letters or not, or maybe we should allow both (also, I'm removing 'int' and 'uint' since they don't mean anything, but maybe we could keep them as aliases for 'Int64'/'UInt64')? I'll leave as is for now, it shouldn't matter much either way, I imagine this parameter is not used frequently.

sdv/data_processing/numerical_formatter.py
```diff
@@ -0,0 +1,101 @@
+"""Formatter for numerical data."""
```
fealho, line 1: I wrote this in it's own file, but we could move it together with the DataProcessor, since it's the only place it's used. I don't have strong opinions either way.

amontanez24: I think it's fine to leave in its own file

amontanez24 (commented):
I mainly have one comment about using `Series` instead of `np.array` for the `format_data` method. Besides that this looks good!

sdv/data_processing/data_processor.py
```diff
@@ -294,14 +315,20 @@ def reverse_transform(self, data):
...
             if pd.api.types.is_integer_dtype(self._dtypes[column_name]):
                 column_data = column_data.round()
 
             dtype = self._dtypes[column_name]
             reversed_data[column_name] = column_data[column_data.notna()].astype(dtype)
 
+        # reformat numerical columns using the NumericalFormatter
+        for column in original_columns:
+            if column in self.formatters:
+                data_to_format = reversed_data[column].to_numpy()
```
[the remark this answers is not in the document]
fealho: Nope, I just left it the same as in RDT because then I didn't need to change the tests.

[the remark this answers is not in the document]
amontanez24: maybe we should change it. I think the only change that will have to be made is replacing the numpy arrays with series

amontanez24 (approved):
Thanks for addressing the feedback! :shipit:

PR #1008, opened 2022-09-14
Add anonymization to DataProcessor

### LINKED ISSUE #950: Add anonymization to DataProcessor
### Problem Description

_As a user, it would be useful if data that I mark as `pii` in the metadata gets anonymized._ 
The new `DataProcessor` class should utilize the `AnonymizedFaker` class to handle anonymization.

### Expected behavior

- When configuring which transformers should be used on each field and the `sdtype` is not one of the basic `sdtypes` with a predefined transformer, then we should do the following:
    - If the field is marked as `pii`, we should configure a `AnonymizedFaker` class based on the details in the Additional Context section.
    - Otherwise we should use the default categorical transformer.

### Additional context

The transformer we select if the field is marked as `pii` depends on the `sdtype`. We should follow the rules below

Data Type | Transformer (if PII)
-- | --
"physical_address" | AnonymizedFaker(provider_name="address", function_name="address")
"email" | AnonymizedFaker(provider_name="internet", function_name="email")
"latitude" | AnonymizedFaker(provider_name="geo", function_name="latitude")
"longitude" | AnonymizedFaker(provider_name="geo", function_name="longitude")
"ipv4_address" | AnonymizedFaker(provider_name="internet", function_name="ipv4")
"ipv6_address" | AnonymizedFaker(provider_name="internet", function_name="ipv6")
"mac_address" | AnonymizedFaker(provider_name="internet", function_name="mac_address")
"name" | AnonymizedFaker(provider_name="person", function_name="name")
"phone_number" | If premium: AnonymizedGeoExtractor()If not premium: AnonymizedFaker(provider_name="phone_number", function_name="phone_number")
"ssn" | AnonymizedFaker(provider_name="ssn", function_name="ssn")
"user_agent_string" | AnonymizedFaker(provider_name="user_agent", function_name="user_agent")
<Other Fakers>"<function_name>"OR<"provider"."function">eg."lorem.sentence" | AnonymizedFaker(provider_name="lorem", function_name="sentence")

If another `sdtype` is provided, then we should attempt to search through the Faker function names for one that matches the sdtype and use that if possible. If we cannot do a search, then we will require users to provide the sdtype in the following format: `provider_name.function_name`

pvk-developer:
Resolves #950

### DIFF
--- a/sdv/data_processing/anonymization.py
+++ b/sdv/data_processing/anonymization.py
@@ -6,7 +6,7 @@ from faker import Faker
 from rdt.transformers import AnonymizedFaker
 
 SDTYPE_ANONYMIZERS = {
-    'phisical_address': {
+    'physical_address': {
         'provider_name': 'address',
         'function_name': 'address'
     },
--- a/sdv/data_processing/anonymization.py
+++ b/sdv/data_processing/anonymization.py
@@ -1,0 +1,90 @@
+"""Anonymization module for the ``DataProcessor``."""
+
+import inspect
+
+from faker import Faker
+from rdt.transformers import AnonymizedFaker, get_default_transformers
+
+
+SDTYPE_ANONYMIZERS = {
+    'phisical_address': {
+        'provider_name': 'address',
+        'function_name': 'address'
+    },
+    'email': {
+        'provider_name': 'internet',
+        'function_name': 'email'
+    },
+    'latitude': {
+        'provider_name': 'geo',
+        'function_name': 'latitude'
+    },
+    'longitude': {
+        'provider_name': 'geo',
+        'function_name': 'longitude'
+    },
+    'ipv4_address': {
+        'provider_name': 'internet',
+        'function_name': 'ipv4'
+    },
+    'ipv6_address': {
+        'provider_name': 'internet',
+        'function_name': 'ipv6'
+    },
+    'mac_address': {
+        'provider_name': 'internet',
+        'function_name': 'mac_address'
+    },
+    'name': {
+        'provider_name': 'person',
+        'function_name': 'name'
+    },
+    'phone_number': {
+        'provider_name': 'phone_number',
+        'function_name': 'phone_number'
+    },
+    'ssn': {
+        'provider_name': 'ssn',
+        'function_name': 'ssn'
+    },
+    'user_agent_string': {
+        'provider_name': 'user_agent',
+        'function_name': 'user_agent'
+    },
+}
+
+
+def _detect_provider_name(function_name):
+    function_name = getattr(Faker(), function_name)
+    module = inspect.getmodule(function_name).__name__
+    module = module.split('.')
+    if len(module) == 2:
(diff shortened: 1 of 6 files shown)


sdv/data_processing/anonymization.py
```diff
@@ -0,0 +1,89 @@
...
+    },
+    'user_agent_string': {
+        'provider_name': 'user_agent',
+        'function_name': 'user_agent'
+    },
+}
+
+
+def _detect_provider_name(function_name):
+    function_name = getattr(Faker(), function_name)
```
amontanez24, line 57: I think you're right, if we can move this to `RDT` at some point that would be great

sdv/data_processing/anonymization.py
```diff
@@ -0,0 +1,89 @@
+"""Anonymization module for the ``DataProcessor``."""
+
+import inspect
+
+from faker import Faker
+from rdt.transformers import AnonymizedFaker
+
+SDTYPE_ANONYMIZERS = {
+    'phisical_address': {
```
amontanez24, line 9: typo: `physical_address`

sdv/data_processing/anonymization.py
```diff
@@ -0,0 +1,89 @@
...
+
+def _detect_provider_name(function_name):
+    function_name = getattr(Faker(), function_name)
+    module = inspect.getmodule(function_name).__name__
+    module = module.split('.')
+    if len(module) == 2:
+        return 'BaseProvider'
+
+    else:
+        return module[-1]
```
pvk-developer, line 64: Reminder: 
If we move this functionality into the `AnonymizedFaker` we can remove `faker` from the requirements.

amontanez24 (changes requested):
One typo, but could we also add an integration test?

tests/integration/data_processing/test_data_processor.py
```diff
@@ -0,0 +1,59 @@
...
+    # Reverse Transform
+    reverse_transformed = dp.reverse_transform(transformed)
+
+    # Assert
+    data_unique_occupations = data['occupation'].unique()
+    reverse_transformed_unique_occupations = reverse_transformed['occupation'].unique()
+    substracted = set(reverse_transformed_unique_occupations) - set(data_unique_occupations)
+
+    assert len(reverse_transformed_unique_occupations) == len(substracted)
+    assert 'occupation' not in transformed.columns
```
amontanez24, line 59: we could also assert that none of the values in the reversed are in the original

amontanez24 (approved):
One minor comment, but besides that looks good to me!

tests/integration/data_processing/test_data_processor.py
pvk-developer: Yes that is being tested with the `len` basically, if the unique values are `639` and after we rest the original ones they still are `639` (the substracted) it means that no old values are in the new ones 🤔 right?

fealho (approved):

PR #1017, opened 2022-09-20
Add method that generates keys to the `DataProcessor`

### LINKED ISSUE #994: Add method to generate keys to DataProcessor
### Problem Description

For certain multi-table algorithms, it would be good if the primary keys were available after the transform step. This means they can't be created at the end during the `reverse_transform`.

### Expected behavior

- Add a method `generate_primary_key` to the `DataProcessor`.
    - It should take one parameter: `num_rows` - the number of rows to generate
    - This method should use the `create_anonymized_column` method that will be added to the `HyperTransformer` in this [issue](https://github.com/sdv-dev/RDT/issues/546).
    - It should use the metadata to figure out how to initialize the `AnonymizedFaker` based on the `sdtype` for the primary key and then generate the column(s)

- This method should be added to the `transform` process when `is_condition` is False.

pvk-developer:
Resolves #994

### DIFF
--- a/sdv/data_processing/data_processor.py
+++ b/sdv/data_processing/data_processor.py
@@ -244,7 +244,7 @@ class DataProcessor:
 
         self._primary_keys = []
         if self.metadata._primary_key:
-            self._primary_keys = cast_to_iterable(self.metadata._primary_key)
+            self._primary_keys = list(cast_to_iterable(self.metadata._primary_key))
 
         for column in set(data.columns) - columns_created_by_constraints:
             column_metadata = self.metadata._columns.get(column)
--- a/sdv/data_processing/data_processor.py
+++ b/sdv/data_processing/data_processor.py
@@ -183,7 +183,29 @@ class DataProcessor:
         return data
 
     @staticmethod
-    def create_primary_key_transformer(sdtype, column_metadata):
+    def create_anonymized_transformer(sdtype, column_metadata):
+        """Create an instance of an ``AnonymizedFaker``.
+
+        Read the extra keyword arguments from the ``column_metadata`` and use them to create
+        an instance of an ``AnonymizedFaker`` transformer.
+
+        Args:
+            sdtype (str):
+                Sematic data type or a ``Faker`` function name.
+            column_metadata (dict):
+                A dictionary representing the rest of the metadata for the given ``sdtype``.
+
+        Returns:
+            Instance of ``rdt.transformers.pii.AnonymizedFaker``.
+        """
+        kwargs = {}
+        for key, value in column_metadata.items():
+            if key not in ['pii', 'sdtype']:
+                kwargs[key] = value
+
+        return get_anonymized_transformer(sdtype, kwargs)
+
+    def create_primary_key_transformer(self, sdtype, column_metadata):
         """Create an instance for the primary key.
 
         Read the keyword arguments from the ``column_metadata`` and use them to create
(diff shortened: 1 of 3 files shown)


sdv/data_processing/data_processor.py
```diff
@@ -202,20 +205,62 @@ def create_anonymized_transformer(self, sdtype, column_metadata):
...
+        return transformer
+
     def _create_config(self, data, columns_created_by_constraints):
         sdtypes = {}
         transformers = {}
         self._anonymized_columns = []
+
+        self._primary_keys = []
+        if self.metadata._primary_key:
+            self._primary_keys = list(cast_to_iterable(self.metadata._primary_key))
```
amontanez24, line 247: why do we need this? Can't we just reference the primary key through the metadata wherever we need it

sdv/data_processing/data_processor.py
```diff
@@ -202,20 +205,62 @@ def create_anonymized_transformer(self, sdtype, column_metadata):
...
+            column_metadata (dict):
+                A dictionary representing the rest of the metadata for the given ``sdtype``.
+
+        Returns:
+            transformer:
+                Instance of ``rdt.transformers.text.RegexGenerator`` or
+                ``rdt.transformers.pii.AnonymizedFaker`` with ``enforce_uniqueness`` set to
+                ``True``.
+        """
+        regex_format = column_metadata.get('regex_format')
```
amontanez24, line 227: why not check that the sdtype is `text`?

sdv/data_processing/data_processor.py
```diff
@@ -202,20 +205,62 @@ def create_anonymized_transformer(self, sdtype, column_metadata):
...
+        if regex_format:
+            transformer = rdt.transformers.RegexGenerator(
+                regex_format=regex_format,
+                enforce_uniqueness=True
+            )
+
+        else:
+            column_metadata['enforce_uniqueness'] = True
```
amontanez24, line 235: Idk if we should alter the original metadata. Maybe we should copy the dict if we are going to add to it

sdv/data_processing/data_processor.py
```diff
@@ -202,20 +205,62 @@ def create_anonymized_transformer(self, sdtype, column_metadata):
...
+
+        self._primary_keys = []
+        if self.metadata._primary_key:
+            self._primary_keys = list(cast_to_iterable(self.metadata._primary_key))
+
         for column in set(data.columns) - columns_created_by_constraints:
             column_metadata = self.metadata._columns.get(column)
             sdtype = column_metadata.get('sdtype')
+            sdtypes[column] = sdtype
             if column_metadata.get('pii'):
```
amontanez24, line 253: this is the same check as on line 259. Why not just move the line underneath to there?

amontanez24 (changes requested):
This looks mostly good! Just a few small comments and questions

sdv/data_processing/data_processor.py
```diff
@@ -202,20 +205,62 @@ def create_anonymized_transformer(self, sdtype, column_metadata):
...
+                ``True``.
+        """
+        regex_format = column_metadata.get('regex_format')
+        if regex_format:
+            transformer = rdt.transformers.RegexGenerator(
+                regex_format=regex_format,
+                enforce_uniqueness=True
+            )
+
+        else:
```
amontanez24, line 234: what if the primary key is an int? We should be able to generate a range like we used to

tests/integration/data_processing/test_data_processor.py
```diff
@@ -53,3 +55,125 @@ def test_data_processor_with_anonymized_columns(tmpdir):
...
+        - Anonymize the field ``occupation``.
+        - Create ``id`` and ``ssn`` columns.
+
+    Run:
+        - Create a ``DataProcessor`` with the new metadata.
+        - Fit the ``DataProcessor`` instance.
+        - Transform the data.
+        - Reverse transform the data.
+
+    Side effects:
```
amontanez24, line 80: I thought we were going to alabel this section `Assert` to match the sections in the test

amontanez24 (approved):
This looks good, but can we update the docstrings for the tests to match our new standard?

sdv/data_processing/data_processor.py
```diff
@@ -292,6 +336,25 @@ def fit(self, data):
...
+        Returns:
+            pandas.DataFrame:
+                A data frame with the newly generated primary keys of the size ``num_rows``.
+        """
+        if self._hyper_transformer.field_transformers.get(self._primary_key) is None:
+            return pd.DataFrame({self._primary_key: np.arange(num_rows)})
+
+        return self._hyper_transformer.create_anonymized_columns(
+            num_rows=num_rows,
+            column_names=[self._primary_key]
```
fealho, line 355: I believe `self._primary_key` can be a tuple, since it can be set in line 248. In such cases we would want to pass each of them separately no?

fealho (commented):
Looks good! I just have one question regarding what happens when primary_keys is a tuple.

fealho (approved):

sdv/data_processing/data_processor.py
```diff
@@ -336,23 +338,30 @@ def fit(self, data):
...
+            reset_primary_key (bool):
+                Whether or not reset the primary keys generators. Defaults to ``False``.
 
         Returns:
             pandas.DataFrame:
                 A data frame with the newly generated primary keys of the size ``num_rows``.
         """
         if self._hyper_transformer.field_transformers.get(self._primary_key) is None:
-            return pd.DataFrame({self._primary_key: np.arange(num_rows)})
+            if reset_primary_key:
```
amontanez24, line 355: do we need to add a unit test for this case?

amontanez24 (commented):
Changes look good to me! Once the test coverage for the unit tests is 100% :shipit:

PR #1027, opened 2022-09-23
Add `validate` to `BaseSynthesizer`

### LINKED ISSUE #1014: Add validate method to synthesizer
### Problem Description

_As a user, it would be helpful if I could check if my data was valid according to my metadata._

### Expected behavior

- Add a `validate` method to the `BaseSynthesizer`
- If the method finds an errors, it should raise an `InvalidDataError` in the following format

```python 
synthesizer.validate(data)

Error: Invalid values found for numerical column 'age': ('a', 'b', 'c', +more)
```
- The method should do the following checks
    - Columns described as sdtype=numerical should only have data that is numerical (or missing) – no other strings, etc. are allowed
`Error: Invalid values found for numerical column 'age': ('a', 'b', 'c', +more)`
    - Columns described as sdtype=datetime should only have data that datetime or missing – try to convert everything to the datetime format and error if it fails
`Error: Invalid values found for datetime column 'start_date': (0.0, 30.0, 4.4, +more)`
    - Columns described as sdtype=boolean should only have values that are: True, False or missing
`Error: Invalid values found for boolean column 'is_subscribed': (0.0, 30.0, 4.4, +more)`
    - Columns marked as a key (primary, alternate, sequence, or foreign) should not have any missing values
`Error: Key column 'user_id' contains missing values`
    - Columns marked as a primary or alternate keys should be unique in the table
`Error: Primary key column 'user_id' contains repeating values: ('UID_000', 'UID_001', 'UID_002', +more)`
    - (Sequential only) Context columns (stored in the model's parameters) should be fixed for each sequence – ie if you group by sequence key, the context columns should not vary
`Error: Context column 'patient_address' is changing inside sequence ('Patient_ID'='ID_004').`

### Additional context

- For the sequential case, we should override the method in the `PARSynthesizer`

fealho:
Resolve #1014.

**Note**: as described in the issue, the validate only checks for `numerical`, `boolean` and `datetime` sdtypes. `categorical`, `text` or any others are not being validated.

### DIFF
--- a/sdv/single_table/base.py
+++ b/sdv/single_table/base.py
@@ -144,6 +144,7 @@ class BaseSynthesizer:
         return errors
 
     def _validate_context_columns(self, data):
+        # NOTE: move method to PARSynthesizer when it has been implemented
         errors = []
         context_column_names = self._data_processor._model_kwargs.get('context_columns')
         if context_column_names:
--- a/sdv/single_table/base.py
+++ b/sdv/single_table/base.py
@@ -170,42 +170,44 @@ class BaseSynthesizer:
     def _is_numerical(value):
         return pd.isna(value) | pd.api.types.is_float(value) | pd.api.types.is_integer(value)
 
-    def _validate_column(self, data, column):
+    @staticmethod
+    def _is_boolean(value):
+        return True if pd.isna(value) | (value is True) | (value is False) else False
+
+    def _validate_column(self, column):
         """Validate values of the column satisfy its sdtype properties."""
         errors = []
-        sdtype = self.metadata._columns[column]['sdtype']
+        sdtype = self.metadata._columns[column.name]['sdtype']
 
         # boolean values must be True/False, None or missing values
         # int/str are not allowed
         if sdtype == 'boolean':
-            valid = data[column].apply(
-                lambda x: True if pd.isna(x) | (
-                    x is True) | (
-                    x is False) else False)
-            invalid_values = list(set(data[column][~valid]))
+            valid = column.apply(self._is_boolean)
+            invalid_values = set(column[~valid])
             if invalid_values:
                 invalid_values = self._update_invalid_values(invalid_values)
                 errors.append(
-                    f"Invalid values found for boolean column '{column}': {invalid_values}.")
+                    f"Invalid values found for boolean column '{column.name}': {invalid_values}.")
 
         # numerical values must be int/float, None or missing values
         # str/bool are not allowed
         if sdtype == 'numerical':
-            valid = data[column].apply(self._is_numerical)
-            invalid_values = list(set(data[column][~valid]))
+            valid = column.apply(self._is_numerical)
+            invalid_values = set(column[~valid])
             if invalid_values:
                 invalid_values = self._update_invalid_values(invalid_values)
                 errors.append(
-                    f"Invalid values found for numerical column '{column}': {invalid_values}.")
+                    f"Invalid values found for numerical column '{column.name}': {invalid_values}."
+                )
 
         # datetime values must be castable to datetime, None or missing values
         if sdtype == 'datetime':
-            valid = data[column].apply(lambda x: pd.isna(x) | is_datetime_type(x))
-            invalid_values = list(set(data[column][~valid]))
+            valid = column.apply(lambda x: pd.isna(x) | is_datetime_type(x))
+            invalid_values = set(column[~valid])
             if invalid_values:
                 invalid_values = self._update_invalid_values(invalid_values)
                 errors.append(
-                    f"Invalid values found for datetime column '{column}': {invalid_values}.")
+                    f"Invalid values found for datetime column '{column.name}': {invalid_values}.")
 
         return errors
 
@@ -226,7 +228,6 @@ class BaseSynthesizer:
                     * primary or alternate keys are not unique
                     * context columns vary for a sequence key
                     * values of a column don't satisfy their sdtype
-                    * values of a column don't satisfy their kwargs
         """
         if not isinstance(data, pd.DataFrame):
             raise ValueError(f'Data must be a DataFrame, not a {type(data)}.')
@@ -246,7 +247,7 @@ class BaseSynthesizer:
 
         # Every column must satisfy the properties of their sdtypes
         for column in data:
-            errors += self._validate_column(data, column)
+            errors += self._validate_column(data[column])
 
         if errors:
             raise InvalidDataError(errors)
--- a/sdv/single_table/base.py
+++ b/sdv/single_table/base.py
@@ -1,6 +1,8 @@
 """Base Synthesizer class."""
 
 import inspect
+import pandas as pd
+from sdv.constraints.utils import cast_to_datetime64
 
 from sdv.data_processing.data_processor import DataProcessor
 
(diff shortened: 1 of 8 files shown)


sdv/single_table/base.py
```diff
@@ -42,3 +46,175 @@ def get_parameters(self):
...
+            repeated_values = set(data[key][data[key].duplicated()])
+            if repeated_values:
+                repeated_values = self._update_invalid_values(repeated_values)
+                errors.append(f"Key column '{key}' contains repeating values: {repeated_values}")
+
+        return errors
+
+    def _validate_context_columns(self, data):
+        errors = []
+        context_column_names = self._data_processor._model_kwargs.get('context_columns')
```
fealho, line 118: This is currently incorrect. I was unsure where we want to put it (inside the PAR model I presume?), so I just implemented/tested it here.

sdv/single_table/base.py
```diff
@@ -42,3 +46,175 @@ def get_parameters(self):
     def get_metadata(self):
         """Return the ``SingleTableMetadata`` for this synthesizer."""
         return self.metadata
+
+    def _validate_metadata_matches_data(self, columns):
+        errors = []
+        metadata_columns = self.metadata._columns if self.metadata._columns else []
```
amontanez24, line 52: I think you can just do `metadata_columns = self.metadata._columns or  []`

sdv/single_table/base.py
```diff
@@ -42,3 +46,175 @@ def get_parameters(self):
...
+        invalid_data_columns = set(columns).difference(metadata_columns)
+        if invalid_data_columns:
+            errors.append(
+                f'The columns {sorted(invalid_data_columns)} are not present in the metadata.')
+
+        invalid_metadata_columns = set(metadata_columns).difference(columns)
```
amontanez24, line 58: I think this should be called `missing_columns` based on the error message. They aren't invalid

sdv/single_table/base.py
```diff
@@ -42,3 +46,175 @@ def get_parameters(self):
...
+
+        if errors:
+            raise InvalidDataError(errors)
+
+    def _get_primary_plus_alternate_keys(self):
+        keys = set()
+        if self.metadata._primary_key:
+            keys = {self.metadata._primary_key}
+
+        if self.metadata._alternate_keys:
```
amontanez24, line 73: I think by default the alternate keys is an empty list so this check is unnecessary

sdv/single_table/base.py
```diff
@@ -42,3 +46,175 @@ def get_parameters(self):
...
+            keys.update(set(self.metadata._alternate_keys))
+
+        return keys
+
+    def _get_sequence_keys(self):
```
amontanez24, line 78: can we add to the name that it returns a set? Otherwise I would expect it to just return whatever `self._sequence_key` is

sdv/single_table/base.py
```diff
@@ -42,3 +46,175 @@ def get_parameters(self):
...
+        keys = self._get_primary_plus_alternate_keys()
+        keys.update(self._get_sequence_keys())
+        for key in sorted(keys):
+            if pd.isna(data[key]).any():
+                errors.append(f"Key column '{key}' contains missing values.")
+
+        return errors
+
+    @staticmethod
+    def _update_invalid_values(invalid_values):
```
amontanez24, line 98: Can we rename this to `_format_invalid_values_string` or something like that?

sdv/single_table/base.py
```diff
@@ -42,3 +46,175 @@ def get_parameters(self):
...
+            if invalid_values:
+                invalid_values = self._update_invalid_values(invalid_values)
+                errors.append(
+                    f"Invalid values found for boolean column '{column.name}': {invalid_values}.")
+
+        # numerical values must be int/float, None or missing values
+        # str/bool are not allowed
+        if sdtype == 'numerical':
+            valid = column.apply(self._is_numerical)
+            invalid_values = set(column[~valid])
```
amontanez24, line 163: Any reason not to have the methods just return the set of invalid values themselves?

tests/unit/single_table/test_base.py
```diff
@@ -61,3 +69,328 @@ def test_get_metadata(self, mock_data_processor):
 
         # Assert
         assert result == metadata
+
+    def test_validate_type(self):
+        """When data is not of type ``pd.DataFrame``, an error should be raised."""
```
amontanez24, line 74: I think eventually we might run the lint on tests as well and they require docstrings to be in an imperative mood.

sdv/single_table/base.py
```diff
@@ -42,3 +46,175 @@ def get_parameters(self):
...
+
+        # boolean values must be True/False, None or missing values
+        # int/str are not allowed
+        if sdtype == 'boolean':
+            valid = column.apply(self._is_boolean)
+            invalid_values = set(column[~valid])
+            if invalid_values:
+                invalid_values = self._update_invalid_values(invalid_values)
+                errors.append(
+                    f"Invalid values found for boolean column '{column.name}': {invalid_values}.")
```
amontanez24, line 157: Maybe this whole chunk can be a method that takes in the sdtype and invalid values since it is used in three places

sdv/single_table/base.py
```diff
@@ -2,7 +2,11 @@
 
 import inspect
 
+import pandas as pd
+
+from sdv.constraints.utils import is_datetime_type
```
amontanez24, line 7: Maybe the `is_datetime_type` function should be moved to the `sdv.utils` level

amontanez24 (changes requested):
This looks pretty good! Mostly comments about restructuring and renaming methods

sdv/single_table/base.py
fealho: Seemed a little easier to read to me, though I don't have a preference. I can change it if you want?

amontanez24: I just notice they all follow the pattern of checking if they are the type, filtering out the ones that aren't and then raising an error with them. You could have methods that say `_get_invalid_numerical_values`, `_get_invalid_datetime_values`... It's minor though

amontanez24 (approved):
LGTM! Thanks for addressing :shipit:

sdv/single_table/base.py
```diff
@@ -42,3 +46,166 @@ def get_parameters(self):
...
+        if missing_metadata_columns:
+            errors.append(
+                f'The metadata columns {sorted(missing_metadata_columns)} '
+                'are not present in the data.'
+            )
+
+        if errors:
+            raise InvalidDataError(errors)
+
+    def _get_primary_plus_alternate_keys(self):
```
pvk-developer, line 68: Colud we use `and` instead of `plus` ?

sdv/single_table/base.py
```diff
@@ -42,3 +46,166 @@ def get_parameters(self):
...
+
+        return errors
+
+    @staticmethod
+    def _is_numerical(value):
+        return pd.isna(value) | pd.api.types.is_float(value) | pd.api.types.is_integer(value)
+
+    @staticmethod
+    def _is_boolean(value):
+        return True if pd.isna(value) | (value is True) | (value is False) else False
```
pvk-developer, line 139: Maybe move this to `utils` as `is_datetime`.

sdv/single_table/base.py
```diff
@@ -42,3 +46,166 @@ def get_parameters(self):
...
+
+        return []
+
+    def _validate_column(self, column):
+        """Validate values of the column satisfy its sdtype properties."""
+        errors = []
+        sdtype = self.metadata._columns[column.name]['sdtype']
+
+        # boolean values must be True/False, None or missing values
+        # int/str are not allowed
```
pvk-developer, line 156: I think we can move this in the `docstring` or just remove it.

tests/unit/single_table/test_base.py
```diff
@@ -61,3 +69,328 @@ def test_get_metadata(self, mock_data_processor):
...
+            "\nInvalid values found for numerical column 'num1': ['10', False, True, '+ 1 more']."
+            '\n'
+            "\nInvalid values found for numerical column 'num2': "
+            '[datetime.datetime(1, 1, 1, 0, 0)].'
+        )
+        with pytest.raises(InvalidDataError, match=err_msg):
+            instance.validate(data)
+
+    def test_validate(self):
+        """Test the method doesn't crash when the passed data is vaild.
```
pvk-developer, line 364: typo: `vaild` -> `valid`

tests/unit/single_table/test_base.py
```diff
@@ -61,3 +69,328 @@ def test_get_metadata(self, mock_data_processor):
...
+        """Test method doesn't raise when data is an empty dataframe."""
+        data = pd.DataFrame()
+        metadata = SingleTableMetadata()
+        instance = BaseSynthesizer(metadata)
+
+        # Run
+        instance.validate(data)
+
+    def test_validate_sdtypes(self):
+        """Test error is raised if column values don't satistfy their sdtype.
```
pvk-developer, line 319: typo: `satistfy` -> `satisfy`

pvk-developer (changes requested):
LGTM! I left some minor suggestions and a couple of typos! Let me know what you think :shipit:

sdv/single_table/base.py
fealho: If you don't have strong preferences, I'd prefer to leave it here. It's not that relevant to put in the docstrings, but I think it's useful for whoever is reading the code in the future (I, at least, was very confused what exactly each datatype means, and this comment avoids me having to check the utils file to figure it out).

pvk-developer: No strong preferences

PR #1035, opened 2022-09-27
Upgrade sdmetrics version

### LINKED ISSUE #1010: Patch release: Bump required version for `SDMetrics`
### Problem Description

As we're preparing for the SDMetrics `0.7.0` release, it would be good do a patch release in SDV to keep the ecosystem compatible. The only change is to increment the upper bound version by 1 minor release.

```
'sdmetrics>=0.6.0,<0.8'
```

This SDV version would be `0.17.1` -- because there is no other functionality that's changing.

### Timeframe

This can happen independently of SDMetrics release. SDV code shouldn't break if we do it before or after the SDMetrics release.

katxiao:
Resolves #1010

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -28,7 +28,7 @@ requirements:
     - ctgan >=0.5.2,<0.6
     - deepecho >=0.3.0.post1,<0.4
     - rdt >=1.2.0,<2
-    - sdmetrics >=0.6.0,<0.7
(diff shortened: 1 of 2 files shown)


pvk-developer (approved):

amontanez24:
Can we link and resolve this issue with this PR? #1010 
The issue only bumps the upper bound since 0.6 should still work

PR #1036, opened 2022-09-27
Only allow datetime and numerical sdtypes to be set as the sequence index

### LINKED ISSUE #1030: Only allow datetime and numerical sdtypes to be set as the sequence index
### Problem Description

_It only makes sense for sequence index to be datetimes or numbers. As a user, it would be helpful if I couldn't accidentally set the wrong type as a sequence index._

#### Warnings:
If the `set_sequence_index` method is called and a `sequence_index` already exists, raise a warning similar to the one for primary keys.

#### Errors:
If a user attempts to set a sequence index of a sdtype that is not `numerical` or `datetime`, raise an error.

amontanez24:
resolves #1030

### DIFF
--- a/sdv/metadata/single_table.py
+++ b/sdv/metadata/single_table.py
@@ -383,6 +383,10 @@ class SingleTableMetadata:
                 f'Unknown sequence index value {column_name}.'
                 ' Keys should be columns that exist in the table.'
             )
+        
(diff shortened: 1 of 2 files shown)


pvk-developer (approved):
Just a minor comment but it's good to go! 🚢

fealho (approved):

PR #1044, opened 2022-09-29
Add fit to synthesizers

### LINKED ISSUE #1019: Add fit and fit_processed_data methods to synthesizers
### Problem Description

_As a user, it would be useful to be able to fit a model to data that has already been transformed. This way, I can refit without doing all the transformations again. I can also have other people handle the transformations and just fit the processed data myself._

#### Add `fit_processed_data` method
- Parameters:
    - data (`pandas.DataFrame`): Data that has already been cleaned (nulls handled and converted to numeric). If constraints apply, the constraint transformations should also have been done here.
- The method should fit the underlying model in the synthesizer to the data.

#### Add  `fit` method.
- Parameters:
    - data (`pd.DataFrame`): The raw data (before any transformations)
- This method should do the following:
```python
cleaned_data = self.preproccess(data)
self.fit_processed_data(cleaned_data)
```

pvk-developer:
Resolves #1019

### DIFF
--- a/sdv/single_table/copulagan.py
+++ b/sdv/single_table/copulagan.py
@@ -152,7 +152,7 @@ class CopulaGANSynthesizer(CTGANSynthesizer):
         for column in processed_data.columns:
             column_name = column.replace('.value', '')
             sdtype = columns.get(column_name, {}).get('sdtype')
-            if column_name in columns and sdtype != 'categorical':
+            if column_name in columns and sdtype not in ['categorical', 'boolean']:
                 sdtypes[column] = 'numerical'
                 distribution = self._numerical_distributions.get(
                     column_name,
--- a/sdv/single_table/copulagan.py
+++ b/sdv/single_table/copulagan.py
@@ -149,9 +149,9 @@ class CopulaGANSynthesizer(CTGANSynthesizer):
         columns = self.metadata._columns
         transformers = {}
         sdtypes = {}
-        for column in processed_data:
+        for column in processed_data.columns:
             column_name = column.replace('.value', '')
-            sdtype = columns.get(column, {}).get('sdtype')
+            sdtype = columns.get(column_name, {}).get('sdtype')
             if column_name in columns and sdtype != 'categorical':
                 sdtypes[column] = 'numerical'
                 distribution = self._numerical_distributions.get(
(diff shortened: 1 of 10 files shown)


sdv/single_table/copulas.py
```diff
@@ -84,6 +86,25 @@ def __init__(self, metadata, enforce_min_max_values=True, enforce_rounding=True,
...
         }
 
+    def _fit(self, processed_data):
+        """Fit the model to the table.
+
+        Args:
+            processed_data (pandas.DataFrame):
+                Data to be learned.
+        """
+        numerical_distributions = deepcopy(self._numerical_distributions)
```
amontanez24, line 96: why do we have to copy here?

sdv/single_table/copulagan.py
```diff
@@ -144,4 +144,43 @@ def __init__(self, metadata, enforce_min_max_values=True, enforce_rounding=True,
...
         }
+
+    def _create_gaussian_normalizer_config(self, processed_data):
+        columns = self.metadata._columns
+        transformers = {}
+        sdtypes = {}
+        for column in processed_data.columns:
+            column_name = column.replace('.value', '')
+            sdtype = columns.get(column_name, {}).get('sdtype')
+            if column_name in columns and sdtype != 'categorical':
```
amontanez24, line 155: should we be doing the same with boolean columns? Perhaps `CopulaGANSynthesizer` should mark booleans as categorical as well.

sdv/single_table/utils.py
```diff
@@ -0,0 +1,38 @@
+"""Utility functions for tabular models."""
+
+import numpy as np
+
+
+def detect_discrete_columns(metadata, data):
```
amontanez24, line 6: we should test this method

amontanez24 (changes requested):
This is looking very good so far! I just left a couple questions and a request for more testing on the utility function.

sdv/single_table/copulagan.py
pvk-developer: Yes I forgot to ask about this, I left it as the legacy code, but I think that the intention was to use booleans as well. Also, should booleans arrive as `True - False` rather than numerical `0. - 1.` ? I think that was the intention of the legacy code but that was not really achived.

sdv/single_table/copulas.py
pvk-developer: I think that copy is good as we alter the `numerical_distributions` when fitting, this can allow us to re-fit the same instance with the same numerical distributions that we set prior (as you can see further in the code the `numerical_distributions` get updated, so just for sanity I wanted to use a copy.

sdv/single_table/copulagan.py
amontanez24: I think it's fine either way. You can pass as `True` and `False`

sdv/single_table/copulas.py
amontanez24: The thing is we are going to in the future expose methods that let users see the fitted distributions. As long as there is a way to retrieve them from the column name I think it's fine

sdv/single_table/utils.py
```diff
@@ -1,7 +1,5 @@
 """Utility functions for tabular models."""
 
-import numpy as np
-
 
 def detect_discrete_columns(metadata, data):
     """Detect th discrete columns in a dataset.
```
amontanez24, line 5: I know it's not your code, but typo: th

amontanez24 (approved):
I think this looks ready to go! :shipit:

fealho (approved):

PR #1049, opened 2022-10-03
adding manual trigger to workflow for docs

amontanez24:


.github/workflows/docs.yml
```diff
@@ -3,6 +3,7 @@ name: Generate Docs
 on:
   push:
     branches: [ stable ]
+  workflow_dispatch:
```
pvk-developer, line 6: No need to add anything else ?

amontanez24: nope

pvk-developer (approved):
:shipit:

PR #1068, opened 2022-10-18
Create PARSynthesizer

### LINKED ISSUE #1055: Create PARSynthesizer
### Problem Description

The `PAR` model needs to be migrated to the new synthesizer structure.

### Expected behavior

- There should be a new module called `sequential` in sdv (the same level as `single_table`)
- There should be a class called `PARSynthesizer`
- The `PARSynthesizer` should have the following init parameters
    - metadata: A SingleTableMetadata object
    - enforce_min_max_values: A boolean describing whether or not to strictly enforce the observed min/max values in numerical columns
    - enforce_rounding: A boolean describing whether or not to round the synthetic data based on the input data
    - `context_columns`: A list of strings, representing the columns that do not vary in a sequence
    - `context_synthesizer`: A string with the name of the model to use the context values
        - (default) 'GaussianCopulaSynthesizer'
        - Available options: 'GaussianCopulaSynthesizer', 'CTGANSynthesizer', 'CopulaGANSynthesizer', 'TVAESynthesizer'
    - `context_synthesizer_parameters`: A dictionary that maps each parameter name to a parameter values. Refer to the context model for the parameters that are allowed.
    - `segment_size`
    - `epochs`
    - `cuda`
    - `sample_size`
    - `verbose` 

### Additional context

There are a few main changes from the current `PAR` model implementation
1. The `context_synthesizer` now only accepts a string, but will work with any single-table synthesizer. It is now accompanied by the new `context_synthesizer_parameters` parameter which allows users to specify the configuration for the `context_model`.
2. The `PARSynthesizer` inherits directly from the `BaseSynthesizer` class. 
    - If there end up being too many differences, another option could be to just make `PARSynthesizer` not inherit from anything. It will have different sampling methods from the other synthesizers. Configuring the `DataProcessor` for it might also be challenging since it has to only transform the columns for the context model.
    - A third option could be to move the sampling methods to a `BaseSingleTableSynthesizer` class that inherits from the `BaseSynthesizer`
4. The current [BaseTimeseriesModel](https://github.com/sdv-dev/SDV/blob/master/sdv/timeseries/base.py) and `PAR` model can be combined into one class now since there aren't any other sequential models.

amontanez24:
resolves #1055


amontanez24:
The issue states to add parameters to allow the specification of the model to use for the context columns, but for now we'll just hide this, so I didn't add those parameters

sdv/sequential/par.py
```diff
@@ -0,0 +1,94 @@
...
+        self._data_processor = DataProcessor(metadata)
+        self.context_columns = context_columns
+        self.segment_size = segment_size
+        self._model_kwargs = {
+            'epochs': epochs,
+            'sample_size': sample_size,
+            'cuda': cuda,
+            'verbose': verbose,
+        }
+        self._verbose = verbose
```
fealho, line 71: Is there a need for the verbose attribute when it's already stored in the _model_kwargs?

amontanez24: good point. Looks like in the old timeseries model we had another reference to it because it gets used in sampling, but you're right, we can just access it through the model kwargs

pvk-developer (approved):

PR #1076, opened 2022-10-20
Create `auto_assign_transformers` and `get_transformers` for `Synthesizers`

### LINKED ISSUE #1020: Add assign_transformers and get_transformers methods to synthesizers
### Problem Description

_As a user, it would be useful if I could call a method that autonomously assigns transformers to my data using the metadata. It would also be helpful if I could examine those transformers afterwards._

#### Add `auto_assign_transformers` method to `BaseSynthesizer`
- Parameters:
    - data (`pd.DataFrame`): The raw data.
- This method should use the `DataProcessor` to do the following:
    1. Transform the constraints
    2. Create the config for the `HyperTransformer` and set it
- This method **SHOULD NOT** fit the `HyperTransformer`

#### Add `get_transformers` method to `BaseSynthesizer`
- This method should return a dictionary mapping the column name to the **actual transformer object** that will be used. This is essentially the `transformers` section of the `HyperTransformer`'s config.
- Errors:
    - If the `auto_assign_transformers` has not already been called, then this should crash
`Error: No transformers were returned in 'get_transformers'. Use 'assign_transformers' or 'fit' to create them.`

### Additional context

- The `auto_assign_transformers` method can be used in some circumstances in the `preprocess_data` method in #1018

pvk-developer:
Resolves #1020

### DIFF
--- a/sdv/data_processing/data_processor.py
+++ b/sdv/data_processing/data_processor.py
@@ -314,6 +314,10 @@ class DataProcessor:
 
         self._hyper_transformer.update_transformers(column_name_to_transformer)
 
+    def _create_hyper_transformer_instance(self):
+        if self._hyper_transformer is None:
+            self._hyper_transformer = rdt.HyperTransformer()
+
     def _fit_hyper_transformer(self, data, columns_created_by_constraints):
         """Create and return a new ``rdt.HyperTransformer`` instance.
 
@@ -331,13 +335,12 @@ class DataProcessor:
         Returns:
             rdt.HyperTransformer
         """
-        if self._hyper_transformer is None:
-            self._hyper_transformer = rdt.HyperTransformer()
-
-        if not self._hyper_transformer._fitted:
+        self._create_hyper_transformer_instance()
+        if self._hyper_transformer.field_transformers == {}:
             config = self._create_config(data, columns_created_by_constraints)
             self._hyper_transformer.set_config(config)
 
+        if not self._hyper_transformer._fitted:
             if not data.empty:
                 self._hyper_transformer.fit(data)
 
--- a/sdv/data_processing/data_processor.py
+++ b/sdv/data_processing/data_processor.py
@@ -349,7 +349,18 @@ class DataProcessor:
                 )
                 self.formatters[column_name].learn_format(data[column_name])
 
-    def _prepare_fitting(self, data):
+    def prepare_for_fitting(self, data):
+        """Prepare the ``DataProcessor`` for fitting.
+
+        This method will learn the ``dtypes`` of the data, fit the numerical formatters,
+        fit the constraints and create the configuration for the ``rdt.HyperTransformer``.
+        If the ``rdt.HyperTransformer`` has already been updated, this will not perform the
+        actions again.
+
+        Args:
+            data (pandas.DataFrame):
+                Table data to be learnt.
+        """
         if self._hyper_transformer.field_transformers == {}:
             LOGGER.info(f'Fitting table {self.table_name} metadata')
             self._dtypes = data[list(data.columns)].dtypes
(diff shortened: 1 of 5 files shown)


sdv/data_processing/data_processor.py
```diff
@@ -314,6 +314,10 @@ def update_transformers(self, column_name_to_transformer):
 
         self._hyper_transformer.update_transformers(column_name_to_transformer)
 
+    def _create_hyper_transformer_instance(self):
```
amontanez24, line 317: why aren't we just using the `HyperTransformer` in the `DataProcessor`? We shouldn't need to make another one

sdv/single_table/base.py
```diff
@@ -270,6 +270,44 @@ def get_metadata(self):
...
+
+        Args:
+            data (pandas.DataFrame):
+                The raw data (before any transformations) that will be used to fit the model.
+        """
+        constrained = self._data_processor._fit_transform_constraints(data)
+        columns_created_by_constraints = set(constrained.columns) - set(data.columns)
+        config = self._data_processor._create_config(constrained, columns_created_by_constraints)
+        self._data_processor._create_hyper_transformer_instance()
+        self._data_processor._hyper_transformer.set_config(config)
```
amontanez24, line 287: here we are missing that the formatters should also be fit. Can we instead make some methods in the `DataProcessor` that do these steps and then call them here?

amontanez24 (changes requested):

sdv/single_table/base.py
```diff
@@ -270,6 +270,44 @@ def get_metadata(self):
...
+        """Get a dictionary mapping of ``column_name``  and ``rdt.transformers``.
+
+        A dictionary representing the column names and the transformers that will be used
+        to transform the data.
+
+        Returns:
+            dict:
+                A dictionary mapping with column names and transformers.
+        """
+        field_transformers = {}
```
fealho, line 299: The logic can be tidied up a little:

```python3
if self._data_processor._hyper_transformer is not None:
    ft = self._data_processor._hyper_transformer.field_transformers
    if ft: return ft
raise
```

pvk-developer: Actually I forgot to change this logic, as `hyper_transformer` will no longer be `None`!

sdv/single_table/base.py
```diff
@@ -270,6 +270,41 @@ def get_metadata(self):
...
+
+        Args:
+            data (pandas.DataFrame):
+                The raw data (before any transformations) that will be used to fit the model.
+        """
+        constrained = self._data_processor._fit_transform_constraints(data)
+        columns_created_by_constraints = set(constrained.columns) - set(data.columns)
+        self._data_processor._fit_numerical_formatters(data)
+        config = self._data_processor._create_config(constrained, columns_created_by_constraints)
+        self._data_processor._hyper_transformer.set_config(config)
```
amontanez24, line 287: These steps are very similar to what happens in the `DataProcessor.fit`. Should we make another method that sets up the hypertransformer as well as does the constraints and formatting in the `DataProcessor` and then call that method both here and in `DataProcessor.fit`?

sdv/data_processing/data_processor.py
```diff
@@ -355,25 +349,35 @@ def _fit_numerical_formatters(self, data):
                 )
                 self.formatters[column_name].learn_format(data[column_name])
 
+    def _prepare_fitting(self, data):
```
amontanez24, line 352: we could make this method public honestly

amontanez24 (approved):
LGTM! Thanks for addressing the comments

sdv/data_processing/data_processor.py
pvk-developer: Sounds good 👍🏻

fealho (approved):

Grayangelz:
https://indico.math.cnrs.fr/

PR #1088, opened 2022-10-28
Add preprocess and fit methods to PARSynthesizer

amontanez24:


sdv/sequential/par.py
```diff
@@ -88,6 +113,102 @@ def get_parameters(self):
...
+
+    def _fit_sequence_columns(self, timeseries_data):
+        self._model = self._build_model()
+
+        # handle output name from rdt
+        if self._sequence_index:
+            modified_name = self._sequence_index + '.value'
+            if modified_name in timeseries_data.columns:
+                timeseries_data = timeseries_data.rename(columns={
+                    modified_name: self._sequence_index
```
amontanez24, line 160: I didn't test this line because it should be irrelevant after the RDT changes

sdv/sequential/par.py
```diff
@@ -88,6 +113,102 @@ def get_parameters(self):
 
         return instantiated_parameters
 
-    def get_metadata(self):
-        """Return the ``SingleTableMetadata`` for this synthesizer."""
-        return self.metadata
+    def preprocess(self, data):
+        """Transform the raw data to numerical space.
+
+        For PAR, none of the sequence keys are transformed.
```
pvk-developer, line 119: Missing `Args` and `Return`

sdv/sequential/par.py
```diff
@@ -88,6 +113,102 @@ def get_parameters(self):
...
+        else:
+            context = transformed[self._sequence_key].copy()
+            # Add constant column to allow modeling
+            context[str(uuid.uuid4())] = 0
+
+        context = context.groupby(self._sequence_key).first().reset_index()
+        self._context_synthesizer.fit(context)
+
+    def _build_model(self):
+        return PARModel(**self._model_kwargs)
```
pvk-developer, line 141: This could be removed, I don't think there is a need of a `_build_model` method unless you have a `self._model_class` which makes this abstraction worth it.

pvk-developer (approved):
I think that this looks good, just left couple of minor nits.

sdv/sequential/par.py
```diff
@@ -45,21 +54,37 @@ class PARSynthesizer:
...
+    def __init__(self, metadata, enforce_min_max_values=True, enforce_rounding=False,
+                 context_columns=None, segment_size=None, epochs=128, sample_size=1, cuda=True,
+                 verbose=False):
+        super().__init__(
+            metadata=metadata,
+            enforce_min_max_values=enforce_min_max_values,
+            enforce_rounding=enforce_rounding,
+        )
+        sequence_key = self.metadata._sequence_key
+        self._sequence_key = list(cast_to_iterable(sequence_key)) if sequence_key else None
```
fealho, line 84: cast_to_iterable already returns a list

sdv/sequential/par.py
```diff
@@ -88,6 +113,107 @@ def get_parameters(self):
...
+            diffs = np.diff(sequence_index).tolist()
+            data[sequence_index_idx] = diffs[0:1] + diffs
+            data.append(sequence_index[0:1] * len(sequence_index))
+
+    def _fit_sequence_columns(self, timeseries_data):
+        self._model = PARModel(**self._model_kwargs)
+
+        # handle output name from rdt
+        if self._sequence_index:
+            modified_name = self._sequence_index + '.value'
```
fealho, line 162: we don't add .value anymore?

fealho (approved):
👍

sdv/sequential/par.py
amontanez24: I'll get rid of this when we upgrade the rdt version to the one with the changes

sdv/sequential/par.py
amontanez24: `sequence_key` can be a tuple and if it is then `cast_to_iterable` returns a tuple

PR #1092, opened 2022-11-04
Add validate to `BaseMultiTableSynthesizer`

### LINKED ISSUE #1071: Add validate method to `BaseMultiTableSynthesizer`
### Problem Description

_As a user, it would be useful to see if the data I provide matches my metadata at the multi-table level._

### Acceptance criteria

- Add a method called `validate` to the `BaseMultiTableSynthesizer` class
    - This method should have the following parameters:
        - data: a dictionary mapping table name -> the pandas DataFrame for that table
    - The method should call the `SingleTableSynthesizer.validate` to accumulate those errors
    - Additionally, it should perform the following check: Columns marked as foreign keys should be referencing a primary key that exists in the parent table – ie no unknown parent references.
        - If this fails raise the following error:
        `Error: Foreign key column 'purchaser_id' contains unknown references: ('Unknown', 'USER_999', 'ZZZ', +more). All the values in must reference a primary key.`
### Expected behavior

```python
>>> synthesizer = HMASynthesizer(metadata)
>>> synthesizer.validate(data)

Error: Foreign key column 'purchaser_id' contains unknown references: ('Unknown', 'USER_999', 'ZZZ', +more). All the values in must reference a primary key.
```

pvk-developer:
Resolves #1071

### DIFF
--- a/sdv/multi_table/base.py
+++ b/sdv/multi_table/base.py
@@ -129,15 +129,19 @@ class BaseMultiTableSynthesizer:
             try:
                 self._table_synthesizers[table_name].validate(table_data)
 
-            except (InvalidDataError, ValueError) as e:
-                errors.append(e)
+            except (InvalidDataError, ValueError) as error:
+                error_msg = str(error)
+                error_msg = error_msg.replace(
+                    'The provided data does not match the metadata:',
+                    f"The provided data for table '{table_name}' does not match the metadata:"
+                )
+                errors.append(str(error_msg))
 
         if errors:
             error_msg = '\n'.join(errors)
-            raise InvalidDataError(error_msg)
+            raise InvalidDataError(errors)
 
         foreign_key_errors = self._validate_foreign_keys(data)
-
         if foreign_key_errors:
             error_msg = '\n'.join(foreign_key_errors)
             raise InvalidDataError(error_msg)
--- a/sdv/multi_table/base.py
+++ b/sdv/multi_table/base.py
@@ -105,7 +105,8 @@ class BaseMultiTableSynthesizer:
 
                     errors.append(
                         f"Error: foreign key column '{relation['child_foreign_key']}' contains "
-                        f'unknown references: {message}'
+                        f'unknown references: {message} All the values in this column must '
+                        'reference a primary key.'
                     )
             if errors:
                 error_msg = 'Relationships:\n'
@@ -142,12 +143,16 @@ class BaseMultiTableSynthesizer:
                 self._table_synthesizers[table_name].validate(table_data)
 
             except (InvalidDataError, ValueError) as error:
-                error_msg = str(error)
-                error_msg = error_msg.replace(
-                    'The provided data does not match the metadata:',
-                    f"Table: '{table_name}'"
-                )
+                if isinstance(error, InvalidDataError):
+                    error_msg = f"Table: '{table_name}'"
+                    for _error in error.errors:
+                        error_msg += f'\nError: {_error}'
+
+                else:
+                    error_msg = str(error)
+
                 errors.append(error_msg)
+
             except KeyError:
                 continue
 
--- a/sdv/multi_table/base.py
+++ b/sdv/multi_table/base.py
@@ -3,6 +3,8 @@
 from collections import defaultdict
 from copy import deepcopy
 
+import pandas as pd
+
 from sdv.single_table.copulas import GaussianCopulaSynthesizer
 from sdv.single_table.errors import InvalidDataError
 
@@ -85,26 +87,31 @@ class BaseMultiTableSynthesizer:
         return self.metadata
 
     def _validate_foreign_keys(self, data):
+        error_msg = None
         errors = []
         for relation in self.metadata._relationships:
-            child_table = data[relation['child_table_name']]
-            parent_table = data[relation['parent_table_name']]
-            child_row = child_table[relation['child_foreign_key']]
-            parent_row = parent_table[relation['parent_primary_key']]
-            missing_values = child_row[~child_row.isin(parent_row)].unique()
-            if any(missing_values):
-                message = ', '.join(missing_values[:5].astype(str))
-                if len(missing_values) > 5:
-                    message = f'({message}, + more).'
-                else:
-                    message = f'({message}).'
+            child_table = data.get(relation['child_table_name'])
+            parent_table = data.get(relation['parent_table_name'])
+            if isinstance(child_table, pd.DataFrame) and isinstance(parent_table, pd.DataFrame):
+                child_column = child_table[relation['child_foreign_key']]
+                parent_column = parent_table[relation['parent_primary_key']]
+                missing_values = child_column[~child_column.isin(parent_column)].unique()
+                if any(missing_values):
+                    message = ', '.join(missing_values[:5].astype(str))
+                    if len(missing_values) > 5:
+                        message = f'({message}, + more).'
+                    else:
+                        message = f'({message}).'
 
-                errors.append(
-                    f"Error: foreign key column '{relation['child_foreign_key']}' contains "
-                    f'unknown references: {message}'
-                )
+                    errors.append(
+                        f"Error: foreign key column '{relation['child_foreign_key']}' contains "
+                        f'unknown references: {message}'
+                    )
+            if errors:
+                error_msg = 'Relationships:\n'
+                error_msg += '\n'.join(errors)
 
-        return errors
+        return error_msg
 
     def validate(self, data):
         """Validate data.
@@ -128,13 +135,7 @@ class BaseMultiTableSynthesizer:
         errors = []
         missing_tables = set(self.metadata._tables) - set(data)
         if missing_tables:
-            if len(missing_tables) > 1:
-                raise InvalidDataError([
-                    f'The provided data is missing the tables {missing_tables}.'
-                ])
-            else:
-                raise InvalidDataError([
-                    f'The provided data is missing the table {missing_tables}.'])
+            errors.append(f'The provided data is missing the tables {missing_tables}.')
 
         for table_name, table_data in data.items():
             try:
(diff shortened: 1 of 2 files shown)


sdv/multi_table/base.py
```diff
@@ -82,3 +83,74 @@ def update_table_parameters(self, table_name, table_parameters):
...
+                    f'unknown references: {message}'
+                )
+
+        return errors
+
+    def validate(self, data):
+        """Validate data.
+
+        Args:
+            data (dict):
```
fealho, line 113: I'd call it tables. Data makes me think it's just a dataframe

sdv/multi_table/base.py
```diff
@@ -82,3 +83,74 @@ def update_table_parameters(self, table_name, table_parameters):
...
+                    * values of a column don't satisfy their sdtype
+        """
+        errors = []
+        missing_tables = set(self.metadata._tables) - set(data)
+        if missing_tables:
+            if len(missing_tables) > 1:
+                raise InvalidDataError([
+                    f'The provided data is missing the tables {missing_tables}.'
+                ])
+            else:
```
fealho, line 135: I don't think we need to be this careful with error messages (adding another error message just for the case where it's plural). I at least always use plural for a set of items. @amontanez24 any opinions on it?

tests/unit/multi_table/test_base.py
```diff
@@ -136,3 +142,223 @@ def test_get_metadata(self):
...
+            }),
+            'upravna_enota': pd.DataFrame({
+                'id_upravna_enota': np.arange(10),
+            }),
+        }
+
+        instance = BaseMultiTableSynthesizer(metadata)
+
+        # Run and Assert
+        with pytest.raises(InvalidDataError):
```
fealho, line 276: Add error message

tests/unit/multi_table/test_base.py
```diff
@@ -136,3 +142,223 @@ def test_get_metadata(self):
...
+            'upravna_enota': pd.DataFrame({
+                'id_upravna_enota': np.arange(10).astype(str),
+            }),
+        }
+
+        instance = BaseMultiTableSynthesizer(metadata)
+
+        # Run and Assert
+        error_msg = re.escape(
+            'The provided data does not match the metadata:\n'
```
fealho, line 326: This error message is a little weird. Maybe the colons at the end are supposed to be dots?

pvk-developer: This message comes from the `InvalidDataError` 🤔 I'm not sure if a new class is desired to be created with a different message. @amontanez24 any suggestions on that?

sdv/multi_table/base.py
pvk-developer: I'm fine with tables, I just followed the issue specification, @amontanez24 green light on the `tables` rather than `data`?

sdv/multi_table/base.py
```diff
@@ -82,3 +83,74 @@ def update_table_parameters(self, table_name, table_parameters):
...
+            if any(missing_values):
+                message = ', '.join(missing_values[:5].astype(str))
+                if len(missing_values) > 5:
+                    message = f'({message}, + more).'
+                else:
+                    message = f'({message}).'
+
+                errors.append(
+                    f"Error: foreign key column '{relation['child_foreign_key']}' contains "
+                    f'unknown references: {message}'
```
amontanez24, line 104: This message is missing the second half of the error: `All the values in this column must reference a primary key.`

sdv/multi_table/base.py
amontanez24: @npatki I think this was set to `data` to keep consistency with the `SingleTableSynthesizer` method, but since it does take a different type as input maybe it would make sense to change?

sdv/multi_table/base.py
```diff
@@ -82,3 +83,74 @@ def update_table_parameters(self, table_name, table_parameters):
     def get_metadata(self):
         """Return the ``MultiTableMetadata`` for this synthesizer."""
         return self.metadata
+
+    def _validate_foreign_keys(self, data):
+        errors = []
+        for relation in self.metadata._relationships:
+            child_table = data[relation['child_table_name']]
+            parent_table = data[relation['parent_table_name']]
+            child_row = child_table[relation['child_foreign_key']]
```
amontanez24, line 92: should this be called `child_column`? We are getting the entire column that matches the foreign key right? Not just a single row

sdv/multi_table/base.py
amontanez24: Yeah I think it's fine to just use plural and print it as a list or set

sdv/multi_table/base.py
```diff
@@ -82,3 +83,74 @@ def update_table_parameters(self, table_name, table_parameters):
...
+
+            except (InvalidDataError, ValueError) as error:
+                error_msg = str(error)
+                error_msg = error_msg.replace(
+                    'The provided data does not match the metadata:',
+                    f"The provided data for table '{table_name}' does not match the metadata:"
+                )
+                errors.append(error_msg)
+
+        if errors:
```
amontanez24, line 151: should we raise these before doing the foreign keys check, or just raise them all together at the bottom

amontanez24: @npatki I think the desire is to have all errors returned at once right?

tests/unit/multi_table/test_base.py
amontanez24: I think it's helpful to have the label for each individual table since that helps the user know where to look. Maybe we can get rid of the top line. @npatki do you have thoughts on this?

amontanez24 (changes requested):

sdv/multi_table/base.py
npatki: I see the same issue with `tables` though -- it does not imply anything about this being a dictionary. In fact, I'd assume it's a **list** if I just saw the word.

To me, `data` does not imply a `DataFrame` but rather some concept of the collected data. The nice thing about using `data` is that whenever we do a demo to future customers, etc. it's very clear what the concept is.

tests/unit/multi_table/test_base.py
npatki: Is this for validating data against the metadata? The idea was to collect and print all the errors at once, which is why there was probably a colon?

```python
>>> synthesizer.validate(data)

Error: Invalid values found for numerical column 'age': ('a', 'b', 'c', +more)
Error: Key column 'user_id' contains missing values
Error: Foreign key column 'purchaser_id' contains unknown references: ('Unknown', 'USER_999', 'ZZZ', +more). All the values in must reference a primary key.
```

npatki: The top line is for the overall error we throw (afaik, Python only allows you to throw 1 error at a time). The rest of the rows are the message.

Breaking up by table makes sense for cleanly summarizing where the errors happen. This could be similar to what we do in `metadata.validate()`: We split it up by table and then add a section for relationships.

```python
>>> synthesizer.validate(data)

Table: 'sessions'
Error: Invalid values found for numerical column 'age': ('a', 'b', 'c', +more)

Table: 'users'
Error: Key column 'user_id' contains missing values
...

Relationships:
Error: Foreign key column 'purchaser_id' contains unknown references: ('Unknown', 'USER_999', 'ZZZ', +more). All the values in must reference a primary key.
```

For context, I'm pasting below what we do for `metadata.validate()`:
```python

Table: 'users'
Error: Invalid values ("pii") for datetime column "start_date".
Error: A Unique constraint is being applied to column "age". This column is already a key for that table.

Table: 'transactions'
Error: Invalid regex format string "[A-{6}" for text column "transaction_id"
Error: Unknown key value 'ttid'. Keys should be columns that exist in the table.
Error: Invalid increment value (0.5) in a FixedIncrements constraint. Increments must be positive integers.

Relationships:
Error: Relationship between tables ('users', 'transactions') contains an unknown primary key 'userr_id'.
Error: The relationships in the dataset are disjointed. Tables ('sessions') are not connected to any of the other tables.

```

sdv/multi_table/base.py
npatki: The outcome is to throw a single `InvalidDataError` whose message contains all the problems that we can possibly fine.

So this means we should throw it after doing _all_ the checks (including any primary/foreign key checks.). This is similar to what we are doing in `metadata.validate()`.

sdv/multi_table/base.py
```diff
@@ -85,26 +87,31 @@ def get_metadata(self):
...
-
-                errors.append(
-                    f"Error: foreign key column '{relation['child_foreign_key']}' contains "
-                    f'unknown references: {message}'
-                )
-
-        return errors
+            child_table = data.get(relation['child_table_name'])
+            parent_table = data.get(relation['parent_table_name'])
+            if isinstance(child_table, pd.DataFrame) and isinstance(parent_table, pd.DataFrame):
```
amontanez24, line 95: why did we add the check for if they're dataframes?

sdv/multi_table/base.py
```diff
@@ -142,12 +143,16 @@ def validate(self, data):
...
-                    'The provided data does not match the metadata:',
-                    f"Table: '{table_name}'"
-                )
+                if isinstance(error, InvalidDataError):
+                    error_msg = f"Table: '{table_name}'"
+                    for _error in error.errors:
+                        error_msg += f'\nError: {_error}'
+
+                else:
+                    error_msg = str(error)
```
amontanez24, line 152: Instead of using an if statement, can we just have separate catch statements for the different error type? ie.
```python
except InvalidDataError as error:
    error_msg = f"Table: '{table_name}'"
    for _error in error.errors:
        error_msg += f'\nError: {_error}'
except ValueError as error:
    error_msg = str(error)

File: sdv/multi_table/base.py [hunk 10]
```diff
@@ -105,7 +105,8 @@ def _validate_foreign_keys(self, data):
 
                     errors.append(
                         f"Error: foreign key column '{relation['child_foreign_key']}' contains "
-                        f'unknown references: {message}'
+                        f'unknown references: {message} All the values in this column must '
```
amontanez24, line 108: Add a period after `{message}`

amontanez24 (approved):
A question and some minor comments but otherwise LGTM!

sdv/multi_table/base.py
pvk-developer: Before I was raising the error if the check failed on the `SingleTableSynthesizer` , however, since we want to have all the errors raised at once, I have to check this otherwise it will fail as key error or something else.

fealho (approved):

PR #1093, opened 2022-11-07
Add preprocess and fit to MultiTable synthesizers

### LINKED ISSUE #1074: Add preprocess and fit methods to multi-table synthesizers
### Problem Description

_As a user, it would be useful to be able to preprocess my data in a separate step from modeling. It would also be helpful to do this from the multi-table level._

### Acceptance criteria

Add the following methods:

- `preprocess(data)`
    - `data` is a dictionary mapping each table name to a pandas.dataFrame
    - This method should essentially loop through the single table synthesizers for each table and call preprocess on them with the proper data
    - It should return a dictionary mapping each table name to the transformed data
    - This method can be added to the `BaseMultiTableSynthesizer`
    - It should only raise one warning if any of the synthesizers have been fit. The warning should read:
    `Warning: This synthesizer has already been fit. To use the new preprocessed data, please refit the synthesizer using 'fit' or 'fit_processed_data'`
- `fit_processed_data(processed_data)`
    - `processed_data` is a dictionary mapping each table name to a pandas.dataFrame. This data should have already been ran through he data processor.
    - This method will be specific to each `MultiTableSynthesizer`, so for now only needs to be implemented in the `HMASynthesizer`.
- `fit(data)`
    - `data` is a dictionary mapping each table name to a pandas.dataFrame
    - should call `preprocess` and then `fit_processed_data`

### Expected behavior

- `preprocess`
    - This method should essentially loop through each table and call `SingleTableSynthesizer.preprocess` with the correct data
-  `fit_processed_data(processed_data)`
    - This is where the current `HMA` algorithm should take place. Each child table should be modeled and then the parameters for that model should be used to extend the table of the parent until eventually the parent is modeled. The code in [hma](https://github.com/sdv-dev/SDV/blob/V1.0.0.dev/sdv/relational/hma.py) should be reviewed as influence.

### Additional context

- It is a requirement that the primary keys be available to the `MultiTableSynthesizer` before it fits the models. This should be satisfied as the `DataProcessor` now makes the primary key the index during `transform`
- There is a slight change in the workflow from what happens in [hma](https://github.com/sdv-dev/SDV/blob/V1.0.0.dev/sdv/relational/hma.py). We now transform each table first, and then will be calling the fit method for each model and extending the tables with model parameters of the child table.

pvk-developer:
Resolves #1074


Grayangelz (approved):

sdv/single_table/copulas.py
```diff
@@ -86,6 +88,7 @@ def __init__(self, metadata, enforce_min_max_values=True, enforce_rounding=True,
             field: self._validate_distribution(distribution)
             for field, distribution in (numerical_distributions or {}).items()
         }
+        self._num_rows = None
```
amontanez24, line 91: why is this just always being set to None? Shouldn't we set this value when we fit?

sdv/multi_table/base.py
```diff
@@ -161,3 +163,59 @@ def validate(self, data):
...
+
+        Returns:
+            dict:
+                A dictionary with the preprocessed data.
+        """
+        self.validate(data)
+        if self._fitted:
+            warnings.warn(
+                'This model has already been fitted. To use the new preprocessed data, '
+                "please refit the model using 'fit' or 'fit_processed_data'."
```
amontanez24, line 191: I think right now if someone runs `fit` twice in a row, they will see this warning even though they shouldn't

sdv/multi_table/hma.py
```diff
@@ -20,5 +27,181 @@ def __init__(self, metadata, synthesizer_kwargs=None):
...
+        """
+        foreign_keys = self._get_all_foreign_keys(table_name)
+        keys = {}
+        for fk in foreign_keys:
+            keys[fk] = table_data.pop(fk).to_numpy()
+
+        for column in table_data.columns:
+            column_data = table_data[column]
+            fill_value = 0 if column_data.isna().all() else column_data.mean()
+            table_data[column] = table_data[column].fillna(fill_value)
```
amontanez24, line 157: since this is now after rdt, what are the possible scenarios where there are missing values?

amontanez24 (commented):
This is looking good so far! I left a few comments and questions, but I'll continue to look through for more

sdv/single_table/copulas.py
pvk-developer: Yes I think I forgot to add it on the `fit`, I added it now.

sdv/single_table/utils.py
```diff
@@ -148,3 +152,216 @@ def validate_file_path(output_file_path):
         pass
 
     return output_path
+
+
+def get_nearest_correlation_matrix(matrix):
```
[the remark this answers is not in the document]
pvk-developer: Sure I will add them back to the class as originally.

sdv/multi_table/hma.py
pvk-developer: Splitting, I'm leaving this since we are introducing new columns with the parameters of each univariate.

sdv/single_table/copulas.py
```diff
@@ -141,3 +147,164 @@ def _sample(self, num_rows, conditions=None):
...
+           [[c_{2,1}], [c_{3,1}, c_{3,2}], ..., [c_{n,1},...,c_{n,n-1}]]
+
+        Corresponding to the values from the lower half of the original correlation matrix,
+        **excluding** the diagonal.
+
+        The output is the complete correlation matrix reconstructed using the given values
+        and scaled to the :math:`[-1, 1]` range if necessary.
+
+        Args:
+            triangle_covariange (list[list[float]]):
```
fealho, line 242: typo, triangular_covariance. Although since we are computing the correlation matrix, made it would be better to call it triangular_correlation instead

sdv/single_table/copulas.py
```diff
@@ -141,3 +147,164 @@ def _sample(self, num_rows, conditions=None):
...
+                A list that contains lists of floats of size 1, 2, 3... up to ``n-1``,
+                where ``n`` is the size of the target covariance matrix.
```
fealho, line 244: Same here, correlation, not covariance matrix.

tests/unit/single_table/test_utils.py
```diff
@@ -38,3 +44,336 @@ def test_detect_discrete_columns():
...
+    # Setup
+    error_msg = (
+        'Error: Sampling terminated. Partial results are stored in a temporary file: test.csv. '
+        'This file will be overridden the next time you sample. Please rename the file if you '
+        'wish to save these results.'
+        '\n'
+        'Test error'
+    )
+
+    # Run and assert
```
fealho, line 198: For all of these errors, I think we usually put `error_msg` in the `Run and Assert` section (and Assert is usually capitalized). I don't think you should bother changing it here, since it doesn't really matter, just noting so we are consistent in the future.

fealho (commented):
Looks good! I only have one question, besides that I have no complaints.

sdv/single_table/copulas.py
```diff
@@ -141,3 +147,164 @@ def _sample(self, num_rows, conditions=None):
...
+                univariate['scale'] = max(0, univariate['scale'])
+
+            univariates.append(univariate)
+
+        model_parameters['univariates'] = univariates
+        model_parameters['columns'] = columns
+
+        covariance = model_parameters.get('covariance')
+        if covariance:
+            model_parameters['covariance'] = self._rebuild_correlation_matrix(covariance)
```
fealho, line 294: I'm confused here, why does the `_rebuild_correlation_matrix` take in a covariance matrix, and returns one as well (or at least is stored in a parameter called covariance).

tests/unit/multi_table/test_base.py
```diff
@@ -342,3 +342,149 @@ def test_validate_missing_foreign_keys(self):
...
+                'upravna_enota': np.arange(10),
+            }),
+            'oseba': pd.DataFrame({
+                'upravna_enota': np.arange(10),
+                'id_nesreca': np.arange(10),
+            }),
+            'upravna_enota': pd.DataFrame({
+                'id_upravna_enota': np.arange(10),
+            }),
+        }
```
amontanez24, line 436: should we move this to a `get_data` method? Seems like it's used a lot

tests/unit/multi_table/test_hma.py
```diff
@@ -19,8 +23,231 @@ def test___init__(self):
...
+            '__oseba__id_nesreca__univariates__upravna_enota__scale': [np.nan] * 4,
+            '__oseba__id_nesreca__num_rows': [1.] * 4,
+        })
+
+        pd.testing.assert_frame_equal(expected_result, result)
+        assert instance._modeled_tables == ['oseba']
+        assert instance._max_child_rows['__oseba__id_nesreca__num_rows'] == 1
+
+    def test__pop_foreign_keys(self):
+        """Remove the foreign keys from the ``table_data``."""
```
amontanez24, line 138: minor: should we say `Test that ...`

tests/unit/multi_table/test_hma.py
```diff
@@ -19,8 +23,231 @@ def test___init__(self):
...
+        pd.testing.assert_frame_equal(expected_data, data)
+
+    def test__model_table(self):
+        """Test that ``_model_table`` performs the modeling.
+
+        Modeling consists in getting the table for the given table name,
+        learn the size of this, remove the foreign keys and clear any null values
+        by using the ``_clear_nans`` method. Then, fit the table model calling the
+        ``fit_processed_data``, add again the foreign keys, update the ``tables`` and
+        mark the table name as modeled within the ``instance._modeled_tables``.
```
amontanez24, line 181: minor: can we keep the tenses the same throughout this summary? For example:
```
Modeling consists of getting the table for the given table name,
learning the size of this table, removing the foreign keys and clearing
any null values by using the ``_clear_nans`` method. Then, fitting the table model by calling
``fit_processed_data``,  adding back the foreign keys, updating the ``tables`` and
marking the table name as modeled within the ``instance._modeled_tables``.
```

tests/unit/single_table/test_utils.py
```diff
@@ -38,3 +44,336 @@ def test_detect_discrete_columns():
...
+
+    Expect that the error is raised at the end of the function.
+
+    Input:
+        - False
+        - a temp file
+        - the sampling error
+
+    Side Effects:
+        - the error is raised.
```
amontanez24, line 214: should we change all the tests in this file to match the new testing guidelines?

amontanez24 (approved):
A few minor comments but I think this is good!

sdv/single_table/copulas.py
pvk-developer: I think this comes from `Copulas`, there is a deprecation warning for renaiming 'covariance' to 'correlation', but I can't recall why. 
```
./copulas/multivariate/gaussian.py:        warnings.warn('`covariance` will be renamed to `correlation` in v0.4.0',
```

fealho (approved):

PR #1094, opened 2022-11-09
Add sampling methods to PARSynthesizer

### LINKED ISSUE #1052: Sampled context columns in PAR must be in the same order
### Environment Details

* SDV version: 0.17.1

### Error Description

When using the PAR model's `sample` method and providing a context, there is an error if the context columns aren't in the same order as the original data.

**Expected Behavior**: I expect that it's ok if the columns are in a different order. They are named so it's possible to know which column is which.

### Steps to reproduce

```python
from sdv.demo import load_timeseries_demo
from sdv.timeseries import PAR
import pandas as pd

data = load_timeseries_demo()

model = PAR(
    entity_columns=['Symbol'],
    context_columns=['MarketCap', 'Sector', 'Industry'],
    sequence_index='Date',
    epochs=1
)

model.fit(data)

# these are in the wrong order
# original data has MarketCap, Sector then Industry
context = pd.DataFrame(data={
    'Symbol': ['Apple', 'Google'],  
    'Sector': ['Technology', 'Health Care'],
    'MarketCap': [1.2345e+11, 4.5678e+10],
    'Industry': ['Electronic Components', 'Medical/Nursing Services']
})

model.sample(context=context)
```

The above code will work if I just move the columns around
```python
context = pd.DataFrame(data={
    'Symbol': ['Apple', 'Google'],  
    'MarketCap': [1.2345e+11, 4.5678e+10],
    'Sector': ['Technology', 'Health Care'],
    'Industry': ['Electronic Components', 'Medical/Nursing Services']
})

model.sample(context=context)
```

### LINKED ISSUE #1083: Add sampling methods to PARSynthesizer
### Problem Description

_As a user, it would be helpful to have a clear understanding of how to sample sequential data whether I have to provide the context or not._

### Acceptance criteria

Add the following methods to the `PARSynthesizer`
- `sample`
    - Parameters:
        - (required) num_sequences(int): The number of sequences to sample
        - sequence_length(int): An integer that describes the length of every sequence
            - (default) None: The model will determine the length of every sequence
        - randomize_samples (bool):
            - True: Every sample() call will have randomly generated data that is different
            - (default) False: Every sample() call has the same data generated
- `sample_sequential_columns`
    - Parameters:
        - (required) context_columns (pandas.DataFrame): A DataFrame containing the rows for the unchanging context 
        - sequence_length (int): An integer that describes the length of every sequence
            - (default) None: The model will determine the length of every sequence
        - randomize_samples(bool):
            - True: Every sample() call will have randomly generated data that is different
            - (default) False: Every sample() call has the same data generated

### Expected behavior

```python
>>> synthesizer.sample(5, sequence_length=10, randomize_samples=True)
>>> context = pd.DataFrame({'name': ['sam', 'sam', 'sam']})
>>> synthesizer.sample_sequential_columns(context, 5, True)
```

### Additional context

- The current PAR model only has one sampling method with a bunch of different arguments. The functionality in that method can essentially be split up into the two methods above.

amontanez24:
resolves #1083 
resolves #1052

### DIFF
--- a/sdv/data_processing/data_processor.py
+++ b/sdv/data_processing/data_processor.py
@@ -60,7 +60,8 @@ class DataProcessor:
         'datetime': rdt.transformers.UnixTimestampEncoder(
             missing_value_replacement='mean',
             model_missing_values=True,
-        )
+        ),
+        'text': rdt.transformers.RegexGenerator()
     }
     _DTYPE_TO_SDTYPE = {
         'i': 'numerical',
@@ -284,7 +285,7 @@ class DataProcessor:
                 self._anonymized_columns.append(column)
 
             else:
-                transformers[column] = self._transformers_by_sdtype.get(sdtype)
+                transformers[column] = deepcopy(self._transformers_by_sdtype.get(sdtype))
 
         for column in columns_created_by_constraints:
             dtype_kind = data[column].dtype.kind
--- a/sdv/sequential/par.py
+++ b/sdv/sequential/par.py
@@ -277,11 +277,11 @@ class PARSynthesizer(BaseSynthesizer):
 
         return output
 
-    def _sample(self, context_columns, sequence_length=None, randomize_samples=False):
+    def _sample(self, context_columns, sequence_length=None):
         sampled = self._sample_from_par(context_columns, sequence_length)
         return self._data_processor.reverse_transform(sampled)
 
-    def sample(self, num_sequences, sequence_length=None, randomize_samples=False):
+    def sample(self, num_sequences, sequence_length=None):
         """Sample new sequences.
 
         Args:
@@ -290,8 +290,6 @@ class PARSynthesizer(BaseSynthesizer):
             sequence_length (int):
                 If passed, sample sequences of this length. If ``None``, the sequence length will
                 be sampled from the model.
-            randomize_samples (bool):
-                Whether or not to use a fixed seed when sampling. Defaults to False.
 
         Returns:
             pandas.DataFrame:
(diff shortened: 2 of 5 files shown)


sdv/data_processing/data_processor.py
```diff
@@ -284,7 +285,7 @@ def _create_config(self, data, columns_created_by_constraints):
                 self._anonymized_columns.append(column)
 
             else:
-                transformers[column] = self._transformers_by_sdtype.get(sdtype)
+                transformers[column] = deepcopy(self._transformers_by_sdtype.get(sdtype))
```
amontanez24, line 288: adding this for when the rdt changes come through

sdv/single_table/base.py
```diff
@@ -374,6 +366,32 @@ def _set_random_state(self, random_state):
         """
         self._model.set_random_state(random_state)
 
+    def _randomize_samples(self, randomize_samples):
```
amontanez24, line 369: I moved this originally so that PAR could use it, but the code seems to call `set_random_state` on the underlying model. `PAR` doesn't have this ability right now. If it is hard to add I might just remove all randomization related stuff from PAR until we can support it

sdv/sequential/par.py
```diff
@@ -217,3 +221,112 @@ def _fit(self, processed_data):
...
+            for column, value in zip(self.context_columns, context_values):
+                sequence_df[column] = value
+
+            output.append(sequence_df)
+
+        output = pd.concat(output)
+        output = output[self._output_columns].reset_index(drop=True)
+        if self._sequence_index:
+            output = output.rename(columns={
+                self._sequence_index: self._sequence_index + '.value'
```
pvk-developer, line 275: Will this `.value` go away after RDT changes?

sdv/sequential/par.py
```diff
@@ -217,3 +221,112 @@ def _fit(self, processed_data):
...
+                be sampled from the model.
+            randomize_samples (bool):
+                Whether or not to use a fixed seed when sampling. Defaults to False.
+
+        Returns:
+            pandas.DataFrame:
+                Table containing the sampled sequences in the same format as the fitted data.
+        """
+        context_columns = self._context_synthesizer._sample_with_progress_bar(
+            num_sequences, output_file_path='disable', show_progress_bar=False)
```
pvk-developer, line 301: minor nit, could we put them in a line per arg

pvk-developer (approved):
Great! :shipit:

sdv/sequential/par.py
amontanez24: yes it will!

fealho (approved):

PR #1098, opened 2022-11-14
Create SingleTablePreset

### LINKED ISSUE #1079: Create SingleTablePreset
### Problem Description

The `TabularPreset` class should follow the new naming conventions.

- Create a new file in the `lite` module called `single_table.py`
- Move the code in `lite/tabular.py` to this new file
- Rename the `TabularPreset` to `SingleTablePreset`
- Rewrite the code to wrap around the `GaussianCopularSynthesizer` instead of the old model
- It should now take in a `SingleTableMetadata` object
- Reorder the `__init__` params so that the metadata comes first
- Migrate all tests as well

### Expected behavior

```python
from sdv.lite import SingleTablePreset

synthesizer = SingleTablePreset(metadata, name='FAST_ML')
```

amontanez24:
resolves #1079

### DIFF
--- a/sdv/lite/__init__.py
+++ b/sdv/lite/__init__.py
@@ -1,7 +1,9 @@
 """SDV lite package that contains model presets."""
 
+from sdv.lite.single_table import SingleTablePreset
 from sdv.lite.tabular import TabularPreset
 
 __all__ = (
+    'SingleTablePreset',
     'TabularPreset',
 )
--- a/sdv/lite/single_table.py
+++ b/sdv/lite/single_table.py
@@ -5,7 +5,7 @@ import sys
 
 import cloudpickle
 import numpy as np
-from rdt.transformers import BinaryEncoder, FloatFormatter, FrequencyEncoder, UnixTimestampEncoder
+import rdt.transformers
 
 from sdv.single_table import GaussianCopulaSynthesizer
 from sdv.utils import get_package_versions, throw_version_mismatch_warning
@@ -19,7 +19,7 @@ PRESETS = {
 
 
 class SingleTablePreset():
-    """Class for all tabular model presets.
+    """Class for all single table synthesizer presets.
 
     Args:
         metadata (sdv.metadata.SingleTableMetadata):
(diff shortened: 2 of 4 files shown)


sdv/lite/single_table.py
```diff
@@ -0,0 +1,250 @@
...
+                'boolean': rdt.transformers.BinaryEncoder(
+                    missing_value_replacement=-1 if self._null_column else 'mode',
+                    model_missing_values=transformer_model_missing_values,
+                ),
+                'datetime': rdt.transformers.UnixTimestampEncoder(
+                    missing_value_replacement='mean' if self._null_column else 'mode',
+                    model_missing_values=transformer_model_missing_values,
+                ),
+            }
+            self._synthesizer._data_processor._transformers_by_sdtype.update(sdtype_transformers)
```
amontanez24, line 75: @katxiao Do you know if this is still necessary?

katxiao: My understanding is that we wanted to only model the null values in the transformers in certain cases (when constraints are present) to optimize performance, but that this shouldn't be a consideration using the newest RDT (cc @npatki). With the newest RDT, we should always model the null values using the new transformers.

sdv/lite/single_table.py
```diff
@@ -0,0 +1,250 @@
...
+    """
+
+    _synthesizer = None
+    _null_percentages = None
+    _null_column = False
+    _default_synthesizer = GaussianCopulaSynthesizer
+
+    def __init__(self, metadata, name):
+        if name not in PRESETS:
+            raise ValueError(f'`name` must be one of {PRESETS}.')
```
pvk-developer, line 38: 'name' instead of `name`

sdv/lite/single_table.py
pvk-developer, line 75: I think that we should move this logic outside the `__init__`, otherwise if we add more presents in the future this will get bigger and bigger.
I would recommend a function that accepts the `name` and then proceeds with the preset.

pvk-developer (changes requested):

sdv/lite/__init__.py
```diff
@@ -1,7 +1,9 @@
 """SDV lite package that contains model presets."""
 
+from sdv.lite.single_table import SingleTablePreset
 from sdv.lite.tabular import TabularPreset
 
 __all__ = (
+    'SingleTablePreset',
     'TabularPreset',
```
katxiao, line 8: Are we going to remove `TabularPreset`?

katxiao (approved):
lgtm

sdv/lite/__init__.py
amontanez24: yes, but I think in a separate pr

sdv/lite/single_table.py
```diff
@@ -0,0 +1,197 @@
...
+
+LOGGER = logging.getLogger(__name__)
+
+FAST_ML_PRESET = 'FAST_ML'
+PRESETS = {
+    FAST_ML_PRESET: 'Use this preset to minimize the time needed to create a synthetic data model.'
+}
+
+
+class SingleTablePreset():
```
pvk-developer, line 20: can remove the `()`

sdv/lite/single_table.py
```diff
@@ -0,0 +1,197 @@
...
+                  'Have any requests for custom presets? Contact the SDV team to learn '
+                  'more an SDV Premium license.\n')
+
+    def __repr__(self):
+        """Represent single table preset instance as text.
+
+        Returns:
+            str
+        """
+        return f'SingleTablePreset(name={self.name})'
```
pvk-developer, line 197: This misses the metadata and may end up confusing a bit. Usually if we override this like that we should be able to   `eval(repr(instance))` and get a new instance, in this case this is not possible because missing metadata (which makes no sense to be represented in here). However I think that a different approach may be desired. For example:
```
SingleTablePreset:
    preset name: 'FAST_ML'
    metadata: {
        ...
    }
```

pvk-developer (approved):
Left an open ended discussion, I think the pr looks good as it is, just the representation of the class feels a little bit odd.

sdv/lite/single_table.py
amontanez24: I think in this case the significance of printing the name is to show what type of preset was used. I think it would be unnecessary to print out the entire metadata. As you mentioned, the purpose is not to make it so that a new instance could be created by copying the text

PR #1100, opened 2022-11-16
Add methods

### LINKED ISSUE #1072: Add transformer related methods to BaseMultiTableSynthesizer
### Problem Description

_As a user, it would be useful to be able to update, learn or view transformers for each synthesizer in my dataset._

### Acceptance criteria

Add the following methods to the `BaseMultiTableSynthesizer`
- `auto_assign_transformers(data)`
    - `data` is a dictionary mapping the table name to a `pandas.DataFrame`
    - This method should go through all of the single table synthesizers for each table and call the underlying `auto_assign_transformers` with the appropriate DataFrame.
- `get_transformers(table_name)`
    - `table_name` is string
    - This method should access the single table synthesizer for the table and call its underlying `get_transformers` method.
- `update_transformers(table_name, column_name_to_transformer)`
    - `table_name` is a string
    - `column_name_to_transformer` is a dict mapping column names to transformer instances
    - This method should once again use the single table synthesizer method for the specified table

### Expected behavior

```python
>>> synthesizer.assign_transformers(data)

>>> synthesizer.get_transformers(table_name='users')
{
  'student_id': RegexGenerator(regex_format='patient_id[0-9]{4}')
  'ssn': AnonymizedFaker(),
  'child_age': FloatFormatter(),
  'parent_age#child_age#diff': FormatterFormater()
  ...
}

>>> from rdt.transformers as rt

>>> paf = rt.PeudoAnonymizedFaker(provider_name='address', function_name='address')
>>> pn = rt.AnonymizedGeoExtractor()
>>> pn2 = rt.AnonymizedGeoExtractor()

>>> model.update_transformers(
  column_name_to_transformer={
    'student_id': rt.RegexGenerator(regex_format='patient_id[0-9]{4}'),
    'ssn': paf,
    'child_age': rt.FloatFormatter(model_missing_values=True),
    'parent_age#child_age#diff': rt.FloatFormatter()
    'cell_phone': pn,
    'work_phone': pn2
})

```

fealho:
Resolve #1072.

### DIFF
--- a/sdv/multi_table/base.py
+++ b/sdv/multi_table/base.py
@@ -162,6 +162,10 @@ class BaseMultiTableSynthesizer:
         if errors:
             raise InvalidDataError(errors)
 
+    def _validate_table_name(self, table_name):
+        if table_name not in self._table_synthesizers:
+            raise InvalidDataError([f"Table '{table_name}' is not present in the metadata."])
+
     def auto_assign_transformers(self, data):
         """Automatically assign the required transformers for the given data and constraints.
 
@@ -177,9 +181,7 @@ class BaseMultiTableSynthesizer:
                 If a table of the data is not present in the metadata.
         """
         for table_name, table_data in data.items():
-            if table_name not in self._table_synthesizers:
-                raise InvalidDataError([f"Table '{table_name}' is not present in the metadata."])
-
+            self._validate_table_name(table_name)
             self._table_synthesizers[table_name].auto_assign_transformers(table_data)
 
     def get_transformers(self, table_name):
@@ -200,9 +202,7 @@ class BaseMultiTableSynthesizer:
             ValueError:
                 If ``table_name`` is not present in the metadata.
         """
-        if table_name not in self._table_synthesizers:
-            raise ValueError(f"Table '{table_name}' is not present in the metadata.")
-
+        self._validate_table_name(table_name)
         return self._table_synthesizers[table_name].get_transformers()
 
     def update_transformers(self, table_name, column_name_to_transformer):
@@ -218,7 +218,5 @@ class BaseMultiTableSynthesizer:
             ValueError:
                 If ``table_name`` is not present in the metadata.
         """
-        if table_name not in self._table_synthesizers:
-            raise ValueError(f"Table '{table_name}' is not present in the metadata.")
-
+        self._validate_table_name(table_name)
         self._table_synthesizers[table_name].update_transformers(column_name_to_transformer)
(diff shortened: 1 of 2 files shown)


sdv/multi_table/base.py
```diff
@@ -161,3 +161,64 @@ def validate(self, data):
...
+        Returns:
+            dict:
+                A dictionary mapping with column names and transformers.
+
+        Raises:
+            ValueError:
+                If ``table_name`` is not present in the metadata.
+        """
+        if table_name not in self._table_synthesizers:
+            raise ValueError(f"Table '{table_name}' is not present in the metadata.")
```
amontanez24, line 204: let's make it an `InvalidDataError` for all of them

tests/unit/multi_table/test_base.py
```diff
@@ -342,3 +342,93 @@ def test_validate_missing_foreign_keys(self):
...
+        # Run
+        instance.auto_assign_transformers(data)
+
+        # Assert
+        instance._table_synthesizers['nesreca'].auto_assign_transformers.assert_called_once_with(
+            table1)
+        instance._table_synthesizers['oseba'].auto_assign_transformers.assert_called_once_with(
+            table2)
+
+    def test_auto_assign_transformers_error(self):
```
amontanez24, line 369: minor: but for all these can we make the suffix `missing_table` instead of `error`

amontanez24 (commented):
Couple of minor comments but after that should be good to go 👍

sdv/multi_table/base.py
```diff
@@ -161,3 +161,64 @@ def validate(self, data):
...
+
+        Raises:
+            InvalidDataError:
+                If a table of the data is not present in the metadata.
+        """
+        for table_name, table_data in data.items():
+            if table_name not in self._table_synthesizers:
+                raise InvalidDataError([f"Table '{table_name}' is not present in the metadata."])
+
+            self._table_synthesizers[table_name].auto_assign_transformers(table_data)
```
pvk-developer, line 183: We can use `self.validate(data)` on the first lane, then we remove the `if - raise` statement.
Which leads me to this question:
Should we move `validate` as a decorator, which will allow us to do this (Its fine to just call `self.validate` but it's fancier I guess). @amontanez24 @fealho 
```
@validate
def method_that_uses_data(self, data):
   # logic that doesn't need to handle anything from the data
```

pvk-developer (changes requested):

sdv/multi_table/base.py
fealho: Seems reasonable to me.

amontanez24: yeah I think it's fine, I just don't know if we need to do all the validation every time

pvk-developer (approved):

amontanez24 (approved):
:shipit:

PR #1101, opened 2022-11-17
Add validate to PARSynthesizer

### LINKED ISSUE #1081: Add validate to PARSynthesizer
### Problem Description

_As a user, it would be helpful if there was a way to see if my sequential data is correct according to my metadata._

### Acceptance criteria

- Add a method called `validate` to the `PARSynthesizer`
- This method should make the same checks as the `BaseSynthesizer.validate` plus the following one
    - Context columns (stored in the model's parameters) should be fixed for each sequence – ie if you group by sequence key, the context columns should not vary
`Error: Context column 'patient_address' is changing inside sequence ('Patient_ID'='ID_004').`

### Expected behavior

```python
synthesizer.validate(data)

Error: Context column 'patient_address' is changing inside sequence ('Patient_ID'='ID_004').
```

### Additional context

- The `PARSynthesizer` does not inherit from the `BaseSynthesizer` class. This is mostly because the sampling methods for these classes are different. It might be worth investigating ways to refactor the code so that there isn't too much repetition. One option could be [mixins](https://coderbook.com/@marcus/deep-dive-into-python-mixins-and-multiple-inheritance/). We could have a mixin for most of the repeated functionality that both `PARSynthesizer` and `BaseSynthesizer` inherit from.

amontanez24:
resolves #1081

### DIFF
--- a/sdv/sequential/par.py
+++ b/sdv/sequential/par.py
@@ -4,7 +4,6 @@ import inspect
 import logging
 import uuid
 import warnings
-from copy import deepcopy
 
 import numpy as np
 import pandas as pd
@@ -16,7 +15,6 @@ from sdv.errors import SamplingError, SynthesizerInputError
 from sdv.metadata.single_table import SingleTableMetadata
 from sdv.single_table import GaussianCopulaSynthesizer
 from sdv.single_table.base import BaseSynthesizer
-from sdv.single_table.errors import InvalidDataError
 from sdv.utils import cast_to_iterable
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


sdv/sequential/par.py
```diff
@@ -82,8 +84,21 @@ def __init__(self, metadata, enforce_min_max_values=True, enforce_rounding=False
...
+                'constraints in the metadata.'
+            )
+            self._data_processor._constraints = []
+
         sequence_key = self.metadata._sequence_key
         self._sequence_key = list(cast_to_iterable(sequence_key)) if sequence_key else None
+        if context_columns and not self._sequence_key:
+            raise SynthesizerInputError(
+                'No sequence_keys are specified in the metadata. The PARSynthesizer cannot '
+                'model context_columns in this case.'
```
pvk-developer, line 99: Should `sequence_keys` and `context_columns` be highlighted with ' ?

sdv/single_table/base.py
```diff
@@ -176,6 +155,13 @@ def _validate_column(self, column):
 
         return errors
 
+    def _validate_synthesizer_specific_rules(self, data):
```
pvk-developer, line 158: Feels a bit too verbose, maybe with just `_validate` similar to what we do with `RDT._transform` and other `_` methods is good enough to specify that is 'synthesizer specific'.

pvk-developer (changes requested):

fealho (approved):
👍

pvk-developer (approved):
LGTM ! :shipit:

PR #1118, opened 2022-11-28
Add transformer methods to `PARSynthesizer`

### LINKED ISSUE #1082: Add transformer methods to PARSynthesizer
### Problem Description

_As a user, it would be useful to be able to update, learn or view transformers for my sequential data._

### Acceptance criteria

Add the following methods to the `PARSynthesizer`
- `auto_assign_transformers(data)`
    - `data` is a `pandas.DataFrame`
    - By default, this method shouldn't assign any transformers.
- `get_transformers()`
    - This should return the transformers for every column in the data.
- `update_transformers(column_name_to_transformer)`
    - `column_name_to_transformer` is a dict mapping column names to transformer instances
    - If the user tries to update the transformer for a context column, we should raise an error

### Expected behavior

```python
>>> synthesizer.assign_transformers(data)

>>> synthesizer.get_transformers(table_name='users')
{
  'patient_id': None
  'ssn': None,
  'time': None
  ...
}

>>> from rdt.transformers as rt

>>> paf = rt.PeudoAnonymizedFaker(provider_name='address', function_name='address')
>>> pn = rt.AnonymizedGeoExtractor()
>>> pn2 = rt.AnonymizedGeoExtractor()

>>> model.update_transformers(
  column_name_to_transformer={
    'patient_id': rt.RegexGenerator(regex_format='[0-9]{4}'),
})

Error: Transformers for context columns are not allowed to be updated.

```

### Additional context
- We should not return or allow the modification of any of the transformers in the context synthesizer. For now, that will be hidden from the user.

fealho:
Resolve #1082.

### DIFF
--- a/sdv/lite/single_table.py
+++ b/sdv/lite/single_table.py
@@ -32,10 +32,10 @@ class SingleTablePreset:
 
     def _setup_fast_preset(self, metadata):
         self._synthesizer = GaussianCopulaSynthesizer(
-                metadata=metadata,
-                default_distribution='norm',
-                enforce_rounding=False
-            )
+            metadata=metadata,
+            default_distribution='norm',
+            enforce_rounding=False
+        )
         self._synthesizer._data_processor._update_transformers_by_sdtypes(
             'categorical',
             rdt.transformers.FrequencyEncoder(add_noise=True)
--- a/sdv/sequential/par.py
+++ b/sdv/sequential/par.py
@@ -167,6 +167,23 @@ class PARSynthesizer(BaseSynthesizer):
         self.update_transformers(sequence_key_transformers)
         return super().preprocess(data)
 
+    def update_transformers(self, column_name_to_transformer):
+        """Update any of the transformers assigned to each of the column names.
+
+        Args:
+            column_name_to_transformer (dict):
+                Dict mapping column names to transformers to be used for that column.
+
+        Raises:
+            ValueError:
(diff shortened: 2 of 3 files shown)


sdv/sequential/par.py
```diff
@@ -167,6 +167,23 @@ def preprocess(self, data):
         self.update_transformers(sequence_key_transformers)
         return super().preprocess(data)
 
+    def update_transformers(self, column_name_to_transformer):
```
fealho, line 170: I'm only adding `update_transformers` because the other two methods are already implemented in base.

amontanez24: I think this should be good. Can you double check that the correct transformers are assigned when calling `auto_assign_transformers`? For PAR, everything should be `None` except for datetimes

amontanez24 (approved):
Looks good pending the answer to one question

sdv/lite/single_table.py
```diff
@@ -32,10 +32,10 @@ class SingleTablePreset:
 
     def _setup_fast_preset(self, metadata):
         self._synthesizer = GaussianCopulaSynthesizer(
-                metadata=metadata,
-                default_distribution='norm',
-                enforce_rounding=False
-            )
```
pvk-developer, line 38: 🤔 wondering why `lint` did not capture it

pvk-developer (approved):
LGTM!

PR #1133, opened 2022-12-06
Change `print` to `logging`

### LINKED ISSUE #1107: Use logging instead of printing in detect methods
### Problem Description

_As a user, it can be overwhelming if I get a full print out of the metadata every time the `detect_from_dataframe` or `detect_from_csv` methods are run._

### Expected behavior

- Instead of printing, we can use the logger to show info statements for the metadata. This way we can view the detected metadata if needed, but users can turn it off.

fealho:
Resolve #1107.

### DIFF
--- a/tests/integration/lite/test_single_table.py
+++ b/tests/integration/lite/test_single_table.py
@@ -26,9 +26,6 @@ def test_sample():
     assert len(samples) == 10
 
 
-test_sample()
-
-
 def test_sample_with_constraints():
     """Test sampling for the ``SingleTablePreset``."""
     # Setup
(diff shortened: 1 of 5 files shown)


tests/integration/lite/test_single_table.py
```diff
@@ -26,6 +26,9 @@ def test_sample():
     assert len(samples) == 10
 
 
+test_sample()
```
amontanez24, line 29: should this be deleted?

amontanez24 (changes requested):
Can we update this method as well?
https://github.com/sdv-dev/SDV/blob/47b1d11f4a2830d08418e08770a82f015fe4c912/sdv/metadata/multi_table.py#L285

amontanez24 (approved):

PR #1144, opened 2022-12-12
SingleTableSynthesizer fit doesn't update rounding

### LINKED ISSUE #1104: SingleTableSynthesizer fit doesn't update rounding
### Environment Details

* SDV version: 1.0.0.dev
* Python version: any
* Operating System: any

### Error Description

_As a user, if I refit the synthesizer, the rounding scheme doesn't update to the new data._

Additionally, the warning message if you refit says `This model has already been fitted. To use the new preprocessed data, please refit the model using 'fit' or 'fit_processed_data'.` This doesn't make any sense since `fit` is being called. It should only raise that warning if `preprocess` is called directly.

### Steps to reproduce

```python
data = pd.DataFrame(data={'Data': np.random.normal(4,1,size=1000)})

# first fit: the syntheszier correctly learns a rounding scheme of 4 digits
synthesizer.fit(data.round(4))
synthesizer.sample(5)

# second fit: the synthesizer still uses rounding scheme of 4 digits
synthesizer.fit(data.round(2))
synthesizer.sample(5)
```

### Technical Details
The rounding issue seems to be because of this line:
https://github.com/sdv-dev/SDV/blob/35c22114d44b7f04ec159c3743cda1c09a6d1b9f/sdv/data_processing/data_processor.py#L367-L372

We are only refitting if the `HyperTransformer` is empty. Instead we should refit everything if `preprocess` or `fit` are called since the data may be different.

amontanez24:
resolves #1104 

This PR does two main changes:
1. The `DataProcessor` refits everything (formatters, constraints and transformers) with every call to `preprocess` or `fit`, but it skips the fitting of the formatters and constraints when `auto_assign_transformers` is called after they've already been fit
2. The config for the `HyperTransformer` is only created if it doesn't already exist. This is because if users update the config using one of the update transformer methods, those changes should persist through consecutive fit calls. We only need to refit the formatters and constraints as certain values in the data may have changed, but the columns and sdtypes will not

### DIFF
--- a/sdv/data_processing/data_processor.py
+++ b/sdv/data_processing/data_processor.py
@@ -103,6 +103,7 @@ class DataProcessor:
         self._anonymized_columns = []
         self._primary_key = self.metadata._primary_key
         self._primary_key_generator = None
+        self._prepared_for_fitting = False
 
     def get_model_kwargs(self, model_name):
         """Return the required model kwargs for the indicated model.
@@ -364,7 +365,7 @@ class DataProcessor:
             data (pandas.DataFrame):
                 Table data to be learnt.
         """
-        if self._hyper_transformer.field_transformers == {}:
+        if not self._prepared_for_fitting:
             LOGGER.info(f'Fitting table {self.table_name} metadata')
             self._dtypes = data[list(data.columns)].dtypes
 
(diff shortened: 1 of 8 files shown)


fealho (commented):
Looks good to me, I'll approve after it passes the tests. Only thing I have to add is maybe you could add the code described in the issue as a test case, to show you actually fixed the original issue.

amontanez24:
> Looks good to me, I'll approve after it passes the tests. Only thing I have to add is maybe you could add the code described in the issue as a test case, to show you actually fixed the original issue.

I just added a test case with multiple fits

fealho (approved):

pvk-developer (approved):

PR #1148, opened 2022-12-14
Add `download_demo` method

### LINKED ISSUE #1128: Add demo loading functionality
### Problem Description

_As a user, it would be useful to be able to access demo data that I could use to test the SDV out with in a Sandbox like environment._

We want to have functionality similar to the current demo module that allows users to see available demo datasets, and download them.

### Acceptance criteria
- Add a file called `demo.py` to the `datasets` module
- Add a function called `download_demo`
    - The function should have the following parameters:
        - (required) `modality`: One of: 'single_table', 'multi_table', 'sequential'
        - (required) `dataset_name`: A string with the name of a dataset
        - `output_folder_name`: The name of the local folder where the metadata and data should be stored
            - (default) None: Do not save the data locally. Just load it as Python objects.
            - '<name>': Create a subfolder in the desired location to store the data. Note: only store metadata_v1.json and not metadata_v0.json.
    - Returns tuple of (data, metadata)
        - data:
            - (single table and sequential) A pandas DataFrame object
            - (multi table) A dictionary mapping table name (string) to a pandas DataFrame object
        - metadata
            - (single table and sequential) A `SingleTableMetadata` object
            - (multi table) A `MultiTableMetadata` object
    - Errors
        - If dataset name isn't provided:
`Error: Missing required parameter 'dataset_name'.`
        - If the dataset name exists in our bucket but under a different modality
`Error: Dataset name '<name>' is a <modality> dataset. Use 'load_<modality>_demo' to load this dataset.`
for eg.
`Error: Dataset name 'heart_rate' is a sequential dataset. Use 'load_sequential_demo' to load this dataset.`
        - If the dataset name doesn't exist in our bucket:
`Error: Invalid dataset name 'dataset_1'. Use 'list_available_demos' to get a list of demo datasets.`
        - If there is already a folder that exists
`Error: Folder 'my_datasets/student_placements/' already exists. Please specify a different name or use 'load_from_csv' to load from an existing folder.`

### Expected behavior

```python

  modality='single_table',
  dataset_name='student_placements',
  output_folder_name='my_datasets/student_placements')
```

### Additional context

- We should no longer be using the bucket that is currently used to get demo data. Instead we want to switch to the new demo data bucket. The new bucket has a designated folder for each modality which should make downloading the correct data easier.
- We also no longer want to save the csv every time and load data from there. Instead, we should download the data directly from the bucket and obtain the DataFrame from the bits that S3 returns.

fealho:
Resolve #1128.

Changes compared to the issue:
1. The error "if dataset name isn't provided" shouldn't be necessary, since it crashes automatically if you don't pass a required param.
2. Adding another error if modality is not one of the 3 required ones.
3. I'm assuming you meant load_csvs instead of load_from_csv in the final error message.


fealho:
@amontanez24 could you double check that the names of the functions mentioned in the errors are correct? That is, `load_csvs`, `load_{other_modality}_demo` and `list_available_demos` .

sdv/datasets/demo.py
```diff
@@ -0,0 +1,140 @@
...
+            data[filename] = pd.read_csv(data_path)
+
+    if modality != 'multi_table':
+        data = data.popitem()[1]
+
+    return data
+
+
+def _get_metadata(modality, output_folder_name):
+    metadata_path = os.path.join(output_folder_name, 'metadata.json')
```
amontanez24, line 92: the metadatas are all titled `metadata_v1.json`

sdv/datasets/errors.py
```diff
@@ -0,0 +1,5 @@
+"""Datasets Exceptions."""
+
+
+class InvalidArgumentError(Exception):
```
amontanez24, line 4: typo: arrguments

sdv/datasets/demo.py
```diff
@@ -0,0 +1,140 @@
...
+    return os.path.join(BUCKET_URL, modality.upper(), dataset_name + '.zip')
+
+
+def _validate_args_download_demo(modality, dataset_name, output_folder_name):
+    possible_modalities = ['single_table', 'multi_table', 'sequential']
+    if modality not in possible_modalities:
+        raise InvalidArgumentError(f"'modality' must be in {possible_modalities}.")
+
+    if output_folder_name and os.path.exists(output_folder_name):
+        raise InvalidArgumentError(
```
amontanez24, line 31: maybe we can just use `ValueError` for these

sdv/datasets/demo.py
```diff
@@ -0,0 +1,140 @@
...
+                    f"Use 'load_{other_modality}_demo' to load this dataset."
+                )
+            except urllib.error.HTTPError:
+                pass
+
+        # If the dataset doesn't exist at all, raise different error
+        raise InvalidArgumentError(
+            f"Invalid dataset name '{dataset_name}'. "
+            "Use 'list_available_demos' to get a list of demo datasets."
+        )
```
amontanez24, line 57: I think we should get rid of this whole section. Instead, during the actual `_download` method, we should throw that in a try catch and raise an error if it can't be downloaded suggesting that possibly it is a different modality.

sdv/datasets/demo.py
```diff
@@ -0,0 +1,140 @@
...
+        os.rename(
+            os.path.join(output_folder_name, 'metadata_v1.json'),
+            os.path.join(output_folder_name, 'metadata.json')
+        )
+
+
+def _get_data(modality, output_folder_name):
+    data = {}
+    for filename in os.listdir(output_folder_name):
+        if filename != 'metadata.json':
```
amontanez24, line 81: again I think the filename should be `metadata_v1.json`. We should just pull this up into a constant global variable

amontanez24 (changes requested):

sdv/datasets/demo.py
fealho: Yeah, I implemented like this to validate whether the dataset is in another modality. Since we removed it, this is unnecessary.

PS: also, I originally implemented it using the `requests` library, so I could make a HEAD request, which doesn't cost much. But I forgot that we don't have `requests`, and I don't know how to do it with `urllib`...

sdv/datasets/demo.py
fealho: I renamed them to `metadata` because it looked a little silly to have a `metadata_v1` if we don't have a `metadata_v0`. If you think it could cause some issue/confusion though, I'll just change it.

amontanez24: Is this for downloading or uploading? I may have been confused. I just mean when downloading the metadata, the filename is `metadata_v1` and if you look for just `metadata` it won't exist

fealho: Further down the code I rename it:

```python3
os.rename(
            os.path.join(output_folder_name, 'metadata_v1.json'),
            os.path.join(output_folder_name, 'metadata.json')
        )
```

sdv/datasets/demo.py
```diff
@@ -0,0 +1,122 @@
...
+
+    LOGGER.info(f'Extracting dataset into {output_folder_name}')
+    bytes_io = io.BytesIO(response.read())
+    with ZipFile(bytes_io) as zf:
+        os.makedirs(output_folder_name, exist_ok=True)
+        zf.extractall(output_folder_name)
+        os.remove(os.path.join(output_folder_name, 'metadata_v0.json'))
+        os.rename(
+            os.path.join(output_folder_name, 'metadata_v1.json'),
+            os.path.join(output_folder_name, 'metadata.json')
```
amontanez24, line 56: minor: use the constant here

sdv/datasets/demo.py
```diff
@@ -0,0 +1,122 @@
...
+
+def _download(modality, dataset_name, output_folder_name):
+    dataset_url = _get_dataset_url(modality, dataset_name)
+    LOGGER.info(f'Downloading dataset {dataset_name} from {dataset_url}')
+    try:
+        response = urllib.request.urlopen(dataset_url)
+    except urllib.error.HTTPError:
+        raise ValueError(
+            f"Invalid dataset name '{dataset_name}'. "
+            "Use 'list_available_demos' to get a list of demo datasets."
```
amontanez24, line 45: minor: I think we could add a line here saying "make sure you have the correct modality for the dataset name" or something like that

sdv/datasets/demo.py
```diff
@@ -0,0 +1,123 @@
...
+
+    LOGGER.info(f'Extracting dataset into {output_folder_name}')
+    bytes_io = io.BytesIO(response.read())
+    with ZipFile(bytes_io) as zf:
+        os.makedirs(output_folder_name, exist_ok=True)
+        zf.extractall(output_folder_name)
+        os.remove(os.path.join(output_folder_name, 'metadata_v0.json'))
+        os.rename(
+            os.path.join(output_folder_name, 'metadata_v1.json'),
+            os.path.join(output_folder_name, METADATA_FILENAME)
```
amontanez24, line 57: would it be possible to extract the objects directly in memory instead of storing them to disk in a temp file? something like what's discussed here? https://stackoverflow.com/questions/5710867/downloading-and-unzipping-a-zip-file-without-writing-to-disk
And then we'd only save to disk if requested by the user

amontanez24 (commented):
I left a comment about streaming the files directly into objects in memory. Also seems like some tests are failing

sdv/datasets/demo.py
fealho: I thought about it, but the code would be a little more complicated (I need to pass a path to `load_from_json`) and a little less readable, while also potentially increasing memory usage (since, if you don't do something fancy, you have a copy of the zip file and the object in memory at the same time). It just doesn't seem worth it.

pvk-developer: As @amontanez24 pointed, it is possible to read the object in memory and once you exit the `with` indentation the garbage collector will close the `zip` file and unload it from memory, only the assigned variable in it will stay.

I'm not sure about the format that the `json` will come out, but if it's a `dict` just use the private method `_load_from_dict`:  https://github.com/sdv-dev/SDV/blob/V1.0.0.dev/sdv/metadata/single_table.py#L512

sdv/datasets/demo.py
```diff
@@ -0,0 +1,119 @@
...
+            * If the ``dataset_name`` doesn't exist in the bucket.
+            * If there is already a folder named ``output_folder_name``.
+            * If ``modality`` is not ``'single_table'``, ``'multi_table'`` or ``'sequential'``.
+    """
+    _validate_args_download_demo(modality, output_folder_name)
+
+    use_temp_dir = False
+    if output_folder_name is None:
+        use_temp_dir = True
+        output_folder_name = tempfile.mkdtemp()
```
pvk-developer, line 110: This can be removed if using the `memory` instead.

sdv/datasets/demo.py
```diff
@@ -0,0 +1,119 @@
...
+
+    _download(modality, dataset_name, output_folder_name)
+    data = _get_data(modality, output_folder_name)
+    metadata = _get_metadata(modality, output_folder_name)
+
+    if use_temp_dir:
+        shutil.rmtree(output_folder_name)
```
pvk-developer, line 117: Same here.

pvk-developer (changes requested):

sdv/metadata/multi_table.py
```diff
@@ -610,7 +610,7 @@ def load_from_json(cls, filepath):
             A ``MultiTableMetadata`` instance.
         """
         metadata = read_json(filepath)
-        return cls._load_from_dict(metadata)
+        return cls._load_from_dict(metadata)  # (1 <- check here)
```
pvk-developer, line 613: I think that the comment can be removed

pvk-developer (approved):
LGTM, just make sure to remove the comment
Good job!

amontanez24 (approved):
Thanks for addressing! LGTM :shipit:

PR #1151, opened 2022-12-20
Update email and author

pvk-developer:


conda/meta.yaml
```diff
@@ -45,9 +45,8 @@ requirements:
 
 about:
   home: "https://sdv.dev"
-  license: MIT
```
amontanez24, line 48: I think we can just delete this file. We don't do these conda releases anymore

npatki: +1 we could just delete the entire `conda/` folder while we're at it right? Seems we did this for [Copulas](https://github.com/sdv-dev/Copulas) already.

npatki (changes requested):
Let's delete the `conda/` folder altogether.

npatki (approved):
LGTM!

amontanez24 (approved):
:shipit:

PR #1157, opened 2022-12-27
Add ability to reset random sampling

### LINKED ISSUE #1130: Add ability to reset random sampling
### Problem Description

_As a user, it would be nice if I could get the same sampled results as a previous sample call. It would also be nice to control when since sometimes I want new sampled data. On top of this, the consistency should be for all columns._

### Acceptance criteria
- We want to change the way we handle randomness to make the above request possible. This means doing the following steps:
    - Remove the `randomize_samples` parameter from all sample calls
    - Set the seed for the underlying model on the initial call.
    - Add a method called `reset_sampling` that resets the seed for the model back to the original state, as well as resets that random state of the `HyperTransformer`.
    - We should add the same method to `MultiTableSynthesizers` as well. This would just loop through each `SingleTableSynthesizer` and call the method.

### Expected behavior

```python

synthesizer.fit(data)
synthetic_data_1 = synthesizer.sample(10)

synthesizer.reset_sampling()
synthetic_data_2 = synthesizer.sample(10)
```
`synthetic_data_2` should be the same as `synthetic_data_1`

### Additional context

- We may need to add some logic to track the state of randomization for the underlying model. Currently, we have a method to set the random state on them. We should set the random state in the beginning and then let the model continue to use that state until `reset_sampling` is called

pvk-developer:
Resolves #1130

### DIFF
--- a/sdv/data_processing/data_processor.py
+++ b/sdv/data_processing/data_processor.py
@@ -409,7 +409,11 @@ class DataProcessor:
 
     def reset_sampling(self):
         """Reset the sampling state for the anonymized columns and primary keys."""
-        self._hyper_transformer.reset_randomization()
+        # Resetting the transformers manually until fixed on RDT
+        for transformer in self._hyper_transformer.field_transformers.values():
+            if transformer is not None:
+                transformer.reset_randomization()
+
         self._keys_generators = {
             key: itertools.count()
             for key in self._keys_generators
@@ -494,7 +498,7 @@ class DataProcessor:
 
         try:
             transformed = self._hyper_transformer.transform_subset(data)
-        except (rdt.errors.NotFittedError, rdt.errors.TransformerProcessingError):
+        except (rdt.errors.NotFittedError, rdt.errors.Error):
             transformed = data
 
         return transformed
--- a/sdv/data_processing/data_processor.py
+++ b/sdv/data_processing/data_processor.py
@@ -494,7 +494,7 @@ class DataProcessor:
 
         try:
             transformed = self._hyper_transformer.transform_subset(data)
-        except (rdt.errors.NotFittedError, rdt.errors.Error):
+        except (rdt.errors.NotFittedError, rdt.errors.TransformerProcessingError):
             transformed = data
 
         return transformed
@@ -562,7 +562,7 @@ class DataProcessor:
                 column_data = reversed_data[column_name]
 
             dtype = self._dtypes[column_name]
-            if pd.api.types.is_integer_dtype(dtype) and column_data.dtype == dtype:
+            if pd.api.types.is_integer_dtype(dtype) and column_data.dtype != 'O':
                 column_data = column_data.round()
 
             reversed_data[column_name] = column_data[column_data.notna()].astype(dtype)
--- a/sdv/data_processing/data_processor.py
+++ b/sdv/data_processing/data_processor.py
@@ -409,11 +409,7 @@ class DataProcessor:
 
     def reset_sampling(self):
         """Reset the sampling state for the anonymized columns and primary keys."""
-        # Resetting the transformers manually until fixed on RDT
-        for transformer in self._hyper_transformer.field_transformers.values():
-            if transformer is not None:
-                transformer.reset_randomization()
-
+        self._hyper_transformer.reset_randomization()
         self._keys_generators = {
             key: itertools.count()
             for key in self._keys_generators
@@ -566,7 +562,7 @@ class DataProcessor:
                 column_data = reversed_data[column_name]
 
             dtype = self._dtypes[column_name]
-            if pd.api.types.is_integer_dtype(dtype):
+            if pd.api.types.is_integer_dtype(dtype) and column_data.dtype == dtype:
                 column_data = column_data.round()
 
             reversed_data[column_name] = column_data[column_data.notna()].astype(dtype)
--- a/sdv/data_processing/data_processor.py
+++ b/sdv/data_processing/data_processor.py
@@ -407,6 +407,18 @@ class DataProcessor:
         self._fit_hyper_transformer(constrained)
         self.fitted = True
 
+    def reset_sampling(self):
+        """Reset the sampling state for the anonymized columns and primary keys."""
+        # Resetting the transformers manually until fixed on RDT
+        for transformer in self._hyper_transformer.field_transformers.values():
+            if transformer is not None:
+                transformer.reset_randomization()
+
+        self._keys_generators = {
+            key: itertools.count()
+            for key in self._keys_generators
+        }
+
     def generate_keys(self, num_rows, reset_keys=False):
         """Generate the columns that are identified as ``keys``.
 
--- a/sdv/multi_table/base.py
+++ b/sdv/multi_table/base.py
@@ -104,16 +104,26 @@ class BaseMultiTableSynthesizer:
         """Return the ``MultiTableMetadata`` for this synthesizer."""
         return self.metadata
 
+    def _get_all_foreign_keys(self, table_name):
+        foreign_keys = []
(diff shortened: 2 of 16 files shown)


sdv/single_table/base.py
```diff
@@ -56,18 +56,22 @@ def _update_default_transformers(self):
...
+    def _initialize_synthesizer(self):
+        self._data_processor = DataProcessor(
+            metadata=self.metadata,
+            enforce_rounding=self.enforce_rounding,
+            enforce_min_max_values=self.enforce_min_max_values
+        )
+        self._update_default_transformers()
+        self._fitted = False
+        self._random_state_set = False
+
```
pvk-developer, line 68: When fitting was called over again it wasn't resetting properly the rest of the items. That is why I decided to create this function which should reset the `data_processor` and other states.

amontanez24: why does the `DataProcessor` need to be re-initialized? Can't it just refit?

sdv/multi_table/base.py
```diff
@@ -23,6 +24,21 @@ class BaseMultiTableSynthesizer:
     """
 
     _synthesizer = GaussianCopulaSynthesizer
+    _numpy_seed = 73251
+
+    @contextlib.contextmanager
+    def _temporary_numpy_seed(self):
```
amontanez24, line 30: can we rename to `_set_temp_numpy_seed`?

sdv/multi_table/base.py
```diff
@@ -292,13 +314,11 @@ def sample(self, scale=1.0, randomize_samples=False):
...
                 than the original tables. Defaults to ``1.0``.
-            randomize_samples (bool):
-                Whether or not the data should change on each sample call. If ``True``
-                every time ``sample`` is called, different data will be generated, if ``False``
-                every time ``sample`` is called, the same data will be generated. Defaults to
-                ``False``.
         """
-        return self._sample(scale=scale, randomize_samples=randomize_samples)
+        with self._temporary_numpy_seed():
+            sampled_data = self._sample(scale=scale)
```
amontanez24, line 319: any reason not to do this at the single table level?

sdv/data_processing/data_processor.py
```diff
@@ -407,6 +407,18 @@ def fit(self, data):
         self._fit_hyper_transformer(constrained)
         self.fitted = True
 
+    def reset_sampling(self):
+        """Reset the sampling state for the anonymized columns and primary keys."""
+        # Resetting the transformers manually until fixed on RDT
+        for transformer in self._hyper_transformer.field_transformers.values():
+            if transformer is not None:
+                transformer.reset_randomization()
```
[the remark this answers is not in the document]
pvk-developer: Yes, but we need a release candidate or point to the branch.

sdv/multi_table/base.py
pvk-developer: Yes, the `numpy.random.choice` is being used in the `hma.py` https://github.com/sdv-dev/SDV/blob/fdbafd4d8a9fffe00b5bfca169c44c07b48199c3/sdv/relational/hma.py#L470

With this we ensure that the primary keys and relations between tables is maintained the same.

sdv/single_table/base.py
pvk-developer: Yes I think we can actually re-fit, I can't find the test that I did that was causing this to fail. Maybe I used the wrong data at some point and thats why it didn't work at first glance. But so far it should be working by just refitting, applied the changes.

tests/unit/multi_table/test_base.py
```diff
@@ -217,6 +217,17 @@ def test_validate_missing_table(self):
         with pytest.raises(InvalidDataError, match=error_msg):
             instance.validate(data)
 
+    def test_validate_key_error(self):
```
amontanez24, line 220: minor: what will continue?

amontanez24 (commented):
This is looking good! For a sanity check, can we add integration tests that use all the random causing items (eg. data with anonymized columns, a model that uses randomness and both multi and single table cases)

tests/integration/multi_table/test_hma.py
```diff
@@ -31,3 +35,44 @@ def test_hma(tmpdir):
...
+    """End to end integration test that uses ``reset_sampling``.
+
+    This test uses ``reset_sampling`` to ensure that the model will generate the same data
+    as the first sample after this method has been called.
+    """
+    # Setup
+    faker = Faker()
+    metadata, data = load_demo(metadata=True)
+    metadata.to_json(tmpdir / 'old.json')
+    MultiTableMetadata.upgrade_metadata(tmpdir / 'old.json', tmpdir / 'new.json')
```
amontanez24, line 50: instead if doing this can we use the new [download_demo](https://github.com/sdv-dev/SDV/blob/V1.0.0.dev/sdv/datasets/demo.py#L98) method and avoid conversions?

amontanez24 (approved):
Awesome! One minor comment but otherwise looks good

tests/integration/multi_table/test_hma.py
pvk-developer: I think we will have to doublecheck the datasets before doing that.

amontanez24: The datasets should be good. The problem is we're going to remove the old code so we can't use the old `load_demo`

setup.py
```diff
@@ -22,7 +22,7 @@
     'copulas>=0.7.0,<0.8',
     'ctgan>=0.5.3.dev0,<0.6',
     'deepecho>=0.3.0.post1,<0.4',
-    'rdt>=1.3.0.dev0,<2.0',
```
amontanez24, line 25: we can just pin it, otherwise when we do the actual release it'll grab that

sdv/multi_table/hma.py
```diff
@@ -277,7 +270,10 @@ def _extract_parameters(self, parent_row, table_name, foreign_key):
         num_rows_key = f'{prefix}num_rows'
         if num_rows_key in flat_parameters:
             num_rows = flat_parameters[num_rows_key]
-            flat_parameters[num_rows_key] = min(self._max_child_rows[num_rows_key], num_rows)
+            flat_parameters[num_rows_key] = min(
+                self._max_child_rows[num_rows_key],
+                math.ceil(num_rows)
```
amontanez24, line 275: why do we need to do this? `num_rows` should be an int right?

sdv/multi_table/base.py
```diff
@@ -253,6 +280,13 @@ def preprocess(self, data):
         processed_data = {}
         for table_name, table_data in data.items():
             synthesizer = self._table_synthesizers[table_name]
+            synthesizer.auto_assign_transformers(table_data)
+            foreign_key_columns = self._get_all_foreign_keys(table_name)
+            column_name_to_transformers = {
+                column_name: None
+                for column_name in foreign_key_columns
+            }
+            synthesizer.update_transformers(column_name_to_transformers)
```
amontanez24, line 289: I think this is a fine solution but can we call attention to what we're doing here by pulling it out into a separate method called `_skip_foreign_key_transformations` or something like that?

sdv/multi_table/hma.py
pvk-developer: Yes, but in some scenarios which has to generate the primary key it goes `0.3` for example, and that rounded goes to `0` which then ends up failing from `rdt` -> https://github.com/sdv-dev/RDT/blob/e4fb0e5960feb8e829c8d8f5f64abdade8f8eadf/rdt/hyper_transformer.py#L712

setup.py
pvk-developer: Sure!

amontanez24 (approved):
:shipit: 🥳

PR #1162, opened 2023-01-06
Fix conditioning on negative float values.

### LINKED ISSUE #1161: Conditional sampling with negative float values doesn't work
### Environment Details

* SDV version: 0.17.2
* Python version: Python 3.9.10
* Operating System: Mac OS 12.6.1

### Error Description

When sampling records from the model with negative float as a condition, no records are being returned, as the [`_filter_conditions`](https://github.com/sdv-dev/SDV/blob/783db48a68b824f3755d105748ef2ec20b60c12e/sdv/tabular/base.py#L168-L196) method filters them all out.

That's because it compares negative `distance` value, with an absolute value (which will always be >= 0).

### Steps to reproduce

Sample code that triggers this issue:
```python
data = pd.DataFrame({
    'column1': [-float(x) for x in list(range(100))],
    'column2': list(range(100))
})

model = GaussianCopula()
model.fit(data)

sampled = model.sample_conditions([Condition({'column1': -50.0})])

# this assertion fails, as the dataframe is empty
assert len(sampled) == 1
```

pimlock:
Fixes #1161

The problem was that for negative condition values, the `distance` was negative, but it was compared against the result of `np.abs(a - b)`, which will always be >= 0.

The fix is to use `abs` on the distance as well.

### DIFF
--- a/sdv/tabular/base.py
+++ b/sdv/tabular/base.py
@@ -187,7 +187,7 @@ class BaseTabularModel:
         for column, value in conditions.items():
             column_values = sampled[column]
             if column_values.dtype.kind == 'f':
-                distance = value * float_rtol
+                distance = abs(value * float_rtol)
                 sampled = sampled[np.abs(column_values - value) <= distance]
                 sampled[column] = value
             else:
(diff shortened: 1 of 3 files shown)


katxiao (approved):

npatki:
Hi @pimlock thanks for contributing to the SDV codebase. We are in the process of deprecating our old, CLA assistant bot (seen up above) and moving over to a new Contributor Licensing Agreement. 

To request our new Contributor Licensing Agreement, we ask that you please fill out the information in the form: [https://bit.ly/sdv-cla-form](https://bit.ly/sdv-cla-form). Once we receive your submission, we'll send you a PDF to the email that you give us.

You will only have to do this once for contributing to any of the libraries in the SDV ecosystem. Thanks for bearing with us!

PR #1174, opened 2023-01-18
Migrate adding and validating constraints

### LINKED ISSUE #1163: Migrate adding and validating constraints to BaseSynthesizer
### Problem Description

_As a user, it would be nice if the metadata used through the `SDV` ecosystem was the same. For some tools, constraints are irrelevant and so having them in the metadata is confusing._

In order to make metadata consistent throughout our ecosystem, we want to separate constraints from it.

### Acceptance criteria

- Move all functionality involving constraints in the metadata into the `BaseSynthesizer` class. This includes
    - [Validating constraints](https://github.com/sdv-dev/SDV/blob/V1.0.0.dev/sdv/metadata/single_table.py#L217)
    - [Adding constraints](https://github.com/sdv-dev/SDV/blob/V1.0.0.dev/sdv/metadata/single_table.py#L233)
- Instead of `add_constraint`, create a new `add_constraints` method
    - Parameters
        - `constraints` - List of constraints described as dicts in the following format: 
        `{'constraint_class': NAME, 'constraint_parameters': {'param_1': ..., 'param_2': ...}}`
    - Error Handling: Perform the same validation & error handling that we are already doing. Note that we always have access to the metadata because it is required when instantiating the synthesizer.
        - If the user has already fit the synthesizer, warn them that they need to refit it.
`Warning: For these constraints to take effect, please refit the synthesizer using 'fit'`
    - If called multiple times, continue to add to the existing list of constraints.

### Expected behavior

```python
age_is_positive = {
  'constraint_class': 'Positive',
  'constraint_parameters': {
    'column': 'age',
    'strict_boundaries': True
  }
}

synthesizer.add_constraints(constraints=[age_is_positive])
```

### Additional context

- On top of moving the constraint logic to the `BaseSynthesizer` there are some other key changes. The new method for adding constraints take in a list instead of an individual constraint dictionary. Additionally, the format of the constraint dictionary is different as the parameters are now in a nested dict. This means that some logic around the validation code will likely need to change.
- The `metadata_upgrader` will no longer need to [update constraints](https://github.com/sdv-dev/SDV/blob/V1.0.0.dev/sdv/metadata/metadata_upgrader.py#L296). The code for converting constraints can stay but remove the conversion from `convert_metadata`

### LINKED ISSUE #1168: Add get_constraints method to synthesizers
### Problem Description

_As a user, it would be helpful to see the constraints already added to my synthesizer_

### Acceptance Criteria
- Add a method called `get_constraints` to the `BaseSingleTableSynthesizer`
    - It should return a list of the constraints in dictionary format

### Expected behavior

```python
# synthesizer with no constraints
synthesizer.get_constraints()
[]

# synthesizer with constraints
synthesizer.get_constraints()
[{
  'constraint_name': 'Positive',
  'constraint_parameters': {
    'column': 'age',
    'strict_boundaries': True
  }
},
{
  'constraint_name': 'MyConstraintClass',
  'constraint_parameters': {
    'custom_param_1': 34,
    'custom_param_2': True
  }
}]
```

### Additional Notes

- Return copy

pvk-developer:
Resolves #1163 
Resolves #1168


sdv/constraints/base.py
```diff
@@ -354,10 +354,10 @@ def from_dict(cls, constraint_dict):
                 New constraint instance.
         """
         constraint_dict = constraint_dict.copy()
```
amontanez24, line 356: do we still need to copy it now that we aren't popping off the item anymore?

sdv/single_table/base.py
```diff
@@ -69,6 +69,7 @@ def __init__(self, metadata, enforce_min_max_values=True, enforce_rounding=True)
         self._fitted = False
         self._random_state_set = False
         self._update_default_transformers()
+        self._constraints = []
```
amontanez24, line 72: Is this used anywhere?

amontanez24 (approved):
This looks really good! I just left two minor comments, but after addressing :shipit:

sdv/single_table/base.py
```diff
@@ -269,6 +270,26 @@ def get_metadata(self):
...
+        """
+        if self._fitted:
+            warnings.warn(
+                "For these constraints to take effect, please refit the synthesizer using 'fit'."
+            )
+
+        self._data_processor.add_constraints(constraints)
+
+    def get_constraints(self):
+        """Return a list of the current constraints that will be used."""
```
fealho, line 290: Add `Returns:` section, since it's a public method.

fealho (commented):
👍

sdv/single_table/base.py
pvk-developer: 👍🏻

sdv/single_table/base.py
pvk-developer: 👍🏻

fealho (approved):

PR #1196, opened 2023-01-27
Misc bug fixes for SDV `1.0.0`

### LINKED ISSUE #1193: Misc. bug fixes for SDV 1.0.0
### Environment Details

* SDV version: 1.0.0.dev
* Python version: Any
* Operating System: Any

### Error Description

The following changes should be made:
1. If a float column has more than 14 decimals places, sometimes we are unable to round. This raises the following warning:
` usr/local/lib/python3.8/dist-packages/sdv/data_processing/numerical_formatter.py:72: UserWarning: No rounding scheme detected for column 'max(y_wgs84)'. Synthetic data will not be rounded.`
The warning can be overwhelming and there isn't anything the user can really do about it, so we should make it logger info statement instead.
2. The default for the `model_missing_values` parameter is currently set to True when calling `auto_assign_transformers`. It should be False instead
3. The `synthesizer.load` method should change the name of the `path` paramter to `filepath`
4. The `sample_conditions` method should be renamed to `sample_from_conditions`
5. The following methods should be added to the `SingleTablePreset`
    * `get_metadata` - Wrapper around the `SingleTableSynthesizer.get_metadata` method
    * `get_parameters` - Wrapper around the `SingleTableSynthesizer.get_parameters` method

pvk-developer:
Resolves #1193

### DIFF
--- a/sdv/lite/single_table.py
+++ b/sdv/lite/single_table.py
@@ -77,8 +77,7 @@ class SingleTablePreset:
         """
         self._synthesizer.fit(data)
 
-    def sample(self, num_rows, max_tries_per_batch=100, batch_size=None,
-               output_file_path=None, conditions=None):
+    def sample(self, num_rows, max_tries_per_batch=100, batch_size=None, output_file_path=None):
         """Sample rows from this table.
 
         Args:
@@ -91,9 +90,6 @@ class SingleTablePreset:
             output_file_path (str or None):
                 The file to periodically write sampled rows to. If None, does not
                 write rows anywhere.
-            conditions:
-                Deprecated argument. Use the ``sample_from_conditions`` method with
-                ``sdv.sampling.Condition`` objects instead.
 
         Returns:
             pandas.DataFrame:
@@ -104,7 +100,6 @@ class SingleTablePreset:
             max_tries_per_batch,
             batch_size,
             output_file_path,
-            conditions
         )
 
         return sampled
--- a/sdv/lite/single_table.py
+++ b/sdv/lite/single_table.py
@@ -84,7 +84,7 @@ class SingleTablePreset:
                 The file to periodically write sampled rows to. If None, does not
                 write rows anywhere.
             conditions:
-                Deprecated argument. Use the ``sample_conditions`` method with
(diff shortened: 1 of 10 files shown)


sdv/lite/single_table.py
```diff
@@ -84,7 +92,7 @@ def sample(self, num_rows, max_tries_per_batch=100, batch_size=None,
                 The file to periodically write sampled rows to. If None, does not
                 write rows anywhere.
             conditions:
```
amontanez24, line 94: Since we will be releasing a new major version I don't think we need to support this deprecated argument anymore

amontanez24 (approved):
Just one comment about removing an unnecessary parameter. After that, :shipit:

tests/unit/data_processing/test_numerical_formatter.py
```diff
@@ -23,21 +22,23 @@ def test___init__(self):
         assert formatter.enforce_min_max_values is True
         assert formatter.computer_representation == 'Int8'
 
-    def test__learn_rounding_digits_more_than_15_decimals(self):
+    @patch('sdv.data_processing.numerical_formatter.LOGGER')
+    def test__learn_rounding_digits_more_than_15_decimals(self, log_mock):
         """Test the ``_learn_rounding_digits`` method with more than 15 decimals.
 
         If the data has more than 15 decimals, return None and raise warning.
```
fealho, line 29: doesnt raise warning anymore

tests/unit/multi_table/test_hma.py
```diff
@@ -645,10 +645,7 @@ def test_get_learned_distributions(self):
         """Test that ``get_learned_distributions`` returns a dict.
 
         Test that it returns a dictionary with the name of the columns and the learned
-        distribution and it's parameters. In this case we ensure that the columns generated
-        by the ``rdt.transformers`` are returned as well (``a_value.is_null``) but the extended
-        columns are not being returned. Can't ensure the ``a_value.is_null`` distribution thats
-        why no validation is being done to check the values of it.
+        distribution and it's parameters.
```
fealho, line 648: its

fealho (approved):
👍

PR #1217, opened 2023-01-30
Sampling should reset after retraining the model

### LINKED ISSUE #1201: Sampling should reset after retraining the model
### Problem Description
Every time I call sample, the indices and other random generators progress.

Eg. 
```python
sample1 = synthesizer.sample(5) # has primary key IDs 0, 1, 2, 3, 4
sample2 = synthesizer.sample(2) # has primary key IDs 5, 6
```

This is working as expected. However, if I ever refit the synthesizer, I am seeing that the values are also progressing.

```python
synthesizer.fit(data)

# alternatively
processed = synthesizer.preprocess(data)
synthesizer.fit_processed_data(processed)

sample3 = synthesizer.sample(2) # primary keys are: 7, 8
```

### Expected behavior

I expect that if I ever call `fit` or `fit_processed_data` all the generators should be reset.  So in the above example, `sample3` should be reset so that the primary keys start back at `0, 1`

### Additional context
A simple workaround for now is just to reset the keys myself. But it would be nice if the `fit` call already did this.

```python
synthesizer.fit(data)
synthesizer.reset_sampling()
synthesizer.sample(2)
```

pvk-developer:
Resolves #1201


katxiao:
Just wanted to double check - was it intentional to remove the single table python notebooks?

pvk-developer:
> Just wanted to double check - was it intentional to remove the single table python notebooks?

All the `tutorial notebooks` have been removed. The reason why is that we will move them into integration tests and guide the users towards our documentation.

amontanez24 (approved):
:shipit:

katxiao (approved):
Thanks! lgtm

PR #1218, opened 2023-01-31
Rename `update_table_parameters` to `set_table_parameters`

### LINKED ISSUE #1200: Change function name `HMASynthesizer.update_table_parameters` --> `set_table_parameters`
### Problem Description
Currently, the method to reset the table parameters is named `HMASynthesizer.update_table_parameters`.

This does not accurately reflect the functionality:
1. If I am _updating_ parameters, I would expect that any partial changes override the original values but the rest remain
2. If I am _setting_ parameters, I would expect that all the parameters are getting reset

From looking at the functionality, it appears that we are in the 2nd scenario. So the method is not actually updating, but setting.

### Expected behavior

Change the method name to: `HMASynthesizer.set_table_parameters`

### Additional context

Try running the following code:
```python
s = HMASynthesizer(metadata)
s.update_table_parameters(
    table_name='users',
    table_parameters={
        'enforce_min_max_values': False,
        'numerical_distributions': {
            'A': 'gamma',
            'B': 'beta' }})

s.update_table_parameters(
    table_name='users',
    table_parameters={
        'numerical_distributions': {
            'A': 'uniform'}})
```

Observe that the end state is:
- `enforce_min_max_values=True` and
- `A='uniform'`

This means that the second call for `update_table_parameters` completely overrides the first -- i.e. this is **(re)setting** parameters not **updating** them.

pvk-developer:
Resolves #1200


amontanez24 (approved):
Hopefully these files stop sneaking back in 😅

pvk-developer:
> Hopefully these files stop sneaking back in 😅

🙏🏻 I think those are all

fealho (approved):

PR #1225, opened 2023-02-02
Add  `DatetimeFormatter` and fix Datetime in constraints

### LINKED ISSUE #1203: Cannot apply `Inequality` constraint on demo dataset's datetime columns
### Environment Details
* SDV version: 1.0.0 (in progress) + existing versions (0.18.0)
* Python version: 3.8

### Error Description
The `student_placements_pii` demo dataset contains `start_date` and `end_date` columns that represent dates. They include some missing values.

If I apply an `Inequality` constraint to designate that `start_date < end_date`, then the synthesizer crashes during `fit`.

### Steps to reproduce
```python

    modality='single_table',
    dataset_name='student_placements_pii'
)

start_lessthan_end = {
    'constraint_parameters': {
      'low_column_name': 'start_date',
      'high_column_name': 'end_date'  
    }   
}

synthesizer.add_constraints(constraints=[start_lessthan_end])
synthesizer.fit(data)
```

```
AggregateConstraintsError: 
ufunc 'isnan' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

### Notes
Both `start_date` and `end_date` columns are being represented as `object` dtypes in pandas. **This should be OK.**

I also tried casting them myself to see if this helps:
```python
data['start_date'] = pd.to_datetime(data['start_date'])
data['end_date'] = pd.to_datetime(data['end_date'])
```

If I do this, then the `fit` calls works, but there is still an error during `sample`:

```
IntCastingNaNError: Error: Sampling terminated. Partial results are stored in a temporary file: .sample.csv.temp. This file will be overridden the next time you sample. Please rename the file if you wish to save these results.
Cannot convert non-finite values (NA or inf) to integer
```

pvk-developer:
Resolves #1203

### DIFF
--- a/sdv/data_processing/data_processor.py
+++ b/sdv/data_processing/data_processor.py
@@ -452,11 +452,12 @@ class DataProcessor:
             if not data.empty:
                 self._hyper_transformer.fit(data)
 
-    def _fit_numerical_formatters(self, data):
-        """Fit a ``NumericalFormatter`` for each column in the data."""
+    def _fit_formatters(self, data):
+        """Fit ``NumericalFormatter`` and ``DatetimeFormatter`` for each column in the data."""
         for column_name in data:
             column_metadata = self.metadata._columns.get(column_name)
-            if column_metadata.get('sdtype') == 'numerical' and column_name != self._primary_key:
+            sdtype = column_metadata.get('sdtype')
+            if sdtype == 'numerical' and column_name != self._primary_key:
                 representation = column_metadata.get('computer_representation', 'Float')
                 self.formatters[column_name] = NumericalFormatter(
                     enforce_rounding=self._enforce_rounding,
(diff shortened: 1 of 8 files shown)


sdv/data_processing/data_processor.py
```diff
@@ -465,6 +465,15 @@ def _fit_numerical_formatters(self, data):
                 )
                 self.formatters[column_name].learn_format(data[column_name])
 
+    def _fit_datetime_formatters(self, data):
```
amontanez24, line 468: This method is so similar to `_fit_numerical_formatters` that we could probably combine them into a `_fit_formatters` and just have different if statements and one loop

amontanez24 (approved):
One minor comment but otherwise I think this is good to go!

sdv/data_processing/datetime_formatter.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class DatetimeFormatter:
+    """Formatter for datetime data.
+
+    Args:
+        datetime_format (str):
+            The strftime to use for parsing time. For more information, see
+            https://docs.python.org/3/library/datetime.html#strftime-and-strptime-behavior.
+            If ``None`` it will attempt to learn it by itself.
```
fealho, line 14: Add that it defaults to None.

sdv/data_processing/datetime_formatter.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def format_data(self, column):
+        """Format a column according to the learned format.
+
+        Args:
+            column (pd.Series):
+                Data to format.
+
+        Returns:
+            numpy.ndarray containing the formatted data.
```
fealho, line 39: I think we usually format it like this:
```
numpy.ndarray:
   Contains the formatted data.
```

fealho (commented):
Looks good! I'd only double check which lines you are not testing, because the coverage is going down right now.

sdv/data_processing/data_processor.py
pvk-developer: I was doubting between that option too. Will address it asap!

pvk-developer:
> Looks good! I'd only double check which lines you are not testing, because the coverage is going down right now.

Thanks for noticing, actually important lines to test since those were the major changes to the `constraints`.

PR #1233, opened 2023-02-05
Validate constraint parameters in `add_constraints` 

### LINKED ISSUE #1210: Improve error handling when validating `constraints`
### Problem Description
In the new version of SDV 1.0, I will be able to provide constraints using a dictionary format such as:

```python
{
  'table_name': 'hotel_logs', # required for multi table
  'constraint_parameters': {
    'low_column_name': 'start_date'
    'high_column_name': 'end_date'
  }
}
```

Currently, there is a confusing errors if I don't provide the `constraint_parameters` key:
```
InvalidConstraintsError: The provided constraint is invalid:
Missing required values {'column_names'} in a CustomConstraint constraint.
'NoneType' object is not iterable
```

### Expected behavior
We should explicitly check for the following:

- In a multi table case, the top-level dictionary keys should be `'constraint_class'`, `'constraint_parameters'` and `'table_name'`
- In single table, the top-level dictionary keys should be `'constraint_class'` and `'constraint_parameters'`

If the required top-level keys are not found, throw the following error:
```
SynthesizerInputError: A constraint is missing required parameters (<missing values>). Please add these parameters to your constraint definition.

Example:
SynthesizerInputError: A constraint is missing required parameters ('constraint_class'). Please add these parameters to your constraint definition.
SynthesizerInputError: A constraint is missing required parameters ('constraint_parameters', 'table_name'). Please add these parameters to your constraint definition.
```

If there are extra keys that are found, throw the following error:
```
SynthesizerInputError: Unrecognized constraint parameter (<example>). Please remove this parameter from your constraint definition.

Example
SynthesizerInputError: Unrecognized constraint parameter ('column_name'). Please remove this parameter from your constraint definition.
```

fealho:
Resolve #1210.

### DIFF
--- a/sdv/multi_table/base.py
+++ b/sdv/multi_table/base.py
@@ -368,6 +368,35 @@ class BaseMultiTableSynthesizer:
         synthesizer = self._table_synthesizers[table_name]
         return synthesizer.get_learned_distributions()
 
+    def _validate_constraints(self, constraints):
+        for constraint_dict in constraints:
+            params = {'constraint_class', 'constraint_parameters', 'table_name'}
+            keys = constraint_dict.keys()
+            missing_params = params - keys
+            if missing_params:
+                raise SynthesizerInputError(
+                    f'A constraint is missing required parameters {missing_params}. '
+                    'Please add these parameters to your constraint definition.'
+                )
+
+            extra_params = keys - params
+            if extra_params:
+                raise SynthesizerInputError(
+                    f'Unrecognized constraint parameter {extra_params}. '
+                    'Please remove these parameters from your constraint definition.'
+                )
+
+            if constraint_dict['constraint_class'] == 'Unique':
+                raise SynthesizerInputError(
+                    "The constraint class 'Unique' is not currently supported for multi-table"
+                    ' synthesizers. Please remove the constraint for this synthesizer.'
+                )
+
+        if self._fitted:
+            warnings.warn(
+                "For these constraints to take effect, please refit the synthesizer using 'fit'."
+            )
+
     def add_constraints(self, constraints):
         """Add constraints to the synthesizer.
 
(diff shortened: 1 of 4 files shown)


sdv/multi_table/base.py
```diff
@@ -368,6 +368,35 @@ def get_learned_distributions(self, table_name):
...
+                    f'A constraint is missing required parameters {missing_params}. '
+                    'Please add these parameters to your constraint definition.'
+                )
+
+            extra_params = keys - params
+            if extra_params:
+                raise SynthesizerInputError(
+                    f'Unrecognized constraint parameter {extra_params}. '
+                    'Please remove these parameters from your constraint definition.'
+                )
```
pvk-developer, line 387: It feels like we are doing this twice, I think we can just check if `table_name` is not in `constraint_dict` raise the error stating that you are missing `table_name`, if not then the data processor will raise the error.

pvk-developer (changes requested):

sdv/multi_table/base.py
fealho: Yeah, I did it like that so the errors would show up together (if you don't pass any parameters, it will crash saying only `table_name` is missing, and only after you add `table_name` will it tell you the other errors).

But I think your way is better, it's too much duplicated code otherwise.

fealho:
@amontanez24 @npatki Note that these errors are not being aggregated with the other constraint errors (`AggregateConstraintsError, InvalidConstraintsError`). So these will show up first, and only after it is fixed the other aggregate errors will show up.

This seems to be what the issue was describing, but I wanted to double check this is indeed the intended approach.

npatki:
@fealho 

> Note that these errors are not being aggregated with the other constraint errors

The aggregated constraints errors are thrown when are doing _fit_ right? Ie they happen in the middle of execution when trying to model the constraint.

This issue is just when specifying the constraint before any sort of processing or fitting happens.

fealho:
@npatki They are thrown in the `add_constraints` as an `InvalidConstraintsError`: https://github.com/sdv-dev/SDV/blob/5f586ec4fcc5e13f64d770f301f4d1ac7170b39d/sdv/data_processing/data_processor.py#L233-L254

npatki:
@fealho oh got it!

In any case, I think the answer remains the same: We want to throw the errors described in _this_ issue first. And then if that all passes, we can then do additional validation and throw an _AggregateConstraintsError` as needed.

pvk-developer (approved):

amontanez24 (approved):
:shipit:

PR #1234, opened 2023-02-06
Validate `numerical_distrubutions` parameter

fealho:
Resolve #1211.

PR #1235, opened 2023-02-06
Log columns in `numerical_distributions` which have been renamed

### LINKED ISSUE #1212: Log if any `numerical_distributions` will not be applied
_Nice to have. We can remove this from the milestone if needed._

### Problem Description
In SDV 1.0, I will be able to specify numerical distributions for any original column name:

```python
synthesizer = GaussianCopulaSynthesizer(
  metadata,
  numerical_distributions={
    'age': 'uniform',
    'start_date': 'beta',
    'end_date': 'beta',
  }
)
```

However (due to constraints and transformations), the column may no longer exist once we hit the fitting stage -- meaning that my specified numerical distribution won't be applied. For example, if I apply an Inequality constraint between `start_date` and `end_date`, then `start_date` may be dropped.

### Expected behavior
It would be nice to log that this is happening (`logger.info`):

After `preprocess`, we should have a new list of column names. If any user-provided name in `numerical_distributions` is no longer available, then log it during the `fit_processed_data` method:

```
INFO: Requested distribution ('beta') cannot be applied to column ('start_date')
because it no longer exists after preprocessing.
```

fealho:
Resolve #1212.

### DIFF
--- a/sdv/single_table/copulagan.py
+++ b/sdv/single_table/copulagan.py
@@ -3,6 +3,7 @@ from copy import deepcopy
 
 import rdt
 
+from sdv.errors import SynthesizerInputError
 from sdv.single_table.copulas import GaussianCopulaSynthesizer
 from sdv.single_table.ctgan import CTGANSynthesizer
 from sdv.single_table.utils import validate_numerical_distributions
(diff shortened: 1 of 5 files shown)


amontanez24 (approved):
This looks good! One thing I noticed for both this PR and #1234 is that there are very similar validations being done in both `CopulaGANSynthesizer` and `CopulasSynthesizer`. In the future maybe we should consider places where validation is repetitive and think about pulling that out into it's own module or something like that

pvk-developer (approved):

sdv/single_table/copulagan.py
```diff
@@ -179,6 +182,14 @@ def _fit(self, processed_data):
             processed_data (pandas.DataFrame):
                 Data to be learned.
         """
+        unseen_columns = self._numerical_distributions.keys() - set(processed_data.columns)
+        for column in unseen_columns:
+            LOGGER.info(
+                f"Requested distribution '{self.numerical_distributions[column]}' "
+                f"cannot be applied to column '{column}' because it no longer "
+                'exists after preprocessing.'
+            )
```
amontanez24, line 191: I think we can move this validation out to utils and pass the logger

pvk-developer (approved):
LGTM :shipit:

PR #1239, opened 2023-02-07
Fix order in which transformers are returned

### LINKED ISSUE #1222: Order the output of `get_transformers()` based on the metadata
### Problem Description
In the new SDV version 1.0, I am able to inspect the transformers that will be used if I run:

```python
synthesizer.auto_assign_transformers(real_data)
synthesizer.get_transformers()
```

This returns a dictionary mapping each column name to the transformer object.

```python
{
  'age': FloatFormatter(),
  'weight': FloatFormatter(),
  ...
}
```

### Expected behavior
In the returned dictionary from `get_transformers()`, I observed that the keys were out of order.

It would be nice if the keys were **in the same order as the metadata**, as the metadata is considered the ground truth. This will make it easier for me to inspect the transformers.

pvk-developer:
Resolves #1222

### DIFF
--- a/sdv/single_table/base.py
+++ b/sdv/single_table/base.py
@@ -352,6 +352,11 @@ class BaseSynthesizer:
                 "Use 'auto_assign_transformers' or 'fit' to create them."
             )
 
+        field_transformers = {
+            column_name: field_transformers.get(column_name)
+            for column_name in self.metadata._columns
+        }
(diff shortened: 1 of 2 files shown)


fealho (approved):

amontanez24 (approved):
:shipit:

PR #1244, opened 2023-02-08
Improve PII handling and add tests

### LINKED ISSUE #1226: Improve `update_column` validation for `pii` attribute
_For SDV 1.0 (in progress)_

### Problem Description
The `update_column` method is used to update parts of the metadata that may be incorrect. This method should also do some validation to make sure the user does not provide invalid data (that will cause issues later). 

When passing in `pii`, the method currently accepts any input:

```python
metadata.update_column(
    'column_name': 'user_address',
    'sdtype': 'address',
    'pii': 'a_string_input_type'
```

Or 

```python
metadata.update_column(
    'column_name': 'user_address',
    'sdtype': 'address',
    'pii': 40.0
```

**

### Expected behavior
Only `True` and `False` values are allowed for the `pii` attribute. If the user passes in an incorrect value, we should throw an error.

```
InvalidMetadataError: Parameter 'pii' is set to an invalid attribute (40.0). Expected a value of True or False.
```

pvk-developer:
Resolves #1226

### DIFF
--- a/sdv/metadata/single_table.py
+++ b/sdv/metadata/single_table.py
@@ -170,7 +170,8 @@ class SingleTableMetadata:
             - ``ValueError`` if the ``kwargs`` do not contain ``sdtype``.
             - ``ValueError`` if the column has unexpected values or ``kwargs`` for the given
               ``sdtype``.
-            - ``ValueError`` if the ``pii`` value is not ``True`` or ``False`` when present.
+            - ``InvalidMetadataError`` if the ``pii`` value is not ``True`` or ``False`` when
+               present.
         """
         if column_name in self._columns:
             raise ValueError(
@@ -210,7 +211,8 @@ class SingleTableMetadata:
             - ``ValueError`` if the column doesn't already exist in the ``SingleTableMetadata``.
             - ``ValueError`` if the column has unexpected values or ``kwargs`` for the current
               ``sdtype``.
-            - ``ValueError`` if the ``pii`` value is not ``True`` or ``False`` when present.
+            - ``InvalidMetadataError`` if the ``pii`` value is not ``True`` or ``False`` when
+               present.
         """
         self._validate_column_exists(column_name)
         _kwargs = deepcopy(kwargs)


tests/unit/metadata/test_single_table.py
```diff
@@ -481,6 +481,58 @@ def test__validate_column_text(self, mock__validate_text, mock__validate_kwargs)
...
+    @patch('sdv.metadata.single_table.SingleTableMetadata._validate_unexpected_kwargs')
+    def test__validate_column_pii(self, mock__validate_kwargs):
+        """Test ``_validate_column`` method when ``pii`` is ``True``or ``False``."""
+        # Setup
+        instance = SingleTableMetadata()
+
+        # Run and Assert
+        instance._validate_column('address', 'address', pii=True)
+
+    def test_update_column_sdtype(self):
```
fealho, line 507: Just curious, why did you add this and the test below?

fealho (commented):
Update the `Raises` sections of `update_column` and `add_column` and I think this is good.

tests/unit/metadata/test_single_table.py
pvk-developer: because they were missing and I just realized that while looking into the unit tests (I was going to use `update_column` to test the implementation but then I saw that those are missing and that I have to unit test it on the validate).

sdv/metadata/single_table.py
```diff
@@ -144,6 +144,13 @@ def _validate_column(self, column_name, sdtype, **kwargs):
             self._validate_datetime(column_name, **kwargs)
         elif sdtype == 'text':
             self._validate_text(column_name, **kwargs)
+        if 'pii' in kwargs:
+            pii_value = kwargs['pii']
+            if not isinstance(pii_value, bool):
+                raise InvalidMetadataError(
+                    f"Parameter 'pii' is set to an invalid attribute ({pii_value}). "
+                    'Expected a value of True or False.'
+                )
```
amontanez24, line 153: extremely minor:
I think for consistency we could put this part in a `_validate_pii` method. You wouldn't have to change tests or anything for this.

amontanez24 (approved):
After addressing Felipe's comments :shipit:

fealho (approved):

sdv/metadata/single_table.py
```diff
@@ -166,6 +170,7 @@ def add_column(self, column_name, **kwargs):
             - ``ValueError`` if the ``kwargs`` do not contain ``sdtype``.
             - ``ValueError`` if the column has unexpected values or ``kwargs`` for the given
               ``sdtype``.
+            - ``ValueError`` if the ``pii`` value is not ``True`` or ``False`` when present.
```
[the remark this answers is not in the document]
pvk-developer: I was actually about to rename all value errors in this file to `InvalidMetadataError` however I think its a good first task for  newcomers.

PR #1252, opened 2023-02-09
Issue 1241 data val HMA sample

### LINKED ISSUE #1241: Better input validation on `HMASynthesizer.sample`
Version: SDV 1.0.0 (in progress)

### Problem Description
In this version, the `HMASynthesizer` will have a parameter, `scale` that controls how large to make the synthetic data. It should only be possible to input floats or integers that are >0.0.

However, we are currently accepting other inputs such as `True, False, '1', etc.` -- we are actually producing synthetic data for this, even though it's unclear what the parameter value really means.

### Expected behavior
Throw an error if the `scale` parameter is not an integer or float that is >0.0.

```python
>>> synthesizer.sample(scale=True)

SamplingError: Invalid parameter for 'scale' (True). Please provide a number that is >0.0.
```

Same error should be thrown if the user provides a negative number.

R-Palazzo:
Resolves #1241


sdv/multi_table/base.py
```diff
@@ -345,6 +345,10 @@ def sample(self, scale=1.0):
                 If ``scale`` is lower than ``1.0`` create fewer rows by the factor of ``scale``
                 than the original tables. Defaults to ``1.0``.
         """
+        if not isinstance(scale, (float, int)) or type(scale) is bool:
+            raise SynthesizerInputError('scale parameter can only be float or a int instance.')
+        elif scale < 0 or np.isnan(scale):
+            raise ValueError('scale parameter must be >0.0.')
```
frances-h, line 351: Should this also be a `SynthesizerInputError` instead of a `ValueError`?

R-Palazzo: Made the change in the last commit

tests/unit/multi_table/test_base.py
```diff
@@ -638,6 +638,32 @@ def test__sample(self):
...
+            SynthesizerInputError, SynthesizerInputError, SynthesizerInputError,
+            SynthesizerInputError
+        ]
+        err_msg_1 = re.escape(
+            'scale parameter can only be float or a int instance.'
+        )
+        err_msg_2 = re.escape('scale parameter must be >0.0.')
+        err_msg = ['', '', err_msg_1, err_msg_1, err_msg_2, err_msg_2]
+
+        for (scale, err, msg) in zip(scales, errors, err_msg):
```
pvk-developer, line 663: Just make sure that a `SynthesizerInputError` occurs here and matches the message that we are expecting. For instance you can have multiple tests of this but with different inputs.

The one that passes you can just use `instance.sample(scale=scale)` without the need of `pytest.Raises`

pvk-developer (changes requested):

sdv/multi_table/base.py
```diff
@@ -345,6 +345,10 @@ def sample(self, scale=1.0):
                 If ``scale`` is lower than ``1.0`` create fewer rows by the factor of ``scale``
                 than the original tables. Defaults to ``1.0``.
         """
+        if not isinstance(scale, (float, int)) or type(scale) is bool:
```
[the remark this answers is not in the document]
R-Palazzo: Good comments, I just modified it.

pvk-developer (approved):

frances-h (approved):

PR #1255, opened 2023-02-10
Set multi-table foreign key transformers as None

### LINKED ISSUE #1249: Multi-table should show foreign key transformers as None
### Environment Details

* SDV version: v1.0.0
* Python version: Any
* Operating System: Any

### Error Description

In the multi-table case, all foreign keys are assigned `None` as a transformer. However this change is done in [preprocess](https://github.com/sdv-dev/SDV/blob/9324fcdcaba99452a81227e2397898ef582ed2b3/sdv/multi_table/base.py#L299), and not in `auto_assign_transformers`. For this reason, if a user calls `get_transformers` after auto-assigning them, it won't show `None` for the foreign key. The proposed fix is to just move the code in [_skip_foreign_key_transformations](https://github.com/sdv-dev/SDV/blob/9324fcdcaba99452a81227e2397898ef582ed2b3/sdv/multi_table/base.py#L268) to inside of `auto_assign_transformers`.  This will require some refactoring to loop over all tables and assign the transformers for foreign keys to `None`.

### Steps to reproduce

```python

data, metadata = download_demo('multi_table', 'world_v1')
synth = HMASynthesizer(metadata=metadata)
synth.get_transformers(table_name='City')
```
Observe the following output
```
{'Name': LabelEncoder(add_noise=True),
 'add_numerical': FloatFormatter(computer_representation='Int64'),
 'ID': None,
 'CountryCode': RegexGenerator(),
 'District': LabelEncoder(add_noise=True),
 'Population': FloatFormatter(computer_representation='Int64')}
```

```python
synth.fit(data)
synth.get_transformers('City')
```

Notice the foreign key `CountryCode` now has `None` as a transformer.
```
{'Name': LabelEncoder(add_noise=True),
 'add_numerical': FloatFormatter(computer_representation='Int64'),
 'ID': None,
 'CountryCode': None,
 'District': LabelEncoder(add_noise=True),
 'Population': FloatFormatter(computer_representation='Int64')}
```

frances-h:
Resolve #1249

### DIFF
--- a/sdv/multi_table/base.py
+++ b/sdv/multi_table/base.py
@@ -217,7 +217,8 @@ class BaseMultiTableSynthesizer:
         """
         for table_name, table_data in data.items():
             self._validate_table_name(table_name)
-            self._table_synthesizers[table_name].auto_assign_transformers(table_data)
+            synthesizer = self._table_synthesizers[table_name]
+            self._skip_foreign_key_transformations(synthesizer, table_name, table_data)
 
     def get_transformers(self, table_name):
         """Get a dictionary mapping of ``column_name`` and ``rdt.transformers``.


sdv/multi_table/base.py
```diff
@@ -217,7 +217,8 @@ def auto_assign_transformers(self, data):
         """
         for table_name, table_data in data.items():
             self._validate_table_name(table_name)
-            self._table_synthesizers[table_name].auto_assign_transformers(table_data)
+            synthesizer = self._table_synthesizers[table_name]
+            self._skip_foreign_key_transformations(synthesizer, table_name, table_data)
```
amontanez24, line 221: This method calls `self.auto_assign_transformers` so I think this will just recurse forever. I think instead we could move the logic inside of `_skip_foreign_key_transformations` that sets the foreign key's transformers to None inside of this method, and just call this method everywhere that `_skip_foreign_key_transformations` is currently called

amontanez24 (changes requested):

sdv/multi_table/base.py
amontanez24, line 221: minor: could we rename `_skip_foreign_key_transformations` to make it clear that it also assigns the transformers for a table? Maybe just `_assign_table_transformers` or something along those lines

amontanez24 (changes requested):
Can we add a small unit test that checks the transformer assigned to a foreign key column is now None after calling `auto_assign_transformers`?

sdv/multi_table/base.py
pvk-developer: Also @frances-h since you are calling it up here, it would be nice to move the method definition above this one

amontanez24 (approved):
Looks good! Thanks for addressing :shipit:

pvk-developer (approved):
LGTM :shipit:

PR #1265, opened 2023-02-14
Validate `enforce` parameters

### LINKED ISSUE #1242: Better input validation when creating single and multi table synthesizers
Version SDV 1.0 

### Problem Description
All synthesizers have `enforce_min_max_values` and `enforce_rounding` parameters that must be a boolean (`True` or `False`)

Currently, we are accepting other values that are not boolean. It is not clear what these parameters mean.

### Expected behavior
If I input anything other than a boolean for these parameters, raise an error.

```python
>>> synthesizer = GaussianCopulaSynthesizer(metadata, enforce_min_max_values='test')

SynthesizerInputError: Invalid value for parameter 'enforce_min_max_values' ('test'). Please provide True or False.
```

For multi table synthesizers, the same error can be thrown in the `set_table_parameters` method:
```python
>>> synthesizer.set_table_parameters(
    table_name='parent_table',
    table_parameters={
        'enforce_min_max_values': 'Test',
        'enforce_rounding':'Test2',
        'default_distribution': 'beta',
        'numerical_distributions': { 'test3': 'beta' }
    }
)

SynthesizerInputError: Invalid value for parameter 'enforce_min_max_values' ('Test'). Please provide True or False.
```

### Additional context
Just throw a single error as soon as we find one. If multiple parameters have an error, the user will just see an error for the first one we notice.

fealho:
Resolve #1242

### DIFF
--- a/tests/unit/constraints/test_tabular.py
+++ b/tests/unit/constraints/test_tabular.py
@@ -762,10 +762,8 @@ class TestFixedCombinations():
         assert instance._uuids_to_combinations is not None
         expected_out_a = pd.Series(['a', 'b', 'c'], name='a')
         pd.testing.assert_series_equal(expected_out_a, out['a'])
-        try:
-            [uuid.UUID(u) for c, u in out['b#c'].items()]
-        except ValueError:
-            pytest.fail('ValueError')
+
+        [uuid.UUID(u) for c, u in out['b#c'].items()]
 
     def test_transform_non_string(self):
         """Test the ``FixedCombinations.transform`` method with non strings.
@@ -799,10 +797,8 @@ class TestFixedCombinations():
         assert instance._uuids_to_combinations is not None
         expected_out_a = pd.Series(['a', 'b', 'c'], name='a')
         pd.testing.assert_series_equal(expected_out_a, out['a'])
-        try:
-            [uuid.UUID(u) for c, u in out['b#c#d'].items()]
-        except ValueError:
-            pytest.fail('ValueError')
+
+        [uuid.UUID(u) for c, u in out['b#c#d'].items()]
 
     def test_transform_not_all_columns_provided(self):
         """Test the ``FixedCombinations.transform`` method.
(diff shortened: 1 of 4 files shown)


tests/unit/constraints/test_tabular.py
```diff
@@ -765,7 +765,7 @@ def test_transform(self):
         try:
             [uuid.UUID(u) for c, u in out['b#c'].items()]
         except ValueError:
-            pytest.fail('ValueError')
+            pytest.fail('ValueError')  # noqa: PT009
```
fealho, line 768: `flake8-pytest-style` update doesn't let you to use these anymore

frances-h: Why does this need to be in a try/except block? Wouldn't the `ValueError` cause the test to fail anyways?

amontanez24: Does this line usually get hit? I'm not sure why we need this anyway

fealho: @frances-h I believe it's for clarity: it's weird to just create a list and not use it. This makes it clear we are expecting `uuid.UUID()` to not raise a ValueError.

@amontanez24 It never gets reached, since the test (hopefully) never fails. The list itself should stay there though, since we need to verify that every value was correctly converted into an id.

I really don't think this matters though, I'm just deleting it as you suggested so we don't have to add the `noqa`.

frances-h: @fealho re: clarity, another option could be to add an `assert` before the list. Assuming the list is not empty, it'll always pass and maybe be clearer that we want to explicitly test creating a list of UUIDs.

tests/unit/constraints/test_tabular.py
```diff
@@ -762,10 +762,8 @@ def test_transform(self):
         assert instance._uuids_to_combinations is not None
         expected_out_a = pd.Series(['a', 'b', 'c'], name='a')
         pd.testing.assert_series_equal(expected_out_a, out['a'])
-        try:
-            [uuid.UUID(u) for c, u in out['b#c'].items()]
-        except ValueError:
-            pytest.fail('ValueError')
+
+        [uuid.UUID(u) for _, u in out['b#c'].items()]
```
amontanez24, line 766: It seems like this is just trying to check that the outputs are indeed all UUIDs, which is what they're supposed to be. Tbh I don't know if that is necessary for a unit test. At this point, the map of uuids to combinations should have already been made and we really just need to assert that the values are from that map like expected

frances-h (approved):

amontanez24 (approved):
:shipit:

PR #1266, opened 2023-02-15
Fix `HMASynthesizer` not able to fit when only primary keys and foreign keys are within a table.

### LINKED ISSUE #1257: `HMASynthesizer` errors out when fitting a dataset that has a table which holds primary key and foreign keys only
### Environment Details

* SDV version: 1.0.0.dev
* Python version: 3.8
* Operating System: MacOS

### Error Description
The `HMASynthesizer` fails to fit a dataset if it has a table that is contains only a primary key and foreign keys. This error occurs because all the columns have been dropped for the given table therefore it can't fit the model.

```python
File ~/Projects/sdv-dev/SDV/sdv/multi_table/hma.py:188, in HMASynthesizer._model_table(self, table_name, tables)
    184 self._clear_nans(table)
    185 LOGGER.info('Fitting %s for table %s; shape: %s', self._synthesizer.__name__,
    186             table_name, table.shape)
--> 188 self._table_synthesizers[table_name].fit_processed_data(table)
    190 for name, values in keys.items():
    191     table[name] = values

File ~/Projects/sdv-dev/SDV/sdv/single_table/base.py:421, in BaseSynthesizer.fit_processed_data(self, processed_data)
    414 def fit_processed_data(self, processed_data):
    415     """Fit this model to the transformed data.
    416 
    417     Args:
    418         processed_data (pandas.DataFrame):
    419             The transformed data used to fit the model to.
    420     """
--> 421     self._fit(processed_data)
    422     self._fitted = True
    423     self._fitted_date = datetime.datetime.today().strftime('%Y-%m-%d')

File ~/Projects/sdv-dev/SDV/sdv/single_table/copulas.py:132, in GaussianCopulaSynthesizer._fit(self, processed_data)
    130 with warnings.catch_warnings():
    131     warnings.filterwarnings('ignore', module='scipy')
--> 132     self._model.fit(processed_data)

....

ValueError: need at least one array to concatenate
``` 

This is what the Metadata structure looks like:
image

As seen in the image the error occurs on the `gmember` table.

### Steps to reproduce

```python
from sdv.datasets.demo import download_demo, get_available_demos

data, metadata = download_demo('multi_table', 'Biodegradability_v1')

hmas = HMASynthesizer(metadata)
hmas.fit(data)
```

pvk-developer:
Resolves #1257

### DIFF
--- a/sdv/multi_table/base.py
+++ b/sdv/multi_table/base.py
@@ -29,18 +29,16 @@ class BaseMultiTableSynthesizer:
 
     _synthesizer = GaussianCopulaSynthesizer
     _numpy_seed = 73251
-    _numpy_generator = None
 
     @contextlib.contextmanager
     def _set_temp_numpy_seed(self):
         initial_state = np.random.get_state()
         if isinstance(self._numpy_seed, int):
             np.random.seed(self._numpy_seed)
-            self._numpy_generator = np.random.PCG64(self._numpy_seed)
-            np.random.default_rng(self._numpy_generator)
+            np.random.default_rng(self._numpy_seed)
         else:
             np.random.set_state(self._numpy_seed)
-            np.random.default_rng(self._numpy_generator)
+            np.random.default_rng(self._numpy_seed[1])
         try:
             yield
         finally:
--- a/sdv/multi_table/base.py
+++ b/sdv/multi_table/base.py
@@ -29,19 +29,22 @@ class BaseMultiTableSynthesizer:
 
     _synthesizer = GaussianCopulaSynthesizer
     _numpy_seed = 73251
+    _numpy_generator = 73251
 
     @contextlib.contextmanager
     def _set_temp_numpy_seed(self):
         initial_state = np.random.get_state()
         if isinstance(self._numpy_seed, int):
             np.random.seed(self._numpy_seed)
+            np.random.default_rng(self._numpy_generator)
         else:
             np.random.set_state(self._numpy_seed)
-
+            np.random.default_rng(self._numpy_generator)
         try:
             yield
         finally:
             self._numpy_seed = np.random.get_state()
+            self._numpy_generator = np.random.get_bit_generator()
             np.random.set_state(initial_state)
 
     def _initialize_models(self):
--- a/sdv/multi_table/base.py
+++ b/sdv/multi_table/base.py
@@ -29,13 +29,14 @@ class BaseMultiTableSynthesizer:
 
     _synthesizer = GaussianCopulaSynthesizer
     _numpy_seed = 73251
-    _numpy_generator = 73251
+    _numpy_generator = None
 
     @contextlib.contextmanager
     def _set_temp_numpy_seed(self):
         initial_state = np.random.get_state()
         if isinstance(self._numpy_seed, int):
             np.random.seed(self._numpy_seed)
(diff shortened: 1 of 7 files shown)


sdv/multi_table/hma.py
```diff
@@ -140,7 +148,7 @@ def _pop_foreign_keys(self, table_data, table_name):
                 The name representing the table.
 
         Returns:
-            keyes (dict):
```
amontanez24, line 143: good catch

sdv/multi_table/base.py
```diff
@@ -28,15 +28,18 @@ class BaseMultiTableSynthesizer:
...
     _synthesizer = GaussianCopulaSynthesizer
     _numpy_seed = 73251
+    _numpy_generator = None
 
     @contextlib.contextmanager
     def _set_temp_numpy_seed(self):
         initial_state = np.random.get_state()
         if isinstance(self._numpy_seed, int):
             np.random.seed(self._numpy_seed)
+            self._numpy_generator = np.random.PCG64(self._numpy_seed)
```
amontanez24, line 38: What exactly is going on here with the numpy generator?

pvk-developer: I found that at some point `scipy.stats` uses `numpy.random.default_rng` to control randomness
rather than the seeded one which is working widely across our projects.

Here is how `numpy` proposes the new randomness: https://numpy.org/doc/stable/reference/random/generator.html
which I think that down the line somewhere `scipy.stats` started to use it rather than using
the `numpy.random`.

The problem with the `default_rng` is that I didn't find a way to properly set a `numpy.random.get_state`
object, which ended up leading me to this solution.

Also I found this interesting, but it will require us to re-work part of `RDT`: [Good RNG practices](https://albertcthomas.github.io/good-practices-random-number-generators/),
TLDR; pass down the line a `generator` object rather than setting a `random.seed` and use `generator.random`
to generate random values etc...

sdv/multi_table/hma.py
```diff
@@ -431,13 +444,18 @@ def _get_likelihoods(self, table_rows, parent_rows, table_name, foreign_key):
...
+
         for parent_id, row in parent_rows.iterrows():
             parameters = self._extract_parameters(row, table_name, foreign_key)
             table_meta = self._table_synthesizers[table_name].get_metadata()
             synthesizer = self._synthesizer(table_meta, **self._synthesizer_kwargs)
-            synthesizer.set_parameters(parameters)
+            synthesizer._set_parameters(parameters)
             try:
-                likelihoods[parent_id] = synthesizer.get_likelihood(table_rows)
+                synthesizer._numpy_seed = np.random.get_state()[1]
```
amontanez24, line 457: why do we need to set the numpy seed here but also set the numpy generator on the method we call on the next line? Shouldn't one be enough?

pvk-developer: just pass the seed to the method you mean?

amontanez24: No I mean we call `synthesizer._get_likelihood` on the next line and inside that method you say 
```python
with np.random.default_rng(self._numpy_seed):
    ....
```
I'm just wondering why we need to do both things and why we don't do them in the same place

fealho (commented):
Looks good, it's just missing some testing (it's -0.18% in coverage).

amontanez24 (approved):
I think this looks good now!

PR #1286, opened 2023-02-28
Add single table integration tests

### LINKED ISSUE #1276: Add integration tests for 'Synthesize a Table' demo
### Problem Description

### Expected behavior

- Make integration tests for all of the `SDV` processes ran in the "Synthesize a Table" demo. There are two of these demos, one for `GaussianCopulaSynthesizer` and one for `CTGANSynthesizer`. Write the tests for both

### Technical Details

    - Validate the sampled data

frances-h:
Resolve #1276

### DIFF
--- a/tests/integration/single_table/test_copulas.py
+++ b/tests/integration/single_table/test_copulas.py
@@ -1,16 +1,11 @@
-from pathlib import Path
-
 from sdv.datasets.demo import download_demo
 from sdv.evaluation.single_table import evaluate_quality, get_column_pair_plot, get_column_plot
 from sdv.sampling import Condition
 from sdv.single_table import GaussianCopulaSynthesizer
 
 
-def test_synthesize_table_gaussian_copula(tmpdir):
-    """Test synthesize a table demo.
-
-    Runs and tests the Synthesize a Table demo using Gaussian Copula.
-    """
+def test_synthesize_table_gaussian_copula(tmp_path):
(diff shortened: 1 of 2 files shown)


pvk-developer (approved):
I like the structure I may change the other tests that I did to follow the same one 👍🏻

tests/integration/single_table/test_copulas.py
```diff
@@ -0,0 +1,121 @@
...
+    assert model_path.exists()
+    assert model_path.is_file()
+    loaded_synthesizer = GaussianCopulaSynthesizer.load(model_path)
+    assert isinstance(synthesizer, GaussianCopulaSynthesizer)
+    assert loaded_synthesizer.get_info() == synthesizer.get_info()
+    assert loaded_synthesizer.metadata.to_dict() == metadata.to_dict()
+    loaded_synthesizer.sample(20)
+
+    # Assert - custom synthesizer
+    assert custom_quality_report.get_score() > 0
```
amontanez24, line 111: Is it ever possible for this to be 0? I'd hope not but just curious

amontanez24 (approved):
LGTM, very thorough!

tests/integration/single_table/test_copulas.py
frances-h: I think technically yes if every score was zero or `nan` since it's just the average

PR #1293, opened 2023-03-02
Add integration test for adding constraints demo

### LINKED ISSUE #1280: Add integration tests for 'Adding Constraints' demo
### Problem Description

### Expected behavior

- Make integration tests for all of the `SDV` processes ran in the "Adding Constraints" demo.

### Technical Details

    - Validate the sampled data
- Test loading from a different file. You can download the file with the custom constraint [here](https://docs.sdv.dev/sdv/reference/constraint-logic/custom-logic/example-iftruethenzero#download-the-example)

pvk-developer:
Resolves #1280

### DIFF
--- a/tests/integration/single_table/test_copulas.py
+++ b/tests/integration/single_table/test_copulas.py
@@ -1,3 +1,5 @@
+import pandas as pd
+
 from sdv.datasets.demo import download_demo
 from sdv.evaluation.single_table import evaluate_quality, get_column_pair_plot, get_column_plot
 from sdv.sampling import Condition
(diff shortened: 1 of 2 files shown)


frances-h (approved):
Looks good!

tests/integration/single_table/test_copulas.py
```diff
@@ -0,0 +1,83 @@
...
+    assert model_path.exists()
+    assert model_path.is_file()
+    loaded_synthesizer = GaussianCopulaSynthesizer.load(model_path)
+
+    assert isinstance(loaded_synthesizer, GaussianCopulaSynthesizer)
+    assert loaded_synthesizer.get_info() == synthesizer.get_info()
+    assert loaded_synthesizer.metadata.to_dict() == metadata.to_dict()
+    sampled_data = loaded_synthesizer.sample(100)
+    validation = sampled_data[sampled_data['has_rewards']]
+    assert validation['amenities_fee'].sum() == 0.0
```
amontanez24, line 83: can we validate the sampled data with the `synthesizer.validate` method?

tests/integration/single_table/test_copulas.py
```diff
@@ -0,0 +1,83 @@
...
+    # Re-Fit the model
+    synthesizer.fit(real_data)
+    synthetic_data_custom_constraint = synthesizer.sample(500)
+
+    # Assert
+    validation = synthetic_data_custom_constraint[synthetic_data_custom_constraint['has_rewards']]
+    assert validation['amenities_fee'].sum() == 0.0
+
+    # Save and Load
+    temp_dir = TemporaryDirectory()
```
amontanez24, line 69: minor: but I see in some places we use `temp_dir = TemporaryDirectory()` and in other tests we pass the `tempdir` as a parameter on the test. Is there a reason to use one over the other and if not should we unify these and just use one throughout our tests?

amontanez24 (approved):
A couple minor comments but otherwise LGTM!

tests/integration/single_table/test_copulas.py
pvk-developer: `tempdir` as a parameter is a fixture that comes with `pytest`. I think it's pretty much the same on how we would like to proceed with this. Only pro I see is one less import but a little bit confusing for newcomers if they don't know where the `tmpdir` came from, other than that there is no strong preferences.

amontanez24 (approved):

PR #1302, opened 2023-03-09
Add custom processing and anonymization integration test

### LINKED ISSUE #1279: Add integration tests for 'Custom Processing & Anonymization' demo
### Problem Description

### Expected behavior

- Make integration tests for all of the `SDV` processes ran in the "Custom Processing & Anonymization" demo

### Technical Details

    - Validate the sampled data

frances-h:
Resolve #1279

### DIFF
--- a/tests/integration/single_table/test_copulas.py
+++ b/tests/integration/single_table/test_copulas.py
@@ -1,4 +1,8 @@
+from uuid import UUID
+
 import pandas as pd
+from rdt.transformers import (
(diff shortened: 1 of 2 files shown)


amontanez24 (approved):
LGTM! Completely unrelated to this issue, but while you're updating the tests, can you change [this line](https://github.com/sdv-dev/SDV/blob/9e62b05b0afc9bb00a92c0244c48ad7fc9780f0b/tests/integration/single_table/test_ctgan.py#L19) to be more like 100 epochs? This test takes a while

PR #1305, opened 2023-03-10
Update Readme to introduce `SDV` 1.0 beta

### LINKED ISSUE #1301: Update README to introduce Beta
Now that it's almost ready, let's update the README asking users to try the Beta version and give us feedback.

The changes be merged to `master` since that is the default branch that shows up for anyone visiting our GitHub page -- but we **should not** merge them into stable yet.

### Additional Context
For the actual release of SDV 1.0, we should update the README once again with new code snippets. The newer version will be the one that's actually merged into `stable`

pvk-developer:
Resolves #1301

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 [![Tests](https://github.com/sdv-dev/SDV/workflows/Run%20Tests/badge.svg)](https://github.com/sdv-dev/SDV/actions?query=workflow%3A%22Run+Tests%22+branch%3Amaster)
 [![Coverage Status](https://codecov.io/gh/sdv-dev/SDV/branch/master/graph/badge.svg)](https://codecov.io/gh/sdv-dev/SDV)
 [![Downloads](https://pepy.tech/badge/sdv)](https://pepy.tech/project/sdv)
-[![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/sdv-dev/SDV/master?filepath=tutorials)
+[![Binder](https://mybinder.org/badge_logo.svg)](https://bit.ly/sdv-tutorials)
 [![Slack](https://img.shields.io/badge/Slack%20Workspace-Join%20now!-36C5F0?logo=slack)](https://bit.ly/sdv-slack-invite)
 
 <div align="left">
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -89,7 +87,14 @@ hierarchical generative modeling and recursive sampling techniques.
 
 If you want to quickly discover **SDV**, simply click the button below and follow the tutorials!
 
-[![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/sdv-dev/SDV/master?filepath=tutorials)
+[![Binder](https://mybinder.org/badge_logo.svg)](https://bit.ly/sdv-tutorials)
+
+**New! Try the SDV 1.0 Beta**
```
frances-h, line 92: Should this also be a third level heading instead of just bolded?

pvk-developer: As discussed during a call with @npatki this can stay as it is for now.

frances-h (approved):

README.md
```diff
@@ -41,26 +41,24 @@ Underneath the hood it uses several probabilistic graphical modeling and deep le
...
-| :octocat: **[Repository]**                    | The link to the Github Repository of this library.                   |
-| :keyboard: **[Development Status]**           | This software is in its Pre-Alpha stage.                             |
-| [![][Slack Logo] **Community**][Community]    | Join our Slack Workspace for announcements and discussions.          |
-| [![][MyBinder Logo] **Tutorials**][Tutorials] | Run the SDV Tutorials in a Binder environment.                       |
+| Important Links                               |                                                                                                     |
+| --------------------------------------------- | ----------------------------------------------------------------------------------------------------|
+| :computer: **[Website]**                      | Check out the SDV Website for more information about the project.                                   |
+| :book: **[Documentation]**                    | Learn how to use the SDV library with tutorials, demos and API references.                          |
```
npatki, line 47: @pvk-developer I wonder why this is wrapping around to a new line right now? Is there any way that the word "Documentation" can be right next to the book icon on the same line?

image

npatki: If we can't get it to fit, I'd suggest we:
- Shorten "Documentation" to "Docs"
- Shorten "DataCebo Blog" to "Blog"

npatki (commented):
Left a few minor comments. Rest LGTM

README.md
pvk-developer: @npatki it's because of the length of the text on the description. Here is how it looks after the changes:
Screenshot from 2023-03-13 13-20-50
As you can see now the `Important Links` get wrapped. However, if you open the `README.md` only it shows as this: (ps this is the old version with `datacebo blog` and `documentation`):
Screenshot from 2023-03-13 13-23-29

The reason is because of github styling. I did some experiments trying to style it with `html` but still getting similar results. I don't really want to hard code the style as it may look different on different devices as well.

amontanez24 (approved):

README.md
npatki: Interesting. Let's keep the top version. Even if `REAMDE.md` shows the full version, I think most people will read it on the main branch.

PR #1307, opened 2023-03-13
Add end to end test for synthesizing sequences.

### LINKED ISSUE #1295: Add integration tests for 'Synthesize Sequences' demo
### Problem Description

### Expected behavior
- Make integration tests for all of the SDV processes ran in the "Synthesize Sequences" demo.

### Technical Details
- Put the tests in the appropriate files
   - Validate the sampled data

pvk-developer:
Resolves #1295

### DIFF
--- a/tests/integration/sequential/test_par.py
+++ b/tests/integration/sequential/test_par.py
@@ -3,6 +3,7 @@ import datetime
 import pandas as pd
 from deepecho import load_demo
 
+from sdv.datasets.demo import download_demo
 from sdv.metadata import SingleTableMetadata
 from sdv.sequential import PARSynthesizer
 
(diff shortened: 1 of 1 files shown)


fealho (approved):

tests/integration/sequential/test_par.py
```diff
@@ -103,3 +104,69 @@ def test_save_and_load(tmp_path):
...
+    })
+
+    # Run - Fit
+    synthesizer.fit(real_data)
+    custom_synthesizer.fit(real_data)
+
+    # Run - Sample
+    synthetic_data = synthesizer.sample(num_sequences=10)
+    custom_synthetic_data = custom_synthesizer.sample(num_sequences=3, sequence_length=2)
+    custom_synthetic_data_conditional = custom_synthesizer.sample_sequential_columns(
```
amontanez24, line 151: Can we test that these sampled rows have the correct provided context?

amontanez24 (approved):
One request but after that looks good!

PR #1308, opened 2023-03-13
Update metadata error for relationship between primary keys

### LINKED ISSUE #1296: Make metadata error more descriptive for invalid relationships
### Problem Description
The SDV only supports multi table schemas that connect a foreign key (of a child table) to a primary key (of a parent table). The schemas _do not_ support the connection between 2 primary keys.

If I try to validate a schema that connects 2 primary keys, the metadata correctly throws an error. However, the error is vague and doesn't help me locate what's wrong. This is especially confusing if there are multiple connections that have issues.

```python

    modality='multi_table',
    dataset_name='Atherosclerosis_v1'
)
metadata.validate()
```

Output
```
Relationships:
A relationship must be specified between a primary key and a non-primary key.
A relationship must be specified between a primary key and a non-primary key. 
```

### Expected behavior
It would be helpful to print out more information about which relationships are incorrect. Let's add this description into the message:

```
Relationships:
Invalid relationship between table 'Death' and table 'Entry'. A relationship must be connect a primary key with a non-primary key.
Invalid relationship between table 'Letter' and table 'Entry'. A relationship must be connect a primary key with a non-primary key.
```

fealho:
Resolve #1296.

### DIFF
--- a/sdv/metadata/multi_table.py
+++ b/sdv/metadata/multi_table.py
@@ -128,12 +128,15 @@ class MultiTableMetadata:
                 f'tables {errors}.'
             )
 
-    def _validate_foreign_child_key(self, child_table_name, child_foreign_key):
+    def _validate_foreign_child_key(self, child_table_name, parent_table_name, child_foreign_key):
         child_primary_key = cast_to_iterable(self.tables[child_table_name].primary_key)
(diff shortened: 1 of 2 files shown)


frances-h (approved):
LGTM 👍

amontanez24 (approved):

PR #1312, opened 2023-03-15
Error if invalid columns passed to FixedCombinations

### LINKED ISSUE #1306: Throw a `SynthesizerInputError` if `FixedCombinations` constraint is applied to a column that is not `boolean` or `categorical`
### Problem Description

The current `FixedCombinations` will not error out when applied to sdtypes that are not `boolean` or `categorical`. This ends up generating confusing results.

### Expected behavior

If applied to a different `sdtype` that is not boolean or categorical it should throw a `SynthesizerInputError`.

### Suggested message

`SynthesizerInputError`: Invalid columns `('age', 'address')` supplied to a `FixedCombinations` constraint. This constraint only supports boolean and categorical columns.

### Additional context

```python
data = pd.DataFrame({'customer_id': ['b'], 'event_id': ['a'], 'id': [1]})
metadata.update_column(
    column_name='customer_id',
    sdtype='uuid4', pii=True
)
metadata.update_column(
    column_name='event_id',
    sdtype='uuid4'
)
constr = {
    'constraint_class': 'FixedCombinations',
    'constraint_parameters': {
        'column_names': ['customer_id', 'event_id']
    }
}
gc = GaussianCopulaSynthesizer(metadata)
gc.add_constraints([constr])
gc.fit(data)
gc.sample(10)
```
This ends up producing the following output which is not as expected:
```
  customer_id event_id  id
0           b        a   1
1           b        a   1
2           b        a   1
3           b        a   1
4           b        a   1
5           b        a   1
6           b        a   1
7           b        a   1
8           b        a   1
9           b        a   1
```

frances-h:
Resolve #1306

### DIFF
--- a/sdv/constraints/tabular.py
+++ b/sdv/constraints/tabular.py
@@ -235,6 +235,19 @@ class FixedCombinations(Constraint):
     _combinations_to_uuids = None
     _uuids_to_combinations = None
 
+    @staticmethod
+    def _validate_metadata_specific_to_constraint(metadata, **kwargs):
+        invalid_columns = []
+        column_names = kwargs.get('column_names')
+        for column in column_names:
+            if metadata.columns[column]['sdtype'] not in ['boolean', 'categorical']:
+                invalid_columns.append(column)
+        if invalid_columns:
+            columns = '", "'.join(invalid_columns)
+            raise ConstraintMetadataError(f'Invalid columns ("{columns}") supplied to a '
+                                          'FixedCombinations constraint. This constraint only '
+                                          'supports boolean and categorical columns.')
+
     def __init__(self, column_names):
         if len(column_names) < 2:
             raise ValueError('FixedCombinations requires at least two constraint columns.')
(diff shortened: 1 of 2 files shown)


fealho (changes requested):
You need to call the `_validate_metadata_specific_to_constraint` method in the code, so it's actually used.

sdv/constraints/tabular.py
```diff
@@ -235,6 +235,19 @@ class FixedCombinations(Constraint):
...
     _uuids_to_combinations = None
 
+    @staticmethod
+    def _validate_metadata_specific_to_constraint(metadata, **kwargs):
+        invalid_columns = []
+        column_names = kwargs.get('column_names')
+        for column in column_names:
+            if metadata.columns[column]['sdtype'] not in ['boolean', 'categorical']:
+                invalid_columns.append(column)
+        if invalid_columns:
```
pvk-developer, line 245: Can we have a breakline before this line ? There is still not a plugin to capture this unfortunately, but a rule of thumb is: If the code is going back a level have a breakline.

pvk-developer (approved):
Just a minor style comment but other than that LGTM!

frances-h:
> You need to call the `_validate_metadata_specific_to_constraint` method in the code, so it's actually used.

It's called in the base `Constraint._validate_metadata` method.

fealho (approved):

PR #1323, opened 2023-03-21
Return metadata object when upgrading metadata

### LINKED ISSUE #1319: In `upgrade_metadata`, return the object instead of writing it to a JSON file
### Problem Description
I usually want to upgrade the metadata in order to use it for a new synthesizer in SDV 1.0. Unfortunately, the current [upgrade metadata function](https://docs.sdv.dev/sdv/multi-table-data/data-preparation/multi-table-metadata-api#load-previously-saved-metadata) writes the new JSON to a file. I have to take the extra step to load it in after that.

```python

MultiTableMetadata.upgrade_metadata(
    old_filepath='my_metadata_old.json',
    new_filepath='my_metadata_new.json'
)

# extra step
metadata = MultiTableMetadata.load_from_json(filepath='my_metadata_new.json')
```

Additionally, the upgrade metadata parameters are confusing. I have to make sure that:
- `old_filepath` _does_ exist and has a metadata
- `new_filepath` _does not exist_ so new metadata can be written there

### Expected behavior
It would be nice to modify this function to return the object instead. This can be the case for both single and multi table metadata.

```python
from sdv.metadata import MultiTableMetadata, SingleTableMetadata

multi_table_metadata = MultiTableMetadata.upgrade_metadata(
    filepath='tables_metadata_old.json')

single_table_metadata = SingleTableMetadata.upgrade_metadata(
  filepath='table_metadata_old.json')
```

Later I can save them.

```python
single_table_metadata.save_to_json(filepath='new_metadata.json')
```

frances-h:
Resolve #1319

### DIFF
--- a/sdv/metadata/multi_table.py
+++ b/sdv/metadata/multi_table.py
@@ -673,20 +673,20 @@ class MultiTableMetadata:
         return relationships
 
     @classmethod
-    def upgrade_metadata(cls, old_filepath, new_filepath):
+    def upgrade_metadata(cls, filepath):
         """Upgrade an old metadata file to the ``V1`` schema.
 
         Args:
-            old_filepath (str):
+            filepath (str):
                 String that represents the ``path`` to the old metadata ``json`` file.
-            new_file_path (str):
-                String that represents the ``path`` to save the upgraded metadata to.
 
         Raises:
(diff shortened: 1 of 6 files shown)


amontanez24 (approved):
LGTM!

PR #1326, opened 2023-03-22
Correctly upgrade PII fields in `upgrade_metadata`

### LINKED ISSUE #1317: In `upgrade_metadata`, PII values are being converted to generic categorical columns
### Environment Details

* SDV version: 1.0.0b0
* Python version: 3.7

### Error Description
The `upgrade_metadata` script is not properly converting `pii` values. In the old metadata, PII values were treated as a subset of categorical. For example, this was what I have in the old metadata:

```json
"user_state": {
  "type": "categorical",
  "pii": true,
  "pii_category": "state",
  "pii_locales": ["en_US"]
}
```

**Expected**: In the new SDV 1.0 it should be upgraded to the following:
```json
"user_state": {
  "sdtype": "state",
  "pii": true
}
```

**Observed (Incorrect)**: However, what I observe is this:
```json
"user_state": {
  "sdtype": "categorical"
}
```

### Corner Cases

In the old metadata, it is also possible to pass in a _list_ as `"pii_category"`. In this case the first element is the name of the faker function and the rest are the parameters to it (see [spec](https://sdv.dev/SDV/developer_guides/sdv/metadata.html#categorical-fields-data-anonymization)).

Example:

```json
"column_name": {
  "sdtype": "categorical",
  "pii": true,
  "pii_category": ["bothify", "####"],
}
```

The conversion should still include the faker method name as the sdtype. (We can ignore the parameters, as those can be set separately by the user during `update_transformers()`).

```json
"column_name": {
  "sdtype": "bothify"
  "pii": true
}
```

### Beware
In the old metadata, `'categorical'` by itself is a true categorical column. That should be converted as-is.

```json
"column_name": {
  "type": "categorical"
}
```

Converted to:
```json
"column_name": {
  "sdtype": "categorical"
}
```

frances-h:
Resolve #1317

### DIFF
--- a/sdv/metadata/metadata_upgrader.py
+++ b/sdv/metadata/metadata_upgrader.py
@@ -43,6 +43,14 @@ def _upgrade_columns_and_keys(old_metadata):
             if field != primary_key and field_meta.get('ref') is None:
                 alternate_keys.append(field)
 
+        if field_meta.get('pii'):
+            column_meta['pii'] = True
+            pii_category = field_meta.get('pii_category')
+            if isinstance(pii_category, list):
+                column_meta['sdtype'] = pii_category[0]
+            elif pii_category:
+                column_meta['sdtype'] = pii_category
+
         columns[field] = column_meta
 
     new_metadata['columns'] = columns
--- a/tests/unit/metadata/test_metadata_upgrader.py
+++ b/tests/unit/metadata/test_metadata_upgrader.py
@@ -625,9 +625,7 @@ def test_convert_metadata():
             },
             'student_id': {
                 'type': 'id',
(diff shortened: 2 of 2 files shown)


pvk-developer (approved):

fealho (approved):
Looks good 👍 Can you just double check that the "Beware" section works? You can just do it locally, since it's not really related to the code you changed.

frances-h:
> Looks good 👍 Can you just double check that the "Beware" section works? You can just do it locally, since it's not really related to the code you changed.

Still works! It's also covered by the existing test since the `duration` column is still left as a normal categorical.

PR #1327, opened 2023-03-22
In upgrade_metadata index primary keys should be converted to text

### LINKED ISSUE #1318: In `upgrade_metadata` index primary keys should be converted to `text`
### Environment Details

* SDV version: 1.0.0b0
* Python version: 3.7

### Background
I have the following metadata (written in the old format, pre-SDV 1.0):

```json
{
  "primary_key": "user_id",
  "fields": {
    "user_id": { "type": "id", "subtype": "integer" },
    ...
  }
}
```

This conveys that the user id is an index that increments `0, 1, 2, ...`

### Observed
When I call `updgrade_metadata`, I get the following:

```json
{
  "primary_key": "user_id",
  "columns": {
    "user_id": { "sdtype": "numerical" },
    ...
  }
}
```

_Technically_ this works, but it is weird to make this a `"numerical"` sdtype because it's not actually a statistically valid numerical column. (taking an average, correlation, etc. of this column does not make sense).

### Expected
Instead, I propose the `upgrade_metadata` script should instead make this a `"text"` sdtype with a Regex that generates indices. This follows our expectation: ID columns will have structured text.

```json
{
  "primary_key": "user_id",
  "columns": {
    "user_id": { "sdtype": "text", "regex_format": "\d{30}" },
    ...
  }
}
```

When applied to an int column, the synthesizer will return integer values that increase with every row: `0, 1, 2, ...` (up to a max of 10^30 rows, which is enough for most purposes).

amontanez24:
resolves #1318

### DIFF
--- a/sdv/metadata/metadata_upgrader.py
+++ b/sdv/metadata/metadata_upgrader.py
@@ -31,14 +31,13 @@ def _upgrade_columns_and_keys(old_metadata):
                 column_meta['datetime_format'] = datetime_format
 
         elif old_type == 'id':
+            column_meta['sdtype'] = 'id'
             if subtype == 'integer':
-                column_meta['sdtype'] = 'numerical'
-
+                regex_format = r'\d{30}'
             else:
-                column_meta['sdtype'] = 'text'
                 regex_format = field_meta.get('regex', '[A-Za-z]{5}')
-                if regex_format:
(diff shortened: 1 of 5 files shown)


frances-h (approved):
LGTM 👍

pvk-developer (approved):

PR #1331, opened 2023-03-23
Change `text` sdtype to `id`

### LINKED ISSUE #1324: Change 'text' sdtype to 'id'
### Problem Description

_As a user, it isn't clear which `sdtype` to use for id columns. Text seems like it would be for free text responses._

### Expected behavior

- Change any references and checks in the metadata for 'text' as an `sdtype` to 'id'
- Validate that any keys are either `sdtype` 'id' or one of the natural (faker) types. Keys **cannot** be categorical, boolean, numerical or datetime. Currently we only check for boolean and categorical.
- Change any references in the `DataProcessor`

frances-h:
Resolve #1324


pvk-developer (approved):
LGTM!

PR #1343, opened 2023-03-29
Add version add-on discovery

### LINKED ISSUE #1309: Add functionality to find version add-on
### Problem Description

_As an enterprise user, it would be nice to see the version of public SDV that is being used as well as the enterprise version._

The actual functionality for this will be added in SDV enterprise, however, this library will need to know to add that functionality to the proper name space.

### Expected behavior

- In this[ file](https://github.com/sdv-dev/SDV/blob/master/sdv/__init__.py), add and call a function that uses an entry point called `sdv_versioning` to add the `version` module to the namespace.
- Add any necessary logging and error checking
- Be careful not to override anything in the globals that is needed

### Additional context
The code can look something like the following:

```python
def _add_versioning():
    for entry_point in pkg_resources.iter_entry_points('sdv_version'):
        try:
            version_module = entry_point.load()
        except Exception:
            message = f'Failed to load "{entry_point.name}" from "{entry_point.module_name}". '
            continue

        if 'version' not in globals():
            globals()['version'] = version_module

_add_versioning()
```

In this case, the loaded entry point should return the module itself, so we can just add it directly to the `globals()` under the name `version`. This should enable the following:
```python
>>> sdv.version.public
0.18.0
```

frances-h:
Resolve #1309

### DIFF
--- a/sdv/__init__.py
+++ b/sdv/__init__.py
@@ -7,3 +7,22 @@
 __author__ = 'DataCebo, Inc.'
 __email__ = 'info@sdv.dev'
 __version__ = '1.0.1.dev0'
+
+import warnings
+from importlib.metadata import entry_points
+
+
+def _add_version():
+    for entry_point in entry_points(group='sdv_modules', name='version'):
+        try:
+            module = entry_point.load()
+        except Exception:
+            msg = f'Failed to load "{entry_point.name}" from "{entry_point.module}". '
+            warnings.warn(msg)
+            continue
+
+        if 'version' not in globals():
+            globals()['version'] = module
+
+
+_add_version()
--- a/sdv/__init__.py
+++ b/sdv/__init__.py
@@ -9,11 +9,19 @@ __email__ = 'info@sdv.dev'
 __version__ = '1.0.1.dev0'
 
 import warnings
-from importlib.metadata import entry_points
+from importlib.metadata import entry_points as get_entry_points
 
 
 def _add_version():
-    for entry_point in entry_points(group='sdv_modules', name='version'):
+    try:
+        entry_points = get_entry_points(name='version', group='sdv_modules')
+    except TypeError:
+        entry_points = [
+            entry_point for entry_point in get_entry_points().get('sdv_modules', [])
+            if entry_point.name == 'version'
+        ]
+
+    for entry_point in entry_points:
         try:
             module = entry_point.load()
         except Exception:


sdv/__init__.py
```diff
@@ -7,3 +7,23 @@
...
 __email__ = 'info@sdv.dev'
 __version__ = '1.0.1.dev0'
+
+import warnings
+
+from pkg_resources import iter_entry_points
+
+
+def _add_version():
+    for entry_point in iter_entry_points(name='version', group='sdv_modules'):
```
amontanez24, line 17: I'm wondering if we can make this method a general purpose method that finds all of the new modules to be added in the group `sdv_modules` and then just adds them to the globals based on the `__name__`. What do you think of that?

frances-h: I think that might cause issues if we want to have add-ons for specific submodules instead of at the top level like we do here. Unless we want to have submodule add-ons have a different group?

amontanez24: That's exactly what I was thinking. As a way of limiting the number of functions we have to write here, we can have one that adds modules directly at the current level and it knows that it's safe because they're in the `sdv_modules` group. If we wanted to add a new module to `sdv.metadata`, then we should have another function like this in the `sdv.metadata.__init__.py` that looks for the group `sdv.metadata_modules` or something like that. I just think that adding a full module to a package is something that might happen frequently. I'm open to ideas on this though

frances-h: I think that makes sense, I made the changes. If we want to be able to load add-ons for a submodule, we just need to add `sdv._addons._find_addons` to the `__init__.py` with whatever group name we decide to use for that submodule.

frances-h: If we like this format more, I'll update the other libraries as well

sdv/_addons.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def _find_addons(group, parent_globals):
+    """Find and load add-ons based on the given group.
+
+    Args:
+        group (str): The name of the entry points group to load.
+        parent_globals (dict): The caller's global scope. Modules will be added
+            to the parent's global scope through their name.
+    """
```
amontanez24, line 14: I think this is pretty cool. In some circumstances we may need to add other methods because the grouping might not be as convenient. For example, with transformers, sometimes we want to add them to the `rdt.transformers` globals and sometimes we only want to add them a level in, but for anytime we want to add things at surface level, I think this should work great

sdv/_addons.py
```diff
@@ -0,0 +1,24 @@
...
+import warnings
+
+from pkg_resources import iter_entry_points
+
+
+def _find_addons(group, parent_globals):
+    """Find and load add-ons based on the given group.
+
+    Args:
+        group (str): The name of the entry points group to load.
```
fealho, line 11: New line after variable definition:
``` 
group (str):
     The name...
```

tests/unit/test__addons.py
```diff
@@ -0,0 +1,45 @@
+from unittest.mock import Mock, patch
+
+from sdv._addons import _find_addons
+
+
+@patch('sdv._addons.iter_entry_points')
+def test__add_version(entry_points_mock):
```
fealho, line 7: Needs to contain the name of the method you are running: `test__find_addons`

tests/unit/test__addons.py
```diff
@@ -0,0 +1,45 @@
...
+    _find_addons(group='group', parent_globals=test_dict)
+
+    # Assert
+    entry_points_mock.assert_called_once_with(group='group')
+    assert test_dict['entry_name'] == 'entry_point'
+
+
+@patch('sdv._addons.warnings.warn')
+@patch('sdv._addons.iter_entry_points')
+def test__add_version_bad_addon(entry_points_mock, warning_mock):
```
fealho, line 26: same

fealho (commented):
Looks good, just some nitpicks.

fealho (approved):

PR #1358, opened 2023-04-04
Issue_1341 improve modify error message

### LINKED ISSUE #1341: Improve error message for invalid primary keys
Version: SDV 1.0.0rc0 (release candidate)

### Problem Description
Sometimes, I may accidentally mark a column as a primary key -- even though it is the wrong sdtype. When this happens, the SDV correctly shows me an error.

```python
metadata.set_primary_key(
  table_name='hotels',
  column_name='city')

InvalidMetadataError: The primary_keys ['city'] must be type 'id' or a valid Faker function.
```

However the error is a bit misleading. The sdtype does not have to be a valid _Faker_ function -- it may also be one of the many premium sdtypes that we support.

### Expected behavior
It would be more accurate to update the error message and say the following:

```python
InvalidMetadataError: The primary_keys ['city'] must be type 'id' or another PII type.
```

R-Palazzo:
Resolves #1341.

### DIFF
--- a/sdv/metadata/single_table.py
+++ b/sdv/metadata/single_table.py
@@ -300,7 +300,7 @@ class SingleTableMetadata:
         if bad_keys:
             raise InvalidMetadataError(
                 f"The {key_type}_keys {sorted(bad_keys)} must be type 'id' or "
-                'a valid Faker function.'
+                'another PII type.'
             )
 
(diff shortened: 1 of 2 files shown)


pvk-developer (approved):
LGTM!

PR #1361, opened 2023-04-05
Issue 1354 remove table name parameter

### LINKED ISSUE #1354: Method `load_custom_constraint_classes` does not need `table_name` parameter
### Problem Description
Custom constraint logic is usually written in a separate Python file. In order to apply it, it's necessary to load in the file using the [`load_custom_constraint_classes` method](https://docs.sdv.dev/sdv/multi-table-data/modeling/synthetic-data-workflow/constraints#load_custom_constraint_classes).

As shown in the docs, this method should only accept the file name and class names as parameters. However in practice, I see that it is also requiring `table_name`. 

### Expected behavior
We should remove `table_name` as a parameter for this method. The purpose of the method is to load in the files so that it's possible to use the constraint at a synthesizer-level. There is no real need to specify the table -- in fact, I may wish to use the constraint in several different tables.

### Additional context
The same change should also be made to the [`add_custom_constraint_classes` method](https://docs.sdv.dev/sdv/multi-table-data/modeling/synthetic-data-workflow/constraints#can-i-write-my-custom-logic-in-the-same-file-as-the-synthesizer).

One possible way to make this change: Internally, we can "load" the class in for every possible single table.

R-Palazzo:
Resolves #1354.

### DIFF
--- a/sdv/multi_table/base.py
+++ b/sdv/multi_table/base.py
@@ -429,7 +429,7 @@ class BaseMultiTableSynthesizer:
 
         return constraints
 
-    def load_custom_constraint_classes(self, table_name, filepath, class_names):
+    def load_custom_constraint_classes(self, filepath, class_names):
         """Load a custom constraint class for the specified table's synthesizer.
 
         Args:
@@ -441,9 +441,10 @@ class BaseMultiTableSynthesizer:
             class_names (list):
                 A list of custom constraint classes to be imported.
         """
-        self._table_synthesizers[table_name].load_custom_constraint_classes(filepath, class_names)
+        for t_name in self._table_synthesizers:
+            self._table_synthesizers[t_name].load_custom_constraint_classes(filepath, class_names)
 
-    def add_custom_constraint_class(self, table_name, class_object, class_name):
+    def add_custom_constraint_class(self, class_object, class_name):
         """Add a custom constraint class for the synthesizer to use.
 
         Args:
@@ -454,8 +455,10 @@ class BaseMultiTableSynthesizer:
             class_name (str):
                 The name to assign this custom constraint class. This will be the name to use
                 when writing a constraint dictionary for ``add_constraints``.
+        #self._table_synthesizers[table_name].add_custom_constraint_class(class_object, class_name)
         """
-        self._table_synthesizers[table_name].add_custom_constraint_class(class_object, class_name)
+        for t_name in self._table_synthesizers:
+            self._table_synthesizers[t_name].add_custom_constraint_class(class_object, class_name)
 
     def get_info(self):
         """Get dictionary with information regarding the synthesizer.
--- a/sdv/multi_table/base.py
+++ b/sdv/multi_table/base.py
@@ -433,8 +433,6 @@ class BaseMultiTableSynthesizer:
         """Load a custom constraint class for the specified table's synthesizer.
 
         Args:
-            table_name (str):
-                Table to add constraint to.
             filepath (str):
                 String representing the absolute or relative path to the python file where
                 the custom constraints are declared.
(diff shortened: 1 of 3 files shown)


sdv/multi_table/base.py
```diff
@@ -429,33 +429,32 @@ def get_constraints(self):
 
         return constraints
 
-    def load_custom_constraint_classes(self, table_name, filepath, class_names):
+    def load_custom_constraint_classes(self, filepath, class_names):
         """Load a custom constraint class for the specified table's synthesizer.
```
frances-h, line 433: nitpick: should be changed to something like "Load a custom constraint class for each table's synthesizer"

sdv/multi_table/base.py
```diff
@@ -429,33 +429,32 @@ def get_constraints(self):
...
 
         Args:
-            table_name (str):
-                Table to add constraint to.
             class_object (sdv.constraints.Constraint):
                 A custom constraint class object.
             class_name (str):
                 The name to assign this custom constraint class. This will be the name to use
                 when writing a constraint dictionary for ``add_constraints``.
+        #self._table_synthesizers[table_name].add_custom_constraint_class(class_object, class_name)
```
frances-h, line 454: delete this line

tests/integration/multi_table/test_hma.py
```diff
@@ -178,6 +181,43 @@ def get_custom_constraint_data_and_metadata():
 
 def test_hma_custom_constraint():
```
frances-h, line 182: What's the difference between `test_hma_custom_constraint` and `test_hma_custom_constraint_1_table`?

frances-h (approved):
Just a couple of small changes, otherwise looks good!

sdv/multi_table/base.py
```diff
@@ -429,33 +429,32 @@ def get_constraints(self):
...
-            table_name (str):
-                Table to add constraint to.
             filepath (str):
                 String representing the absolute or relative path to the python file where
                 the custom constraints are declared.
             class_names (list):
                 A list of custom constraint classes to be imported.
         """
-        self._table_synthesizers[table_name].load_custom_constraint_classes(filepath, class_names)
+        for t_name in self._table_synthesizers:
```
amontanez24, line 442: minor: you can loop through the synthesizers directly
```python
for synthesizer in self._table_synthesizers.values():
    ...
```
Also if not too long, can we use full length names (ie. `table_name` instead of `t_name`)

sdv/multi_table/base.py
```diff
@@ -429,33 +429,32 @@ def get_constraints(self):
...
-                Table to add constraint to.
             class_object (sdv.constraints.Constraint):
                 A custom constraint class object.
             class_name (str):
                 The name to assign this custom constraint class. This will be the name to use
                 when writing a constraint dictionary for ``add_constraints``.
+        #self._table_synthesizers[table_name].add_custom_constraint_class(class_object, class_name)
         """
-        self._table_synthesizers[table_name].add_custom_constraint_class(class_object, class_name)
+        for t_name in self._table_synthesizers:
```
amontanez24, line 456: same

tests/integration/multi_table/test_hma.py
```diff
@@ -178,6 +181,43 @@ def get_custom_constraint_data_and_metadata():
...
+
+    # Run - sample
+    sampled = synthesizer.sample(10)
+    assert all(sampled['parent']['numerical_col'] > 1)
+
+
+def test_hma_custom_constraint_1_table():
+    """Test an example of using a custom constraint.
+
+    Here the parent and child table both contains the numerical column.
```
amontanez24, line 217: minor: contain instead of contains

tests/integration/multi_table/test_hma.py
amontanez24: I think he's testing that if the child table has the column name in the constraint, it doesn't get affected somehow by the parent table adding the constraint. Might be unnecessary to check

amontanez24 (approved):
:shipit:

PR #1373, opened 2023-04-14
fix warning

### LINKED ISSUE #1355: Pandas `FutureWarning`: Default dtype for Empty Series will be `'object'`
### Problem Description

Running the `sdv/tests` with `pytests` raised the following `pandas` warning:

`FutureWarning: The default dtype for empty Series will be 'object' instead of 'float64' in a future version. Specify a dtype explicitly to silence this warning.`

### Expected behavior

We want to modify the code to no longer get this warning.

### Additional context

Here is given the code lines in `sdv` that raised the warning. Because this list may be incomplete, it would be nice to check after the correction that the warning is not in the `warning summary` of `pytests`.

#### Lines to be modified:

```
/tests/unit/single_table/test_base.py:839
/tests/unit/single_table/test_base.py:902
/tests/unit/constraints/test_base.py:891
/tests/unit/constraints/test_base.py:942
/tests/unit/single_table/test_base.py:865
/tests/unit/single_table/test_base.py:930
```

R-Palazzo:
Resolve #1355

### DIFF
--- a/tests/unit/single_table/test_base.py
+++ b/tests/unit/single_table/test_base.py
@@ -880,7 +880,6 @@ class TestBaseSingleTableSynthesizer:
         })
         instance = Mock()
         instance._random_state_set = False
-        instance._data_processor._dtypes = pd.Series
         instance._data_processor.filter_valid.return_value = data
 
         # Run
@@ -906,7 +905,6 @@ class TestBaseSingleTableSynthesizer:
             'salary': [90.0, 100.0, 80.0]
         })
         instance = Mock()
-        instance._data_processor._dtypes = pd.Series
         instance._filter_conditions.return_value = data[data.name == 'John Doe']
         conditions = {'salary': 80.}
         transformed_conditions = {'salary': 80.0}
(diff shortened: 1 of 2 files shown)


tests/unit/single_table/test_base.py
```diff
@@ -880,7 +880,7 @@ def test__sample_rows_without_conditions(self):
         })
         instance = Mock()
         instance._random_state_set = False
-        instance._data_processor._dtypes = pd.Series()
+        instance._data_processor._dtypes = pd.Series
```
amontanez24, line 883: pd.Series is a function right? I don't think this makes sense. We should set the dtypes to a series of actual dtype values. If the values don't actually matter for the test then we can set to an empty series of dtype=object

R-Palazzo: Hi @amontanez24. Actually I removed those lines because they were useless in the tests. All is working know and I no longer see the warning in any tests. Mainly, that was the lines in `tests/unit/constraint/test_base.py` that were raising 100 warnings each. [50afae0](https://github.com/sdv-dev/SDV/pull/1373/commits/50afae002e2b55fc69fc045a29fdb24ced2ca04c)

frances-h (approved):
looks good 👍

amontanez24 (approved):
:shipit:

PR #1374, opened 2023-04-14
Pandas FutureWarning: Length 1 tuple will be returned

### LINKED ISSUE #1356: Pandas warning 2
### Problem Description

Running the `sdv/tests` with `pytests` raised the following `pandas` warning:

`FutureWarning: In a future version of pandas, a length 1 tuple will be returned when iterating over a groupby with a grouper equal to a list of length 1. Don't supply a list with a single grouper to avoid this warning.`

### Expected behavior

We want to modify the code to no longer get this warning.

### Additional context

Here is given the code lines in `sdv` that raised the warning. Because this list may be incomplete, it would be nice to check after the correction that the warning is not in the `warning summary` of `pytests`.

#### Lines to be modified:

```
/sdv/constraints/base.py:502
/sdv/single_table/base.py:889
/sdv/sequential/par.py:138
/sdv/single_table/base.py:927
```

R-Palazzo:
Resolve #1356

### DIFF
--- a/sdv/single_table/base.py
+++ b/sdv/single_table/base.py
@@ -21,9 +21,10 @@ from copulas.multivariate import GaussianMultivariate
 from sdv.data_processing.data_processor import DataProcessor
 from sdv.errors import ConstraintsNotMetError, SynthesizerInputError
 from sdv.single_table.errors import InvalidDataError
-from sdv.single_table.utils import check_num_rows, handle_sampling_error, validate_file_path
+from sdv.single_table.utils import (check_num_rows, handle_sampling_error,
+                                    validate_file_path)
 from sdv.utils import (
-    is_boolean_type, is_datetime_type, is_numerical_type, validate_datetime_format)
+    is_boolean_type, is_datetime_type, is_numerical_type, validate_datetime_format, groupby_list)
 
 LOGGER = logging.getLogger(__name__)
 COND_IDX = str(uuid.uuid4())
@@ -893,8 +894,7 @@ class BaseSingleTableSynthesizer(BaseSynthesizer):
         condition_columns = list(conditions.columns)
         conditions.index.name = COND_IDX
         conditions = conditions.reset_index()
-        cond_columns = condition_columns[0] if len(condition_columns) == 1 else condition_columns
-        grouped_conditions = conditions.groupby(cond_columns)
+        grouped_conditions = conditions.groupby(groupby_list(condition_columns))
 
         # sample
         all_sampled_rows = []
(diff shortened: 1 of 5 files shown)


fealho (commented):
Instead of repeating the same logic 4 times, I think it would be better to just create a function in `sdv/utils.py`. Then you can call that function on the code, it should make things more readable that way.

Also, did you check that the warning doesn't show up anymore? I don't think you need to write any tests, just run locally to make sure the warning doesn't happen anymore.

R-Palazzo:
Hi @fealho, thanks for your review. I created a method `sdv/utils.py` to fix this warning and checked that is no longer happening in any tests. [70e6018](https://github.com/sdv-dev/SDV/pull/1374/commits/70e6018e1c9f19e80f158659749ab39109c57a62)

fealho (approved):

sdv/single_table/base.py
```diff
@@ -936,12 +936,9 @@ def _sample_with_conditions(self, conditions, max_tries_per_batch, batch_size,
...
                 all_sampled_rows.append(sampled_rows)
             else:
-                if len(transformed_columns) == 1:
-                    transformed_column = transformed_columns[0]
-                else:
-                    transformed_column = transformed_columns
-
-                transformed_groups = transformed_conditions.groupby(transformed_column)
+                transformed_groups = transformed_conditions.groupby(
+                                                                groupby_list(transformed_columns)
```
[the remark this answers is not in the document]
R-Palazzo: Yes thank you. I changed this in [1a29185](https://github.com/sdv-dev/SDV/pull/1374/commits/1a29185f7788decc3550865fcbd89021045862ea).

sdv/utils.py
```diff
@@ -164,3 +164,8 @@ def load_data_from_csv(filepath, pandas_kwargs=None):
     pandas_kwargs = pandas_kwargs or {}
     data = pd.read_csv(filepath, **pandas_kwargs)
     return data
+
+
+def groupby_list(list_to_check):
+    """Return the list element if the length is 1 and the list otherwise."""
```
pvk-developer, line 170: I would use something more descriptive as -> Return the first element of the list if the length is 1 else the entire list.

pvk-developer (changes requested):
I know this test is a little bit challenging, but could we have a unit test or something to ensure that the warning is not being raised anymore?

sdv/utils.py
R-Palazzo: Yes good idea, done in [41690bd](https://github.com/sdv-dev/SDV/pull/1374/commits/41690bd97a50e0ba17921f6293d4385da17d96a8).

pvk-developer (approved):

PR #1377, opened 2023-04-14
Fix inequality constraint raises RuntimeWarning

### LINKED ISSUE #1275: Inequality constraint raises `RuntimeWarning` (invalid value encountered in log)
### Environment Details

* SDV version: 1.0.0b0 (Beta version)
* Python version: 3.8

### Error Description
When I apply an `Inequality` constraint to the demo data, I see a `RuntimeWarning` when fitting my data. 

```
/usr/local/lib/python3.8/dist-packages/sdv/constraints/tabular.py:451: RuntimeWarning:

invalid value encountered in log
```

This doesn't seem to affect the data quality so I'm ignoring it. But we should check to see if it's indicative of any unintended behavior.

### Steps to reproduce

```python

real_data, metadata = download_demo(
    modality='single_table',
)

checkin_lessthan_checkout = {
    'constraint_parameters': {
        'low_column_name': 'checkin_date',
        'high_column_name': 'checkout_date'
    }
}

synthesizer.add_constraints([
    checkin_lessthan_checkout
])
synthesizer.fit(real_data)
```

### Context
The following line is causing the error in `constraints/tabular.py`:

```python
table_data[self._diff_column_name] = np.log(diff_column + 1)
```

It seems like `np.log` throws this warning if any value is 0 or negative. But this is odd: All values should be `1` or more because:
- We are ensuring that the constraint is _True_ in the real data (so each value in `diff_column` must be >=0 )
- We are adding 1 to each item, so each value must be >= 1

frances-h:
Resolve #1275

### DIFF
--- a/sdv/constraints/tabular.py
+++ b/sdv/constraints/tabular.py
@@ -455,6 +455,32 @@ class Inequality(Constraint):
         valid = np.isnan(low) | np.isnan(high) | self._operator(high, low)
         return valid
 
+    def _get_datetime_diff(self, low, high):
+        """Calculate the difference between two datetime columns.
+
+        When casting datetimes to float using ``astype``, NaT values are not automatically
+        converted to NaN values. This method calculates the difference between the high
+        and low column values, preserving missing values as NaNs. 
+
+        Args:
+            high (numpy.ndarray):
+                The high column values.
+            low (numpy.ndarray):
+                The low column values.
+
+        Returns:
+            numpy.ndarray:
+                The difference between the high and low column values.
+        """
+        if self._dtype == 'O':
+            low = cast_to_datetime64(low)
+            high = cast_to_datetime64(high)
+        diff_column = high - low
+        nan_mask = np.isnan(diff_column)
+        diff_column = diff_column.astype(np.float64)
+        diff_column[nan_mask] = np.nan
+        return diff_column
+    
     def _transform(self, table_data):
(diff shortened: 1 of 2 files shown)


sdv/constraints/tabular.py
```diff
@@ -479,7 +479,9 @@ def _transform(self, table_data):
 
         diff_column = high - low
         if self._is_datetime:
+            nat_mask = np.isnan(diff_column)
```
fealho, line 482: I think you should just call it `nan_mask`, I don't think it's obvious that `nat` stands for `NaT`.

fealho (approved):
Good catch. Do you have any clues why `astype` casts `NaT` to minus infinite? That's such a weird design choice.

amontanez24 (approved):
This looks good! Minor proposal: could we pull all logic for datetime into a separate method and in the docstring for that method explain why we have to add this mask and do these conversions? something like 
```python
def _get_datetime_diff(self, high, low):
    if self._dtype == 'O':
        low = cast_to_datetime64(low)
        high = cast_to_datetime64(high)
    diff_column = high - low
    nan_mask = np.isnan(diff_column)
    diff_column = diff_column.astype(np.float64)
    diff_column[nan_mask] = np.nan

...
if self._is_datetime:
    diff_column = self._get_datetime_diff(high, low)

PR #1383, opened 2023-04-19
Set default locale for synthesizers

### LINKED ISSUE #1371: Setting a locale for all my anonymized (PII) columns
### Problem Description
Sometimes, I have a dataset with many sensitive columns such as `address`, `phone_number`, etc. Usually, the entire dataset comes from a specific region of users so I want to set the locales for all of these values. The SDV 1.0 provides this functionality, but it is cumbersome. 

**Current Functionality**: Currently I can set the locales individually on each of the transformer objects using the [Anonymization Settings](https://colab.research.google.com/drive/1b0W9jpgOHFrkz-i1QynVuaMZXs7BSPb_?usp=sharing#scrollTo=A7I4ZYbT6tI5)

```python
from rdt.transformers.pii import AnonymizedFaker

synth = GaussianCopulaSynthesizer(my_metadata)
synth.auto_assign_transformers(my_data)

# update all PII columns to using the desired locale, such as 'en_CA'
synth.update_transformers(column_name_to_transformer={
    'address': AnonymizedFaker(provider_name='address', function_name='address', locales=['en_CA'],
    'phone_number': AnonymizedFaker(provider_name='phone_number', function_name='phone_number', locales=['en_CA'])
})

synth.fit(data)
synthetic_data = synth.sample(num_rows=100)
```

### Expected behavior
All synthesizers (single table, multi table and sequential) should accept a global `locales` parameter during initialization. This should set the locales for _all_ the relevant column transformers at once.

```python

synth = GaussianCopulaSynthesizer(my_metadata, locales=['en_CA'])
```

Note: if needed, it should still be possible to update the locales on individual columns by using `auto_assign_transformers` and `update_transformers`, as shown above.

### Additional context
Internally, we should make sure that any time we assign an `AnonymizedFaker`, we pass in the `locales` specified in the synthesizer parameters.

If no locales are specified, Faker defaults to `en_US`. So this change does not require backwards compatibility.

frances-h:
Resolve #1371

### DIFF
--- a/sdv/metadata/anonymization.py
+++ b/sdv/metadata/anonymization.py
@@ -3,6 +3,7 @@
 import inspect
 
 from faker import Faker
+from faker.config import AVAILABLE_LOCALES
 from rdt.transformers import AnonymizedFaker
 
 SDTYPE_ANONYMIZERS = {
@@ -56,7 +57,7 @@ def is_faker_function(function_name):
         True if the ``function_name`` is know to ``Faker``, otherwise False.
     """
     try:
-        getattr(Faker(), function_name)
+        getattr(Faker(AVAILABLE_LOCALES), function_name)
     except AttributeError:
         return False
 
--- a/sdv/metadata/anonymization.py
+++ b/sdv/metadata/anonymization.py
@@ -63,8 +63,8 @@ def is_faker_function(function_name):
     return True
 
 
-def _detect_provider_name(function_name):
-    function_name = getattr(Faker(), function_name)
+def _detect_provider_name(function_name, locales=None):
+    function_name = getattr(Faker(locale=locales), function_name)
     module = inspect.getmodule(function_name).__name__
     module = module.split('.')
     if len(module) == 2:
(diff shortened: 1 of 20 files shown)


sdv/metadata/anonymization.py
```diff
@@ -74,26 +75,27 @@ def _detect_provider_name(function_name):
...
         function_kwargs.update(SDTYPE_ANONYMIZERS[function_name])
-        return AnonymizedFaker(**function_kwargs)
+        return AnonymizedFaker(locales=locales, **function_kwargs)
 
-    provider_name = _detect_provider_name(function_name)
+    provider_name = _detect_provider_name(function_name, locales=locales)
     function_kwargs.update({
         'function_name': function_name,
         'provider_name': provider_name
     })
```
pvk-developer, line 99: Overall we can just add `locales` to `function_kwargs` and pass it down the line:
```
function_kwargs.update({
    'locales': locales,
    ...
})
```

pvk-developer (approved):
Left a minor comment but is up to you which approach to take. 
LGTM! :shipit:

sdv/metadata/anonymization.py
frances-h: Looking at it again, I guess I'm a little confused on what `function_kwargs` is actually intended to be. `AnonymizedFaker` takes an argument called `function_kwargs`, which are keyword arguments for the Faker function itself. [This test](https://github.com/sdv-dev/SDV/blob/7194baa81668dc6f56dd0fd1627377b392be9310/tests/unit/metadata/test_anonymization.py#L31) seems to indicate that was the intended purpose of the argument here as well, but in practice we're using it for keyword arguments for `AnonymizedFaker` itself. Actually running this test without mocking `AnonymizedFaker` errors, since it doesn't accept the `domain` keyword argument.

frances-h: I think renaming the variable and using it for init kwargs should help, but I'm not sure if we want to add/fix the ability to pass keyword arguments directly to the Faker function through the `function_kwargs` param in `AnonymizedFaker`

sdv/metadata/anonymization.py
```diff
@@ -56,15 +57,15 @@ def is_faker_function(function_name):
         True if the ``function_name`` is know to ``Faker``, otherwise False.
     """
     try:
-        getattr(Faker(), function_name)
+        getattr(Faker(AVAILABLE_LOCALES), function_name)
```
amontanez24, line 60: What does this do?

amontanez24 (approved):
Just one question, but otherwise looks good!

sdv/metadata/anonymization.py
frances-h: Since some Faker functions are only available in specific locales, this will let us resolve functions from any of the available locales. Previously, if someone tried to use a function not available in en_US, this would error. Since the metadata doesn't have access to the synthesizer locale we want to use, we just have to check all of them to see if it's a valid function.

amontanez24: Can we add a test for such an example and then :shipit:

PR #1399, opened 2023-04-28
Inequality constraint generate all NaNs possibilities

### LINKED ISSUE #1392: `Inequality` constraint does not produce all possibilities of missing values
* SDV version: 1.0.1

### Error Description
In my dataset, I have an `Inequality` constraint between columns `A` and `B`. Both columns may contain nulls in any combinations.
image

The constraint is supposed to hold true whenever both columns are not null.

**Expected**: The synthetic data should create all the different combinations of nulls (eg. both are null, both are not null, only A is null, only B is null).

**Observed**: I never see synthetic data where only A is null.

### Fix
During the forward transform we should:
1. Replace a column with the computed diffs
2. _Create a new categorical column_ that stores which values are missing
3. Replace any missing values in the original columns with the average

image

During the reverse transform, we can use the categorical column to properly assign missing values, ensuring that we create all scenarios.

### Steps to reproduce

```python
import pandas as pd
import numpy as np

data = pd.DataFrame(data={
  'A': [0, 1, np.nan, np.nan], # , 2, 3, 2, 4, 5],
  'B': [2, np.nan, 3, np.nan]# , 5, 5, 3, 6, 6]
})

    'columns': {
        'A': { 'sdtype': 'numerical' },
        'B': { 'sdtype': 'numerical'},
    }
})

synthesizer.add_constraints([{
    'constraint_parameters': {
        'low_column_name': 'A',
        'high_column_name': 'B',
    }
}])

synthesizer.fit(data)
synthetic_data = synthesizer.sample(10000)

# there are no cases where A is null but B is present
synthetic_data[(pd.isna(synthetic_data['A'])) & ~(pd.isna(synthetic_data['B']))]
```

R-Palazzo:
Resolve #1392

### DIFF
--- a/sdv/constraints/tabular.py
+++ b/sdv/constraints/tabular.py
@@ -405,6 +405,7 @@ class Inequality(Constraint):
         self.constraint_columns = (low_column_name, high_column_name)
         self._dtype = None
         self._is_datetime = None
+        self.nan_column_name = None
 
     def _get_data(self, table_data):
         low = table_data[self._low_column_name].to_numpy()
@@ -508,10 +509,10 @@ class Inequality(Constraint):
 
         table_data[self._diff_column_name] = np.log(diff_column + 1)
 
-        nan_dict = compute_nans_column(table_data, [self._low_column_name, self._high_column_name])
-        if nan_dict:
-            [(nan_column_name, nan_column)] = nan_dict.items()
-            table_data[nan_column_name] = nan_column
+        nan_col = compute_nans_column(table_data, [self._low_column_name, self._high_column_name])
+        if nan_col is not None:
+            self.nan_column_name = nan_col.name
+            table_data[self.nan_column_name] = nan_col
             if self._is_datetime:
                 mean_value_low = table_data[self._low_column_name].mode()[0]
             else:
@@ -552,9 +553,8 @@ class Inequality(Constraint):
 
         table_data[self._high_column_name] = pd.Series(diff_column + low).astype(self._dtype)
 
-        nan_col_name = f'{self._diff_column_name}.nan_component'
-        if nan_col_name in table_data.columns:
-            table_data = revert_nans_columns(table_data, nan_col_name)
+        if self.nan_column_name and self.nan_column_name in table_data.columns:
+            table_data = revert_nans_columns(table_data, self.nan_column_name)
 
         return table_data.drop(self._diff_column_name, axis=1)
 
--- a/sdv/constraints/tabular.py
+++ b/sdv/constraints/tabular.py
@@ -39,7 +39,9 @@ import pandas as pd
 from sdv.constraints.base import Constraint
 from sdv.constraints.errors import (
     AggregateConstraintsError, ConstraintMetadataError, FunctionError, InvalidFunctionError)
-from sdv.constraints.utils import cast_to_datetime64, logit, matches_datetime_format, sigmoid
+from sdv.constraints.utils import (
+    add_nans_column, cast_to_datetime64, logit, matches_datetime_format, revert_nans_columns,
+    sigmoid)
 from sdv.utils import convert_to_timedelta, is_datetime_type
 
 INEQUALITY_TO_OPERATION = {
@@ -505,6 +507,17 @@ class Inequality(Constraint):
             diff_column = high - low
 
         table_data[self._diff_column_name] = np.log(diff_column + 1)
+
+        if add_nans_column(table_data, [self._low_column_name, self._high_column_name]):
+            if self._is_datetime:
+                mean_value_low = table_data[self._low_column_name].mode()[0]
+            else:
+                mean_value_low = table_data[self._low_column_name].mean()
+            table_data = table_data.fillna({
+                self._low_column_name: mean_value_low,
+                self._diff_column_name: table_data[self._diff_column_name].mean()
+            })
+
         return table_data.drop(self._high_column_name, axis=1)
 
     def _reverse_transform(self, table_data):
@@ -535,6 +548,11 @@ class Inequality(Constraint):
             low = cast_to_datetime64(low)
 
         table_data[self._high_column_name] = pd.Series(diff_column + low).astype(self._dtype)
+
+        nan_col_name = '#'.join([self._low_column_name, self._high_column_name]) + '.nan_component'
+        if nan_col_name in table_data.columns:
+            table_data = revert_nans_columns(table_data, nan_col_name)
+
         return table_data.drop(self._diff_column_name, axis=1)
 
 
--- a/sdv/constraints/tabular.py
+++ b/sdv/constraints/tabular.py
@@ -40,7 +40,7 @@ from sdv.constraints.base import Constraint
 from sdv.constraints.errors import (
     AggregateConstraintsError, ConstraintMetadataError, FunctionError, InvalidFunctionError)
 from sdv.constraints.utils import (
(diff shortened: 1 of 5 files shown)


sdv/constraints/utils.py
```diff
@@ -104,3 +104,60 @@ def sigmoid(data, low, high):
     data = data * (high - low) + low
 
     return data
+
+
+def check_nans_row(row):
```
R-Palazzo, line 109: I put the following functions in `utils.py` if we want to re-use them for other constraints (ex: `RangeConstraint`)

frances-h (approved):
Left a small style nitpick but otherwise I think it looks good 👍

sdv/constraints/utils.py
```diff
@@ -104,3 +104,60 @@ def sigmoid(data, low, high):
...
+    Args:
+        row (pandas.Series):
+            A pandas row.
+    """
+    columns_with_nans = ''
+    for column, value in row.items():
+        if pd.isna(value):
+            columns_with_nans += f'{column}, '
+
+    columns_with_nans = columns_with_nans.rstrip(', ')
```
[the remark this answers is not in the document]
R-Palazzo: Yes I agree, that's better this way, thanks. Done in [17e0ac2](https://github.com/sdv-dev/SDV/pull/1399/commits/17e0ac2e7bcc0b4e79285fac48de561a9a3c96b6).

sdv/constraints/utils.py
```diff
@@ -104,3 +104,59 @@ def sigmoid(data, low, high):
     data = data * (high - low) + low
 
     return data
+
+
+def check_nans_row(row):
+    """Check for NaNs in a pandas row.
+
+    Outputs a concatenated string of the column names with NaNs.
```
amontanez24, line 112: I think we should add a return section to the docstring

sdv/constraints/tabular.py
```diff
@@ -535,6 +548,11 @@ def _reverse_transform(self, table_data):
             low = cast_to_datetime64(low)
 
         table_data[self._high_column_name] = pd.Series(diff_column + low).astype(self._dtype)
+
+        nan_col_name = '#'.join([self._low_column_name, self._high_column_name]) + '.nan_component'
```
amontanez24, line 552: Can we just name this `f'{self._diff_column_name}.nan_component`?

sdv/constraints/utils.py
```diff
@@ -104,3 +104,59 @@ def sigmoid(data, low, high):
     data = data * (high - low) + low
 
     return data
+
+
+def check_nans_row(row):
```
amontanez24, line 109: minor: can we rename this to `get_nan_component_value`, since this is getting the value to go in the `nan_component` column?

sdv/constraints/utils.py
```diff
@@ -104,3 +104,59 @@ def sigmoid(data, low, high):
...
+        list_column_names (list):
+            The list of column names to check for NaNs.
+    """
+    nan_column_name = '#'.join(list_column_names) + '.nan_component'
+    nan_column = table_data[list_column_names].apply(check_nans_row, axis=1)
+    if not (nan_column == 'None').all():
+        table_data[nan_column_name] = nan_column
+        return True
+    else:
+        return False
```
amontanez24, line 144: In general I think functions should try to limit themselves to doing one thing. Here we are adding the nan column to the data only in certain cases, and return a boolean indicating whether or not we added it. Instead, we can have this function just return the column and if the nan_column is all `None`, we can just return `None`. Then the caller of this method knows to either add the column if it exists, or skip it if it's none.

sdv/constraints/utils.py
```diff
@@ -104,3 +104,59 @@ def sigmoid(data, low, high):
...
+    """
+    columns_with_nans = []
+    for column, value in row.items():
+        if pd.isna(value):
+            columns_with_nans.append(column)
+
+    if columns_with_nans:
+        return ', '.join(columns_with_nans)
+    else:
+        return 'None'
```
amontanez24, line 126: Can we add a test case for this? As in add a unit test that has no nans and make sure everything behaves correctly there. We should also test that `transform` does not add the `.nan_component` column in this case

amontanez24 (changes requested):
This is looking good!

R-Palazzo:
Thanks for your reviews. All the comments have been addressed in [7b2b1be](https://github.com/sdv-dev/SDV/pull/1399/commits/7b2b1be1779a7f5ac5daf523812fdd772e114bfa)

sdv/constraints/utils.py
```diff
@@ -126,22 +129,26 @@ def check_nans_row(row):
...
         table_data (pandas.DataFrame):
             The table data.
         list_column_names (list):
             The list of column names to check for NaNs.
+
+    Returns:
+        A dict with the column name as key and the column indicating where NaNs are as value.
+        Empty dict if there are no NaNs.
     """
     nan_column_name = '#'.join(list_column_names) + '.nan_component'
```
amontanez24, line 145: can we pass the name here so we know it is the same instead of recomputing it?

sdv/constraints/utils.py
```diff
@@ -126,22 +129,26 @@ def check_nans_row(row):
...
-    nan_column = table_data[list_column_names].apply(check_nans_row, axis=1)
-    if not (nan_column == 'None').all():
-        table_data[nan_column_name] = nan_column
-        return True
-    else:
-        return False
+    nan_column_dict = {}
+    column = table_data[list_column_names].apply(get_nan_component_value, axis=1)
+    if not (column == 'None').all():
+        nan_column_dict[nan_column_name] = column
```
amontanez24, line 149: why not just return a series? So if all the values are 'None' return `None`, else returns a series with the proper name and values

amontanez24 (commented):
A couple of small comments but this is almost ready!

sdv/constraints/utils.py
R-Palazzo: Yes, actually I did this first but without returning None when there is no nans, that's why. It's better now, done in [8d39f6a](https://github.com/sdv-dev/SDV/pull/1399/commits/8d39f6a89cf3ba2480e3640d8b87131cd658e7a9)

sdv/constraints/utils.py
R-Palazzo: Yes I pass the name as an attribute of the class, so we no longer need to recompute it. [8d39f6a](https://github.com/sdv-dev/SDV/pull/1399/commits/8d39f6a89cf3ba2480e3640d8b87131cd658e7a9)

sdv/constraints/tabular.py
```diff
@@ -405,6 +405,7 @@ def __init__(self, low_column_name, high_column_name, strict_boundaries=False):
         self.constraint_columns = (low_column_name, high_column_name)
         self._dtype = None
         self._is_datetime = None
+        self.nan_column_name = None
```
amontanez24, line 408: minor: can we make this private ie. (`_nan_column_name`)

amontanez24 (approved):
I think this looks good! Thanks for the changes. I just left one small comment but after that :shipit:

sdv/constraints/tabular.py
R-Palazzo: Yes done in [5c8aa02](https://github.com/sdv-dev/SDV/pull/1399/commits/5c8aa0255bf6e7bbc870ce905c75c43a4e6cd507)

PR #1401, opened 2023-04-28
Silence deprecated locale warning

### LINKED ISSUE #1400: Deprecated locale warning
* SDV version: 1.1.0 (in development)

### Error Description
The SDV version in development is showing a warning every time the metadata is being validated.

```
/usr/local/lib/python3.10/dist-packages/faker/providers/person/fr_QC/__init__.py:10: UserWarning: fr_QC locale is deprecated. Please use fr_CA.
  warnings.warn("fr_QC locale is deprecated. Please use fr_CA.")
```

This is happening because we recently merged changes to detect [Faker functions in different locales](https://github.com/sdv-dev/SDV/blob/9bf8527c65d95db72f19f7f6ebb65972733525d1/sdv/metadata/anonymization.py#L49) for validation purposes. We now iterate through all the locales to check validity -- however, some locales are deprecated so it raises a warning.

The warning is especially confusing for the user, because I have never explicitly put `fr_QC` anywhere in my metadata.

### Proposed Fixes

1. We can just catch the warning in the function, or
2. We can explicitly exclude the `fr_QC` locale since we know it causes a problem

frances-h:
Resolve #1400

### DIFF
--- a/sdv/metadata/anonymization.py
+++ b/sdv/metadata/anonymization.py
@@ -58,9 +58,9 @@ def is_faker_function(function_name):
         True if the ``function_name`` is know to ``Faker``, otherwise False.
     """
     try:
-        warnings.filterwarnings('ignore', module='faker')
-        getattr(Faker(AVAILABLE_LOCALES), function_name)
-        warnings.resetwarnings()
+        with warnings.catch_warnings():
+            warnings.filterwarnings('ignore', module='faker')
+            getattr(Faker(AVAILABLE_LOCALES), function_name)
     except AttributeError:
         return False
 
--- a/sdv/metadata/anonymization.py
+++ b/sdv/metadata/anonymization.py
@@ -1,6 +1,7 @@
 """Anonymization module for the ``DataProcessor``."""
 
 import inspect
+import warnings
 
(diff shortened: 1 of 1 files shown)


sdv/metadata/anonymization.py
```diff
@@ -57,7 +58,9 @@ def is_faker_function(function_name):
         True if the ``function_name`` is know to ``Faker``, otherwise False.
     """
     try:
+        warnings.filterwarnings('ignore', module='faker')
         getattr(Faker(AVAILABLE_LOCALES), function_name)
+        warnings.resetwarnings()
```
pvk-developer, line 63: I think that we should use [catch_warnings](https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings) instead.
```python
        with warnings.catch_warnings():
            warnings.filterwarnings('ignore', module='faker')
            getattr(Faker(AVAILABLE_LOCALES), function_name)
```
If an user has configured something else to the warnings this `warnings.resetwarnings()` may reset that configuration.

fealho (approved):

pvk-developer (approved):

PR #1403, opened 2023-05-01
Refactor Multi-Table modeling

### LINKED ISSUE #1387: Refactor Multi Table Modeling
### Problem Description

_As an engineer, it would be nice if there were clear abstractions in multi table modeling that could be used to add new multi table synthesizers._

Screen Shot 2023-04-22 at 4 47 21 PM

The image above lays out the generalized common modeling steps all multi table synthesizers should use. Currently, the `HMASynthesizer` breaks this abstraction as it models each table in the middle box while doing its extending. This can be seen on the following lines of `HMASynthesizer._model_table`.
https://github.com/sdv-dev/SDV/blob/904aa2d0ef987654d1866c933a8b70a6a504d13e/sdv/multi_table/hma.py#L198-L199

### Expected behavior

To match the desired abstraction we should take the following steps:
1. Add an abstract method to the BaseMultiTableSynthesizer class called `_augment_tables`. This method should be implemented in any child class and be responsible for augmenting tables with the necessary information to model them.
2. Add another abstract method called `_model_tables` to the base. This method should just be used to loop through and model tables.
3. Refactor the `HSASynthesizer` to break up modeling into these two methods.
4. `BaseMultiTableSynthesizer.fit` will now call `preprocess` -> `_augment_tables` -> `_model_tables`

### Additional context

- In HMA, we only need to model the root table. Currently we model every table. This change can be made in the new `_model_tables` implementation.

pvk-developer:
Resolves #1387

### DIFF
--- a/sdv/multi_table/base.py
+++ b/sdv/multi_table/base.py
@@ -288,7 +288,7 @@ class BaseMultiTableSynthesizer:
         if self._fitted:
             warnings.warn(
                 'This model has already been fitted. To use the new preprocessed data, '
-                "please refit the model using 'fit'."
+                "please refit the model using 'fit' or 'fit_processed_data'."
             )
 
         processed_data = {}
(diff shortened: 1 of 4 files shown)


amontanez24 (approved):
This looks really good! Great job :shipit:

frances-h (approved):

PR #1408, opened 2023-05-03
Metadata validates twice for multitable

### LINKED ISSUE #1407: Metadata validates twice for multi-table models.
### Environment Details

* SDV version: 1.0.1
* Python version: Any
* Operating System: Any

### Error Description

Currently the `metadata` is being validated each time an instance of a `Synthesizer` occurs. Within the `BaseMultiTable` synthesizer, an instance of `BaseSingleTableSynthesizer` is  created for each table with its own metadata, which leads to multiple validations of the same metadata. Only one validation should occur per metadata.

### Possible solutions
One solution would be to have a inner flag within the metadata object that is set to `True` each time the metadata has been validated:

```python
metadata._is_valid = False
metadata.validate()
metadata._is_valid -> True
```

If the `metadata` has been modified, the `_is_valid` flag will reset to `False`. This flag should not be stored when saving, however it should be regenerated when loaded, (after loading the `metadata` from dict, call `metdata.validate()`).

pvk-developer:
Fix #1407

### DIFF
--- a/sdv/metadata/multi_table.py
+++ b/sdv/metadata/multi_table.py
@@ -24,6 +24,7 @@ class MultiTableMetadata:
     def __init__(self):
         self.tables = {}
         self.relationships = []
+        self._is_valid = False
 
     def _validate_missing_relationship_keys(self, parent_table_name, parent_primary_key,
                                             child_table_name, child_foreign_key):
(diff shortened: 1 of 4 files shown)


pvk-developer:
Partial code has been merged on #1410 may be retaken later.

PR #1409, opened 2023-05-03
Modify is_valid method in Range constraint to consider NaNs

### LINKED ISSUE #1396: Fix validity check for `Range` constraint
* SDV version: 1.0.1 (latest)

### Error Description
A Range constraint is supposed to check for instances where `low < middle < high`. If there are null values, we should be ignoring whatever is null and check the remaining columns.

1. If everything is present, then check `low < middle` and `middle < high` (implies that `low < high` too)
2. Else if `low` is null, then check `middle < high`
3. Else if `middle` is null, then check `low < high`
4. Else if `high` is null, then `low < middle`
5. Else the row is valid

Currently are **not checking case (3)**. This allows for cases where `low > high` to pass and be included in the synthetic data.

### Fix
We should modify the validity check to ensure that `low < high`

### Steps to reproduce

See #1393

R-Palazzo:
Resolve #1396


sdv/constraints/tabular.py
```diff
@@ -892,19 +892,19 @@ def is_valid(self, table_data):
...
+        nan_column = compute_nans_column(table_data, list_columns_nans)
+        validate_column = pd.Series([True]*len(table_data), index=table_data.index)
+        if nan_column is None:
+            validate_column = self._operator(low, middle) * self._operator(middle, high)
+        else:
+            validate_column.loc[nan_column == 'None'] = self._operator(
+                low, middle) * self._operator(middle, high)
+            validate_column.loc[nan_column == self.low_column_name] = self._operator(middle, high)
+            validate_column.loc[nan_column == self.middle_column_name] = self._operator(low, high)
+            validate_column.loc[nan_column == self.high_column_name] = self._operator(low, middle)
```
amontanez24, line 905: I think it might be simpler to do something like 
```python
low_lt_middle = self._operator(low, middle) | np.isnan(low) | np.isnan(middle)
middle_lt_high = self._operator(middle, high) | np.isnan(middle) | np.isnan(high)
low_lt_high = self._operator(low, high) | np.isnan(low) | np.isnan(high)
return low_lt_middle & middle_lt_high & low_lt_high
```
You might have to test which is more performant though

sdv/constraints/tabular.py
```diff
@@ -892,19 +892,11 @@ def is_valid(self, table_data):
...
 
-        satisfy_low_bound = np.logical_or(
-            self._operator(low, middle),
-            np.isnan(low),
-        )
-        satisfy_high_bound = np.logical_or(
-            self._operator(middle, high),
-            np.isnan(high),
-        )
+        low_lt_middle = self._operator(low, middle) | np.isnan(low) | np.isnan(middle)
```
amontanez24, line 895: I think one other small optimization we can make is storing the computation of nans in variables so we only run `np.isnan(column)` once for each column. So like
```python
low_is_nan = np.isnan(low)
middle_is_nan = np.isnan(middle)
low_lt_middle = self._operator(low, middle) | low_is_nan | middle_is_nan
...
```

tests/unit/constraints/test_tabular.py
```diff
@@ -3189,6 +3189,31 @@ def test_is_valid_invalid(self):
         # Assert
         assert not all(result)
 
+    def test_is_valid_with_nans(self):
+        """Test the ``Range.is_valid`` when there is NaNs in the columns."""
```
amontanez24, line 3193: minor: are Nans in the columns

tests/unit/constraints/test_tabular.py
```diff
@@ -3189,6 +3189,31 @@ def test_is_valid_invalid(self):
...
+
+        instance = Range('low', 'middle', 'high')
+
+        # Run
+        result_valid = instance.is_valid(table_data_valid)
+        result_invalid = instance.is_valid(table_data_invalid)
+
+        # Assert
+        assert all(result_valid)
+        assert not all(result_invalid)
```
amontanez24, line 3215: Just to be safe can we directly check that the result_valid and result_invalid are the series we expect? As in compare to the expected outcome instead of jusst checking that any are false

amontanez24 (commented):
Some minor comments but after that should be good!!

fealho (approved):

amontanez24 (approved):
:shipit:

PR #1411, opened 2023-05-08
Add BaseHierarchicalSampler mixin class

### LINKED ISSUE #1394: Add BaseHierarchicalSampler Mixin
### Problem Description

There is not currently a unified abstraction for multi table synthesizers. This has led to diverging implementations for `HMASynthesizer` and `HSASynthesizer`. Additionally it leads to the following issues:

* It is not easy to add new, multi table models: Over the lifetime of DataCebo, we expect to be creating many different multi table synthesizers.
* It is not easy to fix existing bugs: If there is a general bug that affects modeling and sampling for all multi-table synthesizers, we should be able to make the fix in only 1 place.
* It is not easy to add common features between all synthesizers: For example, it may be nice to include some verbosity (progress bars, logger info, etc.) during the fitting and sampling for all synthesizers.

As a solution, we propose creating mixins that unify multi table sampling based on two strategies:
1. Hierarchical
2. Independent

This issue focuses on the Hierarchical strategy.

### Expected behavior

Add a new mixin called `HierarchicalSampler` to the `sdv.sampling` module.

#### Methods

These methods should all have an implementation in the base mixin itself.

* `_sample(self, scale=1.0)`
    * Args
        * scale (float): A float representing how much to scale the data by. 
    * Returns: A dictionary containing as keys the names of the tables and as values the sampled data tables as `pandas.DataFrame`.
    * pseudo-code
    ```python
    sampled_data = {}
    roots = get_tables_with_no_parents()
    for root in roots:
        num_rows = get_num_rows(root, scale)
        synthesizer = self._synthesizers[root]
        self._sample_table(synthesizer, root, num_rows, sampled_data)
    for relationship in relationships:
        self._add_foreign_key_column(child_table, parent_table, child_name, parent_name)
    self._finalize()
    ```
* `_sample_table(self, synthesizer, table_name, num_rows, sampled_data=None)`
    * Args
        * synthesizer (SingleTableSynthesizer): Synthesizer to sample from
        * table_name (string): Name of the table to sample
        * num_rows (int): Number of rows to sample.
        * sampled_data (dict): Dictionary of data sampled so far.
    * pseudo-code
    ```python
    data = synthesizer._sample(num_rows)
    sampled_data[table_name] = data
    for child in self.get_children(table_name):
        for row in data:
            self.add_child_rows(row, child, sampled_data)
    ```
* `_add_child_rows(parent_row, child_name, sampled_data)`
    * Args
        * parent_row (dataframe): Row of parent data to get child rows for
        * child_name(string): Name of child table
        * sampled_data (dict): Dictionary of data sampled so far.
    * pseudo-code
    ```python
        child_synthesizer = self._recreate_child_synthesizer(child_name, parent_rows)
        num_rows = get_num_rows_for_parent()
        self.sample_table(child_synthesizer, child_name, num_rows, sampled_data)
    ```
* `_finalize(sampled_data)`
    * Args
        * sampled_data (dict): Dictionary of data sampled so far.
    * pseudo-code
    ```python
    for table, data in sampled_data.items():
        remove_added_columns(data)
    ```

#### Abstract methods

These methods do not need to be implemented here but any class that uses this mixin must  implement them.

* `_recreate_child_synthesizer(child_name, parent_row)`: This method should create a `SingleTableSynthesizer` for the child based on the values in its parent's row.
* `_add_foreign_key_column(child_table, parent_table, child_name, parent_name)`: This method should add a column for the foreign key that connects the child to the parent. It can use whatever logic it needs to figure out which values to use for each child row.

### Additional context

- This mixin follows the current HMA sampling process pretty closely. The main difference will be breaking the steps into the appropriate methods.

- The goal is that at the end, to add a new hierarchical multi table synthesizer, the only methods needed will be one to link child rows to parent rows, and one to create a child synthesizer from a parent row.

frances-h:
Resolve #1394

### DIFF
--- a/sdv/sampling/hierarchical_sampler.py
+++ b/sdv/sampling/hierarchical_sampler.py
@@ -1,7 +1,6 @@
 """Hierarchical Samplers."""
 import logging
 import math
-from copy import deepcopy
 
 import pandas as pd
 
--- a/sdv/sampling/hierarchical_sampler.py
+++ b/sdv/sampling/hierarchical_sampler.py
@@ -52,28 +52,6 @@ class BaseHierarchicalSampler():
         """
         raise NotImplementedError()
 
-    def _process_samples(self, table_name, sampled_rows):
-        """Process the ``sampled_rows`` for the given ``table_name``.
-
-        Process the raw samples and convert them to the original space by reverse transforming
-        them. Also, when there are synthesizer columns (columns used to recreate an instance
-        of a synthesizer), those will be returned together.
-
-        Args:
-            table_name (str):
-                The name of the table.
-            sampled_rows (pd.DataFrame):
-                The sampled rows from the table to process.
-        """
-        data_processor = self._table_synthesizers[table_name]._data_processor
-        sampled = data_processor.reverse_transform(sampled_rows)
-
-        synthesizer_columns = list(set(sampled_rows.columns) - set(sampled.columns))
-        if synthesizer_columns:
-            sampled = pd.concat([sampled, sampled_rows[synthesizer_columns]], axis=1)
-
-        return sampled
-
     def _sample_rows(self, synthesizer, table_name, num_rows=None):
         """Sample ``num_rows`` from ``synthesizer``.
 
@@ -90,9 +68,7 @@ class BaseHierarchicalSampler():
                 Sampled rows, shape (, num_rows)
         """
         num_rows = num_rows or synthesizer._num_rows
-        sampled_rows = synthesizer.sample(num_rows)
-
-        return self._process_samples(table_name, sampled_rows)
+        return synthesizer._sample_batch(num_rows, remove_extra_columns=False)
 
     def _get_num_rows_from_parent(self, parent_row, child_name, foreign_key):
         """Get the number of rows to sample for the child from the parent row."""
--- a/sdv/sampling/hierarchical_sampler.py
+++ b/sdv/sampling/hierarchical_sampler.py
@@ -22,7 +22,7 @@ class BaseHierarchicalSampler():
             sampler by the multi-table synthesizer using this sampler.
     """
 
-    def __init__(self, metadata, table_synthesizers={}, table_sizes={}):
+    def __init__(self, metadata, table_synthesizers, table_sizes):
         self.metadata = metadata
         self._table_synthesizers = table_synthesizers
         self._table_sizes = table_sizes
@@ -69,6 +69,12 @@ class BaseHierarchicalSampler():
         Process the raw samples and convert them to the original space by reverse transforming
         them. Also, when there are synthesizer columns (columns used to recreate an instance
         of a synthesizer), those will be returned together.
+
+        Args:
+            table_name (str):
+                The name of the table.
+            sampled_rows (pd.DataFrame):
+                The sampled rows from the table to process.
         """
         data_processor = self._table_synthesizers[table_name]._data_processor
         sampled = data_processor.reverse_transform(sampled_rows)
@@ -145,7 +151,7 @@ class BaseHierarchicalSampler():
                 sampled_data[child_name] = pd.concat(
                     [previous, sampled_rows]).reset_index(drop=True)
 
-    def _sample_table(self, synthesizer, table_name, num_rows, sampled_data={}):
+    def _sample_table(self, synthesizer, table_name, num_rows, sampled_data):
         """Sample a single table and all its children.
 
         Args:
--- a/sdv/sampling/hierarchical_sampler.py
+++ b/sdv/sampling/hierarchical_sampler.py
@@ -1,0 +1,217 @@
+"""Hierarchical Samplers."""
+import logging
+import math
+from copy import deepcopy
+
+import pandas as pd
+
+LOGGER = logging.getLogger(__name__)
+
+
+class BaseHierarchicalSampler():
+    """Hierarchical sampler mixin.
+
+    Args:
+        metadata (sdv.metadata.multi_table.MultiTableMetadata):
+            Multi table metadata representing the data tables that this sampler will be used for.
+        table_synthesizers (dict):
+            Dictionary mapping each table to a synthesizer. Should be instantiated and passed to
+            sampler by the multi-table synthesizer using this sampler.
+        table_sizes (dict):
+            Dictionary mapping each table to its size. Should be instantiated and passed to the
+            sampler by the multi-table synthesizer using this sampler.
+    """
+
+    def __init__(self, metadata, table_synthesizers={}, table_sizes={}):
+        self.metadata = metadata
+        self._table_synthesizers = table_synthesizers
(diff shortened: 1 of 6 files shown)


tests/unit/sampling/test_hierarchical_sampler.py
```diff
@@ -0,0 +1,399 @@
...
+        instance = Mock()
+        instance._max_child_rows = {'__sessions__user_id__num_rows': 10}
+
+        # Run
+        result = BaseHierarchicalSampler._get_num_rows_from_parent(
+            instance, parent_row, 'sessions', 'user_id')
+
+        # Assert
+        expected_result = 10.0
+
```
fealho, line 107: delete empty line

tests/unit/sampling/test_hierarchical_sampler.py
```diff
@@ -0,0 +1,399 @@
...
+        expected_result = pd.DataFrame({
+            'session_id': ['a', 'b', 'c'],
+            'os': ['linux', 'mac', 'win'],
+            'country': ['us', 'us', 'es'],
+            'user_id': [1, 2, 3]
+        })
+        pd.testing.assert_frame_equal(sampled_data['sessions'], expected_result)
+
+    def test__add_child_rows_with_sampled_data(self):
+        """Test the adding child rows when sampled data contains values.
```
fealho, line 154: Test adding. No `the`.

fealho (commented):
Looks pretty good! I'll do a second review later today, but for now I only found nitpicks.

sdv/sampling/hierarchical_sampler.py
```diff
@@ -0,0 +1,217 @@
...
+            sampled_rows = synthesizer._sample(num_rows)
+        else:
+            sampled_rows = pd.DataFrame(index=range(num_rows))
+
+        return self._process_samples(table_name, sampled_rows)
+
+    def _get_num_rows_from_parent(self, parent_row, child_name, foreign_key):
+        """Get the number of rows to sample for the child from the parent row."""
+        prefix = f'__{child_name}__{foreign_key}__'
+        num_rows_key = f'{prefix}num_rows'
```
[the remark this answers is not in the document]
frances-h: It's part of the prefix, but I just collapsed them into one declaration since we don't need to keep prefix separate anymore.

sdv/sampling/hierarchical_sampler.py
```diff
@@ -0,0 +1,217 @@
...
+        num_rows = 0
+        if num_rows_key in parent_row.keys():
+            num_rows = parent_row[num_rows_key]
+            num_rows = min(
+                self._max_child_rows[num_rows_key],
+                math.ceil(num_rows)
```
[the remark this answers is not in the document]
frances-h: I pulled this logic from HMA, but my understanding is that we want to round `num_rows` up so that at least 1 row is always created.

sdv/sampling/hierarchical_sampler.py
```diff
@@ -0,0 +1,238 @@
...
+            num_rows (int):
+                Number of rows to sample.
+
+        Returns:
+            pandas.DataFrame:
+                Sampled rows, shape (, num_rows)
+        """
+        num_rows = num_rows or synthesizer._num_rows
+        if synthesizer._model:
+            sampled_rows = synthesizer._sample(num_rows)
```
amontanez24, line 105: I know this is the current implementation, but I'm wondering if constraints are working properly when they require reject sampling. I believe reject sampling is handled in `sample` which we are skipping over here. Maybe we just want to call `sample()`

pvk-developer (approved):

pvk-developer (approved):

pvk-developer (approved):

sdv/sampling/hierarchical_sampler.py
```diff
@@ -101,10 +91,7 @@ def _sample_rows(self, synthesizer, table_name, num_rows=None):
                 Sampled rows, shape (, num_rows)
         """
         num_rows = num_rows or synthesizer._num_rows
-        if synthesizer._model:
-            sampled_rows = synthesizer._sample(num_rows)
-        else:
-            sampled_rows = pd.DataFrame(index=range(num_rows))
+        sampled_rows = synthesizer.sample(num_rows)
 
         return self._process_samples(table_name, sampled_rows)
```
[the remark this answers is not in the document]
pvk-developer: `HMA` shouldn't work if `sample` is called directly. The `synthesizers` should be fitted to the augmented `table`, which contains rows that are not recognized by the `data_processor`.
When we sample we use that data to generate the synthesizer 'state' by using `_extract_parameters` that come from this extra columns, which then are being set to a new instance of a synthesizer.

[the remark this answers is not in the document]
amontanez24: So the issue is that `_sample` doesn't do reject sampling. This means that constraints aren't guaranteed to be met since we sometimes fall back on reject sampling. We need a way to do reject sampling but get the values back here without dropping any extra columns needed to do the algorithm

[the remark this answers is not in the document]
frances-h: Would it generally be the same across the multi-table sampler mixins? Could we make a base sampler mixin class and put it there?

sdv/sampling/hierarchical_sampler.py
```diff
@@ -0,0 +1,210 @@
...
+            synthesizer (SingleTableSynthesizer):
+                Synthesizer to sample from for the table.
+            table_name (string):
+                Name of the table to sample.
+            num_rows (int):
+                Number of rows to sample for the table.
+            sampled_data (dict):
+                A dictionary mapping table names to sampled tables (pd.DataFrame).
+        """
+        LOGGER.info('Sampling %s rows from table %s', num_rows, table_name)
```
amontanez24, line 129: minor: can we switch to f-string

sdv/sampling/hierarchical_sampler.py
```diff
@@ -0,0 +1,210 @@
...
+
+        for relationship in self.metadata.relationships:
+            parent_name = relationship['parent_table_name']
+            child_name = relationship['child_table_name']
+            self._add_foreign_key_columns(
+                sampled_data[child_name],
+                sampled_data[parent_name],
+                child_name,
+                parent_name
+            )
```
amontanez24, line 208: Should we check if the foreign key column already exists?

amontanez24 (commented):
This looks really good! Just a couple minor comments

sdv/sampling/hierarchical_sampler.py
frances-h: I think since we're adding all the foreign key columns in  the `_add_foreign_key_columns` function, it makes sense for it to handle that check. We should make sure we're only calling that function once per relationship though, so I fixed that.

pvk-developer (approved):
LGTM 💯  Great job 🎉

PR #1415, opened 2023-05-09
Update to pandas 2.0

### LINKED ISSUE #1366: Upgrade to `pandas` 2.0
[Pandas 2.0](https://pandas.pydata.org/docs/dev/whatsnew/v2.0.0.html) was released a few days ago.

**Ideal Result:** We should fix any issues in the SDV so that it can work on `pandas 2.0`. After this, we can consider uncapping the max version of pandas.

Note that for this to work, we may have to fix `FutureWarnings` that are currently showing up throughout the SDV ecosystem. See:
#1355, #1356, #1357 (and potentially more)

pvk-developer:
Resolves #1366


sdv/utils.py
```diff
@@ -15,6 +15,17 @@ def cast_to_iterable(value):
     return [value]
 
 
+def get_fisrt_valid_value(values):
+    """Return the first not ``nan`` value or ``nan`` if all are ``nan``."""
+    for item in values:
+        if pd.isna(item):
+            continue
+        elif item != 'nan':
+            return item
```
amontanez24, line 24: minor but you can rewrite as:
```python
for item in values:
    if !pd.isna(item):
        return item

return item

File: sdv/utils.py [hunk 2]
```diff
@@ -15,6 +15,17 @@ def cast_to_iterable(value):
     return [value]
 
 
+def get_fisrt_valid_value(values):
```
amontanez24, line 18: typo: fisrt

sdv/utils.py
```diff
@@ -31,9 +42,14 @@ def get_datetime_format(value):
...
-    if not isinstance(value, (list, np.ndarray)):
-        value = [value]
+        value = value.astype(str).to_numpy()
+    elif isinstance(value, list):
+        value = np.array(value)
+    if not isinstance(value, np.ndarray):
+        value = np.array([value])
+
+    #  Fixes a bug that occurs in pandas as numpy casts to np.str_
+    value = np.array([get_fisrt_valid_value(value)], dtype=object)
```
amontanez24, line 52: If we're only going to use the first value, why go through the trouble of converting all the lists/series above?

amontanez24 (commented):
A couple of minor comments and a question

sdv/utils.py
pvk-developer: The problem comes with parsed strings, I think I even need to increase the below condition with 'nan', 'NaT', 'None' etc.
The pd.isna does not consider those a nan value as they are string representation of them. But I can simplify it a bit and write it in one if.

amontanez24 (approved):
:shipit:

fealho (approved):
👍

PR #1423, opened 2023-05-12
Add BaseIndependentSampler mixin

### LINKED ISSUE #1395: Add BaseIndependentSampler Mixin
### Problem Description

There is not currently a unified abstraction for multi table synthesizers. This has led to diverging implementations for `HMASynthesizer` and `HSASynthesizer`. Additionally it leads to the following issues:

* It is not easy to add new, multi table models: Over the lifetime of DataCebo, we expect to be creating many different multi table synthesizers.
* It is not easy to fix existing bugs: If there is a general bug that affects modeling and sampling for all multi-table synthesizers, we should be able to make the fix in only 1 place.
* It is not easy to add common features between all synthesizers: For example, it may be nice to include some verbosity (progress bars, logger info, etc.) during the fitting and sampling for all synthesizers.

As a solution, we propose creating mixins that unify multi table sampling based on two strategies:
1. Hierarchical
2. Independent

This issue focuses on the Independent strategy.

### Expected behavior

Add a new mixin called `IndependentSampler` to the `sdv.sampling` module.

#### Methods

These methods should all have an implementation in the base mixin itself.

* `_sample(self, scale=1.0)`
    * Args
        * scale (float): A float representing how much to scale the data by. 
    * Returns: A dictionary containing as keys the names of the tables and as values the sampled data tables as `pandas.DataFrame`.
    * pseudo-code
    ```python
    sampled_data = {}
    for table in tables:
        num_rows = get_num_rows(table, scale)
        synthesizer = self._synthesizers[table]
        self._sample_table(synthesizer, root, num_rows, sampled_data)
    self._connect_tables(sampled_data)
    self._finalize(sampled_data)
    ```
* `_sample_table(self, table_name, num_rows, sampled_data=None)`
    * Args
        * synthesizer (SingleTableSynthesizer): Synthesizer to sample from
        * table_name (string): Name of the table to sample
        * num_rows (int): Number of rows to sample.
        * sampled_data (dict): Dictionary of data sampled so far.
    * pseudo-code
    ```python
    data = synthesizer._sample(num_rows)
    data = reverse_transform(data) ### finalize table here
    sampled_data[table_name] = data
    ```
* `_connect_tables(sampled_data)`
    * Args
        * sampled_data (dict): Dictionary of data sampled so far.
    * pseudo-code
    ```python
    queue = get_root_tables()
    while queue:
        parent = queue.pop()
        for child in self.metadata.get_children(parent):
            self._add_foreign_key_column(sampled_data[child], parent, child_name, parent_name)
            if has_all_foreign_keys(child):
                queue.append(child)
    ```
* `_finalize(sampled_data)`
    * Args
        * sampled_data (dict): Dictionary of data sampled so far.
    * pseudo-code
    ```python
    for table, data in sampled_data.items():
        remove_added_columns(data)
    ```
#### Abstract methods

These methods do not need to be implemented here but any class that uses this mixin must  implement them.

* `_add_foreign_key_column(child_table_rows, parent_table_rows, child_name, parent_name)`: This method should add a column for the foreign key that connects the child to the parent. It can use whatever logic it needs to figure out which value to use for that parent row.

### Additional context

- This mixin changes the current sampling strategy that HSA uses in the following ways:
1. Current HSA is depth first (connects children as it samples them). This strategy would sample all tables and then connect.

frances-h:
Resolve #1395

### DIFF
--- a/sdv/sampling/independent_sampler.py
+++ b/sdv/sampling/independent_sampler.py
@@ -51,7 +51,7 @@ class BaseIndependentSampler():
             sampled_data (dict):
                 A dictionary mapping table names to sampled tables (pd.DataFrame).
         """
-        LOGGER.info('Sampling %s rows from table %s', num_rows, table_name)
+        LOGGER.info(f'Sampling {num_rows} rows from table {table_name}')
 
         sampled_rows = synthesizer._sample_batch(num_rows, keep_extra_columns=True)
         sampled_data[table_name] = sampled_rows
@@ -67,7 +67,9 @@ class BaseIndependentSampler():
                 A dictionary mapping table names to the sampled tables (pd.DataFrame).
         """
         queue = [
-            table for table in self.metadata.tables if not self.metadata._get_parent_map()[table]
+            table
+            for table in self.metadata.tables
+            if not self.metadata._get_parent_map()[table]
         ]
         while queue:
             parent = queue.pop(0)
--- a/sdv/sampling/independent_sampler.py
+++ b/sdv/sampling/independent_sampler.py
@@ -1,0 +1,134 @@
+"""Independent Samplers."""
+import logging
+
+LOGGER = logging.getLogger(__name__)
+
+
+class BaseIndependentSampler():
+    """Independent sampler mixin.
+
+    Args:
+        metadata (sdv.metadata.multi_table.MultiTableMetadata):
+            Multi-table metadata representing the data tables that this sampler will be used for.
+        table_synthesizers (dict):
+            Dictionary mapping each table to a synthesizer. Should be instantiated and passed to
+            sampler by the multi-table synthesizer using this sampler.
+        table_sizes (dict):
+            Dictionary mapping each table to its size. Should be instantiated and passed to the
+            sampler by the multi-table synthesizer using this sampler.
+    """
+
+    def __init__(self, metadata, table_synthesizers, table_sizes):
+        self.metadata = metadata
+        self._table_synthesizers = table_synthesizers
+        self._table_sizes = table_sizes
+
+    def _add_foreign_key_columns(self, child_table, parent_table, child_name, parent_name):
+        """Add all the foreign keys that connect the child table to the parent table.
+
+        Args:
+            child_table (pd.DataFrame):
+                The table containing data sampled for the child.
+            parent_table (pd.DataFrame):
+                The table containing data sampled for the parent.
+            parent_name (str):
+                The name of the parent table.
+            child_name (str):
+                The name of the child table.
+        """
+        raise NotImplementedError()
+
+    def _sample_table(self, synthesizer, table_name, num_rows, sampled_data):
+        """Sample a single table and all its children.
+
+        Args:
+            synthesizer (SingleTableSynthesizer):
+                Synthesizer to sample from for the table.
+            table_name (string):
+                Name of the table to sample.
+            num_rows (int):
+                Number of rows to sample for the table.
+            sampled_data (dict):
+                A dictionary mapping table names to sampled tables (pd.DataFrame).
+        """
+        LOGGER.info('Sampling %s rows from table %s', num_rows, table_name)
+
+        sampled_rows = synthesizer._sample_batch(num_rows, keep_extra_columns=True)
+        sampled_data[table_name] = sampled_rows
+
+    def _connect_tables(self, sampled_data):
+        """Connect all related tables.
+
+        This method iterates over all tables and adds all foreign keys to connect
+        related tables.
+
+        Args:
+            sampled_data (dict):
+                A dictionary mapping table names to the sampled tables (pd.DataFrame).
+        """
+        queue = [
+            table for table in self.metadata.tables if not self.metadata._get_parent_map()[table]
+        ]
+        while queue:
+            parent = queue.pop(0)
+            for child in self.metadata._get_child_map()[parent]:
+                self._add_foreign_key_columns(
+                    sampled_data[child], sampled_data[parent], child, parent)
+                if set(self.metadata._get_all_foreign_keys(child)).issubset(
+                        set(sampled_data[child].columns)):
+                    queue.append(child)
+
+    def _finalize(self, sampled_data):
+        """Remove extra columns from sampled tables and apply finishing touches.
+
+        This method reverts the previous transformations to go back
+        to values in the original space.
+
+        Args:
+            sampled_data (dict):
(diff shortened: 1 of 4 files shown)


sdv/sampling/independent_sampler.py
```diff
@@ -0,0 +1,166 @@
...
+        """
+        raise NotImplementedError()
+
+    def _get_all_foreign_keys(self, table_name):
+        foreign_keys = []
+        for relation in self.metadata.relationships:
+            if table_name == relation['child_table_name']:
+                foreign_keys.append(deepcopy(relation['child_foreign_key']))
+
+        return foreign_keys
```
pvk-developer, line 50: Do we need this to be here or in the `Base` ? Is it used for fitting or only for sampling ?

sdv/sampling/independent_sampler.py
```diff
@@ -0,0 +1,166 @@
...
+        data_processor = self._table_synthesizers[table_name]._data_processor
+        sampled = data_processor.reverse_transform(sampled_rows)
+
+        synthesizer_columns = list(set(sampled_rows.columns) - set(sampled.columns))
+        if synthesizer_columns:
+            sampled = pd.concat([sampled, sampled_rows[synthesizer_columns]], axis=1)
+
+        return sampled
+
+    def _sample_table(self, synthesizer, table_name, num_rows, sampled_data={}):
```
pvk-developer, line 68: I think `sampled_data` must be provided to this method (remove the default)

sdv/sampling/independent_sampler.py
```diff
@@ -0,0 +1,166 @@
...
+            Multi-table metadata representing the data tables that this sampler will be used for.
+        table_synthesizers (dict):
+            Dictionary mapping each table to a synthesizer. Should be instantiated and passed to
+            sampler by the multi-table synthesizer using this sampler.
+        table_sizes (dict):
+            Dictionary mapping each table to its size. Should be instantiated and passed to the
+            sampler by the multi-table synthesizer using this sampler.
+    """
+
+    def __init__(self, metadata, table_synthesizers={}, table_sizes={}):
```
pvk-developer, line 24: Overall is better to use `None` as `default` when optional, then do a simple `self._table_synthesizers = table_synthesizers or {}`.
However, I wonder if this are optional here 🤔

pvk-developer (approved):
Really good job, left some questions and some minor nits, other than that great job!

sdv/sampling/independent_sampler.py
frances-h: I think ideally this and `get_foreign_keys` (which returns the foreign keys from one parent/child pair) should be _metadata_ methods. Both of them are used for both sampling and fitting. Not sure if that change should happen in these PRs though, it might make sense to put it off until we actually do the integration.

amontanez24: I think we should move to the metadata

sdv/sampling/independent_sampler.py
```diff
@@ -0,0 +1,172 @@
...
+                Name of the table to sample.
+            num_rows (int):
+                Number of rows to sample for the table.
+            sampled_data (dict):
+                A dictionary mapping table names to sampled tables (pd.DataFrame).
+        """
+        LOGGER.info('Sampling %s rows from table %s', num_rows, table_name)
+
+        if synthesizer._model:
+            sampled_rows = synthesizer._sample(num_rows)
```
amontanez24, line 90: same potential problem about reject sampling here as in [other PR](https://github.com/sdv-dev/SDV/pull/1411#discussion_r1194410703)

sdv/sampling/independent_sampler.py
```diff
@@ -0,0 +1,172 @@
...
+        Returns:
+            dict:
+                A dictionary containing as keys the names of the tables and as values the
+                sampled data tables as ``pandas.DataFrame``.
+        """
+        sampled_data = {}
+        for table in self.metadata.tables:
+            num_rows = int(self._table_sizes[table] * scale)
+            synthesizer = self._table_synthesizers[table]
+            self._sample_table(synthesizer=synthesizer, table_name=table, num_rows=num_rows,
```
amontanez24, line 168: minor: same thing about formatting

sdv/sampling/independent_sampler.py
```diff
@@ -0,0 +1,172 @@
...
+            Dictionary mapping each table to its size. Should be instantiated and passed to the
+            sampler by the multi-table synthesizer using this sampler.
+    """
+
+    def __init__(self, metadata, table_synthesizers, table_sizes):
+        self.metadata = metadata
+        self._table_synthesizers = table_synthesizers
+        self._table_sizes = table_sizes
+
+    def _add_foreign_key_column(self, child_table, parent_table, child_name, parent_name):
```
amontanez24, line 29: As we recently discussed, it seems like there can be multiple foreign keys to the same table. Maybe we should change the name here to make it plural and make the function expected to add all foreign key columns

pvk-developer (approved):

sdv/sampling/independent_sampler.py
```diff
@@ -0,0 +1,134 @@
...
+        This method iterates over all tables and adds all foreign keys to connect
+        related tables.
+
+        Args:
+            sampled_data (dict):
+                A dictionary mapping table names to the sampled tables (pd.DataFrame).
+        """
+        queue = [
+            table for table in self.metadata.tables if not self.metadata._get_parent_map()[table]
+        ]
```
pvk-developer, line 71: Small nit that helps readability if you like it (not required to change up to you):
With list comprehensions you can do:
```python
queue = [
    table
    for table in self.metadata.tables
    if not self.metadata._get_parent_map()[table]
]
``` 
To me it makes it more easy to read since you have `object` that you are storing, the loop that you are performing, the condition that you apply in multiple lines.

pvk-developer (approved):
LGTM!

sdv/sampling/independent_sampler.py
```diff
@@ -0,0 +1,134 @@
...
+            synthesizer (SingleTableSynthesizer):
+                Synthesizer to sample from for the table.
+            table_name (string):
+                Name of the table to sample.
+            num_rows (int):
+                Number of rows to sample for the table.
+            sampled_data (dict):
+                A dictionary mapping table names to sampled tables (pd.DataFrame).
+        """
+        LOGGER.info('Sampling %s rows from table %s', num_rows, table_name)
```
amontanez24, line 54: minor: can we use f-string?

amontanez24 (approved):
One nitpick but otherwise :shipit: 🍾

PR #1426, opened 2023-05-15
Initialize multi-table synthesizers once during __init__

### LINKED ISSUE #1418: Synthesizers created twice during HMA init
### Problem Description

During the `__init__` method of the `BaseMultiTableSynthesizer`, we create a dict of single table synthesizers when we call [_initialize_models](https://github.com/sdv-dev/SDV/blob/master/sdv/multi_table/base.py#L50).
https://github.com/sdv-dev/SDV/blob/3df36ba3296f0632cb9d954eeaf72c7ac27c8b62/sdv/multi_table/base.py#L64-L66

Then later, in the `__init__` for `HAMSynthesizer, we go through this whole process again and recreate a bunch of single table synthesizers but this time with the custom `DEFAULT_SYNTHESIZER_KWARGS`.
https://github.com/sdv-dev/SDV/blob/3df36ba3296f0632cb9d954eeaf72c7ac27c8b62/sdv/multi_table/hma.py#L36-L37

This is not performant since we are initializing each single table synthesizer twice, and each one validates metadata.

### Expected behavior

- Make the `BaseMultiTableSynthesizer` initialize the synthesizers with the custom kwargs and get rid of the statement that redoes it in the `HMASynthesizer`.

frances-h:
Resolve #1418

### DIFF
--- a/sdv/multi_table/base.py
+++ b/sdv/multi_table/base.py
@@ -29,6 +29,7 @@ class BaseMultiTableSynthesizer:
             The default locale(s) to use for AnonymizedFaker transformers. Defaults to ``None``.
     """
 
+    DEFAULT_SYNTHESIZER_KWARGS = None
     _synthesizer = GaussianCopulaSynthesizer
     _numpy_seed = 73251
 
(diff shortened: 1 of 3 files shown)


sdv/multi_table/base.py
```diff
@@ -56,12 +57,17 @@ def _initialize_models(self):
                 **synthesizer_parameters
             )
 
-    def __init__(self, metadata, locales=None):
+    def __init__(self, metadata, locales=None, synthesizer_kwargs=None):
```
amontanez24, line 60: I think the `synthesizer_kwargs` method isn't meant to be part of the new API and we can actually just drop it in the base and HMA
@npatki Is that correct?

amontanez24 (changes requested):

sdv/multi_table/base.py
npatki: @amontanez24 correct!

sdv/multi_table/hma.py
```diff
@@ -72,7 +69,8 @@ def _get_extension(self, child_name, child_table, foreign_key):
                     row = pd.Series({'num_rows': len(child_rows)})
                     row.index = f'__{child_name}__{foreign_key}__' + row.index
                 else:
-                    synthesizer = self._synthesizer(table_meta, **self._synthesizer_kwargs)
+                    synthesizer = self._synthesizer(table_meta,
```
amontanez24, line 72: I think in some places we have the formatting like
```python
function(
    arg1,
    arg2
)
I think we should stay consistent. Also make it easier to meet the 100 char line limit

amontanez24 (commented):
One minor comment but I think this should be good! :shipit:

amontanez24 (approved):
LGTM! Thanks for addressing the comments!

fealho (approved):

PR #1429, opened 2023-05-16
Add validation for constraints 

### LINKED ISSUE #1402: Synthesizers don't validate constraints during validate()
### Environment Details

* SDV version: 1.0.*
* Python version: Any
* Operating System: Any

### Error Description

When running `synthesizer.validate(data)` on data that doesn't meet constraints already set in the synthesizer, we should raise errors for those constraints. 

### Steps to reproduce

``` python

real_data, metadata = download_demo(
    modality='single_table',
)
real_data['checkin_date'][0] = real_data['checkout_date'][1] # make constraint fail

gc = GaussianCopulaSynthesizer(metadata)

checkin_lessthan_checkout = {
    'constraint_parameters': {
        'low_column_name': 'checkin_date',
        'high_column_name': 'checkout_date'
    }
}
gc.add_constraints([
    checkin_lessthan_checkout
])
gc.validate() # Should raise error for constraint
```

pvk-developer:
Resolves #1402 
Fixes a bug with first not nan value for datetimes.

### DIFF
--- a/sdv/constraints/utils.py
+++ b/sdv/constraints/utils.py
@@ -194,3 +194,12 @@ def get_datetime_diff(high, low, dtype='O'):
     diff_column = diff_column.astype(np.float64)
     diff_column[nan_mask] = np.nan
     return diff_column
+
+
+def create_unique_name(name, list_names):
+    """Modify the ``name`` parameter if it already exists in the list of names."""
+    result = name
+    while result in list_names:
+        result += '_'
+
+    return result
--- a/sdv/constraints/utils.py
+++ b/sdv/constraints/utils.py
@@ -194,12 +194,3 @@ def get_datetime_diff(high, low, dtype='O'):
     diff_column = diff_column.astype(np.float64)
     diff_column[nan_mask] = np.nan
     return diff_column
-
-
-def create_unique_name(name, list_names):
-    """Modify the ``name`` parameter if it already exists in the list of names."""
-    result = name
-    while result in list_names:
-        result += '_'
-
-    return result
--- a/sdv/data_processing/data_processor.py
+++ b/sdv/data_processing/data_processor.py
@@ -286,7 +286,23 @@ class DataProcessor:
         """
         return deepcopy(self._constraints_list)
 
+    def _load_constraints(self):
+        loaded_constraints = []
+        default_constraints_classes = list(get_subclasses(Constraint))
+        for constraint in self._constraints_list:
+            if constraint['constraint_class'] in default_constraints_classes:
+                loaded_constraints.append(Constraint.from_dict(constraint))
+
+            else:
+                constraint_class = self._custom_constraint_classes[constraint['constraint_class']]
+                loaded_constraints.append(
+                    constraint_class(**constraint.get('constraint_parameters', {}))
+                )
+
+        return loaded_constraints
+
     def _fit_constraints(self, data):
+        self._constraints = self._load_constraints()
         errors = []
         for constraint in self._constraints:
             try:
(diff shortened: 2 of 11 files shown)


sdv/single_table/base.py
```diff
@@ -204,6 +204,18 @@ def _validate(self, data):
         """
         return []
 
+    def _validate_constraints(self, data):
+        """Validate that the data satisfies the constraints."""
+        errors = []
+        constraints = self._data_processor._load_constraints()
+        for constraint in constraints:
+            try:
+                constraint.fit(data)
```
amontanez24, line 213: So this means we fit constraints twice right? Since they get fit here too 
https://github.com/sdv-dev/SDV/blob/0e39f0f9490b704fe3e9bf48d2795cca8870c341/sdv/data_processing/data_processor.py#L289-L294

Is there a way we can avoid this?

amontanez24 (commented):
It looks like we also don't ever validate the metadata against the constraints with this method 
https://github.com/sdv-dev/SDV/blob/0e39f0f9490b704fe3e9bf48d2795cca8870c341/sdv/constraints/base.py#L156
It doesn't seem like that has any references. We might need to add that too

sdv/single_table/base.py
pvk-developer: > It looks like we also don't ever validate the metadata against the constraints with this method
> 
> https://github.com/sdv-dev/SDV/blob/0e39f0f9490b704fe3e9bf48d2795cca8870c341/sdv/constraints/base.py#L156
> 
> 
> It doesn't seem like that has any references. We might need to add that too

It is being validated when we validate the `constraint dict` when adding a constraint: 
https://github.com/sdv-dev/SDV/blob/master/sdv/data_processing/data_processor.py#L252

sdv/constraints/utils.py
```diff
@@ -194,3 +194,12 @@ def get_datetime_diff(high, low, dtype='O'):
     diff_column = diff_column.astype(np.float64)
     diff_column[nan_mask] = np.nan
     return diff_column
+
+
+def create_unique_name(name, list_names):
```
amontanez24, line 199: One thing to note is that our utils module is technically public, so we should be careful about removing things/moving them around (I know I did that in the past 😅 ). In this case, it should be fine since we haven't done a release with this method yet. However, if we think it will ever be useful to have at a higher level (use outside of constraints), maybe we should leave it in sdv.utils

sdv/data_processing/data_processor.py
```diff
@@ -575,7 +582,10 @@ def prepare_for_fitting(self, data):
             self._fit_formatters(data)
 
             LOGGER.info(f'Fitting constraints for table {self.table_name}')
-            constrained = self._fit_transform_constraints(data)
+            if len(self._constraints_list) != len(self._constraints):
```
amontanez24, line 585: maybe we should add a unit test for this case

amontanez24 (approved):
Couple of minor comments but otherwise LGTM!

sdv/data_processing/data_processor.py
pvk-developer: This test covers it [here](https://github.com/sdv-dev/SDV/pull/1429/files#diff-65e7498d86df62640dd967d34beb38e7db5acaf9b73a2e240158313e61c08328R1300-R1312) ps: Scroll down and you will see it highlighted.
```python
        dp._transform_constraints.return_value = transformed_data
        dp._prepared_for_fitting = False
        dp._hyper_transformer.get_config.return_value = {'sdtypes': {}, 'transformers': {}}
        dp._constraints_list = [1, 2, 3] 
        dp._constraints = []

        # Run
        DataProcessor.prepare_for_fitting(dp, data)

        # Assert
        pd.testing.assert_series_equal(dp._dtypes, pd.Series([np.int64], index=['a']))
        dp._transform_constraints.assert_called_once_with(data)
        dp._fit_constraints.assert_called_once_with(data)
```

sdv/constraints/utils.py
pvk-developer: The reason I did the change is because test where in one place (tests/unit/constraints/test_utils.py) and the code in sdv/utils.py so it had to go either one place or another 👍🏻 now should be fixed.

PR #1441, opened 2023-05-25
Fix child tables being modeled

pvk-developer:
As part of #1387 
Missed to only model root tables (tables that don't have parents). A model was fitted for each table individually as long as it was augmented.


amontanez24 (approved):
I think this should work 👍

fealho (approved):

PR #1446, opened 2023-05-30
Report fitting progress for `HMASynthesizer`

### LINKED ISSUE #1440: Report fitting progress for the `HMASynthesizer`
### Problem Description
The [HMASynthesizer](https://docs.sdv.dev/sdv/multi-table-data/modeling/synthesizers/hmasynthesizer) does not offer a verbosity option today. If there is a large dataset, it may take some time to learn the model (i.e. `fit` the data) and I am just waiting with a blank screen. It would be better if there was a progress report I could see.

### Expected behavior
Add a `verbose` parameter to the initialization of the HMASynthesizer. If set to `True`, then it should print out the progress during all fitting stages. Let's **set the parameter to True by default**. Users can always turn this off.

```python
>>> synth = HMASynthesizer(metadata, verbose=True)
>>> synth.fit(data)

Preprocess Tables: 100%|█████████████████████████████| 3/3 [00:33<00:10, 229.00it/s]

Learning relationships:
(1/2) Tables 'A' and 'B' ('user_id'): 100%|█████████████████████████████| 1000/1000 [00:33<00:10, 229.00it/s]
(2/2) Tables 'B' and 'C' ('session_id'): 100%|█████████████████████████████| 1200/1200 [00:33<00:10, 229.00it/s]

Modeling Tables: 100%|█████████████████████████████| 1/1 [00:33<00:10, 229.00it/s]
```

### Additional context (fit)
Let `r` be the total number of relationships in the multi table schema. Then the total number of steps is `r+2`. Each step will have a new progress bar.

- **Step 1**: The first step will be preprocessing. The progress bar should increment for each table that we preprocess. (https://github.com/sdv-dev/SDV/blob/master/sdv/multi_table/base.py#L307) 
- **Next r steps**: For each relationship, show a new progress bar. The bar should increment as we iterate over each parent row to compute the extension over a child’s foreign key (https://github.com/sdv-dev/SDV/blob/master/sdv/multi_table/hma.py#L63)
- **Final step**: The final step is modeling each root table. The progress bar should increment as we iterate over each root table that is being fitted (https://github.com/sdv-dev/SDV/blob/master/sdv/multi_table/hma.py#L186)

### Additional Context (preprocress)
When preprocessing, only step 1 is performed. 

```python
>>> synth = HMASynthesizer(metadata, verbose=True)
>>> processed_data = synth.preprocess(data)

Preprocess Tables: 100%|█████████████████████████████| 3/3 [00:33<00:10, 229.00it/s]
```

### Additional Context (fit_processed_data)
When fitting processed data, the preprocessing is already done, so there are only `r+1` remaining steps.

```python
>>> synth.fit_processed_data(processed_data)
Learning relationships:
(1/2) Tables 'A' and 'B' ('user_id'): 100%|█████████████████████████████| 1000/1000 [00:33<00:10, 229.00it/s]
(2/2) Tables 'B' and 'C' ('session_id'): 100%|█████████████████████████████| 1200/1200 [00:33<00:10, 229.00it/s]

Modeling Tables: 100%|█████████████████████████████| 1/1 [00:33<00:10, 229.00it/s]
```

pvk-developer:
Resolves #1440 
image

### DIFF
--- a/sdv/multi_table/base.py
+++ b/sdv/multi_table/base.py
@@ -9,6 +9,7 @@ import cloudpickle
 import numpy as np
 import pandas as pd
 import pkg_resources
+from tqdm.auto import tqdm
 
 from sdv.errors import SynthesizerInputError
 from sdv.single_table.copulas import GaussianCopulaSynthesizer
@@ -27,6 +28,8 @@ class BaseMultiTableSynthesizer:
             for.
         locales (list or str):
             The default locale(s) to use for AnonymizedFaker transformers. Defaults to ``None``.
+        verbose (bool):
+            Whether to print progress for fitting or not.
     """
 
     DEFAULT_SYNTHESIZER_KWARGS = None
@@ -57,7 +60,13 @@ class BaseMultiTableSynthesizer:
                 **synthesizer_parameters
             )
 
-    def __init__(self, metadata, locales=None, synthesizer_kwargs=None):
+    def _get_pbar_args(self, **kwargs):
+        pbar_args = {'disable': not self.verbose}
+        pbar_args.update(kwargs)
+
+        return pbar_args
+
+    def __init__(self, metadata, locales=None, synthesizer_kwargs=None, verbose=True):
         self.metadata = metadata
         self.metadata.validate()
         self.locales = locales
@@ -304,7 +313,8 @@ class BaseMultiTableSynthesizer:
             )
 
         processed_data = {}
-        for table_name, table_data in data.items():
+        pbar_args = self._get_pbar_args(desc='Preprocess Tables')
+        for table_name, table_data in tqdm(data.items(), **pbar_args):
             synthesizer = self._table_synthesizers[table_name]
             self._assign_table_transformers(synthesizer, table_name, table_data)
             processed_data[table_name] = synthesizer._preprocess(table_data)
--- a/sdv/multi_table/base.py
+++ b/sdv/multi_table/base.py
@@ -75,7 +75,7 @@ class BaseMultiTableSynthesizer:
                 print(message, **kwargs)  # noqa: T001
 
             elif message:
-                print(message)  # noqa: T001
+                print(message, **kwargs)  # noqa: T001
 
     def __init__(self, metadata, locales=None, synthesizer_kwargs=None, verbose=True):
         self.metadata = metadata
--- a/sdv/multi_table/base.py
+++ b/sdv/multi_table/base.py
@@ -327,7 +327,7 @@ class BaseMultiTableSynthesizer:
 
         return processed_data
 
-    def _model_tables(self, augmented_data, prefix=None):
+    def _model_tables(self, augmented_data):
         """Model the augmented tables.
 
         Args:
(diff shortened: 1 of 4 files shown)


sdv/multi_table/base.py
```diff
@@ -336,8 +358,8 @@ def fit_processed_data(self, processed_data):
             processed_data (dict):
                 Dictionary mapping each table name to a preprocessed ``pandas.DataFrame``.
         """
-        augmented_data = self._augment_tables(processed_data)
-        self._model_tables(augmented_data)
+        augmented_data = self._augment_tables(processed_data.copy())
+        self._model_tables(augmented_data, prefix='\n')
```
amontanez24, line 362: the base `_model_tables` doesn't have the new prefix arg, so this would crash on HSA right?

sdv/multi_table/hma.py
```diff
@@ -176,26 +195,35 @@ def _pop_foreign_keys(self, table_data, table_name):
...
         return keys
 
-    def _model_tables(self, augmented_data):
+    def _model_tables(self, augmented_data, prefix=None):
         """Model the augmented tables.
 
         Args:
             augmented_data (dict):
                 Dictionary mapping each table name to an augmented ``pandas.DataFrame``.
+            prefix (str):
```
amontanez24, line 204: Why is this parameter necessary? It seems like this method is only called in one place and it always provides the prefix as `'\n'`. Can't we just set it in here?

sdv/multi_table/hma.py
```diff
@@ -47,6 +52,8 @@ def _get_extension(self, child_name, child_table, foreign_key):
                 Data for the child table.
             foreign_key (str):
                 Name of the foreign key field.
+            pbar_desc (str):
```
amontanez24, line 55: minor: can we call `progress_bar_desc`

sdv/multi_table/hma.py
```diff
@@ -136,21 +144,32 @@ def _augment_table(self, table, tables, table_name):
...
-            for index, foreign_key in enumerate(foreign_keys):
-                extension = self._get_extension(child_name, child_table.copy(), foreign_key)
+            for foreign_key in foreign_keys:
+                pbar_desc = (
+                    f'({self._learned_relationships + 1}/{len(self.metadata.relationships)}) '
+                    f"Tables '{table_name}' and '{child_name}' ('{foreign_key}')"
+                )
+                extension = self._get_extension(
+                    child_name,
+                    child_table.copy(),
```
amontanez24, line 163: Is it necessary to copy?

sdv/multi_table/base.py
pvk-developer: Nice catch, let me add it :shipit:

sdv/multi_table/hma.py
pvk-developer: 🤔 yeah we could assumingly we always use `_model_tables` from other methods.

sdv/multi_table/hma.py
pvk-developer: Yes because we alter the data and if that child is used in another parent it ends up being altered.

amontanez24 (approved):
Thanks for addressing! :shipit:

tests/unit/multi_table/test_base.py
```diff
@@ -52,6 +52,50 @@ def test__initialize_models(self):
             call(metadata=instance.metadata.tables['upravna_enota'], locales=locales)
         ])
 
+    def test__get_pbar_args(self):
+        """Test that ``_get_pbar_args`` returns a dictionary with disable oposite to verbose."""
```
fealho, line 56: typo: opposite

fealho (approved):
Looks pretty good!

PR #1449, opened 2023-05-31
Use HierarchicalSampler mixin with HMA

### LINKED ISSUE #1428: Make HMA use hierarchical sampling mixin
### Problem Description

_As an engineer, it would be helpful to have our synthesizers follow clear and well defined abstractions for the sake of modifying or enhancing functionality._

### Expected behavior

- Once #1411 is complete, make the `HMASynthesizer` inherit from that `BaseHierachicalSampler` and implement the required methods.
- Update the tests to match.

### Additional context

- All changes should be private. There shouldn't be a need to alter any user facing methods or classes.

frances-h:
Resolve #1428

### DIFF
--- a/sdv/multi_table/hma.py
+++ b/sdv/multi_table/hma.py
@@ -358,16 +358,15 @@ class HMASynthesizer(BaseHierarchicalSampler, BaseMultiTableSynthesizer):
         """Find parent ids for the given table and foreign key.
 
         The parent ids are chosen randomly based on the likelihood of the available
-        parent ids in the parent table. If the parent table is not sampled, this method
-        will first sample rows for the parent table.
+        parent ids in the parent table.
 
         Args:
             child_table (pd.DataFrame):
-
+                The child table dataframe.
             parent_table (pd.DataFrame):
-
+                The parent table dataframe.
             child_name (str):
-
+                The name of the child table.
             parent_name (dict):
                 Map of table name to sampled data (pandas.DataFrame).
             foreign_key (str):
@@ -377,6 +376,7 @@ class HMASynthesizer(BaseHierarchicalSampler, BaseMultiTableSynthesizer):
             pandas.Series:
                 The parent ids for the given table data.
         """
+        # Create a copy of the parent table with the primary key as index to calculate likilihoods
         primary_key = self.metadata.tables[parent_name].primary_key
         parent_table = parent_table.set_index(primary_key)
         num_rows = parent_table[f'__{child_name}__{foreign_key}__num_rows'].fillna(0).clip(0)
--- a/sdv/multi_table/hma.py
+++ b/sdv/multi_table/hma.py
@@ -9,11 +9,12 @@ import pandas as pd
 from tqdm import tqdm
 
 from sdv.multi_table.base import BaseMultiTableSynthesizer
+from sdv.sampling import BaseHierarchicalSampler
 
 LOGGER = logging.getLogger(__name__)
 
 
-class HMASynthesizer(BaseMultiTableSynthesizer):
+class HMASynthesizer(BaseHierarchicalSampler, BaseMultiTableSynthesizer):
     """Hierarchical Modeling Algorithm One.
 
     Args:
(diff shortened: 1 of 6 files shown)


sdv/sampling/hierarchical_sampler.py
```diff
@@ -68,7 +68,7 @@ def _sample_rows(self, synthesizer, num_rows=None):
                 Sampled rows, shape (, num_rows)
         """
         num_rows = num_rows or synthesizer._num_rows
-        return synthesizer._sample_batch(num_rows, remove_extra_columns=False)
+        return synthesizer._sample_batch(int(num_rows), keep_extra_columns=True)
```
fealho, line 71: @amontanez24 @pvk-developer This line is an example where using an assert could be useful. `num_rows` should always be an integer, and if it is not then wherever it was set did so incorrectly, so the code there should be updated to cast it to an int.

So at this point we should `assert isinstance(num_rows, int)`, or change the docstring to say that `num_rows` can be a float as well. You shouldn't have to write code assuming other methods/inputs are invalid, ie casting to `int` in this case.

fealho (approved):
Looks good to me ⚡

sdv/multi_table/hma.py
```diff
@@ -9,11 +9,12 @@
 from tqdm import tqdm
 
 from sdv.multi_table.base import BaseMultiTableSynthesizer
+from sdv.sampling import BaseHierarchicalSampler
 
 LOGGER = logging.getLogger(__name__)
 
 
-class HMASynthesizer(BaseMultiTableSynthesizer):
+class HMASynthesizer(BaseHierarchicalSampler, BaseMultiTableSynthesizer):
```
amontanez24, line 17: Does this order matter? If not could we flip it 😅

sdv/multi_table/hma.py
```diff
@@ -484,89 +354,44 @@ def _get_likelihoods(self, table_rows, parent_rows, table_name, foreign_key):
...
-        else:
-            ratio = self._table_sizes[parent_name] / self._table_sizes[table_name]
-            num_parent_rows = max(int(round(len(table_rows) * ratio)), 1)
-            parent_model = self._table_synthesizers[parent_name]
-            parent_rows = self._sample_rows(parent_model, parent_name, num_parent_rows)
-
         primary_key = self.metadata.tables[parent_name].primary_key
-        parent_rows = parent_rows.set_index(primary_key)
-        num_rows = parent_rows[f'__{table_name}__{foreign_key}__num_rows'].fillna(0).clip(0)
+        parent_table = parent_table.set_index(primary_key)
```
amontanez24, line 381: @fealho referencing our discussion, this is an area where comments might be useful. Explaining why we set the index either here or in the docstring could be helpful

sdv/single_table/base.py
```diff
@@ -604,10 +604,11 @@ def _sample_rows(self, num_rows, conditions=None, transformed_conditions=None,
                 * int:
                     Number of rows that are considered valid.
         """
-        if not self._random_state_set:
-            self._set_random_state(FIXED_RNG_SEED)
+        if (self._model and (
```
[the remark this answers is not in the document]
amontanez24: +1

sdv/single_table/base.py
```diff
@@ -604,10 +604,11 @@ def _sample_rows(self, num_rows, conditions=None, transformed_conditions=None,
...
+                self._data_processor.get_sdtypes(primary_keys=False) or keep_extra_columns)):
+            if not self._random_state_set:
+                self._set_random_state(FIXED_RNG_SEED)
```
amontanez24, line 610: why can't we do this all the time anymore? Also can we add testing for this case?

amontanez24 (commented):
Mostly minor comments! This is looking good :)

sdv/multi_table/hma.py
frances-h: Order does matter, if we don't have `BaseHierarchicalSampler` first then it'll hit the `BaseMultiTableSynthesizer` methods _first_, so we'll end up hitting `NotImplemented` errors.

sdv/single_table/base.py
frances-h: If we don't have a `_model` then this errors (the default implementation calls the `set_random_state` method on `self._model`.

amontanez24 (approved):
I think this looks good!

PR #1452, opened 2023-06-01
Fix `ValueError` when sampling `PII` columns to a prior numerical column

### LINKED ISSUE #1445: `ValueError` when sampling PII columns
### Environment Details
* SDV version: 1.1.0

### Description
For PII columns only, it should be ok if the input `pandas.dtype` is not the same as the output `pandas.dtype`. For example, the input data may be all 0'd out (since it's sensitive). But I expect the synthetic data data should have strings, based on the `sdtype` that I have selected in the metadata (and ultimately the Faker that is used).

### Steps to reproduce

```python
import pandas as pd

data = pd.DataFrame(data={
    'id': [0, 1, 2, 3, 4],
    'address': [0, 0, 0, 0, 0],
    'numerical': [0.234234, 0.123213, 0.123123, 0.123123, 0.1345435]
})

    'primary_key': 'id',
    'columns': {
        'id': { 'sdtype': 'id' },
        'address': { 'sdtype': 'address' },
        'numerical': { 'sdtype': 'numerical' }
    }
})

synth.fit(data)
synth.sample(10)
```

Output:
```
ValueError: Error: Sampling terminated. Partial results are stored in a temporary file: .sample.csv.temp. This file will be overridden the next time you sample. Please rename the file if you wish to save these results.
invalid literal for int() with base 10: 'PSC 4734, Box 5790\nAPO AA 75326'
```

### Additional Context
For **PII and ID columns only**, we can do a try-catch whenever we try to cast the data back to the original `dtype`. If we cannot do the casting, then just return the data without casting. Log an INFO message when this happens.

```
logger.INFO The real data in 'column_name' was stored as 'int' but the synthetic data could not be cast back
to this type. If this is a problem, please check your input data and metadata settings.
```

pvk-developer:
Resolves #1445

### DIFF
--- a/sdv/data_processing/data_processor.py
+++ b/sdv/data_processing/data_processor.py
@@ -751,17 +751,23 @@ class DataProcessor:
             if is_integer_dtype(dtype) and is_float_dtype(column_data.dtype):
                 column_data = column_data.round()
 
+            reversed_data[column_name] = column_data[column_data.notna()]
             try:
-                reversed_data[column_name] = column_data[column_data.notna()].astype(dtype)
-            except ValueError:
-                LOGGER.info(
-                    f"The real data in '{column_name}' was stored as '{dtype}' but the synthetic "
-                    'data could not be cast back to this type. If this is a problem, please check '
-                    'your input data and metadata settings.'
-                )
-                reversed_data[column_name] = column_data[column_data.notna()]
-                if column_name in self.formatters:
-                    self.formatters.pop(column_name)
+                reversed_data[column_name] = reversed_data[column_name].astype(dtype)
+            except ValueError as e:
+                column_metadata = self.metadata.columns.get(column_name)
+                sdtype = column_metadata.get('sdtype')
+                if sdtype not in self._DTYPE_TO_SDTYPE.values():
+                    LOGGER.info(
+                        f"The real data in '{column_name}' was stored as '{dtype}' but the "
+                        'synthetic data could not be cast back to this type. If this is a '
+                        'problem, please check your input data and metadata settings.'
+                    )
+                    if column_name in self.formatters:
+                        self.formatters.pop(column_name)
+
+                else:
+                    raise ValueError(e)
 
         # reformat columns using the formatters
         for column in sampled_columns:
--- a/tests/integration/single_table/test_copulas.py
+++ b/tests/integration/single_table/test_copulas.py
@@ -323,19 +323,41 @@ def test_validate_with_failing_constraint():
 
 def test_numerical_columns_gets_pii():
     """Test that the synthesizer works when a ``numerical`` column gets converted to ``PII``."""
+    # Setup
     data = pd.DataFrame(data={
         'id': [0, 1, 2, 3, 4],
-        'address': [0, 0, 0, 0, 0],
-        'numerical': [0.234234, 0.123213, 0.123123, 0.123123, 0.1345435]
+        'city': [0, 0, 0, 0, 0],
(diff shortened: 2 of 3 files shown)


sdv/data_processing/data_processor.py
```diff
@@ -751,7 +751,17 @@ def reverse_transform(self, data, reset_keys=False):
...
+                reversed_data[column_name] = column_data[column_data.notna()].astype(dtype)
+            except ValueError:
+                LOGGER.info(
+                    f"The real data in '{column_name}' was stored as '{dtype}' but the synthetic "
+                    'data could not be cast back to this type. If this is a problem, please check '
+                    'your input data and metadata settings.'
+                )
+                reversed_data[column_name] = column_data[column_data.notna()]
+                if column_name in self.formatters:
+                    self.formatters.pop(column_name)
```
amontanez24, line 764: What is this for?

sdv/data_processing/data_processor.py
```diff
@@ -751,7 +751,17 @@ def reverse_transform(self, data, reset_keys=False):
             if is_integer_dtype(dtype) and is_float_dtype(column_data.dtype):
                 column_data = column_data.round()
 
-            reversed_data[column_name] = column_data[column_data.notna()].astype(dtype)
+            try:
+                reversed_data[column_name] = column_data[column_data.notna()].astype(dtype)
+            except ValueError:
```
amontanez24, line 756: Is it possible that this crashes on a column that is not pii, and if so, do we still want to catch the error?

sdv/data_processing/data_processor.py
pvk-developer: We store data formatters in there, numerical and datetime, so if the column failed to be casted as `int` we remove that formatter from there, otherwise it will fail later again.

sdv/data_processing/data_processor.py
```diff
@@ -751,7 +751,17 @@ def reverse_transform(self, data, reset_keys=False):
             if is_integer_dtype(dtype) and is_float_dtype(column_data.dtype):
                 column_data = column_data.round()
 
-            reversed_data[column_name] = column_data[column_data.notna()].astype(dtype)
+            try:
+                reversed_data[column_name] = column_data[column_data.notna()].astype(dtype)
```
fealho, line 755: Can you move the `reversed_data[column_name] = column_data[column_data.notna()]` outside the `try`? So that we are sure the `ValueError` was raised due to the `astype`. You can then remove line 762 as well.

tests/integration/single_table/test_copulas.py
```diff
@@ -318,3 +319,23 @@ def test_validate_with_failing_constraint():
     # Run / Assert
     with pytest.raises(InvalidDataError, match=error_msg):
         gc.validate(real_data)
+
+
+def test_numerical_columns_gets_pii():
+    """Test that the synthesizer works when a ``numerical`` column gets converted to ``PII``."""
+    data = pd.DataFrame(data={
```
fealho, line 326: Add #Setup/Run/Assert.

Also, since this is the only integration test which runs this edge case, maybe add a simple assert like checking samples is a dataframe with the right shape/dtypes.

fealho (commented):
Looks good, some minor comments.

fealho (approved):

sdv/data_processing/data_processor.py
pvk-developer: Great tip 🙏🏻 thanks

PR #1495, opened 2023-07-07
Move progress bar out of base multi table synthesizer

### LINKED ISSUE #1486: Move progress bar out of base multi table synthesizer
### Problem Description

We don't want to show the progress bar for only part of the process on HSA. To handle this, we recommend removing the progress bar from the base for now and moving it into `HMASynthesizer`.

### Expected behavior

- Remove verbose argument from `BaseMultiTableSynthesizer`
- Only show the progress bar in HMA

### Additional context

- Right now the base preprocess method uses the progress bar when applicable.
https://github.com/sdv-dev/SDV/blob/master/sdv/multi_table/base.py#L322-L323
One suggestion could be to make `_get_pbar_args` return a disabled progress bar and just override it in `HMASynthesizer` to use the verbose argument. This way we don't have to move much code around

R-Palazzo:
Resolve #1486

### DIFF
--- a/sdv/multi_table/base.py
+++ b/sdv/multi_table/base.py
@@ -71,11 +71,11 @@ class BaseMultiTableSynthesizer:
         if self.verbose:
             print(text, **kwargs)  # noqa: T001
 
-    def __init__(self, metadata, locales=None, synthesizer_kwargs=None, verbose=False):
+    def __init__(self, metadata, locales=None, synthesizer_kwargs=None):
         self.metadata = metadata
         self.metadata.validate()
         self.locales = locales
-        self.verbose = verbose
+        self.verbose = False
         self._table_synthesizers = {}
         self._table_parameters = defaultdict(dict)
         if synthesizer_kwargs is not None:
--- a/sdv/multi_table/base.py
+++ b/sdv/multi_table/base.py
@@ -28,6 +28,8 @@ class BaseMultiTableSynthesizer:
             for.
         locales (list or str):
             The default locale(s) to use for AnonymizedFaker transformers. Defaults to ``None``.
+        verbose (bool):
+            Whether to print progress for fitting or not.
     """
 
     DEFAULT_SYNTHESIZER_KWARGS = None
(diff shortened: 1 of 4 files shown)


sdv/multi_table/base.py
```diff
@@ -62,20 +60,18 @@ def _initialize_models(self):
...
-        pbar_args = {'disable': not self.verbose}
+        pbar_args = {'disable': True}
         pbar_args.update(kwargs)
 
         return pbar_args
 
     def _print(self, text='', **kwargs):
-        if self.verbose:
-            print(text, **kwargs)  # noqa: T001
+        print(text, **kwargs)  # noqa: T001
```
amontanez24, line 69: do we want to still have this only print when verbose is true? It gets called in the fit still to print an empty line

R-Palazzo: I actually don't know. Because it verbose was True by default I kept it, but maybe we can drop it.

amontanez24: We should still print the new line, but only when the progress bar is actually printed. Maybe the easier fix is to set verbose to False in the base class and set it to what the user provides in HMA.

R-Palazzo: Alright, done in [29c5c16](https://github.com/sdv-dev/SDV/pull/1495/commits/29c5c163cea83bdd0e5fe0b550c18ed58c3f9318)

pvk-developer (approved):

sdv/multi_table/base.py
```diff
@@ -71,7 +71,7 @@ def _print(self, text='', **kwargs):
         if self.verbose:
             print(text, **kwargs)  # noqa: T001
 
-    def __init__(self, metadata, locales=None, synthesizer_kwargs=None, verbose=True):
+    def __init__(self, metadata, locales=None, synthesizer_kwargs=None, verbose=False):
```
amontanez24, line 74: Sorry, I meant remove it as a parameter but set it to false below

sdv/multi_table/hma.py
```diff
@@ -32,12 +32,11 @@ class HMASynthesizer(BaseHierarchicalSampler, BaseMultiTableSynthesizer):
     }
 
     def __init__(self, metadata, locales=None, verbose=True):
-        BaseMultiTableSynthesizer.__init__(self, metadata, locales=locales)
+        BaseMultiTableSynthesizer.__init__(self, metadata, locales=locales, verbose=verbose)
```
amontanez24, line 35: Here we would just set it to true after calling
```python
BaseMultiTableSynthesizer.__init__(self, metadata, locales=locales)
```

amontanez24 (changes requested):

amontanez24 (approved):
:shipit:

PR #1496, opened 2023-07-11
Add support for python 3.11

### LINKED ISSUE #1459: Add support for Python 3.11
### Problem Description
CTGAN should work on Python 3.11

### Background
The SDV library currently runs on Python versions 3.7-3.10. A while ago, we had made the decision to exclude 3.11 because the most recent versions of torch were incompatible with it.

However, torch is now fully upgraded to 2.0 so this should no longer be an issue.

### Dependencies
- CTGAN: https://github.com/sdv-dev/CTGAN/issues/296
- DeepEcho: https://github.com/sdv-dev/DeepEcho/issues/74
- SDMetrics: https://github.com/sdv-dev/SDMetrics/issues/353

fealho:
Resolve #1459.


fealho:
PR is waiting on CTGAN, SDMetrics and DeepEcho releases with python 3.11.

sdv/multi_table/hma.py
```diff
@@ -6,6 +6,7 @@
 
 import numpy as np
 import pandas as pd
+from packaging import version
```
amontanez24, line 9: can we avoid using packaging? It's not listed as an sdv dependency and is probably installed by one of the other libraries. It seems like matplotlib and plotly require it and those are also not direct dependencies for SDV

amontanez24 (approved):
LGTM!

frances-h (approved):
Looks good! 👍

PR #1511, opened 2023-07-21
Reverse transform constraints later in the `DataProcessor`

### LINKED ISSUE #1476: Cannot use custom constraint transforms for certain columns (inconsistent ordering in forward vs. reverse)
### Environment Details
* SDV version: 1.2.0 (latest)
* Python version: Any
* Operating System: Any

### Error Description
Sometimes, I may have a primary key or PII column that is involved in a constraint. If I try to apply a custom constraint for it, it always falls back to the reject sampling strategy (using `is_valid`). It never uses the transform strategy (using `transform` and `reverse_transform`).

### Root Cause
During the Data Processor's reverse transform, we are handling constraints first -- even before the primary keys and anonymized columns are created. This means that the constraints cannot access primary keys or PII columns.

https://github.com/sdv-dev/SDV/blob/50392fad48e6df3553254b5dc72fd0dae2ece30e/sdv/data_processing/data_processor.py#L713-L732

The expected order is **to do the constraints last**.

_Note that the reverse should always be in the opposite order as the forward transform. The forward happens in this order: (constraints -> primary key -> PII columns). See [code](https://github.com/sdv-dev/SDV/blob/50392fad48e6df3553254b5dc72fd0dae2ece30e/sdv/data_processing/data_processor.py#L669-L679)_

fealho:
Resolve #1476.


sdv/data_processing/data_processor.py
```diff
@@ -720,6 +717,13 @@ def reverse_transform(self, data, reset_keys=False):
             generated_keys = self.generate_keys(num_rows, reset_keys)
             sampled_columns.extend(self._keys)
 
+        for constraint in reversed(self._constraints_to_reverse):
+            reversed_data = constraint.reverse_transform(reversed_data)
```
amontanez24, line 721: would it work if you just did sampled_columns = constraint.reverse_transform(sampled_columns)?

fealho: Sorry, I don't get what you mean, you can't apply `reverse_transform` to `sampled_columns`, since that's a list.

amontanez24: Yeah you're right, I got confused. Should we just set `sampled_columns` to be `reversed_data.columns` on line 725 instead of extending. Otherwise that list might contain columns that were created by constraints that we don't want

fealho: That was actually the first thing I tried. It doesn't work because `sampled_columns` has the keys, which `reversed_data.columns` doesn't have.

Your concern is actually not a problem because of line 731, it removes any columns not seen in the metadata.I originally had an extra two lines to remove these extra columns directly, for clarity, but I decided to remove them since they were unnecessary.

frances-h (approved):

sdv/data_processing/data_processor.py
amontanez24: That makes sense. Can you edit the comments before line 731 to add that extra columns will also be dropped?

amontanez24 (approved):
Just a comment about updating a comment. LGTM!

PR #1526, opened 2023-08-02
Allow empty tables to be fitted using `fit_processed_data`

### LINKED ISSUE #1524: Allow empty tables to be fitted using fit_processed_data
### Problem Description

_As a user, it would be great of multi-table synthesizers were able to fit and sample even when child tables don't have statistical columns as that is a common database structure._

### Expected behavior

- `fit_processed_data` should not crash if the dataframe passed is empty.
- If the data is empty, it should:
    - set  `_fitted` to True
    - set `_fitted_date`
    - set `_fitted_sdv_version`
- If the data is not empty, it should only set those parameters if `_fit` runs successfully

fealho:
Resolve #1524.

### DIFF
--- a/tests/unit/multi_table/test_base.py
+++ b/tests/unit/multi_table/test_base.py
@@ -717,6 +717,20 @@ class TestBaseMultiTableSynthesizer:
         instance._augment_tables.assert_called_once_with(data)
         instance._model_tables.assert_called_once_with(instance._augment_tables.return_value)
         assert instance._fitted
+    
+    def test_fit_processed_data_empty_table(self):
+        """Test the fit attributes are properly set when data is empty."""
+        # Setup
+        instance = Mock()
+        data = pd.DataFrame()
+
+        # Run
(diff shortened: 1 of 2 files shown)


tests/unit/multi_table/test_base.py
```diff
@@ -718,6 +718,20 @@ def test_fit_processed_data(self):
...
+    def test_fit_processed_data_empty_table(self):
+        """Test the fit attributes are properly set when data is empty."""
+        # Setup
+        instance = Mock()
+        data = pd.DataFrame()
+
+        # Run
+        BaseMultiTableSynthesizer.fit_processed_data(instance, data)
+
+        # Assert
```
pvk-developer, line 730: could we assert that `instance._fit` has not been called ?

pvk-developer (approved):
LGTM!

amontanez24 (commented):
Looks good! Same comment as @pvk-developer

amontanez24 (approved):
:shipit:

PR #1529, opened 2023-08-03
Improve metadata detection

### LINKED ISSUE #1515: Improve metadata detection
### Problem Description

_After using the metadata detection, I still have to make a lot of manual updates to make the metadata match the data. Some of these changes seem like they could be automated._

### Expected behavior

When calling [detect_from_dataframe](https://github.com/sdv-dev/SDV/blob/c4a52305da59ed589f9d044a696a9791ee2bd1cc/sdv/metadata/single_table.py#L254C20-L254C20) or [detect_from_csv](https://github.com/sdv-dev/SDV/blob/c4a52305da59ed589f9d044a696a9791ee2bd1cc/sdv/metadata/single_table.py#L274), we should now use the following logic in the steps defined below to assign sdtypes:

1. If `dtype==boolean`: The sdtype is categorical
2. If `dtype==pd.datetime` : The sdtype is always datetime. Don't assign a `datetime_format`
3. If  `dtype==int`: The sdtype should be one of: numerical, categorical or id. We can tell them apart based on cardinality (# of unique values).
    1. If there are <=5 rows, we can't properly detect this. Sdtype is numerical.
    2. If the cardinality is less than round(len / 10) (in other words, the number of unique categories is less than 10% the number of rows) and the values are all >=0, sdtype is categorical
    3. Elif there are null values present, sdtype is numerical. (Rationale: an "id" column cannot have null values)
    4. Elif the values are all unique, sdtype is id. Don't assign a `regex_format`.
    5. Else (Fallback) sdtype is numerical
4. If `dtype==float`: The sdtype should be one of: numerical or categorical.
    1. If there are <=5 rows, we can't properly detect this. Sdtype is numerical.
    2. If the cardinality is less than round(len / 10) and all the values are whole numbers (>=0) and there are NaN values, sdtype is categorical.
        - This is for the edge case when the column is supposed to be an int but it has nulls. Pandas does not support int columns with NaNs, so it represents it as a float.
    3. Else (Fallback) sdtype is numerical
5. If `dtype==object`: 
    1. If the values can be cast as a datetime with a consistent datetime format string: sdtype is datetime with the appropriate datetime_format assigned. Right now, we do not support datetime columns with inconsistent formats or with special values such as "12-31-9999". This one might be slow (especially when matching format), so we would probably require subsampling here.
    2. Elif the entire column can be cast to an int, cast it and the follow the logic of dtype==int
    3. Elif the entire column can be cast to a float, cast it and follow the logic of dtype==float
    4. Elif the values are all unique, sdtype is id and don't assign a `regex_format`. 
    5. Elif the cardinality is less than round(len / 5), sdtype is categorical.
    6. Else (Fallback) The sdtype is probably a pii type (we may learn later that it's actually a foreign key) sdtype: "unknown", pii: True
6. Any other dtype is unsupported. We should raise an error in this case

### Additional context

- We may also want to check if the `dtype == 'categorical'`. Right now we're ignoring it because [RDT seems to crash](https://github.com/sdv-dev/RDT/issues/617) with this dtype in some situations
- We are adding a new sdtype called `unknown`. Handling this type will be done in #1516.

R-Palazzo:
Resolve #1515
- Compared to the issue, I just added one condition for `Object` detection:
```
if len(data) <= 10:
    sdtype = 'categorical'
```
I did it because otherwise a lot of test we're failing due to columns detected as `id` rather than `categorical` now. I also feel that this follow a similar statement that we have for numbers which is:
`If there are <=5 rows, we can't properly detect this. Sdtype is numerical.
`
Let me know if this works @npatki 

- One other thing, I noticed there is a lot of `staticmethod` in `SingleTableMetadata`. Following the [google styleguide](https://google.github.io/styleguide/pyguide.html), do we want to move them outside of the class in this PR? @amontanez24, I did it first but then I realised it was confusing which changes are proper to this issue or just cleaning so I removed it. Let me know what we prefer.

### DIFF
--- a/sdv/metadata/single_table.py
+++ b/sdv/metadata/single_table.py
@@ -35,10 +35,8 @@ class SingleTableMetadata:
     }
 
     _DTYPES_TO_SDTYPES = {
-        'i': 'numerical',
-        'f': 'numerical',
-        'O': 'categorical',
-        'b': 'boolean',
+        'b': 'categorical',
+        'c': 'categorrical',
         'M': 'datetime',
     }
 
@@ -252,11 +250,89 @@ class SingleTableMetadata:
 
         return deepcopy(metadata)
 
+    def _determine_sdtype_for_numbers(self, data):
+        """Determine the sdtype for a numerical column.
+
+        Args:
+            data (pandas.Series):
+                The data to be analyzed.
+        """
+        sdtype = 'numerical'
+        if len(data) > 5:
+
+            whole_values = (data == data.round()).loc[~data.isna()].all()
+            positive_values = (data >= 0).loc[~data.isna()].all()
+
+            unique_values = data.nunique()
+            percentage_unique_values = unique_values <= round(len(data) / 10)
+
+            if whole_values and positive_values and percentage_unique_values:
+                sdtype = 'categorical'
+            elif unique_values == len(data) and whole_values:
+                sdtype = 'id'
+
+        return sdtype
+
+    def _determine_sdtype_for_objects(self, data):
+        """Determine the sdtype for an object column.
+
+        Args:
+            data (pandas.Series):
+                The data to be analyzed.
+        """
+        sdtype = None
+        try:
+            pd.to_datetime(data)
+            sdtype = 'datetime'
+
+        except Exception:
+            if len(data) <= 10:
+                sdtype = 'categorical'
+            else:
+                try:
+                    number_data = data.astype('float')
+                    sdtype = self._determine_sdtype_for_numbers(number_data)
+
+                except Exception:
+                    unique_values = data.nunique()
+                    if unique_values == len(data):
+                        sdtype = 'id'
+                    elif unique_values <= round(len(data) / 5):
+                        sdtype = 'categorical'
+                    else:
+                        sdtype = 'unknown'
+
+        return sdtype
+
     def _detect_columns(self, data):
         for field in data:
-            clean_data = data[field].dropna()
-            kind = clean_data.infer_objects().dtype.kind
-            self.columns[field] = {'sdtype': self._DTYPES_TO_SDTYPES.get(kind, 'categorical')}
+            column_data = data[field]
+            clean_data = column_data.dropna()
+            dtype = clean_data.infer_objects().dtype.kind
+
+            sdtype = None
+            if dtype in self._DTYPES_TO_SDTYPES:
+                sdtype = self._DTYPES_TO_SDTYPES[dtype]
+            elif dtype in ['i', 'u', 'f']:
+                sdtype = self._determine_sdtype_for_numbers(clean_data)
+
+            elif dtype == 'O':
+                sdtype = self._determine_sdtype_for_objects(column_data)
+
+            if sdtype is None:
+                raise InvalidMetadataError(
+                    f"Unsupported data type for column '{field}' ({dtype})."
+                    "The valid data types are: 'object', 'int', 'float', 'datetime', 'bool'."
+                )
+            column_dict = {'sdtype': deepcopy(sdtype)}
+
+            if sdtype == 'unknown':
+                column_dict['pii'] = True
+            elif sdtype == 'datetime':
+                datetime_format = get_datetime_format(clean_data)
+                column_dict['datetime_format'] = datetime_format
+
+            self.columns[field] = deepcopy(column_dict)
 
     def detect_from_dataframe(self, data):
         """Detect the metadata from a ``pd.DataFrame`` object.
--- a/sdv/metadata/single_table.py
+++ b/sdv/metadata/single_table.py
@@ -285,7 +285,7 @@ class SingleTableMetadata:
                 sdtype = 'datetime'
 
         except Exception:
-            if len(data) <= 10:
+            if len(data) <= 5:
                 sdtype = 'categorical'
             else:
                 unique_values = data.nunique()
--- a/sdv/metadata/single_table.py
+++ b/sdv/metadata/single_table.py
@@ -290,18 +290,13 @@ class SingleTableMetadata:
             if len(data) <= 10:
                 sdtype = 'categorical'
             else:
-                try:
-                    number_data = data.astype('float')
-                    sdtype = self._determine_sdtype_for_numbers(number_data)
-
-                except Exception:
-                    unique_values = data.nunique()
-                    if unique_values == len(data):
-                        sdtype = 'id'
-                    elif unique_values <= round(len(data) / 5):
-                        sdtype = 'categorical'
-                    else:
-                        sdtype = 'unknown'
+                unique_values = data.nunique()
+                if unique_values == len(data):
+                    sdtype = 'id'
+                elif unique_values <= round(len(data) / 5):
+                    sdtype = 'categorical'
+                else:
+                    sdtype = 'unknown'
 
         return sdtype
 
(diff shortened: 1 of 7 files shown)


amontanez24:
> Resolve #1515
> 
> * Compared to the issue, I just added one condition for `Object` detection:
> 
> ```
> if len(data) <= 10:
>     sdtype = 'categorical'
> ```
> 
> I did it because otherwise a lot of test we're failing due to columns detected as `id` rather than `categorical` now. I also feel that this follow a similar statement that we have for numbers which is: `If there are <=5 rows, we can't properly detect this. Sdtype is numerical. ` Let me know if this works @npatki
> 
> * One other thing, I noticed there is a lot of `staticmethod` in `SingleTableMetadata`. Following the [google styleguide](https://google.github.io/styleguide/pyguide.html), do we want to move them outside of the class in this PR? @amontanez24, I did it first but then I realised it was confusing which changes are proper to this issue or just cleaning so I removed it. Let me know what we prefer.

You can just leave the staticmethods. We can also alter our personal style guide to tolerate them

sdv/metadata/single_table.py
```diff
@@ -245,11 +244,92 @@ def to_dict(self):
...
+        """Determine the sdtype for an object column.
+
+        Args:
+            data (pandas.Series):
+                The data to be analyzed.
+        """
+        sdtype = None
+        try:
+            with warnings.catch_warnings():
+                warnings.simplefilter('ignore', category=UserWarning)
```
pvk-developer, line 280: What warnings are being shown here and why we ignore them? 🤔

sdv/metadata/single_table.py
```diff
@@ -245,11 +244,92 @@ def to_dict(self):
...
+                pd.to_datetime(data)
```
pvk-developer, line 281: There is one more requirement towards this: All `datetimes` should be in the same `format` . My suggestion to this is:
1. Detect the datetime format.
2. Cast to `pd.to_datetime(data, datetime_fromat=detected_format, errors='raise')`
3. If error is raised then we ignore the datetimes.

tests/integration/multi_table/test_hma.py
```diff
@@ -604,6 +604,21 @@ def test_use_own_data_using_hma(tmp_path):
...
+    metadata.update_column(
+        table_name='hotels',
+        column_name='state',
+        sdtype='categorical',
+    )
+    metadata.update_column(
+        table_name='hotels',
+        column_name='classification',
+        sdtype='categorical',
+    )
```
pvk-developer, line 621: Are we updating this because now they are no longer `categories` but `ids` ?

pvk-developer (changes requested):
I think we have to change something on the datetime but other than that looks good ;)

tests/integration/multi_table/test_hma.py
R-Palazzo: They were detected as `unknown` but it was mostly because the demo dataset here is small and so the condition `Elif the cardinality is less than round(len / 5), sdtype is categorical` was not reached. I discussed it with Neha and we decided that it was fine like this.

sdv/metadata/single_table.py
R-Palazzo: It was because we didn't give any `datetime_format` and we ignored it because the user couldn't do anything about it. But now I removed it after the changes of [4aa4a4a](https://github.com/sdv-dev/SDV/pull/1529/commits/4aa4a4ac2c96acb7a1fa5929cc5826512e128fc8)

R-Palazzo:
Thanks for your review @pvk-developer. I addressed the comments in [4aa4a4a](https://github.com/sdv-dev/SDV/pull/1529/commits/4aa4a4ac2c96acb7a1fa5929cc5826512e128fc8).

tests/integration/single_table/test_copulas.py
```diff
@@ -361,3 +362,35 @@ def test_numerical_columns_gets_pii():
...
     pd.testing.assert_frame_equal(expected_sampled, sampled)
+
+
+def test_categorical_column_with_numbers():
+    """Test that categorical column represented with numbers works end to end."""
+    # Setup
+    data = pd.DataFrame({
+        'category_col': [
+            1, 2, 1, 2, 1, 2, np.nan, 1, 1, np.nan, 2, 2, None, 2,
+            1, 1, None, 1, 2, 2
```
amontanez24, line 373: Just so you know, pandas converts the Nones here to np.nan which is why the dtype stays float. They aren't really necessary because of that. If ypou want the `None` values to stay you have to manually set the dtype to object

R-Palazzo: Yes I put `np.nan` and `None` now as a sanity check that everything works for both.

sdv/metadata/single_table.py
```diff
@@ -245,11 +244,90 @@ def to_dict(self):
...
+        """Determine the sdtype for an object column.
+
+        Args:
+            data (pandas.Series):
+                The data to be analyzed.
+        """
+        sdtype = None
+        try:
+            datetime_format = get_datetime_format(data)
+            pd.to_datetime(data, format=datetime_format, errors='raise')
```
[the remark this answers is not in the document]
R-Palazzo: I thought that we need to ensure here that all the data is convertible with the same datetime format. Should we subsample? I didn't look at the computational time.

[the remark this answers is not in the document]
amontanez24: The goal is to check that everything matches, but since it can possibly be slow for very large datasets, we can probably pick a max number of rows and say if at least that many match it's probably fine

sdv/metadata/single_table.py
```diff
@@ -247,11 +246,96 @@ def to_dict(self):
...
+
+            elif dtype == 'O':
+                sdtype = self._determine_sdtype_for_objects(column_data)
+
+            if sdtype is None:
+                raise InvalidMetadataError(
+                    f"Unsupported data type for column '{field}' (kind: {dtype})."
+                    "The valid data types are: 'object', 'int', 'float', 'datetime', 'bool'."
+                )
+            column_dict = {'sdtype': deepcopy(sdtype)}
```
amontanez24, line 330: minor: do we need to deepcopy here? Shouldn't sdtype be a string?

sdv/metadata/single_table.py
```diff
@@ -247,11 +246,96 @@ def to_dict(self):
...
+
+            if sdtype == 'unknown':
+                column_dict['pii'] = True
+            elif sdtype == 'datetime':
+                datetime_format = get_datetime_format(clean_data)
+                column_dict['datetime_format'] = datetime_format
```
amontanez24, line 336: I think we said we'd only set the datetime format in the case where we parsed it from the text. In other words
- If the dtype is datetime
    - make the `sdtype` datetime and the `datetime_format` None
- If the dtype is object
    - if it can be converted to a datetime set the `sdtype` to datetime and set the `datetime_format`

You can maybe have the method that parses the datetime return the format if it finds it or something like that

tests/integration/single_table/test_copulas.py
amontanez24: No I mean that even though you put both in here, pandas converts the Nones to np.nan behind the scenes so None is not actually being tested. To test it you have to set the dtype to object

sdv/metadata/single_table.py
R-Palazzo: alright I see thanks, done in [58e4fcf](https://github.com/sdv-dev/SDV/pull/1529/commits/58e4fcf49eff5120b51b460bd92b594371086c43)

frances-h (approved):
Looks good!

PR #1532, opened 2023-08-07
Support 'unknown' sdtype

### LINKED ISSUE #1516: Support 'unknown' sdtype
### Problem Description

The metadata updates in #1515 will add a new sdtype called `unknown`. This sdtype is when we were unable to pick an sdtype and a generic faker should be used.

### Expected behavior

- When assigning transformers, if the `sdtype` is `unknown`, assign the following transformer
```python
AnonymizedFaker(
    provider_name=None,
    function_name='bothify',
    text='sdv-pii-?????',
    letters='0123456789abcdefghijklmnopqrstuvwxyz'
)
```

### Additional context

- The logic for assigning transformers can be found [here](https://github.com/sdv-dev/SDV/blob/c4a52305da59ed589f9d044a696a9791ee2bd1cc/sdv/data_processing/data_processor.py#L459-L486). We will no longer need the case where we default to categorical transformers if all else fails since every column should have an sdtype or unknown.

R-Palazzo:
Resolve #1516


frances-h (commented):
Should we also change the `else` case on line 149 that defaults to `categorical` now that we're adding the `unknown` sdtype?

pvk-developer (approved):

R-Palazzo:
> Should we also change the `else` case on line 149 that defaults to `categorical` now that we're adding the `unknown` sdtype?

Hi @frances-h, in which file is it? In this [PR](https://github.com/sdv-dev/SDV/pull/1529) we change the metadata detection to detect `unknown` sdtype.

frances-h (approved):

amontanez24 (approved):
LGTM!

PR #1535, opened 2023-08-09
Add SingleTableMetadata visualization

### LINKED ISSUE #1517: Add visualize method to SingleTableMetadata
### Problem Description

_After detecting metadata, it would be helpful to see a visual representation of the sdtypes for each column in the table._

Currently, `visualize` only exists on the `MultiTableMetadata` class, but a single table version could exist. It would be the same except for the relationship labels.

### Expected behavior

- Add `visualize` method to `SingleTableMetadata`. It should be the same as the multi table version except without having to label or draw the relationships.
- It doesn't require a `show_label_parameter`.

### Additional context
- The `show_table_details` parameter should have three options: None, 'full' and 'summarized'. If #1525 isn't completed before this issue, then implement it to work with those options as opposed to True or False. Otherwise it should be easy to copy.

pvk-developer:
Resolves #1517

### DIFF
--- a/sdv/metadata/single_table.py
+++ b/sdv/metadata/single_table.py
@@ -4,6 +4,7 @@ import json
 import logging
 import re
 import warnings
+from collections import defaultdict
 from copy import deepcopy
 from datetime import datetime
 
@@ -11,6 +12,7 @@ from sdv.metadata.anonymization import SDTYPE_ANONYMIZERS, is_faker_function
 from sdv.metadata.errors import InvalidMetadataError
 from sdv.metadata.metadata_upgrader import convert_metadata
 from sdv.metadata.utils import read_json, validate_file_does_not_exist
+from sdv.metadata.visualization import visualize_graph
 from sdv.utils import cast_to_iterable, load_data_from_csv
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


sdv/metadata/single_table.py
```diff
@@ -477,6 +479,60 @@ def validate(self):
                 + '\n'.join([str(e) for e in errors])
             )
 
+    def visualize(self, show_table_details='full', output_filepath=None):
+        """Create a visualization of the multi-table dataset.
```
fealho, line 483: single-table

sdv/metadata/single_table.py
```diff
@@ -477,6 +479,60 @@ def validate(self):
...
+
+        Args:
+            show_table_details (str):
+                If 'full', show the column names, primary, alternate and sequence keys are all
```
fealho, line 487: remove `show` after `If 'full'`. Also, review the docstrings, they mention table_names when I don't think we have them for single_table.

Also, shouldn't None be an option?

sdv/metadata/single_table.py
```diff
@@ -477,6 +479,60 @@ def validate(self):
...
+        """
+        if show_table_details not in ('full', 'summarized'):
+            raise ValueError("'show_table_details' should be 'full' or 'summarized'.")
+
+        if show_table_details == 'full':
+            columns = [
+                fr"{name} : {meta.get('sdtype')}\l"
+                for name, meta in self.columns.items()
+            ]
+            node = ''.join(columns)
```
amontanez24, line 505: do you think we could pull these sections out into helper functions in the `visualize.py` file and then the multi-table and single table metadata could use them? They seem pretty similar.

sdv/metadata/single_table.py
pvk-developer: I did not add it because it didn't make sense (empty box ?) ... In multi table I see a point of `None` since you can see the diagram.

sdv/metadata/single_table.py
pvk-developer: Its slightly different, in this case we always add `\l` meanwhile on multi-table we use `'\l'.join()` which skips one.

fealho (approved):

amontanez24 (approved):
LGTM!

PR #1538, opened 2023-08-14
Use IDGenerator for ID columns

### LINKED ISSUE #1519: Use IDGenerator for ID columns
### Problem Description

_As a user, I'd like to be able to generate id columns that don't all start at 0 and follow the same pattern so that my data could be more realistic._

### Expected behavior

Once the [IDGenerator is added to RDT](https://github.com/sdv-dev/RDT/issues/675), we should use it in SDV.

- If a column is a primary or alternate key, use the following transformer. Set prefix to none if the column is nan int, otherwise set it to "sdv-id-":

```python
IDGenerator(
    prefix="sdv-id-", # if it's an int, no prefix
    starting_value=0,
    suffix=None
)
```
- If an ID is a foreign key, then it should have no assignment (status quo)
- If an ID is NOT a key, then the Data Processor can assign the `AnonymizedFaker`
```python
AnonymizedFaker(
    provider_name=None,
    function_name="bothify",
    text="#####"
)
```

### Additional context
- The code for adding key transformers is here:
https://github.com/sdv-dev/SDV/blob/c4a52305da59ed589f9d044a696a9791ee2bd1cc/sdv/data_processing/data_processor.py#L459-L467

frances-h:
Resolve #1519


sdv/data_processing/data_processor.py
```diff
@@ -458,13 +458,28 @@ def _create_config(self, data, columns_created_by_constraints):
...
-                sdtypes[column] = 'text'
+                if column_metadata.get('regex_format', False):
+                    transformers[column] = self.create_regex_generator(
+                        column,
+                        sdtype,
+                        column_metadata,
+                        is_numeric
+                    )
+                    sdtypes[column] = 'text'
+                elif column == self.metadata.primary_key or column in self.metadata.alternate_keys:
```
amontanez24, line 469: minor: I think you can just check if it's in `self._keys`

amontanez24 (approved):
LGTM!

sdv/data_processing/data_processor.py
```diff
@@ -458,13 +458,28 @@ def _create_config(self, data, columns_created_by_constraints):
...
-                )
-                sdtypes[column] = 'text'
+                if column_metadata.get('regex_format', False):
+                    transformers[column] = self.create_regex_generator(
+                        column,
+                        sdtype,
+                        column_metadata,
+                        is_numeric
+                    )
+                    sdtypes[column] = 'text'
```
fealho, line 468: We usually put a new line after if/else blocks (except if they are only one line).

fealho (approved):
Looks good 👍

PR #1540, opened 2023-08-15
Add `validate_data` for single table

### LINKED ISSUE #1518: Add validate_data method to Metadata
### Problem Description

_As a user, it would be convenient to validate my metadata against a dataset without having to make a synthesizer._

Currently the code for validating data is in the [synthesizer](https://github.com/sdv-dev/SDV/blob/c4a52305da59ed589f9d044a696a9791ee2bd1cc/sdv/single_table/base.py#L218). We can move any code that is specific to metadata validation out into the metadata class itself.

### Expected behavior

- Add `validate_with_data` method to metadata.
- Parameters:
    - `data`:
        - For single table, a pandas.DataFrame with the data
        - For multi table, a dictionary mapping each table name to a pandas.DataFrame

### Additional context

- The synthesizer validation has some code that is specific to certain synthesizers, as well as constraint validation. This method only needs to validate things that can be found in the metadata alone (ie. `_validate_keys_dont_have_missing_values`, `_validate_key_values_are_unique`, `_validate_metadata_matches_data`, `_valdiate_column`).

fealho:
Resolve #1518 for single-table.


frances-h (approved):
LGTM 👍

fealho:
@amontanez24 Wouldn't it be better to call this `validate_data`, instead of `validate_with_data`? You are validating the data matches the metadata, so the `with` doesn't seem necessary (it's also harder to remember, `validate_data` just seems a little more intuitive).

amontanez24:
> @amontanez24 @npatki Wouldn't it be better to call this `validate_data`, instead of `validate_with_data`? You are validating the data matches the metadata, so the `with` doesn't seem necessary (it's also harder to remember, `validate_data` just seems a little more intuitive).

I'm not against it

npatki:
In this method, we _won't_ validate the metadata format right? ie. the checks in the existing `metadata.validate()` and this method are completely different?

If so, then `validate_data()` makes sense to me, as we are only validating that the data matches the metadata. If we all agree, I suggest we update the issue to match!

sdv/metadata/single_table.py
```diff
@@ -471,14 +478,173 @@ def validate(self):
...
+            invalid_values = self._format_invalid_values_string(invalid_values)
+            return [f"Invalid values found for {sdtype} column '{column.name}': {invalid_values}."]
+
+        return []
+
+    def _validate_constraints(self, data):
+        """Validate that the data satisfies the constraints."""
+        errors = []
+        try:
+            self._data_processor._fit_constraints(data)
```
amontanez24, line 609: the data_processor doesn't exist in the metadata. This shouldn't work. I think this got accidentally copied over

sdv/metadata/single_table.py
```diff
@@ -471,14 +478,173 @@ def validate(self):
...
+            if pd.isna(data[key]).any():
+                errors.append(f"Key column '{key}' contains missing values.")
+
+        return errors
+
+    @staticmethod
+    def _format_invalid_values_string(invalid_values):
+        invalid_values = sorted(invalid_values, key=lambda x: str(x))
+        if len(invalid_values) > 3:
+            return invalid_values[:3] + [f'+ {len(invalid_values) - 3} more']
```
amontanez24, line 549: A similar thing happens here:
https://github.com/sdv-dev/SDV/blob/6480f1f066ef302d718d147887adda3b588a495f/sdv/constraints/base.py#L204-L210

Maybe we can generalize this function into something that takes a dataframe and number of rows, and returns the string with the top x rows + y more as a string. Then we can move it into utils

amontanez24 (changes requested):
This looks really good! Couple of small comments

sdv/metadata/single_table.py
fealho: Yeah, this method doesn't get called, I copied it by accident. Good catch!

amontanez24 (approved):
Thanks for addressing the changes! :shipit:

PR #1541, opened 2023-08-15
Add `validate_data` for multi-table

### LINKED ISSUE #1518: Add validate_data method to Metadata

fealho:
Resolve #1518 for multi table.


sdv/metadata/multi_table.py
```diff
@@ -532,6 +533,79 @@ def validate(self):
...
+                    )
+
+            if errors:
+                error_msg = 'Relationships:\n'
+                error_msg += '\n'.join(errors)
+
+        return [error_msg] if error_msg else []
+
+    def validate_with_data(self, data):
+        """Validate the data matches the metadata."""
```
fealho, line 600: Note: improve docstring

sdv/metadata/multi_table.py
```diff
@@ -532,6 +533,79 @@ def validate(self):
...
 
+    def _validate_missing_tables(self, data):
+        errors = []
+        missing_tables = set(self.tables) - set(data)
+        if missing_tables:
+            errors.append(f'The provided data is missing the tables {missing_tables}.')
+
+        return errors
+
+    def _validate_all_tables(self, data, table_synthesizers=None):
```
frances-h, line 544: It might be helpful to add a docstring here explaining what this validation method checks for. Either here or in the docstring for the `validate_data` method. Same with `_validate_foreign_keys`.

sdv/metadata/multi_table.py
```diff
@@ -532,6 +533,79 @@ def validate(self):
...
+        errors = []
+        for table_name, table_data in data.items():
+            try:
+                if table_synthesizers:
+                    table_synthesizers[table_name].validate(table_data)
```
amontanez24, line 549: I think we should keep this out of the metadata. It's a strange cyclic relationship. I think we can just do the case in the else and then do the case in the if only in the `MultiTableSynthesizer`. That would also make it clear why `MultiTableMetadata.validate_data` isn't being called in the synthesizer.

fealho: The issue of doing that is the code gets duplicated, which is why I took this approach. It is confusing though, so I guess I'll just leave the duplicated code there.

sdv/multi_table/base.py
```diff
@@ -233,6 +194,14 @@ def _validate_table_name(self, table_name):
         if table_name not in self._table_synthesizers:
             raise InvalidDataError([f"Table '{table_name}' is not present in the metadata."])
 
+    def _get_all_foreign_keys(self, table_name):
```
amontanez24, line 197: why do we need this method here? It's already in the metadata

sdv/multi_table/base.py
```diff
@@ -201,30 +183,9 @@ def validate(self, data):
...
-
-            except KeyError:
-                continue
-
-        foreign_key_errors = self._validate_foreign_keys(data)
-        if foreign_key_errors:
-            errors.append(foreign_key_errors)
+        errors += self.metadata._validate_missing_tables(data)
+        errors += self._validate_all_tables(data)
+        errors += self.metadata._validate_foreign_keys(data)
```
amontanez24, line 188: alternatively you could call `self.metadata.validate` and then make a separate method that loops through the synthesizers and calls the constraint validation as well as the synthesizer specific validation

sdv/multi_table/base.py
fealho: I'm just moving it down the file because it gets used by the method below, so it was misplaced up there.

amontanez24: why doesn't the method below just use the metadata's version of the method?

amontanez24 (approved):
I think this looks good!

frances-h (approved):
Looks good! 👍

PR #1546, opened 2023-08-15
Update `discrete_detect_columns` logic

### LINKED ISSUE #1450: Inefficient CTGAN modeling when adding categorical transformers
### Environment Details

* SDV version: 1.1.0 (latest)
* Python version: Any
* Operating System: Any

### Error Description
As described in #1433: In order to improve the speed of CTGANSynthesizer (and test quality), I may want to apply categorical pre-processing such as `LabelEncoder`. But even when I do this, the underlying CTGAN model is still treating these values as discrete, categorical values.

### Replication
```python
from rdt.transformers.categorical import LabelEncoder

    modality='single_table',
)

synthesizer = CTGANSynthesizer(metadata)
    'room_type': LabelEncoder(add_noise=True)
})

synthesizer.fit(data)
```

While nothing crashes, it's clear from inspecting the CTGAN ML model that is still treating `room_type` as a discrete column even though it's not!

### Context
We use the method [detect_discrete_columns](https://github.com/sdv-dev/SDV/blob/master/sdv/single_table/utils.py#L15) to identify which columns are categorical, and then pass them along in [CTGAN's fit function](https://github.com/sdv-dev/SDV/blob/master/sdv/single_table/ctgan.py#L117).

The detect_discrete_columns make some assumptions that are not true:

- The metadata can be used to identify what should count as discrete
- The data must match the metadata

These assumptions are not true because if the user has assigned transformers themselves, then the output of those transformers may have changed which columns are discrete. Mainly:

Applying a categorical transformer would change a discrete column into a numerical column

I'm not sure how, but this should be factored in.

### Workaround
In the meantime, it is always possible to use the [RDT library](https://docs.sdv.dev/rdt) outside of the SDV to do your own pre/post processing. Something like this:

```python

from rdt import HyperTransformer
from rdt.transformers.categorical import LabelEncoder

# preprocess the data ourselves using the RDT library
ht = HyperTransformer()
ht.detect_initial_config(df)
ht.update_transformers(column_name_to_transformer={
    'column_name': LabelEncoder(add_noise=True)
})
use_df = ht.fit_transform(df) 

# SDV can just handle the processed, numerical ata
    'columns': {
        'column_name': { 'sdtype': 'numerical' }
    }
})

synthesizer = CTGANSynthesizer(metadata, epochs=1)
synthesizer.fit(use_df)
raw_synthetic_data = synthesizer.sample(num_rows=10)

# post process using the RDT
synthetic_data = ht.reverse_transform(raw_synthetic_data)
```

fealho:
Resolve #1450.

### DIFF
--- a/sdv/single_table/utils.py
+++ b/sdv/single_table/utils.py
@@ -43,12 +43,21 @@ def detect_discrete_columns(metadata, data):
         if column_data.empty:
             continue
 
-        # Non-integer floats cannot be categorical
+        # Non-integer floats and integers with too many unique values cannot be categorical
         try:
             column_data = column_data.astype('float')
-            if list(column_data) != list(column_data.round()):
+            rounded_column_data = list(column_data.round())
+            column_data = list(column_data)
+            is_float = column_data != rounded_column_data
+            is_int = column_data == rounded_column_data
+            num_values = len(column_data)
+            num_categories = len(set(column_data))
+            threshold = max(10, num_values * .1)
+            has_many_categories = num_categories > threshold
+            if is_float or (is_int and has_many_categories):
                 continue
-        except BaseException:
+
+        except (ValueError, TypeError):
             pass
 
         # Everything else is presumed categorical
--- a/sdv/single_table/utils.py
+++ b/sdv/single_table/utils.py
@@ -15,6 +15,10 @@ IGNORED_DICT_KEYS = ['fitted', 'distribution', 'type']
 def detect_discrete_columns(metadata, data):
     """Detect the discrete columns in a dataset.
 
+    Because the metadata doesn't necessarily match the data (we only preprocess the data,
+    while the metadata stays static), this method tries to infer whether the data is
+    discrete.
+
     Args:
         metadata (sdv.metadata.SingleTableMetadata):
             Metadata that belongs to the given ``data``.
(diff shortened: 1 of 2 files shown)


sdv/single_table/utils.py
```diff
@@ -27,24 +31,28 @@ def detect_discrete_columns(metadata, data):
...
+            column_data = column_data.astype('float')
+            if list(column_data) != list(column_data.round()):
+                continue
+        except BaseException:
+            pass
 
-            except Exception:
-                discrete_columns.append(column)
+        # Everything else is presumed categorical
+        discrete_columns.append(column)
```
amontanez24, line 55: I think we also wanted to make sure the number of unique values is less than a threshold percentage of the total number of values in the column. I think we were going to use 10 or 20% as the threshold. Right @npatki

sdv/single_table/utils.py
```diff
@@ -27,24 +31,37 @@ def detect_discrete_columns(metadata, data):
...
+        # Non-integer floats and integers with too many unique values cannot be categorical
+        try:
+            column_data = column_data.astype('float')
+            rounded_column_data = list(column_data.round())
+            column_data = list(column_data)
+            is_float = column_data != rounded_column_data
+            is_int = column_data == rounded_column_data
+            num_values = len(column_data)
+            num_categories = len(set(column_data))
+            threshold = max(10, num_values * .1)
```
R-Palazzo, line 55: Why `max(10, num_values * .1)` here and not `num_values * .1`?

fealho: The main reason not to treat it as categorical is because it's very slow for many categories, because CTGAN one hot encodes them. If you have less than 10 categories, that's not a problem, since you will have at most 10 extra columns, and your dataset is very small (< 100 rows). So I decided to cap it, if your data has less than 10 unique values, it will always be treated as categorical.

R-Palazzo: I agree, thanks for the explanation

sdv/single_table/utils.py
```diff
@@ -27,24 +31,37 @@ def detect_discrete_columns(metadata, data):
...
 
-            except Exception:
-                discrete_columns.append(column)
+        # Non-integer floats and integers with too many unique values cannot be categorical
+        try:
+            column_data = column_data.astype('float')
+            rounded_column_data = list(column_data.round())
+            column_data = list(column_data)
+            is_float = column_data != rounded_column_data
+            is_int = column_data == rounded_column_data
```
amontanez24, line 52: I don't think we need to compute `is_float` and `is_int` since one is the opposite of the other. Also is it more efficient to do these comparisons as lists than dataframes/series? I would've that it would be faster the other way

sdv/single_table/utils.py
```diff
@@ -27,24 +31,37 @@ def detect_discrete_columns(metadata, data):
...
+            num_values = len(column_data)
+            num_categories = len(set(column_data))
```
amontanez24, line 54: is this faster than series.nunique?

amontanez24 (approved):
LGTM!

R-Palazzo (approved):

PR #1555, opened 2023-08-22
Updating rdt and cleaning up install_requires

amontanez24:


setup.py
```diff
@@ -13,6 +13,9 @@
 
 
 install_requires = [
+    'boto3>=1.15.0,<2',
+    'botocore>=1.18,<2',
+    'cloudpickle>=2.1.0,<3.0',
```
amontanez24, line 18: I wanted to alphabetize these to make it easier to parse. I left our own libraries at the bottom but I can alphabetize them as well if we think that's better

frances-h (approved):

PR #1560, opened 2023-08-28
Detect relationships in MultiTableMetadata

### LINKED ISSUE #1522: Detect relationships in MultiTableMetadata
### Problem Description

_As a user, it would be helpful to have the metadata attempt to detect relationships in my dataset because it takes a while to add them manually._

Once #1520 is done, we will have the ability to see all the tables in a dataset at once and infer relationships.

### Expected behavior

- When someone calls `detect_from_dataframes` or `detect_from_csvs`, attempt to detect the relationships at the end of the method by using the following logic:

    1. For the primary key of each table, check all the other tables to see if there is a column with the same name. If there is, assume it is a foreign key.
    2. If the foreign key doesn't create a circular dependency, add the relationship.
    3. If the foreign key isn't marked as sdtype `id`, set it as that.
- After discovering the relationships, raise a warning if the schema is still disconnected.  
- Add a `remove_relationship` method since some of the detected relationships might be wrong.
    - Parameters:
        - `parent_table_name`: The string name of the parent table. If provided, we'll only remove relationships where this table is a parent.
        - `child_table_name`: The string name of the child table. If provided, we'll only remove relationships where this table is a child

### Additional context

- This will depend on #1521 since we'll be able to extract the primary keys from there.

frances-h:
Resolve #1522

### DIFF
--- a/sdv/metadata/multi_table.py
+++ b/sdv/metadata/multi_table.py
@@ -384,24 +384,16 @@ class MultiTableMetadata:
         if len(nodes) == 1:
             return
 
-        queue = [nodes[0]]
-        explored = set()
+        parent_nodes = list(parent_map.keys())
+        queue = [parent_nodes[0]] if parent_map else []
         connected = {table_name: False for table_name in nodes}
 
         while queue:
             node = queue.pop()
-            explored.add(node)
-            children = list(child_map[node]) + list(parent_map[node])
-            if children:
-                connected[node] = True
-                for child in children:
-                    if not connected[child] and child not in queue:
-                        queue.append(child)
-            elif not queue:  # This was an orphan node, pick a new eligible node
-                for eligible_node in nodes:
-                    if eligible_node not in explored:
-                        queue.append(eligible_node)
-                        break
+            connected[node] = True
+            for child in list(child_map[node]) + list(parent_map[node]):
+                if not connected[child] and child not in queue:
+                    queue.append(child)
 
         if not all(connected.values()):
             disconnected_tables = [table for table, value in connected.items() if not value]
--- a/sdv/metadata/multi_table.py
+++ b/sdv/metadata/multi_table.py
@@ -277,9 +277,9 @@ class MultiTableMetadata:
         relationships_to_remove = []
         for relation in self.relationships:
             if (relation['parent_table_name'] == parent_table_name and
-                relation['child_table_name'] == child_table_name):
+                    relation['child_table_name'] == child_table_name):
                 relationships_to_remove.append(relation)
-        
+
         if not relationships_to_remove:
             warning_msg = (
                 f"No existing relationships found between parent table '{parent_table_name}' and "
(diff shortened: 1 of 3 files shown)


sdv/metadata/multi_table.py
```diff
@@ -333,6 +368,87 @@ def _log_detected_table(single_table_metadata):
...
+        if len(nodes) == 1:
+            return
+
+        queue = [nodes[0]]
+        explored = set()
+        connected = {table_name: False for table_name in nodes}
+
+        while queue:
+            node = queue.pop()
+            explored.add(node)
```
amontanez24, line 393: I assume this was added because it's possible to start on a node with no neighbors right? If so can we add a unit test case for this situation?

sdv/metadata/multi_table.py
```diff
@@ -333,6 +368,87 @@ def _log_detected_table(single_table_metadata):
...
+            if children:
+                connected[node] = True
+                for child in children:
+                    if not connected[child] and child not in queue:
+                        queue.append(child)
+            elif not queue:  # This was an orphan node, pick a new eligible node
+                for eligible_node in nodes:
+                    if eligible_node not in explored:
+                        queue.append(eligible_node)
+                        break
```
amontanez24, line 404: I think we could achieve this if we picked are starting node from the parent map right? Then it's guaranteed to be valid

tests/unit/metadata/test_multi_table.py
```diff
@@ -613,6 +613,90 @@ def test_add_relationship_child_key_is_primary_key(self):
...
+                'parent_primary_key': 'id',
+                'child_foreign_key': 'alternate_id',
+            }
+        ]
+
+        # Run
+        instance.remove_relationship('users', 'sessions')
+
+        # Assert
+        instance.relationships == []
```
amontanez24, line 660: I think we have to add `assert`

tests/unit/metadata/test_multi_table.py
```diff
@@ -613,6 +613,90 @@ def test_add_relationship_child_key_is_primary_key(self):
...
+                'child_table_name': 'sessions',
+                'parent_primary_key': 'id',
+                'child_foreign_key': 'user_id',
+            },
+            {
+                'parent_table_name': 'users',
+                'child_table_name': 'sessions',
+                'parent_primary_key': 'id',
+                'child_foreign_key': 'alternate_id',
+            }
```
amontanez24, line 653: Just for sanity, can we add another relationship that doesn't match the parent and child table to make sure it isn't removed

tests/unit/metadata/test_multi_table.py
```diff
@@ -1752,6 +1836,119 @@ def test_update_column_table_does_not_exist(self):
...
+                'child_table_name': 'sessions',
+                'parent_primary_key': 'user_id',
+                'child_foreign_key': 'user_id'
+            }
+        ]
+        assert instance.relationships == expected_relationships
+        assert instance.tables['sessions'].columns['user_id']['sdtype'] == 'id'
+
+    @patch('sdv.metadata.multi_table.warnings')
+    def test__detect_relationships_missing_warning(self, warnings_mock):
```
amontanez24, line 1880: nitpick: can we replace missing with disconnected or something like that?

sdv/metadata/multi_table.py
frances-h: Yeah that makes sense, it also makes it easier to handle the situation where no tables are connected.

amontanez24: Actually after looking again, I'm wondering if it's a problem that we start on a node with no neighbors. If that happens then the check below will still fail since the rest of the tables aren't connected to the first one right?

pvk-developer (approved):
LGTM!

sdv/metadata/multi_table.py
frances-h: It does fail, but I think the error message is confusing. If we have just 1 orphan table and we pick that as the starting node, the error message we get says that every other table is not connected to any of the other tables. If we don't have any relationships, I think it makes sense to list every table as not connected to any of the others.

amontanez24 (approved):
:shipit:

PR #1562, opened 2023-08-29
Add graph traversal to HMA

### LINKED ISSUE #1558: KeyError while executing synthesizer.sample() on HMA Synthesizer.
### Environment Details

* SDV version: 1.3.0
* Python version: 3.8
* Operating System: Windows 11

### Error Description

I got KeyError message when I was trying to fit the HMA Synthesizer on multi table setup. I had a parent table (say A), it had a child table(say B) and B also has a child table(say C).

Like 
A(parent table) 
B(child of A but parent of C)
C(child of B)  

The KeyError message was that it couldn't identify the key 'C' i.e. the last child table.
Thank you

```
Paste the command(s) you ran and the output.
If there was a crash, please include the traceback here.
```

fealho:
Resolve #1558.

This PR adds the ability for HMA to traverse a graph (it used to only handle trees with parent-children relationships).


fealho:
There's probably a way to clean up the traversal logic if we are willing to redesign HMA a bit.

fealho:
There are 3 simple ways to update the code to traverse the graph properly. Below I'll describe why I chose approach (3).

1) Simply add a recursion over the children in the `_sample_table` method. The problem with this approach is that if a child has multiple parents, we need to go through those parents first, so the traversal wouldn't work for multi-parent nodes.

2) BFS. If we wanted to keep the spirit of sampling hierarchy, BFS is the obvious choice. It can run into the same problem as (1) though, and we have to get rid of the `_sample_table` method, since we would just be using a queue and solving the problem iteratively. I actually implemented it before deciding to scratch it and use approach (3).

3) The actual implementation, where we pick a node and go over its parents, and only when a node has no parents left we iterate over the children. While it doesn't quite follow the spirit of the name "HierarchicalSampler", I picked this approach because: (1) it works for any graph, (2) it's simple to update it if we decide to let HMA handle multi-parent relationships in the future, (3) it's very similar to the implementation in the "IndependentSampler", so it should be easier to combine their logic in the future (if we decide to do so).

sdv/sampling/hierarchical_sampler.py
```diff
@@ -113,32 +113,32 @@ def _add_child_rows(self, child_name, parent_name, parent_row, sampled_data):
...
-            num_rows (int):
-                Number of rows to sample for the table.
             sampled_data (dict):
                 A dictionary mapping table names to sampled tables (pd.DataFrame).
         """
-        LOGGER.info(f'Sampling {num_rows} rows from table {table_name}')
+        for parent_name in self.metadata._get_parent_map()[table_name]:
+            if parent_name not in sampled_data:
+                self._sample_table(table_name=parent_name, sampled_data=sampled_data)
+                return  # Optimization to avoid iterating through the same nodes multiple times
```
frances-h, line 130: I think for HMA, it's simpler to skip this check and instead call `_sample_table` for each root node after we sample rows for it. If we have a child we've already sampled, we can skip sampling it (and any of its children) again.

fealho: That runs into the problem I described in (1) [here](https://github.com/sdv-dev/SDV/pull/1562#issuecomment-1707217856), no?

frances-h: +1, for HMA we only use one parent to sample all the rows, then use likelihoods to add foreign keys for any other parents

fealho: Theoretically only 1. The problem with this is, since we are doing DFS, all that really matters is the first root parent (depending on the graph ofc). So we'll first sample everything possible using the first root parent, and only then will we check if there are any extra parents. This is bad because (1) we ideally want all the parents to have influence over the graph and (2) the deeper the branch, the further we get from the original synthesizer in the root parent, so I'd imagine the quality of the results would deteriorate. I'm not sure how important this is in practice, but it's just something to keep in mind.

Because of that I tried the BFS approach (the approach (2) I explained): it's doing this same thing but actually going hierarchically, which minimizes both of those problems.

The other problem is that, if in the future we want to handle multiple foreign keys, this code would have to be completely rewritten, which is why I described approach (3).

amontanez24: Why would the original approach not be able to handle foreign keys?

fealho: You wouldn't have sampled the parents first. If you DFS from one root node, it's possible that the parents of the nodes you encounter along the way have not been sampled yet. So you need to do what I'm doing here, which is going up the graph whenever you encounter a non-sampled parent. And if we are going to be doing that, then there's no point in having that for loop over all the root nodes which we will find either way.

A question though, why would we choose approach (1) instead of (3)? From how I see it, (3) is just a slightly better version of (1).

frances-h (approved):
Looks good 👍

amontanez24 (approved):

PR #1564, opened 2023-08-30
Pandas release causes `is_datetime_type` to fail

fealho:
`pandas 2.1.0` new release makes the `is_datetime64_any_dtype` method fail in the presence of hyphens, which we use to test for datetime values (eg "2020-01-01"). This PR removes that method and improves the logic of `is_datetime_type` to fix this issue.

### DIFF
--- a/sdv/utils.py
+++ b/sdv/utils.py
@@ -1,6 +1,5 @@
 """Miscellaneous utility functions."""
 from collections.abc import Iterable
-from datetime import datetime
 from pathlib import Path
 
 import pandas as pd
@@ -84,12 +83,7 @@ def is_datetime_type(value):
     if isinstance(value, Iterable) and not isinstance(value, str):
         value = get_first_non_nan_value(value)
 
-    return (
-        pd.api.types.is_datetime64_any_dtype(value)
-        or isinstance(value, pd.Timestamp)
-        or isinstance(value, datetime)
-        or bool(get_datetime_format([value]))
-    )
+    return bool(get_datetime_format([value]))
 
 
 def is_numerical_type(value):
--- a/sdv/utils.py
+++ b/sdv/utils.py
@@ -1,5 +1,6 @@
 """Miscellaneous utility functions."""
 from collections.abc import Iterable
+from datetime import datetime
 from pathlib import Path
 
 import pandas as pd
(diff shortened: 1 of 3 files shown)


tests/unit/test_utils.py
```diff
@@ -187,26 +187,6 @@ def test_is_datetime_type_with_invalid_str():
     assert is_datetime is False
 
 
-def test_is_datetime_type_with_string():
```
fealho, line 190: There's a test exactly like this one above.

sdv/utils.py
```diff
@@ -73,23 +73,34 @@ def get_datetime_format(value):
...
-        value (pandas.DataFrame, int, str or datetime):
+        value (array-like iterable, int, str or datetime):
             Input to evaluate.
 
     Returns:
         bool:
             True if the input is a datetime type, False if not.
     """
-    if isinstance(value, Iterable) and not isinstance(value, str):
-        value = get_first_non_nan_value(value)
```
fealho, line 85: Why were we doing this before (selecting only the first value of a list)? Was it because of performance?

pvk-developer: Because it crashed at certain points on certain python versions (I believe it was 3.7). Pandas itself does this, gets first not `nan` value to find the `datetime` but for some reasons in python 3.7 it wasn't working as intended (it was failing with `None` or some other combinations).

pvk-developer (approved):

sdv/utils.py
```diff
@@ -73,23 +41,34 @@ def get_datetime_format(value):
...
-        pd.api.types.is_datetime64_any_dtype(value)
-        or isinstance(value, pd.Timestamp)
-        or isinstance(value, datetime)
-        or bool(get_datetime_format([value]))
-    )
+    if isinstance(value, str) or (not isinstance(value, Iterable)):
+        value = cast_to_iterable(value)
+
+    value = pd.Series(value)
+    value = value[~value.isna()]
```
amontanez24, line 62: nitpick: I think value made sense when it was one item but this should be plural now

sdv/utils.py
```diff
@@ -73,23 +41,34 @@ def get_datetime_format(value):
...
+    for item in value:
```
amontanez24, line 63: This should all work fine, but I do think it might increase the time of the metadata validation. Maybe we should still just grab the first item of the series or a subset

amontanez24 (approved):
LGTM!

PR #1569, opened 2023-09-06
Minor HMA updates

fealho:


sdv/multi_table/base.py
```diff
@@ -93,20 +93,6 @@ def __init__(self, metadata, locales=None, synthesizer_kwargs=None):
         self._fitted_date = None
         self._fitted_sdv_version = None
 
-    def get_table_parameters(self, table_name):
```
fealho, line 96: Do we want both `get_table_parameters` and `get_parameters`? The [docs](https://docs.sdv.dev/sdv/multi-table-data/modeling/synthesizers/hmasynthesizer) only lists `get_table_parameters` but it actually uses `get_parameters`.

fealho:
Implemented in #1578 instead.

PR #1577, opened 2023-09-12
Estimate number of columns to be modeled in `HMA`

### LINKED ISSUE #1572: Make function to estimate number of columns HMA produces.
### Problem Description

_As a user, I'd like to know as soon as possible if HMA will be slow so that I can properly decide whether or not to use it._

### Expected behavior

- Inside `HMASynthesizer`, add a function that can estimate the number or columns that will be modeled for each root table. We will use this later to give alerts.

It is understood that the exact number of rows varies depending on the transformers, constraints etc. For this reason, we only need an estimate. You can make the following assumptions:
- Assume the default transformers are being used
- Assume no constraints

#### Requirements

- The estimate should be the same order of magnitude as the actual answer.
- It should be performant (don't go overboard on trying to make calculations more precise at the cost of performance).

### Additional context

The following tips may help:
- If a parent has a child with n columns
    - A covariance matrix with O(n^2) columns will be added to it
    - A column for the mean and standard deviation of the parent row will be added to it
    - At least n columns will be added for the marginal distribution parameters
    - At least one column will be added for the cardinality

fealho:
Resolve #1572.


sdv/multi_table/hma.py
```diff
@@ -172,6 +250,25 @@ def _augment_table(self, table, tables, table_name):
         self._clear_nans(table)
         return table
 
+    def _augment_tables(self, processed_data):
```
fealho, line 253: Moved this method so it's close to `_augment_table`

sdv/multi_table/hma.py
```diff
@@ -44,6 +44,88 @@ def __init__(self, metadata, locales=None, verbose=True):
...
+            - n*(n-1)/2 correlation columns for each data column
+            - 4 parameters columns for each data column, with:
+                - 1 column for parameter a
+                - 1 column for parameter b
+                - 1 column for parameter scale
+                - 1 column for parameter loc
+        """
+        # The `num_rows` column. Because HMA models at most one relationship between
+        # two tables, this is always 1
+        num_cardinality_columns = 1
```
amontanez24, line 61: Is this right? I thought it modeled all of them but only used one

sdv/multi_table/hma.py
```diff
@@ -44,6 +44,88 @@ def __init__(self, metadata, locales=None, verbose=True):
...
+        """Get the number of columns that will be generated for table_name.
+
+        A table generates:
+            - 1 num_rows column for each for each foreign key with a specific parent
+            - n*(n-1)/2 correlation columns for each data column
+            - 4 parameters columns for each data column, with:
+                - 1 column for parameter a
+                - 1 column for parameter b
+                - 1 column for parameter scale
+                - 1 column for parameter loc
```
amontanez24, line 57: are these the parameters for the default distribution? Maybe we should see what distribution is passed by the user as the default since that happens before fitting, and base the number of parameters off of that

sdv/multi_table/hma.py
```diff
@@ -44,6 +44,88 @@ def __init__(self, metadata, locales=None, verbose=True):
...
+            return num_cardinality_columns
+
+        num_correlation_columns = (num_data_columns - 1) * num_data_columns // 2
+        num_parameters_columns = num_data_columns * 4
+
+        return num_correlation_columns + num_cardinality_columns + num_parameters_columns
+
+    def _estimate_columns_traversal(self, table_name, columns_per_table, visited):
+        """Given a table, estimate how many columns each parent will model.
+
```
amontanez24, line 75: can you add the parameters here. It will be helpful since this method modifies objects in place instead of returning something at the end

sdv/multi_table/hma.py
```diff
@@ -44,6 +44,88 @@ def __init__(self, metadata, locales=None, verbose=True):
             self._table_synthesizers,
             self._table_sizes)
 
+    def _num_extended_columns(self, table_name, parent_name, columns_per_table):
```
amontanez24, line 47: nitpick: can we add `_get` in front here?

sdv/multi_table/hma.py
```diff
@@ -44,6 +44,88 @@ def __init__(self, metadata, locales=None, verbose=True):
...
+    def _get_num_data_columns(self):
+        """Get the number of data columns, ie colums that are not id, for each table."""
+        columns_per_table = {}
+        for table_name, table in self.metadata.tables.items():
+            columns_per_table[table_name] = \
+                sum([1 for col in table.columns.values() if col['sdtype'] != 'id'])
+
+        return columns_per_table
+
+    def estimate_number_of_root_columns(self):
```
amontanez24, line 96: I think we should make this private for now and expose later if we need to. We can also just call it `_estimate_num_columns`

sdv/multi_table/hma.py
```diff
@@ -266,6 +348,7 @@ def _extract_parameters(self, parent_row, table_name, foreign_key):
         return flat_parameters.rename(new_keys).to_dict()
 
     def _recreate_child_synthesizer(self, child_name, parent_name, parent_row):
+        # When more than one foreign key exists between two tables, only the first one is modeled.
```
amontanez24, line 351: I don't think this is right. I think the root is extended with ALL of the parameters for children, but only one of them is used to create the child table synthesizer. See https://github.com/sdv-dev/SDV/blob/8b95b135c65a4e29b3741bcf660dd8d4264eb405/sdv/multi_table/hma.py#L152-L169

amontanez24 (changes requested):

sdv/multi_table/hma.py
fealho: You are right, thanks for catching this!

amontanez24 (approved):
I think this looks good now! Thanks for addressing

frances-h (approved):
Looks good! 👍

PR #1578, opened 2023-09-12
Minor HMA updates

fealho:
Minor updates to improve code readability and remove artifacts.

### DIFF
--- a/sdv/multi_table/base.py
+++ b/sdv/multi_table/base.py
@@ -107,20 +107,6 @@ class BaseMultiTableSynthesizer:
         """
         return self._table_parameters.get(table_name, {})
 
-    def get_parameters(self, table_name):
-        """Return the parameters used to instantiate the table's synthesizer.
-
-        Args:
-            table_name (str):
-                Table name for which the parameters should be retrieved.
-
-        Returns:
-            parameters (dict):
-                A dictionary representing the parameters used to instantiate the table's
-                synthesizer.
-        """
-        return self._table_synthesizers.get(table_name).get_parameters()
-
     def set_table_parameters(self, table_name, table_parameters):
         """Update the table's synthesizer instantiation parameters.
 
--- a/sdv/multi_table/hma.py
+++ b/sdv/multi_table/hma.py
@@ -266,7 +266,8 @@ class HMASynthesizer(BaseHierarchicalSampler, BaseMultiTableSynthesizer):
         return flat_parameters.rename(new_keys).to_dict()
 
     def _recreate_child_synthesizer(self, child_name, parent_name, parent_row):
-        # HMA only supports one foreign key for each parent/child pair
+        # HMA can only handle 1 foreign key per parent/child pair, so it only uses
+        # the first one if multiple keys are passed.
         foreign_key = self.metadata._get_foreign_keys(parent_name, child_name)[0]
         parameters = self._extract_parameters(parent_row, child_name, foreign_key)
         table_meta = self.metadata.tables[child_name]
--- a/sdv/multi_table/hma.py
+++ b/sdv/multi_table/hma.py
@@ -1,7 +1,6 @@
 """Hierarchical Modeling Algorithms."""
 
 import logging
-import math
 from copy import deepcopy
 
 import numpy as np
@@ -122,10 +121,11 @@ class HMASynthesizer(BaseHierarchicalSampler, BaseMultiTableSynthesizer):
             table_data[column] = table_data[column].fillna(fill_value)
 
     def _augment_table(self, table, tables, table_name):
-        """Generate the extension columns for this table.
+        """Recursively generate the extension columns for the tables in the graph.
 
         For each of the table's foreign keys, generate the related extension columns,
-        and extend the provided table.
+        and extend the provided table. Generate them first for the top level tables,
+        then their children, and so on.
 
         Args:
             table (pandas.DataFrame):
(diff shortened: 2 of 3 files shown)


sdv/multi_table/base.py
```diff
@@ -107,20 +107,6 @@ def get_table_parameters(self, table_name):
         """
         return self._table_parameters.get(table_name, {})
 
-    def get_parameters(self, table_name):
```
fealho, line 110: There are both a `get_table_parameters` and `get_parameters`. The docs only lists `get_table_parameters` but it actually uses `get_parameters`, so I'm not sure which one we want to keep.

sdv/multi_table/hma.py
```diff
@@ -122,10 +121,11 @@ def _clear_nans(table_data):
             table_data[column] = table_data[column].fillna(fill_value)
 
     def _augment_table(self, table, tables, table_name):
-        """Generate the extension columns for this table.
+        """Recursively generate the extension columns for the tables in the graph.
```
fealho, line 124: Clarification that the method is actually recursive.

sdv/multi_table/hma.py
```diff
@@ -260,12 +260,13 @@ def _extract_parameters(self, parent_row, table_name, foreign_key):
...
             flat_parameters[num_rows_key] = min(
                 self._max_child_rows[num_rows_key],
-                math.ceil(num_rows)
+                max(0, round(num_rows))
             )
 
         return flat_parameters.rename(new_keys).to_dict()
 
     def _recreate_child_synthesizer(self, child_name, parent_name, parent_row):
+        # HMA only supports one foreign key for each parent/child pair
```
fealho, line 269: I added these comments both to explain why we are only selecting the first foreign key and also so that in the future it's easy to find where the logic needs to be updated to handle multiple foreign keys.

sdv/multi_table/hma.py
```diff
@@ -260,12 +260,13 @@ def _extract_parameters(self, parent_row, table_name, foreign_key):
             num_rows = flat_parameters[num_rows_key]
             flat_parameters[num_rows_key] = min(
                 self._max_child_rows[num_rows_key],
-                math.ceil(num_rows)
+                max(0, round(num_rows))
```
fealho, line 263: If we use math.ceil we can't get 0, because the synthesized data is almost always a small positive value.

I can move this out of this PR if we want to write test cases for this.

sdv/sampling/hierarchical_sampler.py
```diff
@@ -68,7 +67,7 @@ def _sample_rows(self, synthesizer, num_rows=None):
                 Sampled rows, shape (, num_rows)
         """
         num_rows = num_rows or synthesizer._num_rows
-        return synthesizer._sample_batch(int(num_rows), keep_extra_columns=True)
+        return synthesizer._sample_batch(num_rows, keep_extra_columns=True)
```
fealho, line 70: As the docstring says (and I double checked), it's always int here. We either do this or change the docstring.

sdv/sampling/hierarchical_sampler.py
```diff
@@ -60,7 +59,7 @@ def _sample_rows(self, synthesizer, num_rows=None):
         Args:
             synthesizer (copula.multivariate.base):
                 The fitted synthesizer for the table.
-            num_rows (int):
+            num_rows (int or float):
```
fealho, line 62: I think the correct solution would be to always pass `num_rows` as an int to this method, but since that's a more complicated change, I'm just fixing the docstrings.

sdv/multi_table/base.py
amontanez24: what actually uses `get_parameters`?

sdv/multi_table/hma.py
amontanez24: I think this comment is misleading. It sounds like if the schema has 2 foreign keys, HMA would crash, which is false.

sdv/multi_table/base.py
fealho: Nothing in SDV. I thought this was the old way to access the parameters before `get_table_parameters` was added, which people forgot to remove. I can double check if any other libraries use it, but my question was more from a user perspective, which method should they use. I think it's `get_table_parameters`, but I wasn't 100% sure because the docs are wrong.

pvk-developer: They are different methods:
* get_table_parameters(table_name): Get the parameters for a given table name. --> Returns from  `self._table_parameters.get(table_name, {})`
* get_parameters: Returns the parameters that were used to instantiate the model. This should not include the metadata object. Returns from `self._table_synthesizers.get(table_name).get_parameters()`

We may need to file an issue for the docs.

sdv/multi_table/hma.py
pvk-developer: I'm not sure if we need at least 1 here. I think there was a bug when we don't sample any child. If integration tests pass it should be good since it should be covered in there.

fealho: If that value cannot be 0, then it's impossible for HMA to ever sample 0 rows for a primary key. For example, imagine you have two tables, parent and child, where the parent has 100 primary keys, but the every row of the child table only uses the first pk of the parent (so the foreign keys all have the same value). Currently, the sampled data will not be able to learn that, and instead every parent pk will be assigned to at least one child row.

sdv/multi_table/base.py
fealho: That makes sense, thanks. I can file the issue if @amontanez24 is fine with that?

fealho:
This PR was separated into multiple issues, as well as resolved in #1577.

PR #1601, opened 2023-09-21
Only make primary/foreign keys sdtype `id`

### LINKED ISSUE #1598: [Metadata Detection] Only make primary/foreign keys sdtype `id` (leave others as `unknown`)
_This issue is for the in-progress `metadata-improvment` branch, which aims to improve the accuracy of metadata detection from datasets._

### Problem
The new metadata detection logic is described in #1515. This is implemented according to spec but we've found that it's not intuitive.

When the values in a column are unique, the script is very liberal in calling it an `'id'` column. In reality, many PII columns will also have unique values (eg. SSNs, phone numbers, email addresses, etc.). 

### Expected Behavior
To not automatically assume that all unique values are sdtype `'id'`. The end result should be the following:

- Primary keys should be sdtype `'id'`
- Foreign keys should be sdtype `'id'`
- Other columns with unique values should be sdtype `'unkown'` (so they are treated as PII)

### Details
This may be accomplished by a few changes to the logic in #1515:
- 3 (iv): Elif the values are all unique, sdtype is ~id.~  **unknown**. Add it to a list of potential primary keys.
- 5 (iv): Elif the values are all unique, sdtype is ~id~ **unknown**. Add it to a list of potential primary keys. 

Then when detecting relationships:
- Primary key: We can continue to choose the leftmost column of the potential keys we've identified. **Only that column** should have sdtype updated to `'id'`. (See #1521 for initial specs)
- Foreign key: No changes should be needed (See #1522)

amontanez24:
resolves #1598

### DIFF
--- a/sdv/metadata/single_table.py
+++ b/sdv/metadata/single_table.py
@@ -321,10 +321,10 @@ class SingleTableMetadata:
             if dtype in self._DTYPES_TO_SDTYPES:
                 sdtype = self._DTYPES_TO_SDTYPES[dtype]
             elif dtype in ['i', 'f']:
-                sdtype = self._determine_sdtype_for_numbers(column_data, field)
+                sdtype = self._determine_sdtype_for_numbers(column_data)
 
             elif dtype == 'O':
-                sdtype = self._determine_sdtype_for_objects(column_data, field)
+                sdtype = self._determine_sdtype_for_objects(column_data)
 
             if sdtype is None:
                 raise InvalidMetadataError(
@@ -333,7 +333,7 @@ class SingleTableMetadata:
                 )
 
             # Set the first ID column we detect to be the primary key
-            if sdtype is 'id':
+            if sdtype == 'id':
                 if self.primary_key is None:
                     self.primary_key = field
                 else:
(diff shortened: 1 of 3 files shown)


sdv/metadata/single_table.py
```diff
@@ -331,6 +331,14 @@ def _detect_columns(self, data):
                     f"Unsupported data type for column '{field}' (kind: {dtype})."
                     "The valid data types are: 'object', 'int', 'float', 'datetime', 'bool'."
                 )
+
+            # Set the first ID column we detect to be the primary key
+            if sdtype == 'id':
+                if self.primary_key is None:
+                    self.primary_key = field
+                else:
+                    sdtype = 'unknown'
```
amontanez24, line 340: An alternative approach was to assign `unknown` in the `detect_sdtype_for...` methods and keep track of primary key candidates, but I felt those methods didn't need to be responsible for tracking the primary key candidates and I wanted to avoid passing a list along or having it as a instance variable

pvk-developer (approved):
LGTM!

R-Palazzo:
Looks good!
Should we also create a list of potential primary keys as written in the issue?
Also, are the foreign keys defaulted to sdtype `id`?

amontanez24:
> Looks good! Should we also create a list of potential primary keys as written in the issue? Also, are the foreign keys defaulted to sdtype `id`?

The foreign keys get set to id here:
https://github.com/sdv-dev/SDV/blob/792606c5440748ba18768ea88e34f118bd216ed6/sdv/metadata/multi_table.py#L419-L422C51

I chose not to have the list and explained why [here](https://github.com/sdv-dev/SDV/pull/1601#discussion_r1333646304), but I'm open to changing

R-Palazzo (approved):
LGTM! Thanks for the answers

PR #1607, opened 2023-09-25
Add set_address_columns method

### LINKED ISSUE #1593: Add set_address_columns method
### Problem Description

_As a user, I'd like a way to specify that multiple columns in my dataset are used to makeup the same address. This way, the synthetic data created maintains a relationship between those columns and makes a valid address._

### Expected behavior

In the [BaseSynthesizer](https://github.com/sdv-dev/SDV/blob/7eaffba7cf62be78e536ea731c878111d0b358ad/sdv/single_table/base.py#L34) and [BaseMultiTableSynthesizer](https://github.com/sdv-dev/SDV/blob/7eaffba7cf62be78e536ea731c878111d0b358ad/sdv/multi_table/base.py#L17) classes, add a method called `add_address_columns`
- This method should be used to set the transformer to use on the columns provided.

#### Parameters

- (required, multi table only) `table_name` : String that is the name of the table. This must be one of the tables specified in the metadata.
- (required): `column_names`: A list of one or more column names. These must be specified in the metadata for the table.
- `anonymization_level`: String that is the type of anonymization the user wants to see. Must be one of:
    - (default) 'full': Anonymize all components of the address (no learning required)
    - 'street_address': Only anonymized the precise street address (and secondary address). Learn everything else.

#### Behavior

Follow the logic below for transformer assignment
Parameter | Assigned Transformer
-- | --
'full' | RandomLocationGenerator(locales=\<synthesizer locales\>)
'street' | RegionalAnonymizer(locales=\<synthesizer locales\>)

#### Validation

- (multi table only) The table_name must be found in the metadata
`Error: Unknown table name 'userss'. Please choose a table name from the metadata.`
- The column_names must be found in the metadata for the table
`Error: Unknown column names ('A', 'B', 'C'). Please choose column names listed in the metadata for your table.`
- The sdtypes for the columns must be compatible with the address transformer. That is they must be one of : 'country', 'country_code', 'administrative_unit', etc.
`Error: Column 'city_name' has invalid sdtype 'categorical'. Please provide a column that is compatible with address data.`
- You cannot have 2 or more of the same sdtype within an address.
`Error: Columns 'state_name' and 'state' have the same sdtype 'administrative_unit'. Your address data cannot have duplicate fields.`
- If the user has already fit the data, show a warning. The user will need to re-fit in order to get this to work.
`Warning: Please refit your synthesizer for the address changes to appear in your synthetic data.`

```python
synthesizer = HSASynthesizer(metadata)
synthesizer.set_address_columns(
    table_name='users',
    column_names=['line1', 'line2', 'city', 'state', 'zip_code', 'country'],
    anonymization_level='full'
)
```

### Additional context

- This method is meant to replace the `update_transformers` method for address columns. It will happen after the `HyperTransformer` config is created.
- We will want to keep track of columns that are treated as an address, so we may need to store that information in a list/dict somewhere. This will help for error handling when users call `update_transformers` on one of these columns or try to add them to constraints.

R-Palazzo:
Resolve #1593

### DIFF
--- a/sdv/data_processing/data_processor.py
+++ b/sdv/data_processing/data_processor.py
@@ -94,6 +94,7 @@ class DataProcessor:
         del self._transformers_by_sdtype['text']
 
         self._multi_column_transformers = deepcopy(get_multi_column_transformers())
+        self._address_transformers = {}
 
         self._update_numerical_transformer(enforce_rounding, enforce_min_max_values)
         self._hyper_transformer = rdt.HyperTransformer()
@@ -454,9 +455,15 @@ class DataProcessor:
         sdtypes = {}
         transformers = {}
 
+        columns_in_address = {column for key in self._multi_column_transformers for column in key}
         for column in set(data.columns) - columns_created_by_constraints:
             column_metadata = self.metadata.columns.get(column)
             sdtype = column_metadata.get('sdtype')
+
+            if column in columns_in_address:
+                sdtypes[column] = sdtype
+                continue
+
             pii = column_metadata.get('pii', sdtype not in self._transformers_by_sdtype)
             sdtypes[column] = 'pii' if pii else sdtype
 
@@ -514,6 +521,9 @@ class DataProcessor:
                     'categorical',
                     column_metadata
                 )
+
+        for columns, transformer in self._multi_column_transformers.items():
+            transformers[columns] = transformer
 
         config = {'transformers': transformers, 'sdtypes': sdtypes}
         config = self._update_constraint_transformers(data, columns_created_by_constraints, config)
--- a/sdv/data_processing/data_processor.py
+++ b/sdv/data_processing/data_processor.py
@@ -115,13 +115,11 @@ class DataProcessor:
             col for col_tuple in self.columns_to_multi_col_transformer for col in col_tuple
         ]
 
-    def _import_address_transformers(self):
-        """Try to import the address transformers."""
-        try:
-            from rdt.transformers import RandomLocationGenerator, RegionalAnonymizer
-            self.RandomLocationGenerator = RandomLocationGenerator
-            self.RegionalAnonymizer = RegionalAnonymizer
-        except ImportError:
+    def _check_import_address_transformers(self):
+        """Check that the address transformers can be imported."""
+        has_randomlocationgenerator = hasattr(rdt.transformers, 'RandomLocationGenerator')
+        has_regionalanonymizer = hasattr(rdt.transformers, 'RegionalAnonymizer')
+        if not has_randomlocationgenerator or not has_regionalanonymizer:
             raise ImportError(
                 'You must have SDV Enterprise with the address add-on to use the address features'
             )
@@ -133,9 +131,15 @@ class DataProcessor:
             list:
                 A list of columns that are part of the address transformers.
         """
+        self._check_import_address_transformers()
         return [
-            col_tuple for col_tuple, transformer in self.columns_to_multi_col_transformer.items()
-            if isinstance(transformer, (self.RandomLocationGenerator, self.RegionalAnonymizer))
+            item for col_tuple, transformer in self.columns_to_multi_col_transformer.items()
+            if isinstance(
+                transformer, (
+                    rdt.transformers.RandomLocationGenerator,
+                    rdt.transformers.RegionalAnonymizer
+                )
+            ) for item in col_tuple
         ]
 
     def _get_address_transformer_parameters(self, column_names):
@@ -164,11 +168,12 @@ class DataProcessor:
             anonymization_level (str):
                 The anonymization level for the address transformer.
         """
+        self._check_import_address_transformers()
         transformer = None
         if anonymization_level == 'full':
-            transformer = self.RandomLocationGenerator()
+            transformer = rdt.transformers.RandomLocationGenerator()
         elif anonymization_level == 'street_address':
-            transformer = self.RegionalAnonymizer()
+            transformer = rdt.transformers.RegionalAnonymizer()
 
         return transformer
 
--- a/sdv/data_processing/data_processor.py
+++ b/sdv/data_processing/data_processor.py
@@ -133,10 +133,10 @@ class DataProcessor:
         """
         locales = self._locales if self._locales else ['en_US']
         self._check_import_address_transformers()
-        if anonymization_level == 'full':
-            return rdt.transformers.RandomLocationGenerator(locales=locales)
-        elif anonymization_level == 'street_address':
+        if anonymization_level == 'street_address':
             return rdt.transformers.RegionalAnonymizer(locales=locales)
+
+        return rdt.transformers.RandomLocationGenerator(locales=locales)
 
     def set_address_transformer(self, column_names, anonymization_level):
         """Set the address transformer.
@@ -461,10 +461,10 @@ class DataProcessor:
 
     def _get_transformer_instance(self, sdtype, column_metadata):
         transformer = self._transformers_by_sdtype[sdtype]
-        if transformer.__class__.__name__ == 'AnonymizedFaker':
+        if isinstance(transformer, AnonymizedFaker):
             is_lexify = transformer.function_name == 'lexify'
             is_baseprovider = transformer.provider_name == 'BaseProvider'
-            if is_lexify and is_baseprovider:  # Defaults settings
+            if is_lexify and is_baseprovider:  # Default settings
                 return self.create_anonymized_transformer(
                     sdtype, column_metadata, False, self._locales
                 )
--- a/sdv/data_processing/data_processor.py
+++ b/sdv/data_processing/data_processor.py
@@ -460,15 +460,24 @@ class DataProcessor:
         return transformer
 
     def _get_transformer_instance(self, sdtype, column_metadata):
+        transformer = self._transformers_by_sdtype[sdtype]
+        if transformer.__class__.__name__ == 'AnonymizedFaker':
+            is_lexify = transformer.function_name == 'lexify'
+            is_baseprovider = transformer.provider_name == 'BaseProvider'
+            if is_lexify and is_baseprovider:  # Defaults settings
+                return self.create_anonymized_transformer(
+                    sdtype, column_metadata, False, self._locales
+                )
+
         kwargs = {
             key: value for key, value in column_metadata.items()
             if key not in ['pii', 'sdtype']
         }
-        if kwargs and self._transformers_by_sdtype[sdtype] is not None:
-            transformer_class = self._transformers_by_sdtype[sdtype].__class__
+        if kwargs and transformer is not None:
+            transformer_class = transformer.__class__
             return transformer_class(**kwargs)
 
-        return deepcopy(self._transformers_by_sdtype[sdtype])
+        return deepcopy(transformer)
 
     def _update_constraint_transformers(self, data, columns_created_by_constraints, config):
         missing_columns = set(columns_created_by_constraints) - config['transformers'].keys()
--- a/sdv/data_processing/data_processor.py
+++ b/sdv/data_processing/data_processor.py
@@ -182,7 +182,7 @@ class DataProcessor:
                 A dictionary mapping column names to sdtypes.
         """
         list_sdtypes = list(columns_to_sdtypes.values())
-        transformer.locales = self.locales
+        transformer.locales = self._locales
         transformer.columns_to_sdtypes = columns_to_sdtypes
         transformer._list_sdtypes = list_sdtypes
 
@@ -832,6 +832,10 @@ class DataProcessor:
                 )
         except rdt.errors.NotFittedError:
             LOGGER.info(f'HyperTransformer has not been fitted for table {self.table_name}')
+
+        for transformer in self.columns_to_multi_col_transformer.values():
+            if not transformer.output_columns:
+                reversed_data = transformer.reverse_transform(reversed_data)
 
         num_rows = len(reversed_data)
         sampled_columns = list(reversed_data.columns)
--- a/sdv/data_processing/data_processor.py
+++ b/sdv/data_processing/data_processor.py
@@ -9,9 +9,7 @@ from pathlib import Path
 import pandas as pd
 import rdt
 from pandas.api.types import is_float_dtype, is_integer_dtype
-from rdt.transformers import (
-    AnonymizedFaker, IDGenerator, RegexGenerator, get_default_transformers,
-    get_multi_column_transformers)
+from rdt.transformers import AnonymizedFaker, IDGenerator, RegexGenerator, get_default_transformers
 
 from sdv.constraints import Constraint
 from sdv.constraints.base import get_subclasses
@@ -92,9 +90,7 @@ class DataProcessor:
         self._transformers_by_sdtype = deepcopy(get_default_transformers())
         self._transformers_by_sdtype['id'] = rdt.transformers.RegexGenerator()
         del self._transformers_by_sdtype['text']
-
-        self._multi_column_transformers = deepcopy(get_multi_column_transformers())
-        self.columns_to_mutli_col_transformer = {}
+        self.columns_to_multi_col_transformer = {}
 
         self._update_numerical_transformer(enforce_rounding, enforce_min_max_values)
         self._hyper_transformer = rdt.HyperTransformer()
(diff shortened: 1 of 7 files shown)


sdv/single_table/base.py
```diff
@@ -92,6 +92,94 @@ def __init__(self, metadata, enforce_min_max_values=True, enforce_rounding=True,
...
+            )
+
+        if anonymization_level not in {'full', 'street_address'}:
+            raise ValueError(
+                f"Invalid value '{anonymization_level}' for parameter 'anonymization_level'."
+                " Please provide 'full' or 'street_address'."
+            )
+
+        if not isinstance(column_names, tuple):
+            column_names = tuple(column_names)
```
pvk-developer, line 157: Be careful with this:
```python

In [2]: column_names = 'my_column_names'

In [3]: tuple(column_names)
Out[3]: ('m', 'y', '_', 'c', 'o', 'l', 'u', 'm', 'n', '_', 'n', 'a', 'm', 'e', 's')

In [4]: var = (column_names,)

In [5]: var
Out[5]: ('my_column_names',)

In [6]: type(var)
Out[6]: tuple
```

sdv/data_processing/data_processor.py
```diff
@@ -91,6 +93,9 @@ def __init__(self, metadata, enforce_rounding=True, enforce_min_max_values=True,
         self._transformers_by_sdtype['id'] = rdt.transformers.RegexGenerator()
         del self._transformers_by_sdtype['text']
 
+        self._multi_column_transformers = deepcopy(get_multi_column_transformers())
+        self.columns_to_mutli_col_transformer = {}
```
pvk-developer, line 97: typo: mu`t`li -> multi

sdv/data_processing/data_processor.py
```diff
@@ -450,9 +455,17 @@ def _create_config(self, data, columns_created_by_constraints):
         sdtypes = {}
         transformers = {}
 
+        columns_in_multi_col_transformer = {
+            column for key in self.columns_to_mutli_col_transformer for column in key
+        }
```
pvk-developer, line 460: I wouldn't encourage this kind of list comprehension as it makes it a bit complicated to read, there is `chain` to help you achieve the same:
```python
from itertools import chain
columns_in_multi_col_transformer = list(self.columns_to_mutli_col_transformer)
columns_in_multi_col_transformer = set(chain(*columns_in_multi_col_transformer))
```

----
Here is an example
```python

In [7]: x = {
   ...:     ('a', 'b', 'c'): 'Transformer',
   ...:     ('d', 'b', 'k'): 'AnotherTransformer'
   ...: }

In [8]: from itertools import chain

In [9]: d = list(x)

In [10]: set(chain(*d))
Out[10]: {'a', 'b', 'c', 'd', 'k'}
```

sdv/single_table/base.py
```diff
@@ -92,6 +92,94 @@ def __init__(self, metadata, enforce_min_max_values=True, enforce_rounding=True,
...
+        if transformer is None:
+            warnings.warn(
+                'Setting multi-column address data is a premium feature. '
+                'Please contact Datacebo Inc. for more information.'
+            )
+        else:
+            self._update_address_transformer(transformer, columns_to_sdtypes, sdtypes)
+            transformer._validate_sdtypes()
+
+            self._data_processor.columns_to_mutli_col_transformer[column_names] = transformer
```
pvk-developer, line 181: Move all this logic in the `data_processor` since in most of the cases you are accessing the `self._data_processor` and it's attributes.
At the end of the day the transformers live inside the `data_processor` and not in the `synthesizer`. The `synthesizer` should only be a layer between the user and the `data_processor`.
This may require you to move some of the methods for validation there as well, and pass a locales as parameter to the method.

pvk-developer (changes requested):

sdv/single_table/base.py
R-Palazzo: Yes this is what I'm doing today

R-Palazzo: I agree with this

R-Palazzo:
Hi @pvk-developer, @amontanez24, I made the changes we discussed, so it's good for review.

sdv/data_processing/data_processor.py
```diff
@@ -725,6 +833,10 @@ def reverse_transform(self, data, reset_keys=False):
         except rdt.errors.NotFittedError:
             LOGGER.info(f'HyperTransformer has not been fitted for table {self.table_name}')
 
+        for transformer in self.columns_to_multi_col_transformer.values():
```
R-Palazzo, line 836: I needed this to make the RandomLocation work with SDV. I'm down to discuss other options or changes we could do on the `reverse_transform` of the DataProcessor. Basically, without this change, it was not crashing but the `address columns` were not generated because the `hyper transformer` were never called for these columns, mainly because this transformer doesn't have any `output_columns`.

sdv/data_processing/data_processor.py
```diff
@@ -103,6 +104,104 @@ def __init__(self, metadata, enforce_rounding=True, enforce_min_max_values=True,
...
+        """Get the columns that are part of the address transformer.
+
+        Returns:
+            list:
+                A list of columns that are part of the address transformers.
+        """
+        return [
+            col_tuple for col_tuple, transformer in self.columns_to_multi_col_transformer.items()
+            if isinstance(transformer, (self.RandomLocationGenerator, self.RegionalAnonymizer))
+        ]
```
pvk-developer, line 139: is this being called somewhere?

sdv/data_processing/data_processor.py
```diff
@@ -103,6 +104,104 @@ def __init__(self, metadata, enforce_rounding=True, enforce_min_max_values=True,
...
+    def _import_address_transformers(self):
+        """Try to import the address transformers."""
+        try:
+            from rdt.transformers import RandomLocationGenerator, RegionalAnonymizer
+            self.RandomLocationGenerator = RandomLocationGenerator
+            self.RegionalAnonymizer = RegionalAnonymizer
+        except ImportError:
+            raise ImportError(
+                'You must have SDV Enterprise with the address add-on to use the address features'
+            )
```
pvk-developer, line 127: This should be renamed to `check_address_transformers_present` or something like that.
Basically you should ensure that you have access to the `RandomLocationGenerator` which should (I guess) be in `rdt.transformers`. Therefore you should be able to access it trough `rdt.transformers.RandomLocationGenerator`.
On that case you can check like this:
```

In [1]: import rdt
    
In [2]: hasattr(rdt.transformers, 'FloatFormatter')
Out[2]: True

In [3]: 
```
Then the logic is quiet simple:
```python
if not hasattr(rdt.transformers, 'RandomLocationGenerator'):
    raise AttributeError('You must have SDV-Enterprsie ...')
```

sdv/data_processing/data_processor.py
```diff
@@ -103,6 +104,104 @@ def __init__(self, metadata, enforce_rounding=True, enforce_min_max_values=True,
...
+
+        Args:
+            anonymization_level (str):
+                The anonymization level for the address transformer.
+        """
+        transformer = None
+        if anonymization_level == 'full':
+            transformer = self.RandomLocationGenerator()
+        elif anonymization_level == 'street_address':
+            transformer = self.RegionalAnonymizer()
```
pvk-developer, line 171: this should be `rdt.transformers.RandomLocationGenerator`
OR in the worst case use lazy import if that is something that you want to achieve here, but I wouldn't vouch for that.

pvk-developer (changes requested):
Just some structural comments and imports, then it should be good ;)

sdv/data_processing/data_processor.py
```diff
@@ -103,6 +104,112 @@ def __init__(self, metadata, enforce_rounding=True, enforce_min_max_values=True,
...
+    def _set_address_transformer(self, column_names, anonymization_level):
+        """Set the address transformer.
+
+        Args:
+            column_names (tuple[str]):
+                The column names to set the transformer for.
+            anonymization_level (str):
+                The anonymization level for the address transformer.
+        """
+        columns_to_sdtypes = self._get_address_transformer_parameters(column_names)
```
pvk-developer, line 206: no need of this

sdv/data_processing/data_processor.py
```diff
@@ -103,6 +104,112 @@ def __init__(self, metadata, enforce_rounding=True, enforce_min_max_values=True,
...
+        transformer = self._get_address_transformer(anonymization_level)
+        self._update_address_transformer(transformer, columns_to_sdtypes)
+        transformer._validate_sdtypes()
```
pvk-developer, line 209: Both of this are not needed

sdv/data_processing/data_processor.py
```diff
@@ -103,6 +104,112 @@ def __init__(self, metadata, enforce_rounding=True, enforce_min_max_values=True,
...
+                The anonymization level for the address transformer.
+        """
+        self._check_import_address_transformers()
+        transformer = None
+        if anonymization_level == 'full':
+            transformer = rdt.transformers.RandomLocationGenerator()
+        elif anonymization_level == 'street_address':
+            transformer = rdt.transformers.RegionalAnonymizer()
+
+        return transformer
```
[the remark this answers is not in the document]
pvk-developer: Also when giving locales here maybe ensure for it to not be empty or it will fail down the line.

pvk-developer (changes requested):

sdv/data_processing/data_processor.py
R-Palazzo: I know that a lot of logic here will be done by the `HyperTransformer`, but I need them because in the Issue we want to do some `validations` directly when calling `set_address_columns()`, so before creating any config or setting the `HyperTransformer`. If it's fine to let all the validation in a second time to the HyperTransformer, I agree we can remove those methods.

sdv/data_processing/data_processor.py
amontanez24: should we mark the `RandomLocationGenerator` as a generator? Then it should generate the columns when `create_anonymized_columns` is called below

sdv/data_processing/data_processor.py
```diff
@@ -103,6 +104,112 @@ def __init__(self, metadata, enforce_rounding=True, enforce_min_max_values=True,
         if self._primary_key:
             self._keys.append(self._primary_key)
 
+    def _get_column_in_multi_column_transformer(self):
```
amontanez24, line 107: we may want to come out with a nomenclature for columns that are part of multi-column transformers because these name are long and starting to clutter the code. Maybe `_get_grouped_columns` and `self.grouped_columns_to_transformers` or something like that

sdv/data_processing/data_processor.py
```diff
@@ -103,6 +104,112 @@ def __init__(self, metadata, enforce_rounding=True, enforce_min_max_values=True,
...
+        Args:
+            column_names (tuple[str]):
+                The column names to set the transformer for.
+            anonymization_level (str):
+                The anonymization level for the address transformer.
+        """
+        columns_to_sdtypes = self._get_address_transformer_parameters(column_names)
+        transformer = self._get_address_transformer(anonymization_level)
+        self._update_address_transformer(transformer, columns_to_sdtypes)
+        transformer._validate_sdtypes()
```
amontanez24, line 209: The error messages described in the issue are slightly different than that of what the transformer returns. Perhaps we can change those errors to be class methods or static methods on the transformer that take in the column_to_sdtype dict as an argument. Then we can call without initializing the transformers here which seems a little awkward

sdv/multi_table/base.py
```diff
@@ -93,6 +93,19 @@ def __init__(self, metadata, locales=None, synthesizer_kwargs=None):
...
+
+        Args:
+            table_name (str):
+                The name of the table for which the address transformer should be set.
+            column_names (tuple[str]):
+                The column names to be used for the address transformer.
+            anonymization_level (str):
+                The anonymization level to use for the address transformer.
+        """
+        self._table_synthesizers[table_name].set_address_columns(column_names, anonymization_level)
```
amontanez24, line 107: I think the issue said to validate that the table name exists

sdv/data_processing/data_processor.py
```diff
@@ -103,6 +104,112 @@ def __init__(self, metadata, enforce_rounding=True, enforce_min_max_values=True,
...
+    def _check_import_address_transformers(self):
+        """Check that the address transformers can be imported."""
+        has_randomlocationgenerator = hasattr(rdt.transformers, 'RandomLocationGenerator')
+        has_regionalanonymizer = hasattr(rdt.transformers, 'RegionalAnonymizer')
+        if not has_randomlocationgenerator or not has_regionalanonymizer:
+            raise ImportError(
+                'You must have SDV Enterprise with the address add-on to use the address features'
+            )
+
+    def _get_columns_in_address_transformer(self):
```
amontanez24, line 127: this doesn't seem to be used anywhere

amontanez24 (changes requested):
I think we can think through how to better utilize the existing code in the transformers/HyperTransformer

sdv/data_processing/data_processor.py
pvk-developer: Yes I see that, but this is just bad design then and requires changes to the transformer itself in order for it to handle everything. Basically `validate_sdtypes` could be a `classmethod` that takes the `columns_to_sdtypes` that you want to validate or the transformer itself takes those as input on instantiation and you are able to validate them with this same method.

The problem I'm having with this is that we are setting things on the transformer which ends up working like a mock, and those attributes are due to change further in the execution flow (during `fit` for example), that is why I'm a against the `transformer.attribute = new_value`, you can see that this is also more easy to get a bug than doing `Transformer(new_value)` or `transformer.update_attribute(new_value)` as per `transformer._locales` instead of `transformer.locales`.

sdv/data_processing/data_processor.py
pvk-developer: It is being used in #1614 I had the same confusion.

pvk-developer (approved):

sdv/data_processing/data_processor.py
R-Palazzo: That's a good idea but  `create_anonymized_columns` will have to be updated to work with multi-column transformers. Especially we will have to change this [line](https://github.com/sdv-dev/RDT/blob/c2c70646563045f998981e15c4cbe60662cc42ed/rdt/hyper_transformer.py#L716).

R-Palazzo:
Hi @pvk-developer, @amontanez24. I reworked the validation inside the transformers and made the changes accordingly here.

sdv/data_processing/data_processor.py
amontanez24: Maybe we should just move it to that PR. It's kind of confusing here

sdv/multi_table/base.py
```diff
@@ -104,6 +104,11 @@ def set_address_columns(self, table_name, column_names, anonymization_level='ful
             anonymization_level (str):
                 The anonymization level to use for the address transformer.
         """
+        if not self._table_synthesizers.get(table_name):
+            raise ValueError(
+                f"Unknown table name '{table_name}'. Please choose a table name from the metadata."
+            )
```
amontanez24, line 110: I think you could reuse the `_validate_table_name` method for this. Also, I think that method should be changed to raise a `ValueError` like you do here.

amontanez24 (commented):
A couple small comments but this is way cleaner! Thank you for making the adjustments

sdv/data_processing/data_processor.py
amontanez24: Maybe we should?

pvk-developer (approved):

sdv/data_processing/data_processor.py
```diff
@@ -103,6 +104,57 @@ def __init__(self, metadata, enforce_rounding=True, enforce_min_max_values=True,
...
+                The anonymization level for the address transformer.
+        """
+        locales = self._locales if self._locales else ['en_US']
+        self._check_import_address_transformers()
+        if anonymization_level == 'full':
+            return rdt.transformers.RandomLocationGenerator(locales=locales)
+        elif anonymization_level == 'street_address':
+            return rdt.transformers.RegionalAnonymizer(locales=locales)
+
+    def _set_address_transformer(self, column_names, anonymization_level):
```
amontanez24, line 141: I think you can make this public since this is what the other classes call

sdv/single_table/base.py
```diff
@@ -92,6 +92,70 @@ def __init__(self, metadata, enforce_min_max_values=True, enforce_rounding=True,
...
+                " Please provide 'full' or 'street_address'."
+            )
+
+        self._data_processor._check_import_address_transformers()
+
+        if self._fitted:
+            warnings.warn(
+                'Please refit your synthesizer for the address changes to appear in'
+                ' your synthetic data.'
+            )
```
amontanez24, line 108: I don't think we should raise this until after the address transformers actually get set

sdv/single_table/base.py
```diff
@@ -92,6 +92,70 @@ def __init__(self, metadata, enforce_min_max_values=True, enforce_rounding=True,
...
         self._fitted_sdv_version = None
 
+    def _check_address_initialization(self, anonymization_level):
+        if anonymization_level not in {'full', 'street_address'}:
+            raise ValueError(
+                f"Invalid value '{anonymization_level}' for parameter 'anonymization_level'."
+                " Please provide 'full' or 'street_address'."
+            )
+
+        self._data_processor._check_import_address_transformers()
```
amontanez24, line 102: can we just call this in `self._data_processor.set_address_transformers`?

amontanez24 (approved):
LGTM!

PR #1609, opened 2023-09-26
Fix columns filled with `NaN, None or NaT` to be detected as datetime sdtype

### LINKED ISSUE #1589: Columns consisting of only Nones are being detected as datetime
### Environment Details

* SDV version: metadata-improvment branch
* Python version: All
* Operating System: All

### Error Description

If a column is all None, it is being detected as a datetime column. If a column is all `np.nan`, it is being detected as categorical.

Instead, we should detect based on the type of NaN. In other words, if a column is all
- None: we should detect it as categorical
- `np.nan`: we should detect it as numerical
- `pd.NaT`: we should detect it as datetime

### Steps to reproduce

```python
tables = {
    'a': pd.DataFrame({'pk1': [i for i in range(100)], 'pk2': [i for i in range(100)]}),
    'b': pd.DataFrame({
        'pk2': [i for i in range(100)],
        'pk1': [i for i in range(100)],
        'data': [float('nan') for i in range(100)],
        'data2': [None for i in range(100)]
    })
}
metadata.detect_from_dataframes(tables)
```

pvk-developer:
Resolves #1589


sdv/metadata/single_table.py
```diff
@@ -260,6 +260,9 @@ def _determine_sdtype_for_numbers(self, data):
         sdtype = 'numerical'
         if len(data) > 5:
             is_not_null = ~data.isna()
+            if (data == data.round()).loc[is_not_null].empty:
```
fealho, line 263: We use `(data == data.round()).loc[is_not_null]` below, it's better to store it in a variable and reuse it.

sdv/metadata/single_table.py
```diff
@@ -280,28 +283,29 @@ def _determine_sdtype_for_objects(self, data):
...
             with warnings.catch_warnings():
                 warnings.simplefilter('ignore', category=UserWarning)
 
                 data_test = data.sample(10000) if len(data) > 10000 else data
                 datetime_format = get_datetime_format(data_test)
-                pd.to_datetime(data_test, format=datetime_format, errors='raise')
-
-            sdtype = 'datetime'
+                if datetime_format:
+                    pd.to_datetime(data_test, format=datetime_format, errors='raise')
```
fealho, line 304: Is this the only line that raises? If so you could put the try/catch only around it. If that's not the case ignore this comment.

frances-h (approved):

fealho (approved):

sdv/metadata/single_table.py
pvk-developer: I just want to be safe with `get_datetime_format` so I will shorten it there 👍🏻

PR #1623, opened 2023-10-10
Support Numpy 1.25+

### LINKED ISSUE #1501: Investigate how to get numpy >1.25 to pass
### Problem Description

Numpy >= 1.25.0 seems to cause some tests in SDV to fail. We should investigate this and fix it so we can accept the latest version.

### Expected behavior

- Numpy is capped at >= 2.0 instead of 1.25.0

### Additional context

[Here ](https://github.com/sdv-dev/SDV/actions/runs/5327920491/jobs/9651939818)is a link to the failed job.

rwedge:
Resolve #1501

A few tests have hardcoded answers that fail on some CI environments due to the beta distribution fit changing slightly with newer Numpy versions.  These tests were changed to use Truncated Gaussian by default and the expected results updated accordingly.


R-Palazzo (approved):
LGTM!

amontanez24 (approved):
LGTM!

PR #1637, opened 2023-10-18
missing_value_replacement now defaults to random with RDT 1.7.1

lajohn4747:
With the change to RDT:
https://github.com/sdv-dev/RDT/pull/715

Value replacement will default to random which causes a test to fail with RDT: 1.7.1


lajohn4747:
Closed in favor of making an issue, that should be resolved before release:
https://github.com/sdv-dev/SDV/issues/1638

PR #1640, opened 2023-10-18
Issue 1420 num_table pd.type from object to int

### LINKED ISSUE #1420: In `get_available_demos`, the `num_tables` column should be an int
### Problem Description
Currently, if I use `get_available_demos`, it returns a dataframe with the demo name, dataset size and the number of tables. The dtype of `num_tables` is an object instead of an int. 

```python
from sdv.datasets.demo import get_available_demos

demos = get_available_demos(modality='multi_table')
demos.dtypes
```
```
dataset_name     object
size_MB         float64
num_tables       object
dtype: object
```

This makes it difficult for me to sort or filter by `num_tables`.

### Expected behavior
The `num_tables` column should be casted to an int. This should be done in a fault tolerant way so that any non-integer value would show up as NaN.

### Additional context
This applies to all modalities: `single_table`, `multi_table` and `sequential`

lajohn4747:
ClickUp Task: CU-860qunv0v
resolves #1420 

Boto3 Client `head_object` dictionary all values are recorded as strings. It appears pandas has a hard time inferring a int64 type from a string and thus defaults to obj. Using `pd.to_numeric` to correctly cast the type (int/float). If missing data, fill in with pd.nan and leave the column as a float dtype


sdv/datasets/demo.py
```diff
@@ -171,6 +171,9 @@ def get_available_demos(modality):
             headers = client.head_object(Bucket=BUCKET, Key=item['Key'])['Metadata']
             size_mb = headers.get('size-mb', np.nan)
             tables_info['size_MB'].append(round(float(size_mb), 2))
-            tables_info['num_tables'].append(headers.get('num-tables', np.nan))
+            num_tables = headers.get('num-tables', pd.NA)
+            if not pd.isna(num_tables):
+                num_tables = int(num_tables)
+            tables_info['num_tables'].append(num_tables)
```
rwedge, line 177: @npatki seemed fine with the 'num_tables' column being a float in the event a NaN was present which simplifies things.

I think instead of casting to in as we append, we can keep the old append approach from main, and after the 'num_tables' list is finalized, use `pd.to_numeric` to cast the whole list into a numeric array instead.  If there are `NaN` values we'll get a float array and if not we should get an int array. We can use the integration tests to confirm that the output dtype is an integer type

lajohn4747: Sounds reasonable to me. I'll update and add Neha to the review

tests/integration/datasets/test_demo.py
```diff
@@ -24,6 +24,7 @@ def test_get_available_demos_single_table():
         'num_tables': ['1'] * 9
     })
     expected_table['size_MB'] = expected_table['size_MB'].astype(float).round(2)
+    expected_table['num_tables'] = expected_table['num_tables'].astype(int)
```
rwedge, line 27: I tried changing the astype call to cast to float here and the test still passed, so I think to ensure `tables_info['num_tables']` is an integer dtype we should add a new check that tests directly using `pd.api.types.is_integer_dtype` to confirm it. (and same deal for the other test in this file)

lajohn4747: Actually this wasn't even testing the column type, just adding a column as a specific type. Added an assert to confirm that the column is of integer type using `pd.api.types.is_integer_dtype`

rwedge (approved):
LGTM

amontanez24 (approved):
🚢

PR #1645, opened 2023-10-23
Improve debug messages when a constraint falls back to reject sampling approach

### LINKED ISSUE #1478: Improve debug messages when a constraint falls back to reject sampling approach
### Problem Description
When constraints are not working as expected (crashing, producing incorrect data or leading to performance issues), it is hard to debug the root cause. 

When the `transform` or `reverse_transform` strategies fail, the SDV defaults back to the `is_valid` method. Debug message in Logger isn't helpful to understand why this is happening or which exact constraint it is.

```
INFO:sdv.data_processing.data_processor:Fitting constraints for table users
INFO:sdv.data_processing.data_processor:Error transforming CustomConstraint. Using the reject sampling approach instead.
```

### Expected behavior
The logger should craft better debug messages when falling back to reject sampling. There are 2 cases to consider:

1. Overlapping constraints. It is expected that if multiple constraints use the same columns, one of constraints has to fall back to reject sampling. We should check this by seeing if the constraint's `column_names` exist in the dataframe. If not, the debug message should show this.

```
INFO:sdv.data_processing.data_processor:Unable to transform Inequality constraint with columns ['A', 'B', 'C'] because they are not all available in the data. This happens due to multiple, overlapping constraints. Using reject sampling instead.
```

2. Error in transform or reverse transform logic. The other case is that we've caught an exception in these functions so we've had to fall back.
```
INFO:sdv.data_processing.data_processor:Unable to transform CustomConstraint constraint with columns ['A', 'B', 'C'] due to an error in 'reverse_transform': 
ValueError:There are non-numerical values in your data
Using reject sampling instead.
```

### Additional context

- While working on this it might be a good opportunity to improve our personal logging of constraint errors as well. Right now, only the message surfaces, and the stack trace is completely lost. We might want to look into doing a `logger.debug()` with the full stack trace as the message. The stack trace can be extracted using [traceback.format_exc() or sys.exc_info()](https://stackoverflow.com/questions/3702675/catch-and-print-full-python-exception-traceback-without-halting-exiting-the-prog).

amontanez24:
resolves #1478 
CU-85ztc8c9a

This PR addresses the issue above but also introduces a function to log the stack trace of an exception at the `DEBUG` level. There are a few places where we catch errors, accumulate them and then raise them altogether with a large error message later. The problem with this is that we lose the stack trace of those errors. To help remedy this, I call the function in the following places:
1. The initialization of a `AggregateConstraintsError`. This way whenever one is made, we know the log will be accessible.
2. When overlapping constraints cause a `MissingConstraintColumnError`.
3. When a custom constraint raises an unexpected error. This is because in that situation, we don't crash and fall back to reject sampling.

### DIFF
--- a/tests/integration/single_table/test_constraints.py
+++ b/tests/integration/single_table/test_constraints.py
@@ -8,7 +8,8 @@ import pandas as pd
 import pytest
 from copulas.multivariate.gaussian import GaussianMultivariate
 
-from sdv.constraints import create_custom_constraint_class
+from sdv.constraints import Constraint, create_custom_constraint_class
+from sdv.constraints.errors import AggregateConstraintsError
 from sdv.datasets.demo import download_demo
 from sdv.metadata import SingleTableMetadata
 from sdv.sampling import Condition
@@ -917,3 +918,29 @@ def test_custom_and_overlapping_constraint_errors(caplog, demo_data, demo_metada
     log_messages = [record[2] for record in caplog.record_tuples]
     for log in expected_logs:
         assert log in log_messages
+
+
+def test_aggregate_constraint_errors(demo_data, demo_metadata):
+    """Test that if there are multiple constraint errors, they are raised together."""
+    class BadConstraint(Constraint):
+        def __init__(self, column_name):
+            self.column_name = column_name
+
+        def _transform(self, table_data):
+            raise ValueError('Bad constraint')
+
+    synth = GaussianCopulaSynthesizer(demo_metadata)
+    bad_constraint1 = {
+        'constraint_class': 'BadConstraint',
+        'constraint_parameters': {'column_name': 'room_rate'}
+    }
+    bad_constraint2 = {
+        'constraint_class': 'BadConstraint',
+        'constraint_parameters': {'column_name': 'checkin_date'}
+    }
+    synth.add_constraints(constraints=[bad_constraint1, bad_constraint2])
+
+    # Run and Assert
+    message = '\nBad constraint\n\nBad constraint'
+    with pytest.raises(AggregateConstraintsError, match=message):
+        synth.fit(demo_data)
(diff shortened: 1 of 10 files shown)


tests/integration/single_table/test_constraints.py
```diff
@@ -838,3 +853,95 @@ def test_timezone_aware_constraints():
...
+        "Unable to transform CustomConstraint with columns ['room_rate'] due to an error in "
+        'transform: \nTransform error\nUsing the reject sampling approach instead.'
+    ]
+    log_messages = [record[2] for record in caplog.record_tuples]
+    for log in expected_logs:
+        assert log in log_messages
+
+
+def test_aggregate_constraint_errors(demo_data, demo_metadata):
+    """Test that if there are multiple constraint errors, they are raised together."""
```
R-Palazzo, line 925: Add `# Setup`

R-Palazzo (approved):
LGTM!

frances-h (approved):

rwedge:
This PR covers the transform case but do we need to do anything for errors during reverse_transform?

amontanez24:
> This PR covers the transform case but do we need to do anything for errors during reverse_transform?

Currently we don't log anything for the reverse transform. I'm not actually sure if we can fallback to reject sampling in this case but I can test it. If it's doable then I can add that

amontanez24:
> This PR covers the transform case but do we need to do anything for errors during reverse_transform?

@rwedge After looking into it what happens is that for every custom constraint, we actually run the `reverse_transform` in the `transform` function just to make sure that it works. This is because if the transform strategy is used, and reverse transforming fails, we might not be able to get the data into a valid shape for reject sampling, making it impossible. The code where we do this is here:
https://github.com/sdv-dev/SDV/blob/7527fc535e768b6c958b643a1396cdb96caeefc3/sdv/constraints/tabular.py#L175-L188
For simplicity, both `transform` and `reverse_transform` are called in the same try/catch block. After discussing with product, I think we'll leave it that way for now so we don't have to have a bunch of try/catch statements everywhere. Perhaps we will clean it up later

rwedge (approved):

PR #1649, opened 2023-10-24
Add csv options to load_csvs

### LINKED ISSUE #1644: Add options to read CSV files
### Problem Description
In older versions of SDV, it was possible to read a single CSV at a time and pass in extra parameters used to parse out the CSV. For example, perhaps the file was separated by tabs or pipes instead of commas.

Now, we have a method for reading a batch of CSVs at a time (every CSV within a particular folder):

```python
from sdv.datasets.local import load_csvs

datasets = load_csvs(folder_name='my_folder')
```

But I cannot pass in extra args for loading.

### Expected behavior
I expect that I should be able to add keyword args for reading the CSVs. We should accept a dictionary of args to be used by [pands.read_csv](https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html)

```python
from sdv.datasets.local import load_csvs

datasets = load_csvs(
    folder_name='my_folder/',
    read_csv_parameters={
        'sep': '\t', # separators are tabs instead of commas
        'skipinitialspace': True # skip spaces that are at the start of each value
    })
```

Note that this was the intended usage of the singular `load_csv` back when we supported this method.

### Additional context
All CSVs in the folder will be read in with the same arguments as `read_csv_parameters`. If users wish to use different parameters per CSV, then they could directly use the `pandas.read_csv` function.

lajohn4747:
Clickup task: 
CU-86ayj7br6

resolves #1644 

Allow local read csv function to accept pandas csv options as a dictionary, so the user can process the csv in their preferred manner

### DIFF
--- a/sdv/datasets/local.py
+++ b/sdv/datasets/local.py
@@ -6,13 +6,13 @@ from os import path, walk
 from sdv.utils import load_data_from_csv
 
 
-def load_csvs(folder_name, pandas_kwargs=None):
+def load_csvs(folder_name, read_csv_parameters=None):
     """Load csv files from specified folder.
 
     Args:
         folder_name (str):
             The full path of the folder with the data to be loaded.
-        pandas_kwargs (dict):
+        read_csv_parameters (dict):
             A python dictionary of with string and value accepted by ``pandas.read_csv``
             function. Defaults to ``None``.
     """
@@ -26,7 +26,7 @@ def load_csvs(folder_name, pandas_kwargs=None):
         base_name, ext = path.splitext(filename)
         if ext == '.csv':
             filepath = path.join(dirpath, filename)
-            csvs[base_name] = load_data_from_csv(filepath, pandas_kwargs)
+            csvs[base_name] = load_data_from_csv(filepath, read_csv_parameters)
         else:
             other_files.append(filename)
 
--- a/sdv/datasets/local.py
+++ b/sdv/datasets/local.py
@@ -6,12 +6,15 @@ from os import path, walk
 from sdv.utils import load_data_from_csv
 
 
-def load_csvs(folder_name):
+def load_csvs(folder_name, pandas_kwargs=None):
     """Load csv files from specified folder.
 
     Args:
         folder_name (str):
             The full path of the folder with the data to be loaded.
+        pandas_kwargs (dict):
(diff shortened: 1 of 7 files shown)


sdv/datasets/local.py
```diff
@@ -6,12 +6,15 @@
...
 
 
-def load_csvs(folder_name):
+def load_csvs(folder_name, pandas_kwargs=None):
     """Load csv files from specified folder.
 
     Args:
         folder_name (str):
             The full path of the folder with the data to be loaded.
+        pandas_kwargs (dict):
```
amontanez24, line 15: @npatki Should we use this name or the one in the issue? The reason I ask is that the `SingleTableMetadata.detect_from_csv` function calls it `pandas_kwargs` so maybe we should be consistent

sdv/metadata/multi_table.py
```diff
@@ -471,18 +471,21 @@ def detect_from_dataframes(self, data):
 
         self._detect_relationships()
 
-    def detect_table_from_csv(self, table_name, filepath):
+    def detect_table_from_csv(self, table_name, filepath, pandas_kwargs=None):
```
amontanez24, line 474: If we add it here we should also add it to the `ddetect_From_csvs` method

sdv/datasets/local.py
npatki: There is currently no parameter listed in the API docs for `detect_from_csv(s)` for either [Single Table](https://docs.sdv.dev/sdv/single-table-data/data-preparation/single-table-metadata-api#auto-detect-metadata) or [Multi Table](https://docs.sdv.dev/sdv/multi-table-data/data-preparation/multi-table-metadata-api#creation-api) data so we can choose to call it what we want

Back when we had a singular `load_csv` method, we had  called the parameter `read_csv_parameters` so that's why it was carried over into the issue requirements.

So I'm onboard with the reverse: Let's name the parameter `read_csv_parameters` for the single/multi table metadata `detect_from_csv(s)` method (align it to what we describe here in this issue). We could make this a separate issue?

lajohn4747: Changed param name to `read_csv_parameters`

amontanez24 (approved):
LGTM!

R-Palazzo (approved):
LGTM!

PR #1654, opened 2023-10-25
Add print estimate warning when the estimate cols for hma are large

### LINKED ISSUE #1646: Print alert if HMASynthesizer is likely to be slow
### Problem Description

_As a user, it would be helpful to know as soon as possible if my data will take too long to train so that I can make any necessary adjustments._

### Expected behavior

- Using the[ _estimate_num_columns](https://github.com/sdv-dev/SDV/blob/4d4ed3e6a6dc060d471e15fe0b106f1050693065/sdv/multi_table/hma.py#L156), if the total number of columns summed across all tables is > 1000, print the following message:

```python

PerformanceAlert: Using the HMASynthesizer on this metadata schema is not recommended. To model this data, HMA will generate a large number of columns.

Table Name       # Columns in Metadata      Est # Columns (HMA)
transactions     12                         12    
sessions         50                         218
transactions     25                         47985

We recommend simplifying your metadata schema by dropping columns that are not necessary. If this is not possible, contact us at info@sdv.dev for enterprise solutions.
```

- The table must be pretty printed so the tabs/columns line up like above
- Make sure to call it an alert to raise attention

### Additional context

- In this case, it might make sense to just print instead of raising a warning or logging since we don't want it to be disabled and its main intent is to be used live.

lajohn4747:
CU-86ayjengm

resolves #1646 

Print a notice in a tabular format explaining the HMASynthesizer will be slow.

Visual Output:
Screenshot 2023-10-25 at 6 20 39 PM

### DIFF
--- a/sdv/multi_table/hma.py
+++ b/sdv/multi_table/hma.py
@@ -6,6 +6,7 @@ from copy import deepcopy
 
 import numpy as np
 import pandas as pd
+from tabulate import tabulate
 from tqdm import tqdm
 
 from sdv.errors import SynthesizerInputError
@@ -44,6 +45,7 @@ class HMASynthesizer(BaseHierarchicalSampler, BaseMultiTableSynthesizer):
             self.metadata,
             self._table_synthesizers,
             self._table_sizes)
+        self._print_estimate_warning()
 
     def set_table_parameters(self, table_name, table_parameters):
         """Update the table's synthesizer instantiation parameters.
@@ -143,6 +145,37 @@ class HMASynthesizer(BaseHierarchicalSampler, BaseMultiTableSynthesizer):
                 self._get_num_extended_columns(child_name, table_name, columns_per_table)
 
         visited.add(table_name)
+
+    def _print_estimate_warning(self):
+        total_est_cols = 0
+        metadata_columns = self._get_num_data_columns()
+        print_table = []
+        for table, est_cols in self._estimate_num_columns().items():
+            entry = []
+            entry.append(table)
+            entry.append(metadata_columns[table])
+            total_est_cols += est_cols
+            entry.append(est_cols)
+            print_table.append(entry)
+        if total_est_cols > 1000:
+            self._print(
+                'PerformanceAlert: Using the HMASynthesizer on this metadata '
+                'schema is not recommended. To model this data, HMA will '
+                f'generate a large number of columns. ({total_est_cols} columns)\n\n')
+            self._print(
+                tabulate(
+                    print_table,
+                    headers=[
+                        'Table Name',
+                        '# Columns in Metadata',
+                        'Est # Columns'
+                    ]
+                )
+            )
+            self._print(
+                '\nWe recommend simplifying your metadata schema by dropping '
+                'columns that are not necessary. If this is not possible, '
+                'contact us at info@sdv.dev for enterprise solutions.\n')
 
     def _get_num_data_columns(self):
         """Get the number of data columns, ie colums that are not id, for each table."""
--- a/sdv/multi_table/hma.py
+++ b/sdv/multi_table/hma.py
@@ -6,7 +6,6 @@ from copy import deepcopy
 
 import numpy as np
 import pandas as pd
-from tabulate import tabulate
 from tqdm import tqdm
 
 from sdv.errors import SynthesizerInputError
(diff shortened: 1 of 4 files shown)


sdv/multi_table/hma.py
```diff
@@ -144,6 +146,37 @@ def _estimate_columns_traversal(self, table_name, columns_per_table, visited):
...
+            entry.append(est_cols)
+            print_table.append(entry)
+        if total_est_cols > 1000:
+            self._print(
+                'PerformanceAlert: Using the HMASynthesizer on this metadata '
+                'schema is not recommended. To model this data, HMA will '
+                f'generate a large number of columns. ({total_est_cols} columns)\n\n')
+            self._print(
+                tabulate(
+                    print_table,
```
amontanez24, line 167: I think if you convert the `print_table` to a dataframe it should print nicely/ there should be a way to print it. That way we don't have to add the tabulate dependency

amontanez24 (commented):
Looks good! Could you add a unit test and maybe an integration test using the capsys fixture to test things are printing as expected. As you commented before, it probably would be good to automate the testing of our printed outputs

amontanez24 (approved):
Thanks for addressing the changes! LGTM

tests/integration/multi_table/test_hma.py
```diff
@@ -721,6 +721,42 @@ def test_progress_bar_print(self, capsys):
...
+            dataset_name='NBA_v1'
+        )
+
+        key_phrases = [
+            r'PerformanceAlert:',
+            r'large number of columns.',
+            r'contact us at info@sdv.dev for enterprise solutions.'
+        ]
+
+        HMASynthesizer(metadata)
```
R-Palazzo, line 738: Add `# Run` above

tests/integration/multi_table/test_hma.py
```diff
@@ -721,6 +721,42 @@ def test_progress_bar_print(self, capsys):
...
+        # Assert
+        for pattern in key_phrases:
+            match = re.search(pattern, captured.out + captured.err)
+            assert match is not None
+        (_, small_metadata) = download_demo(
+            modality='multi_table',
+            dataset_name='trains_v1'
+        )
+
+        HMASynthesizer(small_metadata)
```
R-Palazzo, line 751: Add `# Run` above

tests/unit/multi_table/test_hma.py
```diff
@@ -85,6 +85,42 @@ def test__get_extension(self):
 
         pd.testing.assert_frame_equal(result, expected)
 
+    def test_warning_message_too_many_cols(self, capsys):
+        """Test that a warning appears if there are more than 1000 expected columns"""
+
```
R-Palazzo, line 90: Add `# Setup`

tests/unit/multi_table/test_hma.py
```diff
@@ -85,6 +85,42 @@ def test__get_extension(self):
...
+        metadata = get_multi_table_metadata()
+
+        key_phrases = [
+            r'PerformanceAlert:',
+            r'large number of columns.',
+            r'contact us at info@sdv.dev for enterprise solutions.'
+        ]
+
+        with patch.object(HMASynthesizer,
```
R-Palazzo, line 99: Add `# Run` above

tests/unit/multi_table/test_hma.py
```diff
@@ -85,6 +85,42 @@ def test__get_extension(self):
...
+
+        captured = capsys.readouterr()
+
+        # Assert
+        for pattern in key_phrases:
+            match = re.search(pattern, captured.out + captured.err)
+            assert match is not None
+
+        small_metadata = get_multi_table_metadata()
+        with patch.object(HMASynthesizer,
```
R-Palazzo, line 112: Usually we use the decorator `@patch` but as you prefer here

R-Palazzo (commented):
Looks good!

tests/unit/multi_table/test_hma.py
```diff
@@ -85,6 +85,42 @@ def test__get_extension(self):
...
+            HMASynthesizer(metadata)
+
+        captured = capsys.readouterr()
+
+        # Assert
+        for pattern in key_phrases:
+            match = re.search(pattern, captured.out + captured.err)
+            assert match is not None
+
+        small_metadata = get_multi_table_metadata()
```
amontanez24, line 111: do we need this? The metadata is the same as the one above right? The only thing that changes is the return value for `_estimate_num_columns`.

tests/unit/multi_table/test_hma.py
lajohn4747: I used the patch.object instead of the annotation because I need to adjust the return value to test both branches of the print statement (low columns vs high columns).

R-Palazzo (approved):
LGTM!

PR #1655, opened 2023-10-26
Fix detection of discrete columns for `CTGANSynthesizer`

### LINKED ISSUE #1647: CTGAN sometimes creates new categories (int data)
_I'm filing this bug on on behalf of @SeasonJJ999, from #1632._

### Environment Details

* SDV version: 1.5.0
* Python version: 3.10

### Error Description
When a column has sdtype `'categorical'`, I expect that the synthetic data will _only_ contain values from the original data. However, CTGAN Synthesizer seems to generate new values that never appeared in the real data. This happens when there are a lot of different category values, and the data type is int.

### Steps to reproduce

```python
import pandas as pd

# dataframe has 2 columns, "A" is categorical
metadata_dict = {
    'METADATA_SPEC_VERSION': 'SINGLE_TABLE_V1',
    'columns': {
        'A': {'sdtype': 'categorical'},
        'B': {'sdtype': 'numerical'},
    }}

metadata = SingleTableMetadata.load_from_dict(metadata_dict)

# all columns contain values from 0 to 49 only
data = pd.DataFrame(data={
    'A': [i for i in range(50)],
    'B': [i for i in range(50)],
})

# fit CTGAN and sample some synthetic data
synth = CTGANSynthesizer(metadata, epochs=10)
synth.fit(data)
synthetic_data = synth.sample(1000)

# test to see if there if the categories are properly being recyled
original_categories = set(data['A'].unique())
synthetic_categories = set(synthetic_data['A'].unique())

recycled_categories = original_categories & synthetic_categories
new_categories = synthetic_categories - original_categories

print('Recycled Categories:', len(recycled_categories))
print('New Categories:', len(new_categories))
```

Output:
```
Recycled Categories: 50
New Categories: 30
```

If I list the new categories, I see values that were not present in the real data:
```
[50, 51, 52, ... 63, -18, -16, ..., -1]
```

### Workaround
One workaround is to pre-process the categorical column first, using something like the `LabelEncoder`. This will guarantee that the synthetic data has the same category values as the real data.

```python
from rdt.transformers.categorical import LabelEncoder

synth = CTGANSynthesizer(metadata, epochs=10)

# update pre-processing for categorical column "A"
synth.update_transformers(column_name_to_transformer={
    'A': LabelEncoder(add_noise=True)
})
synth.fit(data)
synthetic_data = synth.sample(1000)
```

Now the values are all recycled correctly.

### LINKED ISSUE #1648: CTGAN sometimes creates new categories (object data)
This issue is related to #1647. However, it is focused on columns represented as strings (objects). The previous issue was when the data is int.

### Environment Details

* SDV version: 1.5.0 
* Python version: 3.10

### Error Description
When a column has sdtype 'categorical', I expect that the synthetic data will only contain values from the original data. However, CTGAN Synthesizer seems to generate new values that never appeared in the real data.

### Steps to reproduce

```python
import pandas as pd

# dataframe has 2 columns, "A" is categorical
metadata_dict = {
    'METADATA_SPEC_VERSION': 'SINGLE_TABLE_V1',
    'columns': {
        'A': {'sdtype': 'categorical'},
        'B': {'sdtype': 'numerical'},
    }}

metadata = SingleTableMetadata.load_from_dict(metadata_dict)

# all columns contain values from 0 to 49 only
data = pd.DataFrame(data={
    'A': [str(i) for i in range(50)], # represent these as strings (object)
    'B': [i for i in range(50)],
})

# fit CTGAN and sample some synthetic data
synth = CTGANSynthesizer(metadata, epochs=10)
synth.fit(data)
synthetic_data = synth.sample(1000)

# test to see if there if the categories are properly being recyled
original_categories = set(data['A'].unique())
synthetic_categories = set(synthetic_data['A'].unique())

recycled_categories = original_categories & synthetic_categories
new_categories = synthetic_categories - original_categories

print('Recycled Categories:', len(recycled_categories))
print('New Categories:', len(new_categories))
```

Output:
```
Recycled Categories: 0
New Categories: 1000
```

**Problems**:
If you inspect new categories, you'll see that they are **unrounded numbers**:
```
[-0.2985470695455561, 1.7681338424734783, 1.373617579840067, ...]
```

This indicates that CTGAN is not properly treating these values as categorical. It seems to be modeling them as numerical.

 ### Workaround
Same as #1647

pvk-developer:
CU-86ayjjbw0
Resolves #1647 
Resolves #1648

### DIFF
--- a/sdv/single_table/utils.py
+++ b/sdv/single_table/utils.py
@@ -26,6 +26,10 @@ def detect_discrete_columns(metadata, data, transformers):
         data (pandas.DataFrame):
             ``pandas.DataFrame`` that matches the ``metadata``.
 
+        transformers (dict[str: rdt.transformers.BaseTransformer]):
+            A dictionary mapping between column names and the transformers assigned
+            for it.
+
     Returns:
         discrete_columns (list):
             A list of discrete columns to be used with some of ``sdv`` synthesizers.
@@ -46,6 +50,8 @@ def detect_discrete_columns(metadata, data, transformers):
                 discrete_columns.append(column)
                 continue
 
+        # Logic to detect columns produced by transformers outside of the metadata scope
+        # or columns created by constraints.
         column_data = data[column].dropna()
 
         # Ignore columns with only nans and empty datasets
--- a/sdv/single_table/utils.py
+++ b/sdv/single_table/utils.py
@@ -12,7 +12,7 @@ DISABLE_TMP_FILE = 'disable'
 IGNORED_DICT_KEYS = ['fitted', 'distribution', 'type']
 
 
-def detect_discrete_columns(metadata, data):
+def detect_discrete_columns(metadata, data, transformers):
     """Detect the discrete columns in a dataset.
 
     Because the metadata doesn't necessarily match the data (we only preprocess the data,
@@ -32,10 +32,19 @@ def detect_discrete_columns(metadata, data):
     """
     discrete_columns = []
     for column in data.columns:
-        # Numerical and datetime columns never get preprocessed into categorical ones
-        if column in metadata.columns and \
-                metadata.columns[column]['sdtype'] in ['numerical', 'datetime']:
-            continue
+        if column in metadata.columns:
+            sdtype = metadata.columns[column]['sdtype']
+            # Numerical and datetime columns never get preprocessed into categorical ones
+            if sdtype in ['numerical', 'datetime']:
+                continue
+
+            elif sdtype in ['categorical', 'boolean']:
+                transformer = transformers.get(column)
+                if transformer and transformer.get_output_sdtypes().get(column) == 'float':
+                    continue
+
+                discrete_columns.append(column)
+                continue
 
         column_data = data[column].dropna()
 
--- a/sdv/single_table/ctgan.py
+++ b/sdv/single_table/ctgan.py
@@ -218,7 +218,12 @@ class TVAESynthesizer(BaseSingleTableSynthesizer):
             processed_data (pandas.DataFrame):
                 Data to be learned.
         """
-        discrete_columns = detect_discrete_columns(self.get_metadata(), processed_data)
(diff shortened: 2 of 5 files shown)


sdv/single_table/utils.py
```diff
@@ -12,7 +12,7 @@
 IGNORED_DICT_KEYS = ['fitted', 'distribution', 'type']
 
 
-def detect_discrete_columns(metadata, data):
+def detect_discrete_columns(metadata, data, transformers):
```
rwedge, line 15: we should update the docstring args to include `transformers`

sdv/single_table/utils.py
```diff
@@ -32,10 +32,19 @@ def detect_discrete_columns(metadata, data):
...
+            if sdtype in ['numerical', 'datetime']:
+                continue
+
+            elif sdtype in ['categorical', 'boolean']:
+                transformer = transformers.get(column)
+                if transformer and transformer.get_output_sdtypes().get(column) == 'float':
+                    continue
+
+                discrete_columns.append(column)
+                continue
```
amontanez24, line 47: logic looks good but I think we don't need the logic below anymore. Before it was trying to look at columns marked as categorical in the metadata, and decide if we should treat them as discrete based on the unique values. But now since we actually are using the transformers to see what is categorical or not, I don't think it's still useful

pvk-developer: The underlying logic should remain unchanged for the following reason: When the `column` in question is missing from the metadata but is generated by a `transformer`, we rely on the `data` to determine whether it represents a discrete or numerical column. 

For instance, consider the case where we produce the `is_nan` column using the `model_missing_values`. Despite the column being numerical, we categorize it as discrete, given that it only contains two numerical categories.

amontanez24 (approved):
LGTM!

rwedge (approved):

PR #1663, opened 2023-10-31
Make SDV compatible with new `SDMetrics` visualizations

### LINKED ISSUE #1650: Make SDV compatible with SDMetrics 0.12.0
### Problem Description
SDMetrics 0.12.0 hasn't been released yet (as of writing this issue). When it is released, we'll have to make some updates in the SDV visualizations wrappers to support the new functionality.

_Note that SDV has a [pinned requirement](https://github.com/sdv-dev/SDV/blob/main/setup.py#L31) of sdmetrics<0.12. So we'll need to bump this up too when we're done._ 

### Expected behavior
Update the visualization functions for both [Single Table](https://docs.sdv.dev/sdv/single-table-data/evaluation#visualization) and [Multi Table](https://docs.sdv.dev/sdv/multi-table-data/evaluation#visualization).

- `get_column_plot`:
    - Add an optional parameter called `plot_type`
        - By default, we should determine the plot type based on the metadata. (a) if sdtype is numerical/datetime, use `'distplot'`, (b) if sdtype is categorical/boolean, use `'bar'`, (c) otherwise, the data is incompatible so raise an Error.
        - If the user provides a plot type, then their provided value overrides the logic above
    - For datetime columns, SDV should convert the data to a `datetime64` dtype, using the provided `datetime_format` from the metadata. Pass the converted data along to the SDMetrics visualization
- `get_column_pair_plot`
    - Similar to above, add an optional parameter called `plot_type`
        - By default, determine plot type based on meatdata. (a) if both sdtypes are numerical/datetime, then use `'scatter'`, (b) if both sdtypes are categorical/boolean, then use `'heatmap'`, (c) if one is numerical/datetime and the other is categorical/boolean, use `'box'`, (d) otherwise, raise an Error because the data is incompatible
        - If the user provides a plot type, t hen their provided value overrides the logic above
    - For datetime columns, SDV should convert the data to a `datetime64` dtype, using the provided `datetime_format` from the metadata. Pass the converted data along to the SDMetrics visualization
- (multi-table only) `get_cardinality_plot`
    - Add an optional parameter called `plot_type`
        - By default, the plot type should be `'bar'` in all cases
        - If the user provides a plot type, then pass that along instead

### Additional Context

For the error messages, we may want to create a new type of error for visualizations.
```
>>> fig = get_column_plot(
    real_data=real_data,
    column_name='user_id',
    metadata=metadata,
)

VisualizationUnavailableError: The column 'user_id' has sdtype 'id', which does not have a supported visualization. To visualize this data anyways, please add a 'plot_type'.
```

pvk-developer:
CU-86ayjpx8f
Resolves #1650 

Find the notebook with the graphics [here](https://drive.google.com/file/d/1zyq1edk6ExwS0x1YCX3-D2suyRueuWtr/view?usp=sharing).

### DIFF
--- a/sdv/errors.py
+++ b/sdv/errors.py
@@ -54,3 +54,7 @@ class InvalidDataError(Exception):
             'The provided data does not match the metadata:\n' +
             '\n\n'.join(map(str, self.errors))
         )
+
+
+class VisualizationUnavailableError(Exception):
+    """Exception to indicate that a visualization is unavailable."""
--- a/sdv/evaluation/multi_table.py
+++ b/sdv/evaluation/multi_table.py
@@ -67,7 +67,8 @@ def get_column_plot(real_data, synthetic_data, metadata, table_name, column_name
         column_name (str):
             The name of the column.
         plot_type (str or None):
-            The plot type to use to plot the cardinality. Must be either 'bar' or 'distplot'.
+            The plot type to use to plot the cardinality. Must be either 'bar' or 'distplot'. If
+            ``None``, select between 'bar' or displot depending on the data.
             Defaults to ``None``.
 
     Returns:
--- a/sdv/evaluation/multi_table.py
+++ b/sdv/evaluation/multi_table.py
@@ -1,7 +1,9 @@
 """Methods to compare the real and synthetic data for multi-table."""
-import sdmetrics.reports.utils as report
+from sdmetrics import visualization
 from sdmetrics.reports.multi_table.diagnostic_report import DiagnosticReport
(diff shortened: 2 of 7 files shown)


npatki (approved):
LGTM

amontanez24 (approved):
One minor comment, but this looks good!

R-Palazzo (approved):
LGTM! 👍

frances-h (approved):
Looks good! Just one small nitpick, otherwise :shipit:

PR #1664, opened 2023-11-01
Add `_estimate_num_columns` method to ctgan

### LINKED ISSUE #1657: Make function to estimate number of columns CTGAN produces
### Problem Description
_As a user, I'd like to know as soon as possible if CTGAN will be slow so that I can properly decide whether or not to use it._

### Expected behavior

For CTGAN, write a function to estimate the number of columns that will be created. Unlike HMA though, this function will need as input both the metadata _and_ the actual data. And the transformers be already be assigned.

Rough logic:
1. Figure out if the column will be modeled, and whether it would be modeled as continuous or discrete
   - If the sdtype is numerical/datetime, then it will be modeled as continuous
   - Else if the sdtype is categorical/boolean, there is some case logic:
      - If transformers haven't been determined yet, throw an error.
      - Else if transformers have been determine AND there is no transformer assigned to the column, it will be modeled as discrete
      - Else it will be modeled as continuous
   - Else: sdtype is ID or PII. For now, we can consider these columns as being dropped.
2. The number of columns is formulaic
   - Each continuous column will result in 11 columns
   - Each discrete column will result in _n_ columns, where _n_ is the number of unique categories in the data

The function would return a dictionary. Keys will be original columns names, and the values will be the number of columns.

### Example

```python
>>> synth = CTGANSynthesizer(metadata)
>>> synth._estimate_num_columns(data)
SynthesizerInputError: No transformers are assigned. Use 'auto_assign_transformers' before using this function

>>> synth.auto_assign_transformers(data)
>>> synth._estimate_num_columns(data)
# only the statistical columns (numerical, datetime, boolean, categorical) are returned
{
  'age': 11,
  'transactions': 11,
  'ethnicity': 10,
  'marital_status': 5,
  'country_of_origin': 195
}

>>> synth.update_transformers({
        'country_of_origin': UniformEncoder()})

>>> synthesizer._estimate_num_columns(data)
# now notice that country_of_origin has 11 because it's being treated as numerical
{
  'age': 11,
  'transactions': 11,
  'ethnicity': 10,
  'marital_status': 5,
  'country_of_origin': 11
}
```

### Additional Context

This function is for internal use.

fealho:
CU-86ayfuk2q
Resolve #1657.

Note: this doesn't implement the error message described in the issue - "if transformers haven't been determined yet, throw an error" - because `get_transformers` does that internally.

### DIFF
--- a/sdv/single_table/ctgan.py
+++ b/sdv/single_table/ctgan.py
@@ -1,5 +1,5 @@
 """Wrapper around CTGAN model."""
-
+import numpy as np
 from ctgan import CTGAN, TVAE
 
 from sdv.single_table.base import BaseSingleTableSynthesizer
@@ -131,7 +131,7 @@ class CTGANSynthesizer(BaseSingleTableSynthesizer):
 
             elif sdtypes[column] in {'categorical', 'boolean'}:
                 if transformers[column] is None:
-                    num_categories = data[column].nunique(dropna=False)
+                    num_categories = data[column].fillna(np.nan).nunique(dropna=False)
                     num_generated_columns[column] = num_categories
                 else:
                     num_generated_columns[column] = 11
--- a/sdv/single_table/ctgan.py
+++ b/sdv/single_table/ctgan.py
@@ -133,9 +133,6 @@ class CTGANSynthesizer(BaseSingleTableSynthesizer):
                 else:
                     num_generated_columns[column] = 11
 
-            else:
-                num_generated_columns[column] = 0
-
         return num_generated_columns
 
     def _fit(self, processed_data):
--- a/sdv/single_table/ctgan.py
+++ b/sdv/single_table/ctgan.py
@@ -111,6 +111,9 @@ class CTGANSynthesizer(BaseSingleTableSynthesizer):
     def _estimate_num_columns(self, data):
         """Estimate the number of columns that the data will generate.
 
+        Estimates that continuous columns generate 11 columns and categorical ones
+        create n where n is the number of unique categories.
+
         Args:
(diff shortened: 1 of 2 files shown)


sdv/single_table/ctgan.py
```diff
@@ -108,6 +108,33 @@ def __init__(self, metadata, enforce_min_max_values=True, enforce_rounding=True,
...
+        sdtypes = self._data_processor.get_sdtypes()
+        transformers = self.get_transformers()
+        num_generated_columns = {}
+        for column in data.columns:
+            if sdtypes[column] in {'numerical', 'datetime'}:
+                num_generated_columns[column] = 11
+
+            elif sdtypes[column] in {'categorical', 'boolean'}:
+                if transformers[column] is None:
+                    num_categories = data[column].nunique()
```
rwedge, line 131: should we have `dropna=False` to handle NaN values as a category?

frances-h (approved):

sdv/single_table/ctgan.py
```diff
@@ -108,6 +108,33 @@ def __init__(self, metadata, enforce_min_max_values=True, enforce_rounding=True,
             'cuda': cuda
         }
 
+    def _estimate_num_columns(self, data):
+        """Estimate the number of columns that the data will generate.
+
```
frances-h, line 113: Can we add a line in the docstring explaining that continuous columns generate 11 columns and categorical create n where n is the number of unique categories (from the issue)? I think it'd be helpful to keep track of where all the magic numbers are coming from.

rwedge (approved):

PR #1669, opened 2023-11-07
Print alert if `CTGANSynthesizer` is likely to be slow

### LINKED ISSUE #1658: Print alert if CTGANSynthesizer is likely to be slow
### Problem Description
_As a user, it would be helpful to know as soon as possible if my data will take too long to train so that I can make any necessary adjustments._

### Expected behavior

- As the first step of the `preprocess` function, apply the `_estimate_num_columns` function specified in #1657 
- If the total number of columns summed across all tables is > 1000, print the following message
- Continue with on with the method. That is, **the message should not prevent you from ultimately fitting and sampling data**

```
PerformanceAlert:Using the CTGANSynthesizer on this data is not recommended. To model this data, CTGAN will generate a large number of columns.

Original Column Name       Est # of Columns (CTGAN)
age                        11 
transactions               11
country                    195
city                       10234

We recommend preprocessing discrete columns that can have many values, using 'update_transformers'. Or you may drop columns that are not necessary to model. (Exit this script using ctrl-C.)
```

- The table must be pretty printed so the tabs/columns line up like above
- Make sure to call it an alert to raise attention

### Additional context

- In this case, it might make sense to just print instead of raising a warning or logging since we don't want it to be disabled and its main intent is to be used live.
- Because this will happen in `preprocess`, we should also expect this alert to pop up when the user calls `fit`

fealho:
CU-86ayk4h74
Resolve #1658

### DIFF
--- a/sdv/single_table/ctgan.py
+++ b/sdv/single_table/ctgan.py
@@ -138,6 +138,31 @@ class CTGANSynthesizer(BaseSingleTableSynthesizer):
 
         return num_generated_columns
 
+    def _preprocess(self, data):
+        dict_generated_columns = self._estimate_num_columns(data)
+        if sum(dict_generated_columns.values()) > 1000:
+            header = {'Original Column Name  ': 'Est # of Columns (CTGAN)'}
+            dict_generated_columns = {**header, **dict_generated_columns}
+            longest_column_name = len(max(dict_generated_columns, key=len))
+            cap = '<' + str(longest_column_name)
+            lines_to_print = []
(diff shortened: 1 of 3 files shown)


amontanez24:
Task linked: [CU-86ayk4h74 SDV - Print alert if CTGANSynthesizer is likely to be slow #1658](https://app.clickup.com/t/86ayk4h74)

rwedge (commented):
Can we also test this alert does not appear if we are below the columns generated threshold?

rwedge (approved):

pvk-developer (approved):

PR #1670, opened 2023-11-07
Replace invalid symbols in the labels for metadata visualization

### LINKED ISSUE #1625: Unable to visualize metadata (`Error: bad label format` and `CalledProcessError`)
### Environment Details
* SDV version: 1.4.0 (latest)
* Python version: 3.10

### Error Description
When attempting to visualize my metadata for a single table, I am encountering a `CalledProcessError` from the underlying graphing library (graphviz). It appears that there is something misconfigured in the label as the first line is `Error: bad label format`.

### Steps to reproduce
Download the [Income dataset from Kaggle](https://www.kaggle.com/datasets/mastmustu/income). You can use the `train.csv` function to auto detect metadata and attempt to visualize it.

```python
import pandas as pd

real_data = pd.read_csv('train.csv')
real_data['income_>50K'] = real_data['income_>50K'].astype(str)

metadata.detect_from_dataframe(real_data)

metadata.visualize()
```

Output:
```
Error: bad label format 
...
CalledProcessError: '[PosixPath('dot'), '-Kdot', '-Tsvg']' returned non-zero exit status 1. ...
```

Full stack trace is attached below.
[stack_trace.txt](https://github.com/sdv-dev/SDV/files/12863158/stack_trace.txt)

fealho:
CU-86ayfuk2q
Resolve #1625.

### DIFF
--- a/sdv/metadata/visualization.py
+++ b/sdv/metadata/visualization.py
@@ -69,6 +69,8 @@ def _get_graphviz_extension(filepath):
 
     return None, None
 
+def _replace_special_characters(string):
+    return string.replace('<', '_less_than_').replace('>', '_greater_than_')
 
 def visualize_graph(nodes, edges, filepath=None):
     """Plot metadata usign graphviz.
@@ -105,10 +107,10 @@ def visualize_graph(nodes, edges, filepath=None):
     )
 
     for name, label in nodes.items():
-        digraph.node(name, label=graphviz.escape(label))
+        digraph.node(name, label=_replace_special_characters(label))
 
     for parent, child, label in edges:
-        digraph.edge(parent, child, label=graphviz.escape(label), arrowhead='oinv')
+        digraph.edge(parent, child, label=_replace_special_characters(label), arrowhead='oinv')
 
     if filename:
         digraph.render(filename=filename, cleanup=True, format=graphviz_extension)
--- a/sdv/metadata/visualization.py
+++ b/sdv/metadata/visualization.py
@@ -75,7 +75,7 @@ def _replace_special_characters(string):
 
 
 def visualize_graph(nodes, edges, filepath=None):
-    """Plot metadata usign graphviz.
+    """Plot metadata using graphviz.
 
     Try to generate a plot using graphviz.
     If a ``filepath`` is provided save the output into a file.
--- a/sdv/metadata/visualization.py
+++ b/sdv/metadata/visualization.py
@@ -105,11 +105,10 @@ def visualize_graph(nodes, edges, filepath=None):
     )
 
     for name, label in nodes.items():
-        label = label.replace('<', 'lt').replace('>', 'gt')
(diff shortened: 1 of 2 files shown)


amontanez24:
Task linked: [CU-86ayfuk2q SDV - Unable to visualize metadata (`Error: bad label format` and `CalledProcessError`) #1625](https://app.clickup.com/t/86ayfuk2q)

amontanez24 (commented):
Let's add a test but otherwise looks good!

sdv/metadata/visualization.py
```diff
@@ -105,10 +107,10 @@ def visualize_graph(nodes, edges, filepath=None):
     )
 
     for name, label in nodes.items():
-        digraph.node(name, label=graphviz.escape(label))
+        digraph.node(name, label=_replace_special_characters(label))
```
amontanez24, line 112: have you tried using https://graphviz.readthedocs.io/en/stable/api.html#graphviz.nohtml

fealho: Yes, it doesn't address the problem.

amontanez24: I see. It seems like it works if you replace '>' with '\\>'. Then in the output graph the label looks correct 
Screenshot 2023-11-08 at 9 51 51 PM

fealho: Just a note, this approach removes the backslash from the column name. So if a column has `\<` in it, the output will be `<`.

amontanez24: I think that's ok since prior to this change if they had '\\>' in a label, it would only show '>' anyway.

amontanez24 (commented):
Can we add an integration test for this case

tests/integration/metadata/test_visualization.py
```diff
@@ -0,0 +1,31 @@
...
+
+    # Run
+    metadata.visualize()
+
+
+def test_visualize_graph_for_multi_table():
+    """Test it runs when a column name contains `>`."""
+    # Setup
+    data1 = pd.DataFrame({'>': ['a', 'b', 'c']})
+    data2 = pd.DataFrame({'>': ['a', 'b', 'c']})
```
amontanez24, line 21: Can we test both symbols? and maybe the symbols mixed with text

amontanez24 (approved):

tests/integration/metadata/test_visualization.py
```diff
@@ -0,0 +1,42 @@
...
+
+    # Run
+    metadata.visualize()
+    metadata.validate()
+    model.fit(data)
+    model.sample(10)
+
+
+def test_visualize_graph_for_multi_table():
+    """Test it runs when a column name contains symbols."""
```
R-Palazzo, line 24: You can put a different message that the test before to be more specific

R-Palazzo (approved):
LGTM!

PR #1672, opened 2023-11-08
Visualization isn't useful when there are too many data points. Can we subsample?

### LINKED ISSUE #1461: Visualization isn't useful when there are too many data points. Can we subsample?
### Problem Description
The [visualization functions](https://docs.sdv.dev/sdv/single-table-data/evaluation#visualization) are an extremely useful way to see what's happening in my real and synthetic data.

However, if I have a large amount of data, then:
- The visualizations can take a long time to render
- The scatterplots in particular are not informative, as the dots take up the overall screen

Eg. of column pair plot with 50K data points. You cannot even see the real data because the synthetic data is blocking it.
image

### Expected behavior
Adding a new parameter to control sub-sampling. When set to an integer, we can sample the desired number of rows from the real and synthetic data before visualizing. This will make the graphs much faster to create and follow.

**Proposed API**: Add a new parameter called `sample_size`
- (default) `None`: Do not sample. Visualize the full synthetic and real data
- `<integer>`: Sub-sample the real and synthetic data rows (without replacement!) before visualizing. _Note: If there are fewer rows than `<integer>`, then there is no need to sub-sample_

Eg. Taking a sample of 150 data points, it's much easier to see the pattern.
image

### Additional context
This applies to all visualization functions for single and multi-table.

R-Palazzo:
CU-85zt7vbp9
Resolve #1461

### DIFF
--- a/sdv/evaluation/multi_table.py
+++ b/sdv/evaluation/multi_table.py
@@ -88,7 +88,7 @@ def get_column_plot(real_data, synthetic_data, metadata, table_name, column_name
 
 
 def get_column_pair_plot(real_data, synthetic_data, metadata,
-                         table_name, column_names, sample_size=None, plot_type=None):
+                         table_name, column_names, plot_type=None, sample_size=None):
     """Get a plot of the real and synthetic data for a given column pair.
 
     Args:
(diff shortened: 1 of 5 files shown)


amontanez24:
Task linked: [CU-85zt7vbp9 SDV - Visualization isn't useful when there are too many data points. Can we subsample? #1461](https://app.clickup.com/t/85zt7vbp9)

fealho (approved):

sdv/evaluation/multi_table.py
```diff
@@ -88,7 +88,7 @@ def get_column_plot(real_data, synthetic_data, metadata, table_name, column_name
 
 
 def get_column_pair_plot(real_data, synthetic_data, metadata,
-                         table_name, column_names, plot_type=None):
+                         table_name, column_names, sample_size=None, plot_type=None):
```
amontanez24, line 91: can we add this as the last parameter?

R-Palazzo: Yes, done in [588f155](https://github.com/sdv-dev/SDV/pull/1672/commits/588f155d77a097b4601802a3a06ea567d62c2f7d)

amontanez24 (commented):
Looks good! Can we add an integration test?

amontanez24 (approved):
LGTM!

PR #1677, opened 2023-11-15
Set enforce_min_max_values to True for datetime transformers

### LINKED ISSUE #1676: Set enforce_min_max_values to True for datetime transformers
### Problem Description

Sometime datetime values generated by SDV synthesizer are out of bounds.

### Expected behavior

Set the `enforce_min_max_values` parameters of the datetime transformers to `True` by default to ensure that the synthetic data is always in the same range as the real data

R-Palazzo:
CU-86ayq41bk
Resolve #1676

In this PR, RDT is upgraded to 1.9.0
I also noticed that following the logic of this PR, the DataProcessor could be cleaned a bit, mostly concerning the `Floatformatter`. I propose to get rid of the `_update_numerical_transformer` and try to do something similar to this PR, so update the transformer inside `_update_transformers_by_sdtypes`. 
Let me know you thought about it, if we want to do this change and if yes, inside this PR or in a separate issue :wink:


sdv-team:
Task linked: [CU-86ayq41bk SDV - Set enforce_min_max_values to True for datetime transformers #1676](https://app.clickup.com/t/86ayq41bk)

amontanez24 (commented):
Can we add an integration test that asserts the datetime values are in the range?

amontanez24 (approved):
I think this should work!

pvk-developer (approved):
LGTM! Good solution 😉

PR #1679, opened 2023-11-16
Create a unified get_parameters method for all multi-table synthesizers

### LINKED ISSUE #1674: Create a unified `get_parameters` method for all multi-table synthesizers
### Problem Description
The multi-table synthesizers allow users to set and get parameters. Setting parameters must be different for the different synthesizers -- as each has different options based on the underlying algorithm. However, getting the parameters should be more consistent.

### Expected behavior
All synthesizers should have a single method called `get_parameters()` that returns all the parameters used for the synthesizer. It should include:
- Any parameters used during initialization (eg. `locales`, `num_clusters`, etc.)
- A nested dictionary called `'tables'` with settings specific to each table. The format of the dictionary would vary based on the type of multi-table algorithm.

The expected behavior is for the public-facing synthesizer only (HMA).  We'll file a separate issue for HSA, Independent and DayZ (enterprise features).

```python
# HMASynthesizer
>>> hma_synth.get_parameters()
{
  'locales': ['en_US', 'fr_CA'], # add the init params as top level
  'verbose': True,
  'tables': { # the table specific params are now in a nested dictionary
    'users': {
      'table_synthesizer': 'GaussianCopulaSynthesizer', # only possible synthesizer for HMA
      'table_parameters': { # add the table-level parameters here
        'enforce_min_max_values': False,
        'default_distribution': 'truncnorm',
      }
    }
  }
}
```

### Additional context
We can deprecate the `get_table_parameters` method, as this method would return all the parameters.

frances-h:
Resolve #1674

### DIFF
--- a/tests/unit/multi_table/test_base.py
+++ b/tests/unit/multi_table/test_base.py
@@ -236,8 +236,8 @@ class TestBaseMultiTableSynthesizer:
             'numerical_distributions': {}
         }
 
-    def test_get_parameters(self):
-        """Test that the table's synthesizer parameters are being returned."""
+    def test_get_parameters_missing_table(self):
+        """Test that the synthesizer's parameters are being returned."""
         # Setup
         metadata = get_multi_table_metadata()
         instance = BaseMultiTableSynthesizer(metadata, locales='en_CA')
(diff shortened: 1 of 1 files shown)


sdv/multi_table/base.py
```diff
@@ -126,21 +128,39 @@ def get_table_parameters(self, table_name):
...
+
         return self._table_parameters.get(table_name, {})
 
-    def get_parameters(self, table_name):
-        """Return the parameters used to instantiate the table's synthesizer.
-
-        Args:
-            table_name (str):
-                Table name for which the parameters should be retrieved.
+    def get_parameters(self):
```
lajohn4747, line 139: Just curious, would we want to still be able to get single table parameters? Maybe leave `table_name = None` to get all tables otherwise get the specified one?

frances-h: Talked to @npatki about this a bit, and the preferred solution (if we want this functionality) would be to update `get_table_parameters` to do this instead of adding an argument here.

lajohn4747 (approved):
lgtm

rwedge (commented):
looks like we should update the docstring and maybe the test name for `test_get_table_parameters_empty`

sdv/multi_table/base.py
```diff
@@ -126,21 +126,33 @@ def get_table_parameters(self, table_name):
                 A dictionary representing the parameters that will be used to instantiate the
                 table's synthesizer.
         """
-        return self._table_parameters.get(table_name, {})
+        table_synthesizer = self._table_synthesizers.get(table_name)
+        if not table_synthesizer:
+            table_params = {'table_synthesizer': None, 'table_parameters': {}}
```
rwedge, line 131: Codecov reports this line uncovered

rwedge (approved):
Looks good

PR #1681, opened 2023-11-17
Allow me to access loss values for GAN-based synthesizers

### LINKED ISSUE #1671: Allow me to access loss values for GAN-based synthesizers
### Problem Description
The GAN-based synthesizers work by iterating through many different epochs and optimizing a loss function. Currently, all of our core ML models store a DataFrame with the loss values per epoch under a parameter called `.loss_values`.

See: 
- [CTGAN #298](https://github.com/sdv-dev/CTGAN/issues/298)
- [CTGAN #300](https://github.com/sdv-dev/CTGAN/issues/300)
- [DeepEcho #80](https://github.com/sdv-dev/DeepEcho/issues/80)

Now that the underlying ML models are storing loss values, they can be exposed through the SDV synthesizers using a simple function.

### Expected behavior
Each of the GAN-based synthesizers should include a function `get_loss_values()` that returns a copy of the loss values. If the synthesizer has not yet been fit, then this function could raise an error.

```python
# available for CTGAN, TVAE, CopulaGAN and PAR
synthesizer.fit(data)
loss_values_df = synthesizer.get_loss_values()

# if a synthesizer has not yet been fit
unfit_synthesizer.get_loss_values()
SynthesizerProcessingError: Loss values are not available yet. Please fit your synthesizer first.
```

For multi-table synthesizes, it's possible to have a GAN-based model for single tables. (In both HSA and Independent). These synthesizers should then be able to return the loss values given the table name.

```python
# for HSA and Independent, it's possible to use GAN-based synthesizers for individual table modeling
hsa_synthesizer.set_table_parameters(
  table_name='users',
  table_synthesizer='CTGANSynthesizer')

# error if you haven't fit yet
users_loss = hsa_synthesizer.get_loss_values(table_name='users')
SynthesizerProcessingError: Loss values are not available yet. Please fit your synthesizer first.

# after fitting, it should be possible to access the loss values for those tables
hsa_synthesizer.fit(data)
users_loss = has_synthesizer.get_loss_values(table_name='users')

# if you try to get loss values from something that's not GAN-based, then there should be an error
txns_loss = hsa_synthesizer.get_loss_values(table_name='transactions')
SynthesizerInputError: Loss values are not available for table 'transactions' because the table does not use a GAN-based model.
```

### Additional context
In order to do this issue, we'll likely need to make a CTGAN release and a DeepEcho release first. (The PRs are merged, but the releases haven't been made, so the SDV won't be able to access loss values.)

frances-h:
Resolve #1671

### DIFF
--- a/sdv/multi_table/base.py
+++ b/sdv/multi_table/base.py
@@ -406,6 +406,29 @@ class BaseMultiTableSynthesizer:
             f"table because it uses the '{synthesizer.__class__.__name__}'."
         )
 
+    def get_loss_values(self, table_name):
+        """Get the loss values from a model for a table.
+
+        Return a pandas dataframe mapping of the loss values per epoch of GAN
+        based synthesizers
+
+        Args:
+            table_name (str):
+                Table name for which the parameters should be retrieved.
+
+        Returns:
+            pd.DataFrame:
+                Dataframe of loss values per epoch
+        """
+        synthesizer = self._table_synthesizers[table_name]
+        if hasattr(synthesizer, 'get_loss_values'):
+            return synthesizer.get_loss_values()
+
+        raise SynthesizerInputError(
+            f"Loss values are not available for table '{table_name}' "
+            'because the table does not use a GAN-based model.'
+        )
+
     def _validate_constraints_to_be_added(self, constraints):
         for constraint_dict in constraints:
             if 'table_name' not in constraint_dict.keys():
--- a/sdv/multi_table/base.py
+++ b/sdv/multi_table/base.py
@@ -420,6 +420,11 @@ class BaseMultiTableSynthesizer:
             pd.DataFrame:
                 Dataframe of loss values per epoch
         """
+        if table_name not in self._table_synthesizers:
+            raise ValueError(
+                f"Table '{table_name}' is not present in the metadata."
+            )
+
         synthesizer = self._table_synthesizers[table_name]
         if hasattr(synthesizer, 'get_loss_values'):
             return synthesizer.get_loss_values()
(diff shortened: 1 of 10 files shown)


sdv/sequential/par.py
```diff
@@ -266,6 +267,22 @@ def _fit(self, processed_data):
...
+
+        Returns:
+            pd.DataFrame:
+                Dataframe containing the loss values per epoch.
+        """
+        if not self._fitted:
+            err_msg = 'Loss values are not available yet. Please fit your synthesizer first.'
+            raise NotFittedError(err_msg)
+
+        return self._model.loss_values.copy()
```
rwedge, line 284: no longer needed due to mixin

rwedge (approved):
LGTM

sdv/multi_table/base.py
```diff
@@ -406,6 +406,29 @@ def get_learned_distributions(self, table_name):
...
+
+        Args:
+            table_name (str):
+                Table name for which the parameters should be retrieved.
+
+        Returns:
+            pd.DataFrame:
+                Dataframe of loss values per epoch
+        """
+        synthesizer = self._table_synthesizers[table_name]
```
amontanez24, line 423: should we error if the table name is bad? I think there is an error message we use in a lot of places for that

sdv/single_table/utils.py
```diff
@@ -349,3 +349,23 @@ def log_numerical_distributions_error(numerical_distributions, processed_data_co
             f"cannot be applied to column '{column}' because it no longer "
             'exists after preprocessing.'
         )
+
+
+class GANMixin:
```
amontanez24, line 354: How do you feel about putting this class in the ctgan file since all the gans are already there? Also I think VAEs aren't technically GANs. Maybe we should change the name?

sdv/multi_table/base.py
frances-h: I think we usually use `_validate_table_name` but the full error message doesn't really make sense in this context, so I used an edited version.

PR #1710, opened 2023-12-07
Add column relationship validation to single table metadata

### LINKED ISSUE #1698: Add column relationship validation to single table metadata
### Problem Description

_As a user, I'd like to know if I formatted the column relationships in my metadata incorrectly so that I don't run methods that will take longer only to have them crash._

The [SingleTableMetadata ](https://github.com/sdv-dev/SDV/blob/main/sdv/metadata/single_table.py)object is being upgraded to accept a new field called `column_relationships`. The format is as follows:

```json
{
  "columns": {
    "user_lat": { "sdtype": "latitude" },
    "user_lon": { "sdtype": "longitude" },
  },
  "column_relationships": [{
    "type": "gps_coordinate",
    "column_names": ["user_lat", "user_lon" ],
  }]
}
```
### Expected behavior

When running `metadata.validate()` on `SingleTableMetadata`, we should perform the following checks:
- type must be a supported relationship type (only "address" for now)
- column_names must all be listed in the "columns" section of the metadata
- The column_names cannot already be involved in any other column relationship
- The column_names cannot be a primary key or a foreign key (listed in the relationships)
- The columns' sdtypes must be compatible with the relationship type. Each type will have different rules for this.
    - Eg. address supports sdtypes: postcode, state, city, …
If any of the above are invalid, we should raise an `InvalidMetadataError`.

### Additional context
- We should put all of this validation into its own function since it will be reused by the add_column_relationship function in #1699 .
- We should follow the same validation style as below:
https://github.com/sdv-dev/SDV/blob/57feae881c29ea3f591e52d83897c110722137d3/sdv/metadata/single_table.py#L557-L572

frances-h:
CU-86aytvjvb
Resolve #1698

### DIFF
--- a/sdv/metadata/multi_table.py
+++ b/sdv/metadata/multi_table.py
@@ -564,6 +564,26 @@ class MultiTableMetadata:
         warnings.warn('Sequential modeling is not yet supported on SDV Multi Table models.')
         self.tables[table_name].set_sequence_index(column_name)
 
+    def _validate_column_relationships_foreign_keys(self, table_meta, foreign_keys):
+        """Validate that a table's column relationships do not use any foreign keys.
+
+        Args:
+            table_meta (SingleTableMetadata):
+                The table's metadata.
+            foreign_keys (list):
+                The list of foreign keys in the table.
+
+        Raises:
+            - ``InvalidMetadataError`` if foreign keys are used in any column relationships.
+        """
+        for column_relationship in table_meta.column_relationships:
+            column_names = set(column_relationship.get('column_names', []))
+            invalid_columns = column_names.intersection(foreign_keys)
+            if invalid_columns:
+                raise InvalidMetadataError(
+                    f'Cannot use foreign keys {invalid_columns} in column relationship.'
+                )
+
     def _validate_single_table(self, errors):
         for table_name, table in self.tables.items():
             if len(table.columns) == 0:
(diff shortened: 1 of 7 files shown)


sdv-team:
Task linked: [CU-86aytvjvb SDV - Add column relationship validation to single table metadata #1698](https://app.clickup.com/t/86aytvjvb)

sdv/metadata/multi_table.py
```diff
@@ -579,6 +599,14 @@ def _validate_single_table(self, errors):
                 error = str(error).replace(
                     'The following errors were found in the metadata:\n', title)
                 errors.append(error)
+                try:
+                    foreign_key_cols = [
+                        relationship['child_foreign_key'] for relationship in self.relationships
+                        if table_name == relationship.get('child_table_name')
+                    ]
+                    self._validate_column_relationships_foreign_keys(table, foreign_key_cols)
+                except Exception as col_relationship_error:
```
amontanez24, line 608: why does this need to be nested in the previous except? Also, can we calculate the foreign_key_cols once outside the for loop?

sdv/metadata/single_table.py
```diff
@@ -608,6 +616,94 @@ def _append_error(self, errors, method, *args, **kwargs):
...
+        # Validate each individual relationship
+        errors = []
+        for relationship in column_relationships:
+            relationship_type = relationship['type']
+            columns = relationship['column_names']
+            self._append_error(
+                errors,
+                self._validate_column_relationship,
+                relationship_type,
+                columns)
```
amontanez24, line 699: minor: can we move the parenthesis down like below?

amontanez24 (approved):
LGTM!

pvk-developer (approved):

PR #1713, opened 2023-12-11
Fix `BaseIndependentSampler` crashing with ID columns being new `dtype`

### LINKED ISSUE #1712: `BaseIndependentSampler` crashes because it tries to cast id columns
### Environment Details
* SDV version: 1.8.0

### Error Description
`BaseIndependentSampler` crashes with `ValueError: invalid literal for int() with base 10` whenever I have the following scenario:

1. In the metadata, I have specified a regex for an `id` column. The regex implies that the values need to be a string (eg. `[A-Z]{3}`)
2. In the data itself, the `id` columns are actually represented as integers such as `0, 1, 2, ...`

During sampling, the `BaseIndependentSampler` tries to cast the generated regexes (string) back into the original dtype (ints). This causes a crash.

**Expected Behavior**: Any errors during the casting process should be _caught and excepted_ if the sdtype is `id` or any of the PII types (such as `phone_number`, `postcode`, etc.).
- For these types of columns, we should let the user do what they wish. We should always _try_ to cast, but if it isn't possible, just don't cast it. In this case, just return the strings as ids.
- The casting process should only cause an error for statistical types (`datetime`, `boolean`, `categorical`, `numerical`)

pvk-developer:
CU-86ayvkf8c
Resolves #1712

### DIFF
--- a/sdv/sampling/independent_sampler.py
+++ b/sdv/sampling/independent_sampler.py
@@ -102,18 +102,17 @@ class BaseIndependentSampler():
             dtypes_to_sdtype = synthesizer._data_processor._DTYPE_TO_SDTYPE
 
             for name, dtype in dtypes.items():
-
                 try:
                     table_rows[name] = table_rows[name].dropna().astype(dtype)
-
                 except ValueError as e:
                     column_metadata = metadata.columns.get(name)
                     sdtype = column_metadata.get('sdtype')
                     if sdtype not in dtypes_to_sdtype.values():
                         LOGGER.info(
-                            f"The real data in '{name}' was stored as '{dtype}' but the "
-                            'synthetic data could not be cast back to this type. If this is a '
-                            'problem, please check your input data and metadata settings.'
+                            f"The real data in '{table_name}' and column '{name}' was stored as "
+                            f"'{dtype}' but the synthetic data could not be cast back to "
+                            'this type. If this is a problem, please check your input data '
+                            'and metadata settings.'
                         )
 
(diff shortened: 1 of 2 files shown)


sdv/sampling/independent_sampler.py
```diff
@@ -95,10 +95,29 @@ def _finalize(self, sampled_data):
         """
         final_data = {}
         for table_name, table_rows in sampled_data.items():
+
```
amontanez24, line 98: minor: remove blank line

tests/unit/sampling/test_independent_sampler.py
```diff
@@ -196,6 +196,104 @@ def test__finalize(self):
...
+                'session_id': pd.Series(['a', 'b', 'c'], dtype=object),
+                'os': pd.Series(['linux', 'mac', 'win'], dtype=object),
+                'country': pd.Series(['us', 'us', 'es'], dtype=object),
+            }),
+            'transactions': pd.DataFrame({
+                'transaction_id': pd.Series([1, 2, 3], dtype=np.int64),
+                'session_id': pd.Series(['a', 'a', 'b'], dtype=object),
+            }),
+        }
+        for result_frame, expected_frame in zip(result.values(), expected_result.values()):
```
amontanez24, line 294: can we assert the logger works as well?

R-Palazzo (approved):
LGTM!

amontanez24 (approved):
:shipit:

PR #1720, opened 2023-12-15
Fix _get_likelihoods not generating likelihood values

frances-h:
Sampling after calling `reset_sampling` was different because both `fit` and `sample` have to call `transform` in `HMA`. Resetting sampling in the data processor after fitting re-aligns the randomness for sampling.

### DIFF
--- a/sdv/multi_table/base.py
+++ b/sdv/multi_table/base.py
@@ -318,7 +318,6 @@ class BaseMultiTableSynthesizer:
             synthesizer = self._table_synthesizers[table_name]
             self._assign_table_transformers(synthesizer, table_name, table_data)
             processed_data[table_name] = synthesizer._preprocess(table_data)
-            synthesizer._data_processor.reset_sampling()
 
         return processed_data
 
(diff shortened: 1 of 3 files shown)


sdv/multi_table/hma.py
```diff
@@ -507,18 +507,24 @@ def _get_likelihoods(self, table_rows, parent_rows, table_name, foreign_key):
...
+        )
         for parent_id, row in parent_rows.iterrows():
             parameters = self._extract_parameters(row, table_name, foreign_key)
             table_meta = self._table_synthesizers[table_name].get_metadata()
             synthesizer = self._synthesizer(table_meta, **self._table_parameters[table_name])
             synthesizer._set_parameters(parameters)
             try:
-                with np.random.Generator(np.random.get_state()[1]):
-                    likelihoods[parent_id] = synthesizer._get_likelihood(table_rows)
+                likelihoods[parent_id] = synthesizer._get_likelihood(table_rows)
```
amontanez24, line 527: do we still get the randomness from the `probability_density` function?

sdv/multi_table/base.py
```diff
@@ -318,6 +318,7 @@ def preprocess(self, data):
             synthesizer = self._table_synthesizers[table_name]
             self._assign_table_transformers(synthesizer, table_name, table_data)
             processed_data[table_name] = synthesizer._preprocess(table_data)
+            synthesizer._data_processor.reset_sampling()
```
pvk-developer, line 321: If this modification is intended exclusively for `HMA`, I believe overwriting it there would be the better approach.

amontanez24 (approved):
LGTM!

PR #1727, opened 2023-12-22
Make synthesizers work with column_relationships

### LINKED ISSUE #1700: Make synthesizers work with column_relationships
### Problem Description

_As a user, I expect my synthetic data to appropriately maintain column relationships if I specify it in my metadata._

#### Synthesizer Initialization
- Raise a warning if column relationships are provided in the metadata but the user doesn't have permissions to use the advanced transformers
```python

Warning: The metadata contains a column relationship of type 'address'. This relationship will be ignored. For higher quality data in this relationship, please inquire about the SDV Enterprise tier.
```

#### Transformer assignment
- For the `address` type, we should automatically assign the [RandomLocationGenerator](https://docs.sdv.dev/rdt/transformers-glossary/address/randomlocationgenerator) to the set of columns. However, the user would be able to update this using `update_transformers` – just like they would with any other transformer.
```python
>>> synthesizer.auto_assign_transformers(data)
>>> synthesizer.get_transformers()
{
  'age': FloatFormatter(...),
  'credit_card': UniformEncoder(),
  ('vendor_city', 'vendor_state', ...): address.RandomLocationGenerator()
}

>>> from rdt.transformers.address import RegionalAnonymizer
>>> synthesizer.update_transformers({
  ('vendor_city', 'vendor_state', ...): RegionalAnonymizer()
})
```

- Deprecate the `set_address_column` method
    - No need to do any logic in this method. Just raise the warning.

```python
>>> metadata = MultiTableMetadata.load_from_json(...)
>>> synthesizer = HSASynthesizer(metadata)
>>> synthesizer.set_address_columns(...)
Warning: set_address_columns is deprecated. Please add these columns directly to your metadata using 'add_column_relationship'
```

### Additional context
- This is blocked by #1699 
- Make sure that constraints error if any column is in a column relationship

R-Palazzo:
CU-86aytvu6t
Resolve #1700

I wish you a merry Christmas! 🎄

### DIFF
--- a/sdv/data_processing/data_processor.py
+++ b/sdv/data_processing/data_processor.py
@@ -22,6 +22,7 @@ from sdv.data_processing.numerical_formatter import NumericalFormatter
 from sdv.data_processing.utils import load_module_from_path
 from sdv.errors import SynthesizerInputError, log_exc_stacktrace
 from sdv.metadata.single_table import SingleTableMetadata
+from sdv.metadata.validation import _check_import_address_transformers
 
 LOGGER = logging.getLogger(__name__)
 
@@ -75,6 +76,23 @@ class DataProcessor:
         )
         self._transformers_by_sdtype.update({'numerical': custom_float_formatter})
 
+    def _detect_multi_column_transformers(self):
+        """Detect if there are any multi column transformers in the metadata.
+
+        Returns:
+            bool:
+                ``True`` if there are multi column transformers, ``False`` otherwise.
+        """
+        result = {}
+        for relationship in self.metadata.column_relationships:
+            column_names = tuple(relationship['column_names'])
+            relationship_type = relationship['type']
+            if relationship_type == 'address':
+                transformer = rdt.transformers.address.RandomLocationGenerator
+                result[column_names] = transformer(locales=self._locales)
+
+        return result
+
     def __init__(self, metadata, enforce_rounding=True, enforce_min_max_values=True,
                  model_kwargs=None, table_name=None, locales=None):
         self.metadata = metadata
@@ -90,7 +108,7 @@ class DataProcessor:
         self._transformers_by_sdtype = deepcopy(get_default_transformers())
         self._transformers_by_sdtype['id'] = rdt.transformers.RegexGenerator()
         del self._transformers_by_sdtype['text']
-        self.grouped_columns_to_transformers = {}
+        self.grouped_columns_to_transformers = self._detect_multi_column_transformers()
 
         self._update_numerical_transformer(enforce_rounding, enforce_min_max_values)
         self._hyper_transformer = rdt.HyperTransformer()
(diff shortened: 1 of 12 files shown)


sdv-team:
Task linked: [CU-86aytvu6t SDV - Make synthesizers work with column_relationships #1700](https://app.clickup.com/t/86aytvu6t)

sdv/metadata/single_table.py
```diff
@@ -645,8 +647,20 @@ def _validate_column_relationship(self, relationship_type, column_names):
...
         try:
-            self._COLUMN_RELATIONSHIP_TYPES[relationship_type](self.columns, column_names)
+            self._COLUMN_RELATIONSHIP_TYPES[relationship_type](columns_to_sdtypes)
+        except ImportError:
+            warnings.warn(
+                f"The metadata contains a column relationship of type '{relationship_type}'. "
+                f'which requires the {relationship_type} add-on.'
+                'This relationship will be ignored. For higher quality data in this'
+                ' relationship, please inquire about the SDV Enterprise tier.')
+            del self.column_relationships[index]
```
amontanez24, line 662: Not sure if we should actually delete it or just ignore it and leave it. @npatki What do you think? If a user provides a relationship type that they do not have access too (eg. because they don't have enterprise), should we delete it from their metadata or just ignore?

sdv/single_table/base.py
```diff
@@ -134,6 +134,10 @@ def set_address_columns(self, column_names, anonymization_level='full'):
             anonymization_level (str):
                 The anonymization level to use for the address transformer.
         """
+        warnings.warn(
+            '`set_address_columns` is deprecated. Please add these columns directly to your'
+            ' metadata using `add_column_relationship`.', DeprecationWarning
+        )
```
amontanez24, line 140: I think we said we can stop doing the logic in this method and just make it a no-op with a warning

sdv/data_processing/data_processor.py
```diff
@@ -75,6 +76,23 @@ def _update_numerical_transformer(self, enforce_rounding, enforce_min_max_values
...
+            dict:
+                A dictionary mapping column names to the multi column transformer.
+        """
+        result = {}
+        for relationship in self.metadata.column_relationships:
+            column_names = tuple(relationship['column_names'])
+            relationship_type = relationship['type']
+            if relationship_type == 'address':
+                transformer = rdt.transformers.address.RandomLocationGenerator
+                result[column_names] = transformer(locales=self._locales)
```
amontanez24, line 92: should we make a dictionary mapping the relationship types to the default transformer to use?

sdv/metadata/single_table.py
npatki: @amontanez24 we should not delete it.

The metadata is considered a factual, ground source-of-truth. It should be correct regardless of which synthesizers/RDT features you have access to. Or another way to put it: When you get access to SDV Enterprise, you shouldn't have to go back and add to your metadata.

I understand that this is happening in-memory, but it is strange and unexpected behavior for us to be modifying the in-memory metadata object without the user's knowledge. This can lead to some nasty bugs down the line.

sdv/data_processing/data_processor.py
R-Palazzo: I thought of this first but I was getting import errors when defining the dict when the transformer is not available. That's why I kept it in `if`.

pvk-developer (approved):
LGTM!

amontanez24 (approved):
LGTM! Thanks for making the changes

PR #1731, opened 2024-01-05
Add primary key auto-detection for semantic sdtypes

### LINKED ISSUE #1724: Metadata auto-detection should find primary keys of semantic sdtypes
### Problem Description
As of SDV v1.8.0, the metadata auto-detection can identify a wide variety of sdtypes:
- statistical columns such as `'numerical'`', `'datetime'`, `'categorical'`, `'boolean'`
- semantic concepts such as `'email'`, `'phone_number'`, `'latitude'`, etc.
- structured identifiers, `'id'`

However, when detecting a primary key, it only considers columns that are sdtype `'id'`. In reality, *semantic* columns such as `'email'` or `'phone_number'` may also be primary keys and should be considered as possibilities.

### Expected behavior
Consider the default demo dataset. The first column, `'guest_email'` is the primary key. The metadata should continue detect is as sdtype `'email'`, and it should _also_ mark is a primary key.

```python

data, _ = download_demo(
    modality='single_table',
)

```

### Additional context
Note that if the column was named `'guest_id'` and contained random number ids, then the metadata script would correctly identify the sdtype as `'id'` and it would mark it as a primary key.

fealho:
CU-86ayxgv0h, Resolve #1724

### DIFF
--- a/sdv/metadata/single_table.py
+++ b/sdv/metadata/single_table.py
@@ -378,6 +378,7 @@ class SingleTableMetadata:
             data (pandas.DataFrame):
                 The data to be analyzed.
         """
+        first_pii_field = None
         for field in data:
             column_data = data[field]
             clean_data = column_data.dropna()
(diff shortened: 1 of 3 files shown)


sdv-team:
Task linked: [CU-86ayxgv0h SDV - Metadata auto-detection should find primary keys of semantic sdtypes #1724](https://app.clickup.com/t/86ayxgv0h)

sdv/metadata/single_table.py
```diff
@@ -410,12 +411,19 @@ def _detect_columns(self, data):
...
+
             elif sdtype == 'datetime' and dtype == 'O':
                 datetime_format = get_datetime_format(column_data.iloc[:100])
                 column_dict['datetime_format'] = datetime_format
 
             self.columns[field] = deepcopy(column_dict)
 
+        # When no primary key column was set, choose the first pii field
+        if self.primary_key is None and first_pii_field:
+            self.primary_key = first_pii_field
```
frances-h, line 425: Just to clarify, is the logic we want here to set the primary key to the first ID or PII column we find? Or is it to fall back on PII columns only if we don't find an ID column in the table?

fealho: The issue doesn't specify, so I implemented the second approach. I could change this if we have evidence that PII columns are just as likely as ID ones to be selected as primary keys.

frances-h (approved):
LGTM!

rwedge (approved):

PR #1732, opened 2024-01-05
Update `_get_is_datetime` constraint logic to use the metadata

### LINKED ISSUE #1692: `InvalidDataError` for `Inequality` constraint (even though data is valid)
### Environment Details

* SDV version: 1.7.0
* Python version: 3.10

### Error Description
There are certain datetime formats that create an error every time I try to add an `Inequality` constraint. The synthesizer keeps telling me that the constraint is not valid, even though I can verify that it is valid for every single row.

### Steps to reproduce
```python
import pandas as pd

data = pd.DataFrame(data={
    'low_col': ['21 Sep, 15', '23 Aug, 14', '29 May, 12'],
    'high_col': ['02 Nov, 15', '12 Oct, 14', '08 Jul, 12']
})

    'columns': {
        'low_col': { 'sdtype': 'datetime', 'datetime_format': '%d %b, %y' },
        'high_col': { 'sdtype': 'datetime', 'datetime_format': '%d %b, %y' }
    }
})

metadata.validate()
metadata.validate_data(data)

my_constraint = {
    "constraint_class":"Inequality",
    "constraint_parameters":{
        "low_column_name":"low_col",
        "high_column_name":"high_col",
        "strict_boundaries":False
    }
}

synth.add_constraints([my_constraint])
synth.fit(data)
```

Output:
```

Data is not valid for the 'Inequality' constraint:
      low_col    high_col
0  21 Sep, 15  02 Nov, 15
1  23 Aug, 14  12 Oct, 14
2  29 May, 12  08 Jul, 12
```

Note that everything being printed out is actually valid. (Each day in the `low_col` happens earlier than the day in the `high_col`). 

### Additional Context
This seems to happen when the datetime format string includes a `%y`, meaning the last 2 digits of a year (such as 02 for 2002 or 23 for 2023). According to [this reference](https://docs.python.org/3/library/time.html), Python's time module is supposed to interpret `%y` in the following way:
- Values 69–99 are mapped to 1969–1999
- Values 0–68 are mapped to 2000–2068

So there is nothing about the example code above that should be causing an issue. Since all year values are <69, the years should always be interpreted as part of the 2000s.

_[Edit] Upon further investigation, it appears that in the example above, the strings are being compared alpha-numerically. They do not seem to be converted to datetime._

Note that the issue goes away if you update the datetime format to `%Y`, meaning that you use all 4 digits of the year (2002 and 2023).

fealho:
CU-86aytb6w0, Resolve #1692

### DIFF
--- a/sdv/constraints/utils.py
+++ b/sdv/constraints/utils.py
@@ -7,23 +7,28 @@ import numpy as np
 import pandas as pd
 
 
-def cast_to_datetime64(value):
+def cast_to_datetime64(value, datetime_format=None):
     """Cast a given value to a ``numpy.datetime64`` format.
 
     Args:
         value (pandas.Series, np.ndarray, list, or str):
             Input data to convert to ``numpy.datetime64``.
+        datetime_format (str):
+            Datetime format of the `value`.
 
     Return:
         ``numpy.datetime64`` value or values.
     """
+    if datetime_format:
+        datetime_format = datetime_format.replace('%-', '%')
+
     if isinstance(value, str):
-        value = pd.to_datetime(value).to_datetime64()
+        value = pd.to_datetime(value, format=datetime_format).to_datetime64()
     elif isinstance(value, pd.Series):
         value = value.astype('datetime64[ns]')
     elif isinstance(value, (np.ndarray, list)):
         value = np.array([
-            pd.to_datetime(item).to_datetime64()
+            pd.to_datetime(item, format=datetime_format).to_datetime64()
             if not pd.isna(item)
             else pd.NaT.to_datetime64()
             for item in value
@@ -169,7 +174,7 @@ def revert_nans_columns(table_data, nan_column_name):
     return table_data.drop(columns=nan_column_name)
 
 
-def get_datetime_diff(high, low, dtype='O'):
+def get_datetime_diff(high, low, high_datetime_format=None, low_datetime_format=None, dtype='O'):
     """Calculate the difference between two datetime columns.
 
     When casting datetimes to float using ``astype``, NaT values are not automatically
@@ -181,14 +186,18 @@ def get_datetime_diff(high, low, dtype='O'):
             The high column values.
         low (numpy.ndarray):
             The low column values.
+        high_datetime_format (str):
+            Datetime format of the `high` column.
+        low_datetime_format (str):
+            Datetime format of the `low` column.
 
     Returns:
         numpy.ndarray:
             The difference between the high and low column values.
     """
     if dtype == 'O':
-        low = cast_to_datetime64(low)
-        high = cast_to_datetime64(high)
+        low = cast_to_datetime64(low, low_datetime_format)
+        high = cast_to_datetime64(high, high_datetime_format)
 
     diff_column = high - low
     nan_mask = pd.isna(diff_column)
--- a/sdv/constraints/tabular.py
+++ b/sdv/constraints/tabular.py
@@ -618,9 +618,8 @@ class ScalarInequality(Constraint):
         self._dtype = None
         self._operator = INEQUALITY_TO_OPERATION[relation]
 
-    def _get_is_datetime(self, table_data):
-        column = table_data[self._column_name].to_numpy()
-        is_column_datetime = is_datetime_type(column)
+    def _get_is_datetime(self):
+        is_column_datetime = self.metadata.columns[self._column_name]['sdtype'] == 'datetime'
         is_value_datetime = is_datetime_type(self._value)
         is_datetime = is_column_datetime and is_value_datetime
 
@@ -641,7 +640,7 @@ class ScalarInequality(Constraint):
                 The Table data.
         """
         self._validate_columns_exist(table_data)
-        self._is_datetime = self._get_is_datetime(table_data)
+        self._is_datetime = self._get_is_datetime()
         self._dtype = table_data[self._column_name].dtypes
 
     def is_valid(self, table_data):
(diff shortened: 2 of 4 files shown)


sdv-team:
Task linked: [CU-86aytb6w0 SDV - `InvalidDataError` for `Inequality` constraint (even though data is valid) #1692](https://app.clickup.com/t/86aytb6w0)

amontanez24 (commented):
Part of the issue is to use the provided datetime format if applicable. I think the code in #1692 would still fail right now. Either way, can we add an integration test that checks that the datetime format is indeed being used to convert weird datetime columns during the constraints?

fealho:
> Part of the issue is to use the provided datetime format if applicable. I think the code in #1692 would still fail right now. Either way, can we add an integration test that checks that the datetime format is indeed being used to convert weird datetime columns during the constraints?

@amontanez24 I already tested that specific integration test in the [other PR](https://github.com/sdv-dev/SDV/pull/1721/files#diff-0aec02673a42afd4a6c03e9c5032c6bdf922f7a0da87f5bd6009ee9be52f886eR911), but I can add an additional test for the datetime format as well

amontanez24:
> > Part of the issue is to use the provided datetime format if applicable. I think the code in #1692 would still fail right now. Either way, can we add an integration test that checks that the datetime format is indeed being used to convert weird datetime columns during the constraints?
> 
> @amontanez24 I already tested that specific integration test in the [other PR](https://github.com/sdv-dev/SDV/pull/1721/files#diff-0aec02673a42afd4a6c03e9c5032c6bdf922f7a0da87f5bd6009ee9be52f886eR911), but I can add an additional test for the datetime format as well

I thought part of the problem was that we weren't converting the datetime properly because we didn't use the format but I guess I remembered it wrong. Idk if that's still a potential problem that could happen with oddly formed datetime strings though

fealho:
> > > Part of the issue is to use the provided datetime format if applicable. I think the code in #1692 would still fail right now. Either way, can we add an integration test that checks that the datetime format is indeed being used to convert weird datetime columns during the constraints?
> > 
> > 
> > @amontanez24 I already tested that specific integration test in the [other PR](https://github.com/sdv-dev/SDV/pull/1721/files#diff-0aec02673a42afd4a6c03e9c5032c6bdf922f7a0da87f5bd6009ee9be52f886eR911), but I can add an additional test for the datetime format as well
> 
> I thought part of the problem was that we weren't converting the datetime properly because we didn't use the format but I guess I remembered it wrong. Idk if that's still a potential problem that could happen with oddly formed datetime strings though

@amontanez24 The issue 1692 is not really related to the datetime_format, it was just misdiagnosed as such in the beginning, where the true issue was misidentifying the datetime column sdtype.

To assuage your concerns about the datatime_format, we already verify it is valid. `fit` calls `validate`, which in turn calls `validate_datetime_format`:

```python3
    pandas_datetime_format = datetime_format.replace('%-', '%')
    datetime_column = pd.to_datetime(
        column,
        errors='coerce',
        format=pandas_datetime_format
    )
    valid = pd.isna(column) | ~pd.isna(datetime_column)

    return set(column[~valid])
```

If `datetime_format` doesn't match the column, an error will be raised during the validation step.

pvk-developer (approved):
LGTM!

amontanez24:
> > > > Part of the issue is to use the provided datetime format if applicable. I think the code in #1692 would still fail right now. Either way, can we add an integration test that checks that the datetime format is indeed being used to convert weird datetime columns during the constraints?
> > > 
> > > 
> > > @amontanez24 I already tested that specific integration test in the [other PR](https://github.com/sdv-dev/SDV/pull/1721/files#diff-0aec02673a42afd4a6c03e9c5032c6bdf922f7a0da87f5bd6009ee9be52f886eR911), but I can add an additional test for the datetime format as well
> > 
> > 
> > I thought part of the problem was that we weren't converting the datetime properly because we didn't use the format but I guess I remembered it wrong. Idk if that's still a potential problem that could happen with oddly formed datetime strings though
> 
> @amontanez24 The issue 1692 is not really related to the datetime_format, it was just misdiagnosed as such in the beginning, where the true issue was misidentifying the datetime column sdtype.
> 
> To assuage your concerns about the datatime_format, we already verify it is valid. `fit` calls `validate`, which in turn calls `validate_datetime_format`:
> 
> ```python
>     pandas_datetime_format = datetime_format.replace('%-', '%')
>     datetime_column = pd.to_datetime(
>         column,
>         errors='coerce',
>         format=pandas_datetime_format
>     )
>     valid = pd.isna(column) | ~pd.isna(datetime_column)
> 
>     return set(column[~valid])
> ```
> 
> If `datetime_format` doesn't match the column, an error will be raised during the validation step.

I see. No, I'm not concerned about the format being valid. I'm saying that sometimes when converting a string column to datetime with the pd.to_datetime() function, it fails if you don't provide the format. So if we make that call at all in the constraints, I think we should pass the format there

sdv/constraints/utils.py
```diff
@@ -7,23 +7,28 @@
...
         value (pandas.Series, np.ndarray, list, or str):
             Input data to convert to ``numpy.datetime64``.
+        datetime_format (str):
+            Datetime format of the `value`.
 
     Return:
         ``numpy.datetime64`` value or values.
     """
+    if datetime_format:
+        datetime_format = datetime_format.replace('%-', '%')
```
amontanez24, line 23: what is this part for?

amontanez24 (approved):
I think this looks good!

PR #1745, opened 2024-01-18
Tokenize words before determining PII for metadata auto-detection

### LINKED ISSUE #1725: Metadata auto-detection should tokenize words before determining PII
### Problem Description
As of SDV 1.8.0, the metadata can auto-detect semantic (PII) sdtypes such as `vin` (vehicle identifier number) or `administrative_unit` (a state/province). It does this by checking to see if the column name contains important substrings such as `'vin'` or `'state'`.

Unfortunately this can lead to unexpected results:
- A column named `resolving_loans` would be identified as sdtype `'vin'` be cause the word _resolving_ contains the substring `'vin'`
- A column named `RealEstateLoans` would be identified as sdtype `'administrative_unit'` because the word _Estate_ contains the substring `'state'` (which is a type of administrative region). 

### Expected behavior
Instead of checking to see if _any_ substring matches the keywords, the metadata auto-detection script should tokenize the column names first. Then, it should check for exact matches within the tokenized words.

We can tokenize names that contain underscores or camel-case letters. Consider the above examples:
- Column `resolving_loans` would be tokenized into `['resolving', 'loans']`. None of these words exactly match the keyword `vin` so the sdtype cannot be `vin`.
- Column `RealEstateLoans` would be tokenized into `['real', 'estate', 'loans']`. None of these words exactly match the keyword `state` so the sdtype cannot be `state`.

On the other hand:
- Column `vin_number` would be tokenized into `['vin', 'number']`, which exactly matches `'vin'`
- Column `StateDepartment` would be tokenized into `['state', 'department']`, which exactly matches `'state'`

### Additional context
We should be careful with camel-case. 
- All-caps words should not be tokenized (`EXAMPLE`)
- All-caps words with undersscores should be tokenized (`EXAMPLE_COLUMN` --> `['example', 'column']`)

fealho:
CU-86ayxgz4t, Resolve #1725.

### DIFF
--- a/sdv/metadata/single_table.py
+++ b/sdv/metadata/single_table.py
@@ -95,8 +95,14 @@ class SingleTableMetadata:
         'license': 'license_plate',
     }
 
-    # Subset of sdtypes from _REFERENCE_TO_SDTYPE which could be substring of another word
-    _SUBSTRING_SDTYPES = frozenset(['ssn', 'administrative_unit', 'city', 'vin'])
+    _SDTYPES_WITHOUT_SUBSTRINGS = {
+        reference: sdtype
+        for reference, sdtype in _REFERENCE_TO_SDTYPE.items()
+        if sdtype not in {'ssn', 'administrative_unit', 'city', 'vin'}
+    }
+
+    _SDTYPES_WITH_SUBSTRINGS = dict(
+        set(_REFERENCE_TO_SDTYPE.items()) - set(_SDTYPES_WITHOUT_SUBSTRINGS.items()))
 
     _COLUMN_RELATIONSHIP_TYPES = {
         'address': validate_address_sdtypes,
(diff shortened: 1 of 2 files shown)


sdv-team:
Task linked: [CU-86ayxgz4t SDV - Metadata auto-detection should tokenize words before determining PII #1725](https://app.clickup.com/t/86ayxgz4t)

sdv/metadata/single_table.py
```diff
@@ -305,10 +308,34 @@ def _detect_pii_column(self, column_name):
...
                 The column name to be analyzed.
         """
-        cleaned_name = re.sub(r'[^a-zA-Z0-9]', '', column_name).lower()
+        # Subset of sdtypes which are unambiguous, ie aren't a substring of another word
+        # For such cases just check if the word is in the column name
+        sdtypes_subset = {
+            reference: sdtype
+            for reference, sdtype in self._REFERENCE_TO_SDTYPE.items()
+            if sdtype not in self._SUBSTRING_SDTYPES
+        }
```
R-Palazzo, line 317: This can go outside the method

sdv/metadata/single_table.py
```diff
@@ -305,10 +308,34 @@ def _detect_pii_column(self, column_name):
...
+        # To handle the cases where the sdtype could be a substring of another word,
+        # tokenize the column name based on (1) symbols and (2) camelCase
+        tokens = column_name.replace(' ', '_').replace('-', '_').split('_')
+        if len(tokens) == 1:
+            tokens = []
+            if column_name.upper() != column_name and column_name[1:].lower() != column_name[1:]:
+                tokens = re.findall('[A-Z][^A-Z]*', column_name)
+
+        tokens = tokens if tokens else [column_name]
+        tokens = [token.lower() for token in tokens]
```
R-Palazzo, line 332: Can the tokenization logic be wrapped in a separate method?

sdv/metadata/single_table.py
```diff
@@ -305,10 +308,34 @@ def _detect_pii_column(self, column_name):
...
+
+        substr_sdtypes = dict(set(self._REFERENCE_TO_SDTYPE.items()) - set(sdtypes_subset.items()))
+
```
R-Palazzo, line 335: This can also be computed once in the class, not at every call

rwedge (approved):

PR #1764, opened 2024-01-30
Clip HMASynthesizer out-of-bounds parameters

### LINKED ISSUE #1691: HMASynthesizer sometimes creates null values (out-of-bounds parameters synthesized)
### Environment Details

* SDV version: Any

### Error Description
The SDV is only supposed to synthesize null values if the real data also has null values. However, in some cases, the HMA Synthesizer creates erroneous null values (in columns that are not supposed to have these). These incorrect nulls only appear in child tables (i.e. tables with a parent). The root tables are unaffected/do not contain any of these nulls.

### Root Cause
The HMA algorithm works by summarizing the distribution of children -- for eg, using a Beta distribution, it summarizes the childen using the parameters `alpha`, `beta`, `loc` and `scale`. It then models the parameters and creates new ones from scratch during sampling.

Unfortunately, the new parameters that are sampled are not guaranteed to be in-bounds. So there is a chance that the sampled `alpha` or `beta` parameter will be `<0`. This is invalid for a Beta distribution, which is only defined when alpha and beta are >0.

### Fixes
HMA should apply a FloatFormatter to each of the extended columns (for marginal distributions as well as the covariance columns). The FloatFormatter should be set up to clip the synthesized min/max values.

```python
FloatFormatter(enforce_min_max_values=True)
```

Note that these transformers should be accessible after fitting in an easy-to-understand way. For HSA, we are using the parameter `extended_columns`. We should do the same here.

```python
>>> synthesizer.extended_columns['my_table_name']
{
   '<extended_column_name>': FloatFormatter(enforce_min_max_values=True),
   '<extended_column_name>: FloatFormatter(enforce_min_max_values=True),
  ...
}
```

A more robust option for HMA would be to apply some kind of transformer to the extended columns (for each parameter: alpha, beta, etc.). This transformer could be responsible for clipping the min/max values in case they are synthesized to be out-of-bounds.

_There is an issue for this in Copulas (see [issue 367](https://github.com/sdv-dev/Copulas/issues/367)). However, Copulas is not really expected to work with invalid parameter values._

fealho:
CU-86az6fmfu, Resolve #1691.

### DIFF
--- a/sdv/multi_table/hma.py
+++ b/sdv/multi_table/hma.py
@@ -452,7 +452,7 @@ class HMASynthesizer(BaseHierarchicalSampler, BaseMultiTableSynthesizer):
         prefix = f'__{table_name}__{foreign_key}__'
         keys = [key for key in parent_row.keys() if key.startswith(prefix)]
         new_keys = {key: key[len(prefix):] for key in keys}
-        flat_parameters = parent_row[keys].fillna(1e-6)
+        flat_parameters = parent_row[keys].astype(float).fillna(1e-6)
 
         num_rows_key = f'{prefix}num_rows'
         if num_rows_key in flat_parameters:
--- a/sdv/multi_table/hma.py
+++ b/sdv/multi_table/hma.py
@@ -465,7 +465,7 @@ class HMASynthesizer(BaseHierarchicalSampler, BaseMultiTableSynthesizer):
         flat_parameters = flat_parameters.to_dict()
         for parameter_name, parameter in flat_parameters.items():
             float_formatter = self.extended_columns[table_name][parameter_name]
-            flat_parameters[parameter_name] = np.clip(
+            flat_parameters[parameter_name] = np.clip(  # this should be revisited in GH#1769
                 parameter, float_formatter._min_value, float_formatter._max_value)
 
         return {new_keys[key]: value for key, value in flat_parameters.items()}
--- a/sdv/multi_table/hma.py
+++ b/sdv/multi_table/hma.py
@@ -2,7 +2,6 @@
 
 import logging
 from copy import deepcopy
-import math
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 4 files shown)


sdv-team:
Task linked: [CU-86az6fmfu SDV - HMASynthesizer sometimes creates null values (out-of-bounds parameters synthesized) #1691](https://app.clickup.com/t/86az6fmfu)

sdv/multi_table/hma.py
```diff
@@ -349,6 +350,11 @@ def _augment_table(self, table, tables, table_name):
                     foreign_key,
                     progress_bar_desc
                 )
+                for column in extension.columns:
+                    self.extended_columns[child_name][column] = FloatFormatter(
+                        enforce_min_max_values=True)
+                    self.extended_columns[child_name][column].fit(extension, column)
```
frances-h, line 356: I think nan values might be an issue here - if we can't get parameters for a child they end up being nan values here, and then we clear the whole augmented table of nans in line 366. If we move fitting the float formatters to after that we should be good.

frances-h (commented):
Could you also add integration tests? If we have examples that consistently failed previously we could probably just use those

sdv/multi_table/hma.py
fealho: @frances-h I added a line to handle the case where all values are `nan`. The FloatFormatter is only used to store the min/max values, so it should work as long as there's one non-nan number.

fealho:
> Could you also add integration tests? If we have examples that consistently failed previously we could probably just use those

@frances-h It is hard to come up with an integration test that fails. The examples we have take minutes to run, and I don't want to slow down the whole workflow for them. I added an integration/unit test which validates the code works when parameters are out of bounds.

sdv/multi_table/hma.py
frances-h: Right but if there's NaNs in the input, we'll end up adding NaNs back on the reverse as well, no?

frances-h (approved):
Thanks for addressing!

rwedge (approved):

PR #1765, opened 2024-01-31
Improve quality of `sequence_index`

### LINKED ISSUE #1760: Improve quality of `sequence_index`: Move the start dates into the context model
### Problem Description
In multi-sequence data, a sequence index is used to denote both the order of sequences as well as the intervals between them. Usually, this column is `datetime` (but it may also be `numerical` if the user is storing absolute time values). In [this example dataset](https://docs.sdv.dev/sdv/sequential-data/data-preparation), the `sequence_index` is the `Date` column.

The `sequence_index` is generally supposed to adhere to some special rules. However, when using the PARSynthesizer as-is, users have reported that various properties are not being learned and the rules are being broken.

1. The `sequence_index` is supposed to be unique, in order to uniquely determine the sequence order. Users are finding duplicates (see #1723),
2. The `sequence_index` value should be increasing, as most sequential datasets are presented in order. Users are finding that the value sometimes decreases (see #466)
3. The above two rules also imply that for very long sequences, `sequence_index` should continue on into the future. Users are finding that the current index values are very limited in range (see #1752)

### Root Cause
In an investigation with @amontanez24 and @frances-h, we found a potential root cause:

**During Fit**: This function is computing the diffs between rows (as expected) but it is also storing the start sequence index as a single, static value within the sequence. Static values cannot be effectively learned by PAR.
https://github.com/sdv-dev/SDV/blob/e6e508ba197b104dea3fca96fea34d9bbf9a8cd7/sdv/sequential/par.py#L205-L212

**During Sampling**: The diffs between each rows are being added up (as expected) but they are all using a different starting sequence index due to the issue above.
https://github.com/sdv-dev/SDV/blob/e6e508ba197b104dea3fca96fea34d9bbf9a8cd7/sdv/sequential/par.py#L299-L303

### Solution
**During fit**: 
- We should continue computing the differences between rows. But the static value (sequence index) should _not_ be added to the sequence for modeling using PAR.
- Instead, the start value for each sequence should be added to the _context model_. This means that we only model 1 single starting value for each sequence.

**During sample**:
- The context model will synthesize a single starting value per sequence
- Now for each sequence, we can use that constant starting value and sum up the differences

### Additional Notes
- Should the difference column be put through a FloatFormatter with min/max clipping? Ideally, we create differences that are too high or too low
- Note that the `sequence_index` is an optional concept so we should very that PAR can continue to work without it

frances-h:
CU-86az5xcqz
Resolve #1760

The sequence index is now split into a diff column which goes through the sequential model and a context column which is added to the context model. Additionally, `FloatFormatter` is used on the diff column to force sampled columns to be within the given min/max range.

Also, `sample_sequential_columns` had to be adjusted to use conditional sampling to get missing extra context columns before getting sequential samples.

### DIFF
--- a/sdv/sequential/par.py
+++ b/sdv/sequential/par.py
@@ -4,7 +4,6 @@ import inspect
 import logging
 import uuid
 import warnings
-from datetime import datetime
 
 import numpy as np
 import pandas as pd
@@ -15,10 +14,10 @@ from rdt.transformers import FloatFormatter
 
 from sdv.errors import SamplingError, SynthesizerInputError
 from sdv.metadata.single_table import SingleTableMetadata
+from sdv.sampling import Condition
 from sdv.single_table import GaussianCopulaSynthesizer
 from sdv.single_table.base import BaseSynthesizer
 from sdv.single_table.ctgan import LossValuesMixin
-from sdv.sampling import Condition
 from sdv.utils import cast_to_iterable, groupby_list
 
 LOGGER = logging.getLogger(__name__)
@@ -78,7 +77,7 @@ class PARSynthesizer(LossValuesMixin, BaseSynthesizer):
 
         for column in context_columns:
             context_columns_dict[column] = self.metadata.columns[column]
-        
+
         for column, column_metadata in self._extra_context_columns.items():
             context_columns_dict[column] = column_metadata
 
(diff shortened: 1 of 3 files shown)


sdv-team:
Task linked: [CU-86az5xcqz SDV - Improve quality of `sequence_index`: Move the start dates into the context model #1760](https://app.clickup.com/t/86az5xcqz)

sdv/sequential/par.py
```diff
@@ -151,6 +158,41 @@ def _validate_context_columns(self, data):
...
+            fill_value = 0
+        else:
+            fill_value = min(sequence_index_sequence[self._sequence_index].dropna())
+        sequence_index_sequence = sequence_index_sequence.fillna(fill_value)
+
+        data[self._sequence_index] = sequence_index_sequence[self._sequence_index]
+        data = data.merge(
+            sequence_index_context,
+            left_on=self._sequence_key,
+            right_index=True)
```
amontanez24, line 184: It's funny that we merge this back in but just separate it out again later. Not sure if there's a way to avoid that

amontanez24 (commented):
This looks good! Maybe we should add an integration test to show that the issues raised that led to this one were resolved

sdv/sequential/par.py
frances-h: Maybe, but not easily right now. PAR also takes in the context columns when assembling the sequences and fitting the model so we'd need to do the join regardless. We could investigate if the context is really needed there though, and then skip the join here if so.

amontanez24 (approved):
LGTM!

rwedge (approved):

PR #1774, opened 2024-02-05
Add a `version` module to align with SDV Enterprise

### LINKED ISSUE #1761: Add a `version` module to align with SDV Enterprise
### Problem Description
To align the usages of SDV with SDV Enterprise, we should add a `version` module that the user can call to quickly see what they have installed. 

### Expected behavior
The expected behavior is similar to SDV Enterprise: the `version` module should expose an `enterprise` variable and a `public` variable. The only difference is that the `enterprise` variable is set to `None`.  The `public` variable should return the same value as calling `sdv.__version__` (current version).

```python
>>> import sdv
>>> print(sdv.version.enterprise)
None
>>> print(sdv.version.public)
'1.9.0'
>>> print(sdv.__version__)
'1.9.0'
```

### Additional context
This will make it easy for users and customer to troubleshoot which versions they currently have installed.

R-Palazzo:
CU-86az5xxqe
Resolve #1761

### DIFF
--- a/sdv/__init__.py
+++ b/sdv/__init__.py
@@ -17,7 +17,7 @@ from pkg_resources import iter_entry_points
 
 from sdv import (
     constraints, data_processing, datasets, evaluation, lite, metadata, metrics, multi_table,
-    sampling, sequential, single_table)
+    sampling, sequential, single_table, version)
 
 __all__ = [
     'constraints',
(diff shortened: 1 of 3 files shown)


sdv-team:
Task linked: [CU-86az5xxqe SDV - Add a `version` module to align with SDV Enterprise #1761](https://app.clickup.com/t/86az5xxqe)

frances-h (approved):

rwedge (approved):

PR #1786, opened 2024-02-09
Warn users to save their metadata file after auto-detecting/updating it

### LINKED ISSUE #1762: Warn users to save their metadata file after auto-detecting/updating it
### Problem Description
As specified in the [Metadata docs](https://docs.sdv.dev/sdv/multi-table-data/data-preparation/multi-table-metadata-api#creation-api) the metadata auto-detection logic is not meant to be accurate or complete. Furthermore, the metadata auto-detection logic may change in between SDV versions, leading to inconsistent results.

For eg. the following script may not produce the same results in every SDV version because the auto-detection script changes! 

```python

metadata.detect_from_dataframes(my_data) # this logic is not guaranteed to be accurate and may change!!

synthesizer.fit(data)
```

To avoid these issues, the SDV team **strongly recommends** saving the metadata as a separate JSON file. This is not communicated to the user strongly enough, and leads to confusion.

```python
metadata.save_to_json('my_metadata.json')
```

### Expected behavior
When initializing a synthesizer, we should warn the user if they are providing a metadata object that has been auto-detected/modified but has never been saved.

```python
 # any single, multi or sequential synthesizer
```
```
Warning: We strongly recommend saving the metadata using 'save_to_json' for replicability in future SDV versions.
```

### Additional context
The warning does _not_ need to show up if:
- The user has called the `save_to_json()` function on the metadata (meaning that they are following the recommendation) OR
- The user has created the metadata using `load_from_json()` (meaning that they are loading a previously-saved version of it)
- The user retrieved the metadata object from our `download_demo()` function

For all of the above: The warning should reappear if a user update the metadata afterwards using the Python API, or if they call auto-detect on it.

One way to accomplish this would be by setting/unsetting a private flag within the metadata object itself.

R-Palazzo:
CU-86az6hhgm
Resolve #1762


sdv-team:
Task linked: [CU-86az6hhgm SDV - Warn users to save their metadata file after auto-detecting/updating it #1762](https://app.clickup.com/t/86az6hhgm)

sdv/metadata/single_table.py
```diff
@@ -1069,6 +1082,7 @@ def load_from_json(cls, filepath):
                 'class and version.'
             )
 
+        cls._updated = False
```
amontanez24, line 1085: why not make the default false?

sdv/multi_table/base.py
```diff
@@ -69,12 +69,22 @@ def _print(self, text='', **kwargs):
         if self.verbose:
             print(text, **kwargs)  # noqa: T001
 
+    def _check_metadata_updated(self):
+        self.metadata._check_updated_flag()
+        if self.metadata._updated:
```
amontanez24, line 74: Not sure if you need to store the `_updated` flag if you can just directly call `_check_metadata_updated`

amontanez24 (commented):
Looks good, but I have a couple questions to maybe reduce some lines of code

sdv/metadata/single_table.py
R-Palazzo: Yes, done in [18ee971](https://github.com/sdv-dev/SDV/pull/1786/commits/18ee9712445700f04cc8bb51490ebf48a1304237)

sdv/multi_table/base.py
R-Palazzo: Yes, I did it to be consistent between single and multi-table metadata. I removed it in [d20f32a](https://github.com/sdv-dev/SDV/pull/1786/commits/d20f32acf9b84ca2c86522cd6217e0e52c711a9a)

sdv/metadata/multi_table.py
```diff
@@ -482,6 +501,7 @@ def detect_table_from_dataframe(self, table_name, data):
         self._validate_table_not_detected(table_name)
         table = SingleTableMetadata()
         table._detect_columns(data)
+        table._updated = True
```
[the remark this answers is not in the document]
R-Palazzo: Before we seem to be using `table.detect_from_dataframe` from this: https://github.com/sdv-dev/SDV/commit/d1ae6de0d166ce2c62bd41e5c82a0e705a9ba0d1
Maybe @amontanez24 remembers a reason for it.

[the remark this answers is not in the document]
R-Palazzo: Yes, done in [6b0c5cf](https://github.com/sdv-dev/SDV/pull/1786/commits/6b0c5cf828469e255dbf55c99ca7031933ac2bbe)

frances-h (approved):
thanks for addressing!

amontanez24 (approved):
LGTM!

PR #1791, opened 2024-02-14
Provide user-friendly error messages when there are missing values in conditional sampling

### LINKED ISSUE #1758: Validate condition and provide user-friendly messages for NaN/missing values (currently unsupported)
### Problem Description
Currently, our synthesizers do not support conditioning on missing values (see #695). That is it say:
- The [sample_from_conditions](https://docs.sdv.dev/sdv/single-table-data/sampling/conditional-sampling#sample_from_conditions) method does not support missing values in the Conditions object
- The [sample_remaining_columns](https://docs.sdv.dev/sdv/single-table-data/sampling/conditional-sampling#sample_remaining_columns) method does not support any missing values in the `known_columns` DataFrame 

But if I pass in missing values anyways, I get a confusing errors and outputs:

**sample_from_conditions**: I get a `ValueError`
```
ValueError: Error: Sampling terminated. Partial results are stored in a temporary file: .sample.csv.temp. This file will be overridden the next time you sample. Please rename the file if you wish to save these results.
No objects to concatenate
```

**sample_remaining_columns**: It ignores any rows with missing values and creates the rest. If there are no rows that satisfy the requirements, then I get the same error.
```
ValueError: Error: Sampling terminated. Partial results are stored in a temporary file: .sample.csv.temp. This file will be overridden the next time you sample. Please rename the file if you wish to save these results.
No objects to concatenate
```

### Expected behavior
Until we support conditioning on missing values (in #695), we should verify whether the input has any missing values (np.nan, None, empty string, NaT, etc.)

If so, we should provide the user a clearer message:

**sample_from_conditions**
```
SynthesizerInputError: Missing values are not yet supported for conditional sampling. Please include only non-null values in your Condition objects.
```

**sample_remaining_columns**: This case breaks down into two:
- If there is at least 1 row with non-missing data throughout, then we have something to sample. In this case, only show a warning.
```
Warning: Missing values are not yet supported. Rows with any missing values will not be created.
```

- If all of the rows contain at least 1 missing values, then there is nothing to sample. In this case, we can throw an error.
```
SynthesizerInputError: Missing values are not yet supported for conditional sampling. Please include only non-null values in your known columns DataFrame.
```

### Example Code
```python
import numpy as np
import pandas as pd

from sdv.sampling import Condition

    modality='single_table',
    dataset_name='fake_hotel_guests')

synthesizer.fit(data)

# single condition with missing values
my_condition = Condition(
    num_rows=250,
    column_values={'room_type': None, 'has_rewards': False})

synthetic_data = synthesizer.sample_from_conditions(conditions=[my_condition])

# known columns with missing values
# only the final row has all non-null values
known_columns = pd.DataFrame(data={
    'has_rewards': [None, False, True],
    'amenities_fee': [5.00, np.nan, 100.00]})

synthesizer.sample_remaining_columns(known_columns=known_columns)
```

fealho:
CU-86az5d52d, Resolve #1758.

### DIFF
--- a/sdv/single_table/base.py
+++ b/sdv/single_table/base.py
@@ -992,7 +992,7 @@ class BaseSingleTableSynthesizer(BaseSynthesizer):
         self._validate_conditions_unseen_columns(conditions)
         if conditions.dropna().empty:
             self._raise_condition_with_nans()
-        else:
+        elif conditions.isna().any().any():
             warnings.warn(
                 'Missing values are not yet supported. '
                 'Rows with any missing values will not be created.'
--- a/sdv/single_table/base.py
+++ b/sdv/single_table/base.py
@@ -794,13 +794,6 @@ class BaseSingleTableSynthesizer(BaseSynthesizer):
             show_progress_bar=show_progress_bar
         )
 
-    def _validate_conditions(self, conditions):
-        """Validate the user-passed conditions."""
-        for column in conditions.columns:
-            if column not in self._data_processor.get_sdtypes():
-                raise ValueError(f"Unexpected column name '{column}'. "
-                                 f'Use a column name that was present in the original data.')
-
     def _sample_with_conditions(self, conditions, max_tries_per_batch, batch_size,
                                 progress_bar=None, output_file_path=None):
         """Sample rows with conditions.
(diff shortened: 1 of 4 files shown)


sdv-team:
Task linked: [CU-86az5d52d SDV - Validate condition and provide user-friendly messages for NaN/missing values (currently unsupported) #1758](https://app.clickup.com/t/86az5d52d)

sdv/single_table/base.py
```diff
@@ -974,6 +987,17 @@ def sample_from_conditions(self, conditions, max_tries_per_batch=100,
...
+    def _validate_known_columns(self, conditions):
+        """Validate the user-passed conditions."""
+        self._validate_conditions_unseen_columns(conditions)
+        if conditions.dropna().empty:
+            self._raise_condition_with_nans()
+        else:
+            warnings.warn(
+                'Missing values are not yet supported. '
+                'Rows with any missing values will not be created.'
+            )
```
rwedge, line 999: I think we should also check here to see if there are no missing values and only warn if there are some missing values.

fealho: Good catch, thank you!

rwedge (approved):
LGTM!

pvk-developer (approved):
LGTM!

PR #1794, opened 2024-02-14
Bump CTGAN version

frances-h:


setup.py
```diff
@@ -24,7 +24,7 @@
     "pandas>=1.5.0;python_version>='3.11'",
     'tqdm>=4.15,<5',
     'copulas>=0.9.0,<0.10',
-    'ctgan>=0.8,<0.9',
+    'ctgan>=0.8,<0.10',
```
amontanez24, line 27: I think we should just bump the minimum as well

Issue #1792, opened 2024-02-14
Add a utility to drop unknown references (and enforce referential integrity)


npatki:
### Problem Description
For multi-table datasets, SDV currently expects that all the foreign key values must be present in the primary key (aka referential integrity). For various reasons* I may currently be in possession of a datasets that _does not_ have referential integrity. This prevents SDV from being able to model my data and instead gives me an error.

*Reasons may include some messiness in the data source, or having random (incomplete) data from various data sources

### Expected behavior
**Add a utility function called `utils.drop_unknown_references`**:

Parameters:
- (required) `metadata`: A MultiTableMetadata object
- (required) `data`: A dictionary that maps each table name (string) to the data for that table (pandas.DataFrame)
- `drop_missing_values`: A boolean describing whether or not to also drop foreign keys with missing values 
   - (default) `True`: Drop a row if a foreign key has missing values
   - `False`: Allow rows to contain missing values as foreign keys

Output: A dictionary that maps each of the original table names (string) to cleaned data for that table (pandas.DataFrame). The cleaned data should have referential integrity.

```python
from sdv.utils import drop_unknown_references
from sdv.multi_table import HMASynthesizer

cleaned_data = drop_unknown_references(metadata=my_metadata, data=original_data)

synth = HMASynthesizer(metadata)
synth.fit(cleaned_data) # now synthesizers should accept the cleaned data
...
```

Note that if a table has multiple foreign keys, then the script should only keep rows where _all_ foreign keys have references. If any one foreign key has an unknown reference, the entire row should be dropped.

**Update the error message**: If the passed-in data does not have referential integrity, update the error message to point the user towards the `drop_unknown_references` method. _This error check happens in `metadata.validate_data`_.

```python
synth.fit(original_data)
```

```
InvalidDataError: The provided data does not match the metadata:
Relationships:
Error: foreign key column 'parent_id' contains unknown references: <values>. Please use the utility method 'drop_unknown_references' to clean the data.
```

### Additional context

- Do some error validation on the passed in data. It must contain the same table names and the same primary/foreign key column names as described in the metadata.
- After dropping rows with unknown references, there should be at least 1 row remaining in every table. If a table has 0 rows remaining, then throw an error.

```
InvalidDataError: All references in table 'transactions' are unknown and must be dropped. Try providing different data for this table.
```

PR #1798, opened 2024-02-19
Cleanup `utils` module: Make internal functions private

### LINKED ISSUE #1793: Cleanup `utils` module: Make internal functions private
Currently the SDV library has a `utils` module that contains many helper functions. Initially, all the functions in this file were intended to be internal only. However, we may soon add functions that are meant for public use (see #1792). 

To better distinguish internal vs. externally-available functions, we should **prepend an underscore to all the internal helper functions** that are not meant for end-users.

R-Palazzo:
CU-86azc5bc2
Resolve #1793

@npatki is `load_data_from_csv` meant for end-users?


sdv-team:
Task linked: [CU-86azc5bc2 SDV - Add a utility to drop unknown references (and enforce referential integrity) #1792](https://app.clickup.com/t/86azc5bc2)

amontanez24 (commented):
This looks good! Alternatively, should we just move all these private functions to a `_utils.py` to not cause confusion between the functions there and the ones meant as public utility functions? @npatki

npatki:
> This looks good! Alternatively, should we just move all these private functions to a `_utils.py` to not cause confusion between the functions there and the ones meant as public utility functions? @npatki

Yeah this makes sense to me!

npatki (commented):
Hi @R-Palazzo, without going through every line of code here -- my recommendation is to check the API docs and see what is public today (you can search "utils" in the search bar)

You asked before if `load_data_from_csvs` is public -- I don't see it anywhere in the docs. I do see a `load_csvs` though.

pvk-developer (approved):
LGTM; 

We do not have any function on `SDV-Enterprise` that could be affected by this right ?

R-Palazzo:
@pvk-developer, that's a good point, thanks.
The chained inequality feature uses some of the `utils` methods, but it will be a quick fix after the next sdv release.

amontanez24 (approved):
since the whole module is private now we could make the functions public again lol. But I leave that up to you

PR #1841, opened 2024-03-06
Use `AnonymizedFaker` for primary keys

### LINKED ISSUE #1807: Contextual Anonymization transformers shouldn't be used for primary keys
### Environment Details

* SDV version: 1.10.0 (+ SDV Enterprise 0.10.0)
* Python version: Any
* Operating System: Any

### Error Description
My metadata might list a semantic, PII type as a primary key -- for example an `email`.

```json
{
  "primary_key": "user_email",
  "columns": {
    "user_email": { "sdtype": "email" },
    ...
  }
}
```

If I am an SDV Enterprise user _and_ my primary keys are of premium sdtypes (`email`, `phone_number`, ...), then SDV will assign the primary key column to a contextual anonymization transformer such as [DomainBasedAnonymizer](https://docs.sdv.dev/rdt/transformers-glossary/deep-data-understanding/email/domainbasedanonymizer) or [AnonymizedGeoExtractor](https://docs.sdv.dev/rdt/transformers-glossary/deep-data-understanding/phone-number/anonymizedgeoextractor). **Such transformers are not designed to keep uniqueness in mind**, meaning that it may be possible that some primary keys repeat.

### Solution
As a quick solution, we should ensure that we do not assign Contextual Anonymization transformers to primary keys. Instead, we should fall back on AnonymizedFaker like we do for public SDV.

(In the future, we can think about how Contextual Anonymization transformers can support uniqueness. We'd likely need to think through a different algorithm for each one.)

fealho:
CU-86azdzv1n, Resolve #1807.

### DIFF
--- a/tests/unit/data_processing/test_data_processor.py
+++ b/tests/unit/data_processing/test_data_processor.py
@@ -1496,6 +1496,40 @@ class TestDataProcessor:
         }
         assert config == expected_config
 
+    def test__create_config_with_contextual_column_as_primary_key(self):
+        """Test it when a conextual column is a primary key."""
+        # Setup
+        class CustomTransformer:
+            pass
+
+        data = pd.DataFrame({
+            'phone_number': ['+1 (234) 535-2341', '+1 (334) 535-2341'],
+            'email': ['test@gmail.com', 'test2@gmail.com.br'],
+        })
+        metadata = SingleTableMetadata().load_from_dict({
+            'columns': {
+                'email': {'sdtype': 'email'},
+                'phone_number': {'sdtype': 'phone_number'},
+            },
+            'primary_key': 'phone_number',
+        })
+        dp = DataProcessor(metadata)
+        dp._transformers_by_sdtype = {
+            'email': CustomTransformer(),
+            'phone_number': CustomTransformer(),
+        }
+
+        # Run
+        config = dp._create_config(data, set())
+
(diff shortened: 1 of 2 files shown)


sdv-team:
Task linked: [CU-86azdzv1n SDV - Contextual Anonymization transformers shouldn't be used for primary keys #1807](https://app.clickup.com/t/86azdzv1n)

fealho:
Results of the change:
Screenshot 2024-03-07 at 10 06 33 AM (2)

frances-h (commented):
Could we add a test for this as well?

fealho:
> Could we add a test for this as well?

I can't write a test which actually verifies the logic, since that requires SDV-Enterprise. I can add a unit test that runs through the logic though.

tests/unit/data_processing/test_data_processor.py
```diff
@@ -1496,6 +1496,40 @@ def test__create_config_with_address_columns(self):
         }
         assert config == expected_config
 
+    def test__create_config_with_contextual_column_as_primary_key(self):
+        """Test it when a conextual column is a primary key."""
```
frances-h, line 1500: typo: conextual -> contextual

frances-h (approved):
looks good, thanks for addressing!

R-Palazzo (approved):

PR #1854, opened 2024-03-18
Add verbosity to `drop_unknown_references`

### LINKED ISSUE #1845: Add verbosity to `drop_unknown_references`
### Problem Description
Related to #1792: The new utility function for `drop_unknown_references` will be used to drop invalid foreign key rows that do not point to an existing parent. When using it, I receive data with referential integrity but there is no indication as to how many rows were dropped and how many remain.

### Expected behavior
Add an optional parameter `verbose` to the end of this function. By default, it will be `True`.

When set to `True`, there should be a message printed out that tells me how many rows remain per table. This should be pretty printed so that the numbers align -- perhaps as a pandas DatFrame.

```
>>> valid_data = drop_unknown_references(metadata, data, verbose=True)
Success! All foreign keys have referential integrity. 

Table Name    # Rows (Original)    # Invalid Rows   # Rows (New)
sessions      1200                 50               1150     
transactions  5000                 0                5000
```

Note that the # Rows (Original) should be equal to # Invalid Keys + # Rows (New)

If `verbose=False`, then do not show this message.

R-Palazzo:
CU-86aznj0ht
Resolve #1845

@npatki I slightly changed the message printed compared to the issue let me know if it works:

- Data with referential integrity:
Screenshot 2024-03-18 at 14 43 30

- Data without referential integrity:
Screenshot 2024-03-18 at 14 44 03

### DIFF
--- a/sdv/utils/poc.py
+++ b/sdv/utils/poc.py
@@ -1,10 +1,14 @@
 """Utility functions."""
+import sys
+
+import pandas as pd
+
 from sdv._utils import (
     _get_relationship_for_child, _get_rows_to_drop, _validate_foreign_keys_not_null)
 from sdv.errors import InvalidDataError, SynthesizerInputError
 
 
-def drop_unknown_references(metadata, data, drop_missing_values=True):
+def drop_unknown_references(metadata, data, drop_missing_values=True, verbose=True):
     """Drop rows with unknown foreign keys.
 
     Args:
@@ -17,16 +21,24 @@ def drop_unknown_references(metadata, data, drop_missing_values=True):
             Boolean describing whether or not to also drop foreign keys with missing values
             If True, drop rows with missing values in the foreign keys.
             Defaults to True.
+        verbose (bool):
+            If True, print information about the rows that are dropped.
+            Defaults to True.
 
     Returns:
         dict:
             Dictionary with the dataframes ensuring referential integrity.
     """
+    success_message = 'Success! All foreign keys have referential integrity.'
     metadata.validate()
     try:
         metadata.validate_data(data)
         if drop_missing_values:
             _validate_foreign_keys_not_null(metadata, data)
+
+        if verbose:
+            message = [success_message, 'No rows were dropped.']
+            sys.stdout.write('\n'.join(message))
 
         return data
     except (InvalidDataError, SynthesizerInputError):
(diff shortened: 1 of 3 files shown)


sdv-team:
Task linked: [CU-86aznj0ht SDV - Add verbosity to `drop_unknown_references` #1845](https://app.clickup.com/t/86aznj0ht)

sdv-team:
Task linked: [CU-86aznj0ht SDV - Add verbosity to `drop_unknown_references` #1845](https://app.clickup.com/t/86aznj0ht)

tests/integration/utils/test_poc.py
```diff
@@ -75,28 +75,43 @@ def test_drop_unknown_references(metadata, data):
...
+    captured = capsys.readouterr()
 
     # Assert
     pd.testing.assert_frame_equal(result['parent'], data['parent'])
     pd.testing.assert_frame_equal(result['child'], data['child'])
+    expected_message = (
+        'Success! All foreign keys have referential integrity.\n'
+        'No rows were dropped.'
+    )
+    assert captured.out.strip() == expected_message
```
npatki, line 111: I would rather we not have a special case message for this case. In the spec, we said we should print out for every table all the time.

Even if nothing was dropped, I think it's useful reassurance to the user that we checked all tables. Plus, it simplifies our logic.

npatki:
@R-Palazzo I also think that the first case you have added a line `Summary of the number of rows dropped:` which is not really necessary

sdv/utils/poc.py
```diff
@@ -17,22 +21,36 @@ def drop_unknown_references(metadata, data, drop_missing_values=True):
...
+        '# Rows (New)': [len(data[table]) for table in table_names]
+    })
     metadata.validate()
     try:
         metadata.validate_data(data)
         if drop_missing_values:
             _validate_foreign_keys_not_null(metadata, data)
 
+        if verbose:
+            sys.stdout.write('\n'.join([success_message, summary_table.to_string(index=False)]))
```
pvk-developer, line 47: Why `sys.stdout` instead of `print` (just curious).

Should we add one more `\n`, this is how it looks:
image

I think that this is a bit more digestable by the eye:
image

R-Palazzo: Thanks, done in [6067e02](https://github.com/sdv-dev/SDV/pull/1854/commits/6067e02072619dafdc0575c9380174654d94d196). We have a lint error when using `print` but I could also add `# noqa: T001` (in SDMetrics we use sys.stdout)

gsheni (approved):

PR #1857, opened 2024-03-19
Show warning when loading a synthesizer on a previously-saved SDV version

### LINKED ISSUE #1836: Show warning when loading a synthesizer on a previously-saved SDV version
### Problem Description

1. I am on the latest version of SDV (1.10.0) and I encounter a bug
2. I am able to find a workaround so I move on from the bug. I train and save a synthesizer on SDV 1.10.0
3. In a future SDV version (eg. 1.11.0), I am told that the bug is fixed so I upgrade to the new version
4. I load in my saved synthesizer from (3) expecting the bug to be fixed --- but it's not!

The scenario is confusing because it's never communicated to me that I cannot simply re-load my synthesizer expecting a bug to be fixed. I must actually create a new synthesizer and re-fit it in order to see the bug fixes.

### Expected behavior
Upon loading a synthesizer, show a warning if the synthesizer was created on a previous version of SDV. 

```python
SDV Version Warning: You are currently on SDV version 1.11.0 but this 
synthesizer was created on version 1.10.0. The latest bug fixes and features 
may not be available for this synthesizer. To see these enhancements, 
create and train a new synthesizer on this version.
```

### Additional context
Apply this warning to _all_ synthesizers (available to public and enterprise)

Note that the trained version is available under `synthesizer.get_info()` -- (the enterprise version may not be -- perhaps we should add this)

pvk-developer:
Resolves #1836 
CU-86azkp7mu

Added a new attribute to the `synthesizers`  called `_fitted_sdv_enterprise_version` which will only be added to `get_info` if it is not `None`.

### DIFF
--- a/sdv/_utils.py
+++ b/sdv/_utils.py
@@ -316,32 +316,31 @@ def check_sdv_versions_and_warn(synthesizer):
     public_missmatch = current_public_version != synthesizer._fitted_sdv_version
     enterprise_missmatch = current_enterprise_version != synthesizer._fitted_sdv_enterprise_version
     if synthesizer._fitted and (public_missmatch or enterprise_missmatch):
+        static_message = (
+            'The latest bug fixes and features may not be available for this synthesizer. '
+            'To see these enhancements, create and train a new synthesizer on this version.'
+        )
         if public_missmatch and enterprise_missmatch:
             message = (
                 'You are currently on SDV version '
                 f'{current_public_version} and SDV Enterprise version '
                 f'{current_enterprise_version} but this synthesizer was created on '
                 f'SDV version {synthesizer._fitted_sdv_version} and SDV Enterprise version '
-                f'{synthesizer._fitted_sdv_enterprise_version}. The latest bug fixes and features '
-                'may not be available for this synthesizer. To see these enhancements, '
-                'create and train a new synthesizer on this version.'
+                f'{synthesizer._fitted_sdv_enterprise_version}.'
             )
 
         elif public_missmatch:
             message = (
                 'You are currently on SDV version '
                 f'{current_public_version} but this synthesizer was created on '
-                f'version {synthesizer._fitted_sdv_version}. The latest bug fixes and features '
-                'may not be available for this synthesizer. To see these enhancements, '
-                'create and train a new synthesizer on this version.'
+                f'version {synthesizer._fitted_sdv_version}.'
             )
         elif enterprise_missmatch:
             message = (
                 'You are currently on SDV Enterprise version '
                 f'{current_enterprise_version} but this synthesizer was created on '
-                f'version {synthesizer._fitted_sdv_enterprise_version}. The latest bug fixes and '
-                'features may not be available for this synthesizer. To see these '
-                'enhancements, create and train a new synthesizer on this version.'
+                f'version {synthesizer._fitted_sdv_enterprise_version}.'
             )
 
+        message = f'{message} {static_message}'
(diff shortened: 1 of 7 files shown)


sdv-team:
Task linked: [CU-86azkp7mu SDV - Show warning when loading a synthesizer on a previously-saved SDV version #1836](https://app.clickup.com/t/86azkp7mu)

sdv/_utils.py
```diff
@@ -295,3 +297,51 @@ def _validate_foreign_keys_not_null(metadata, data):
...
+                'may not be available for this synthesizer. To see these enhancements, '
+                'create and train a new synthesizer on this version.'
+            )
+        elif enterprise_missmatch:
+            message = (
+                'You are currently on SDV Enterprise version '
+                f'{current_enterprise_version} but this synthesizer was created on '
+                f'version {synthesizer._fitted_sdv_enterprise_version}. The latest bug fixes and '
+                'features may not be available for this synthesizer. To see these '
+                'enhancements, create and train a new synthesizer on this version.'
```
R-Palazzo, line 344: To be less redundant, we should break the message into multiple parts and organize them according to the user situation(public or enterprise)

tests/unit/test__utils.py
```diff
@@ -459,3 +460,89 @@ def side_effect_func(value):
...
+@patch('sdv._utils.warnings')
+def test_check_sdv_versions_and_warn_public_missmatch(mock_warnings):
+    """Test that warnings is raised when public version is missmatched."""
+    # Setup
+    synthesizer = Mock()
+    synthesizer._fitted_sdv_version = '1.0.0'
+    synthesizer._fitted_sdv_enterprise_version = version.enterprise
+
+    # Run
+    check_sdv_versions_and_warn(synthesizer)
```
R-Palazzo, line 489: Can we also use pytest logic:
`with pytest.warns(SDVVersionWarning, match=message):`

R-Palazzo (commented):
Looks good!
Good job, just let few suggestions

tests/unit/test__utils.py
pvk-developer: Addressed in [c6051b8](https://github.com/sdv-dev/SDV/pull/1857/commits/c6051b8c59d29769e49fcf8f729673a03e58fe24)

sdv/_utils.py
pvk-developer: Addressed in [c6051b8](https://github.com/sdv-dev/SDV/pull/1857/commits/c6051b8c59d29769e49fcf8f729673a03e58fe24)

R-Palazzo (approved):
LGTM!

sdv/_utils.py
```diff
@@ -295,3 +297,50 @@ def _validate_foreign_keys_not_null(metadata, data):
...
+
+    Raises:
+        SDVVersionWarning:
+            If the current SDV or SDV Enterprise version does not match the version used to fit
+            the synthesizer.
+    """
+    current_public_version = getattr(version, 'public', None)
+    current_enterprise_version = getattr(version, 'enterprise', None)
+    public_missmatch = current_public_version != synthesizer._fitted_sdv_version
+    enterprise_missmatch = current_enterprise_version != synthesizer._fitted_sdv_enterprise_version
```
frances-h, line 317: I think we can't assume loaded synthesizers will have this attribute since it's being added

pvk-developer: This should be addressed now using `getattr` instead of directly accessing it.

frances-h (approved):

PR #1859, opened 2024-03-20
Add a function to save my multi-table data as CSVs

### LINKED ISSUE #1849: Add a function to save my multi-table data as CSVs
### Problem Description
Currently, I am able to load in a bunch of CSVs at a time from a folder using `load_csvs`. But once I create synthetic data for this multi-table synthesizer, there is no easy way to write it back into CSV form.

### Expected behavior
Create a method called `save_csvs` in the `sdv.datasets.local` module. This would work the same as [`load_csvs`](https://docs.sdv.dev/sdv/multi-table-data/data-preparation/loading-data#load_csvs) but in reverse.

Parameters:
- (required) `data`: The data to save. This is a dictionary of string table names mapped to a pandas.DataFrame containing the data.
- (required) `folder_name`: The name of the folder to write the CSVs into. If the folder does not exist, we will create the folder
- `suffix`: An optional suffix to append to each CSV.
  - (default) If there is no suffix: Each table will be saved as `<table_name>.csv`
  - The suffix is `'-synthetic'`. If a suffix is provided, we'll add it before `.csv`. So if suffix is `'-synthetic'` then each table will be saved as `'<table_name>-synthetic.csv'`.
- `to_csv_parameters`: A dictionary with additional parameters to pass in when saving CSVs. The keys are any of the parameters in [pandas.to_csv](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_csv.html) and the values are the inputs

```python
from sdv.datasets.local import save_csvs

synthetic_data = synthesizer.sample(scale=1.0)

save_csvs(
  data=synthetic_data,
  folder_name='data/'
  suffix='-synthetic',
  to_csv_parameters={ 'sep': '\t' }
)
```

### Additional context
Before attempting to save the CSVs, check to see if any of the exact file names actually exist. If they are, throw an error because we don't want to accidentally override existing files.

R-Palazzo:
CU-86azny031
Resolve #1849

### DIFF
--- a/sdv/datasets/local.py
+++ b/sdv/datasets/local.py
@@ -42,3 +42,48 @@ def load_csvs(folder_name, read_csv_parameters=None):
         )
 
     return csvs
+
+
+def save_csvs(data, folder_name, suffix=None, to_csv_parameters=None):
+    """Save dataframes to csv files in a specified folder.
+
+    Args:
+        data (dict):
+            A dictionary that maps each table name (string) to the data for that
+            table (pandas.DataFrame).
+        folder_name (str):
+            The full path of the folder where the data will be saved.
+        suffix (str):
+            A string to be appended to the name of the file. Defaults to ``None``.
+        to_csv_parameters (dict):
+            A python dictionary of with string and value accepted by ``pandas.DataFrame.to_csv``
+            function. Defaults to ``None``.
+    """
+    if not path.exists(folder_name):
+        raise ValueError(f"The folder '{folder_name}' cannot be found.")
+
+    table_name_to_filepath = {}
+    errors = []
+    for table_name in data:
+        filename = f'{table_name}{suffix}.csv' if suffix else f'{table_name}.csv'
+        filepath = path.join(folder_name, filename)
+        if path.exists(filepath):
+            errors.append(filename)
+
+        table_name_to_filepath[table_name] = filepath
+
+    if errors:
+        end_message = 'Please remove them or specify a different suffix.'
+        filename_to_print = '\n'.join(errors[:3])
+        if len(errors) > 3:
+            end_message = ''.join([f'+ {len(errors) - 3} more files.', end_message])
+
+        raise ValueError(
+            f"The following files already exist in '{folder_name}':\n{filename_to_print}"
+            f'\n{end_message}'
+        )
+
+    to_csv_parameters = to_csv_parameters or {}
+    for table_name, filepath in table_name_to_filepath.items():
+        table = data[table_name]
+        table.to_csv(filepath, **to_csv_parameters)
--- a/sdv/datasets/local.py
+++ b/sdv/datasets/local.py
@@ -1,7 +1,7 @@
 """Methods to load local datasets."""
 
 import warnings
-from os import path, walk
+from os import makedirs, path, walk
 
 from sdv._utils import _load_data_from_csv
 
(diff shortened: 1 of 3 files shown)


sdv-team:
Task linked: [CU-86azny031 SDV - Add a function to save my multi-table data as CSVs #1849](https://app.clickup.com/t/86azny031)

sdv/datasets/local.py
```diff
@@ -42,3 +42,48 @@ def load_csvs(folder_name, read_csv_parameters=None):
...
+        folder_name (str):
+            The full path of the folder where the data will be saved.
+        suffix (str):
+            A string to be appended to the name of the file. Defaults to ``None``.
+        to_csv_parameters (dict):
+            A python dictionary of with string and value accepted by ``pandas.DataFrame.to_csv``
+            function. Defaults to ``None``.
+    """
+    if not path.exists(folder_name):
+        raise ValueError(f"The folder '{folder_name}' cannot be found.")
```
pvk-developer, line 63:
```
(required) folder_name: The name of the folder to write the CSVs into. If the folder does not exist, we will create the folder
```
Instead of raising the error we have to create the folder.

sdv/datasets/local.py
```diff
@@ -42,3 +42,48 @@ def load_csvs(folder_name, read_csv_parameters=None):
...
+
+        table_name_to_filepath[table_name] = filepath
+
+    if errors:
+        end_message = 'Please remove them or specify a different suffix.'
+        filename_to_print = '\n'.join(errors[:3])
+        if len(errors) > 3:
+            end_message = ''.join([f'+ {len(errors) - 3} more files.', end_message])
+
+        raise ValueError(
```
pvk-developer, line 81: Instead of `ValueError` lets use `FileExistsError`

sdv/datasets/local.py
R-Palazzo: Good catch, thanks, done in [fab0aae](https://github.com/sdv-dev/SDV/pull/1859/commits/fab0aaed1dc1d68954cb43d89512b95e297779b7)

sdv/datasets/local.py
R-Palazzo: Yes, done in [fab0aae](https://github.com/sdv-dev/SDV/pull/1859/commits/fab0aaed1dc1d68954cb43d89512b95e297779b7)

sdv/datasets/local.py
```diff
@@ -1,7 +1,9 @@
 """Methods to load local datasets."""
 
 import warnings
-from os import path, walk
+from os import makedirs, path, walk
```
pvk-developer, line 4: Can we import `os` and use `os.path` `os.makedirs` etc.. it is a bit more explicit. Also `path` makes me think of `pathlib` and it  can be a bit confusing.

pvk-developer (approved):
Just one nit pick, but other than that LGTM!

sdv/datasets/local.py
frances-h: Do we have a preference between `os.path` and `pathlib` for situations like this? I usually end up using `pathlib`, but not sure if there's a standard.

pvk-developer: `pathlib` is the way to go when it comes to `paths`, eg: exploring paths, joining paths etc.. but when it comes to interact with the system (like creating a folder etc..) since `pathlib` doesn't offer that functionality `os` is the library of preference.

frances-h (approved):

PR #1865, opened 2024-03-22
Fix assert in PAR integration test

gsheni:
- The assertion was not properly being tested. 
- Added a fix if demo dataset downloaded was created after V1 (in which case there would be no metadata_v0.json present)

### DIFF
--- a/sdv/datasets/demo.py
+++ b/sdv/datasets/demo.py
@@ -67,9 +67,9 @@ def _extract_data(bytes_io, output_folder_name):
         if output_folder_name:
             os.makedirs(output_folder_name, exist_ok=True)
             zf.extractall(output_folder_name)
-            metadata_v1_filepath = os.path.join(output_folder_name, 'metadata_v0.json')
-            if os.path.isfile(metadata_v1_filepath):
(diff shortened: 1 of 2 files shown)


frances-h (approved):

sdv/datasets/demo.py
```diff
@@ -67,7 +67,9 @@ def _extract_data(bytes_io, output_folder_name):
         if output_folder_name:
             os.makedirs(output_folder_name, exist_ok=True)
             zf.extractall(output_folder_name)
-            os.remove(os.path.join(output_folder_name, 'metadata_v0.json'))
+            metadata_v1_filepath = os.path.join(output_folder_name, 'metadata_v0.json')
```
amontanez24, line 70: why is it called metadata_v1_filepath but the suffix has v0?

gsheni: Mistake. Fixed https://github.com/sdv-dev/SDV/pull/1865/commits/04d2083ea057aa9fd10f469b0640dc0d481808f4

amontanez24 (approved):

PR #1867, opened 2024-03-25
HMASynthesizer diagnostic score is not 1.0 when using 'truncnorm' distribution 

### LINKED ISSUE #1831: HMASynthesizer diagnostic score is not 1.0 when using `'truncnorm'` distribution
### Environment Details

* SDV version: 1.10.0
* Python version: (any)
* Operating System: (any)

### Error Description
If I update the default distribution to `'truncnorm'`, then the HMASynthesizer creates synthetic data that is not completely valid. When running the diagnostic report, the Data Validity score is not 100% -- because there are extra NaN/NaT values that appear in the synthetic data.

### Steps to reproduce
Replicate this using the attached metadata and data.

```python
from sdv.datasets.local import load_csvs

data = load_csvs(folder_name='test_data/')
metadata = MultiTableMetadata.load_from_json('test_metadata.json')

for table_name in data.keys():
  synthesizer.set_table_parameters(
    table_name=table_name,
    table_parameters={
      'default_distribution': 'truncnorm'})

synthesizer.fit(data)
synthetic_data = synthesizer.sample()

diagnostic_report = run_diagnostic(
  real_data=data, synthetic_data=synthetic_data, metadata=metadata)
```

[test_data.zip](https://github.com/sdv-dev/SDV/files/14487213/test_data.zip)
[test_metadata.json](https://github.com/sdv-dev/SDV/files/14487215/test_metadata.json)

**OUTPUT**:
At first, you'll see many warnings originating by truncated gaussian during modeling:
```
site-packages/copulas/univariate/truncated_gaussian.py:45: RuntimeWarning: invalid value encountered in scalar divide
site-packages/copulas/univariate/truncated_gaussian.py:46: RuntimeWarning: divide by zero encountered in scalar divide
```
Then during sampling, there are more warnings that the transformed data (coming directly from ML models) contain null values and therefore overall synthetic data (after reverse sampling) will also have null values.
```
site-packages/rdt/transformers/utils.py:217: UserWarning: There are null values in the transformed data. The reversed transformed data will contain null values.
```

Finally, the diagnostic is not 100%:
```
Overall Score: 94.67%

Properties:
- Data Validity: 84.01%
- Data Structure: 100.0%
- Relationship Validity: 100.0%
```

### Additional Context
This was first observed in #1755

frances-h:
CU-86azk5f40
Resolve #1831 

This issue was caused by invalid parameters being sampled for a child synthesizer (in this case, b > a). This generates parameters for the original child table which we can fall back to if the sampled parameters are invalid.

### DIFF
--- a/sdv/multi_table/hma.py
+++ b/sdv/multi_table/hma.py
@@ -39,6 +39,7 @@ class HMASynthesizer(BaseHierarchicalSampler, BaseMultiTableSynthesizer):
         self._max_child_rows = {}
         self._augmented_tables = []
         self._learned_relationships = 0
+        self._default_parameters = {}
         self.verbose = verbose
         BaseHierarchicalSampler.__init__(
             self,
@@ -418,11 +419,9 @@ class HMASynthesizer(BaseHierarchicalSampler, BaseMultiTableSynthesizer):
             augmented_data (dict):
                 Dictionary mapping each table name to an augmented ``pandas.DataFrame``.
         """
-        parent_map = self.metadata._get_parent_map()
         augmented_data_to_model = [
             (table_name, table)
             for table_name, table in augmented_data.items()
-            if table_name not in parent_map
         ]
         self._print(text='\n', end='')
         pbar_args = self._get_pbar_args(desc='Modeling Tables')
@@ -434,6 +433,11 @@ class HMASynthesizer(BaseHierarchicalSampler, BaseMultiTableSynthesizer):
 
             if not table.empty:
                 self._table_synthesizers[table_name].fit_processed_data(table)
+                table_parameters = self._table_synthesizers[table_name]._get_parameters()
+                self._default_parameters[table_name] = {
+                    parameter: value for parameter, value in table_parameters.items()
+                    if 'univariates' in parameter
+                }
 
             for name, values in keys.items():
                 table[name] = values
@@ -475,10 +479,11 @@ class HMASynthesizer(BaseHierarchicalSampler, BaseMultiTableSynthesizer):
         # A child table is created based on only one foreign key.
         foreign_key = self.metadata._get_foreign_keys(parent_name, child_name)[0]
         parameters = self._extract_parameters(parent_row, child_name, foreign_key)
+        default_parameters = getattr(self, '_default_parameters', {}).get(child_name, {})
 
         table_meta = self.metadata.tables[child_name]
         synthesizer = self._synthesizer(table_meta, **self._table_parameters[child_name])
-        synthesizer._set_parameters(parameters)
+        synthesizer._set_parameters(parameters, default_parameters)
         synthesizer._data_processor = self._table_synthesizers[child_name]._data_processor
 
         return synthesizer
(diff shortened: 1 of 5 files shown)


sdv-team:
Task linked: [CU-86azk5f40 SDV - HMASynthesizer diagnostic score is not 1.0 when using `'truncnorm'` distribution #1831](https://app.clickup.com/t/86azk5f40)

sdv/single_table/copulas.py
```diff
@@ -323,12 +324,14 @@ def _rebuild_correlation_matrix(cls, triangular_correlation):
 
         return cls._get_nearest_correlation_matrix(correlation).tolist()
 
-    def _rebuild_gaussian_copula(self, model_parameters):
+    def _rebuild_gaussian_copula(self, model_parameters, default_params={}):
```
pvk-developer, line 327: Can we change `default_pramas={}` to `None` and create the dict inside if required.
Here is why:
image

sdv/single_table/copulas.py
```diff
@@ -362,18 +388,24 @@ def _rebuild_gaussian_copula(self, model_parameters):
     def _get_likelihood(self, table_rows):
         return self._model.probability_density(table_rows)
 
-    def _set_parameters(self, parameters):
+    def _set_parameters(self, parameters, default_params={}):
```
pvk-developer, line 391: Same as -- change `default_params={}` to `default_params=None`

gsheni (approved):
lgtm

sdv/multi_table/hma.py
```diff
@@ -418,11 +419,9 @@ def _model_tables(self, augmented_data):
             augmented_data (dict):
                 Dictionary mapping each table name to an augmented ``pandas.DataFrame``.
         """
-        parent_map = self.metadata._get_parent_map()
```
amontanez24, line 421: why are we getting rid of this check?

sdv/single_table/copulas.py
```diff
@@ -323,25 +324,53 @@ def _rebuild_correlation_matrix(cls, triangular_correlation):
...
+            univariate['type'] = univariate_type
+            model = univariate_type.MODEL_CLASS
+            if hasattr(model, '_argcheck'):
+                to_check = {
+                    parameter: univariate[parameter]
+                    for parameter in inspect.signature(model._argcheck).parameters.keys()
+                    if parameter in univariate
+                }
+                if not model._argcheck(**to_check):
+                    if column in default_params.get('univariates', []):
```
amontanez24, line 361: what is the situatiom where the column won't have default parameters?

amontanez24 (commented):
Looks good! Just a couple questions

sdv/single_table/copulas.py
frances-h: If we load in an old synthesizer and try to sample from it, or in the single table case where we don’t need them.

sdv/multi_table/hma.py
frances-h: We need to model all the tables now to get default parameters, not just the roots.

pvk-developer (approved):

PR #1873, opened 2024-03-26
Implement Version Checking and Raise Error for Downgraded SDV Version Loading

### LINKED ISSUE #1837: Show an error if I accidentally load an SDV synthesizer on an older version of SDV
### Problem Description

1. I have been trying SDV for many months. I have multiple environments running with different SDV versions
2. I train and save a synthesizer on the latest SDV version (1.10.0)
3. Later, I accidentally re-load that synthesizer in an environment with an older version (eg. in my other environment with 1.9.0)
4. Some features don't work and error out. It is difficult to debug why this is happening.

Note that in this case, it is _expected_ that some features may error, as SDV does not offer forwards compatibility. That is to say, if you downgrade your SDV version (in step 3), there is no guarantee that anything will work.

### Expected behavior
Create a new `VersionError` type and show this if I try to load any synthesizer in this scenario.

```python
VersionError: You are currently on SDV version 1.8.0 but this 
synthesizer was created on version 1.10.0. Downgrading your SDV version is not supported.
```

### Additional context
Apply this error to all synthesizers (available to public and enterprise)

Note that the trained version is available under `synthesizer.get_info()` (the enterprise version may not be -- perhaps we should add this)

pvk-developer:
Resolves #1837 
CU-86azkp7me

### DIFF
--- a/sdv/_utils.py
+++ b/sdv/_utils.py
@@ -10,7 +10,7 @@ import pandas as pd
 from pandas.core.tools.datetimes import _guess_datetime_format_for_array
 
 from sdv import version
-from sdv.errors import SDVVersionWarning, SynthesizerInputError
+from sdv.errors import SDVVersionWarning, SynthesizerInputError, VersionError
 
 
 def _cast_to_iterable(value):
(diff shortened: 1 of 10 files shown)


sdv-team:
Task linked: [CU-86azkp7me SDV - Show an error if I accidentally load an SDV synthesizer on an older version of SDV #1837](https://app.clickup.com/t/86azkp7me)

frances-h (commented):
Can we also add an integration test? We could manually modify the `_fitted_sdv_version` on a synthesizer and test saving / loading it.

sdv/_utils.py
```diff
@@ -348,3 +348,83 @@ def check_sdv_versions_and_warn(synthesizer):
...
+            pass
+
+    return False
+
+
+def check_synthesizer_version(synthesizer):
+    """Check if the current synthesizer version is greater than the package version.
+
+    Args:
+        synthesizer (BaseSingleTableSynthesizer or BaseMultiTableSynthesizer):
```
R-Palazzo, line 389: Does it work for PAR also?

R-Palazzo (approved):
LGTM!
I agree with @frances-h about having an integration test for it

sdv/_utils.py
```diff
@@ -348,3 +348,83 @@ def check_sdv_versions_and_warn(synthesizer):
 
             message = f'{message} {static_message}'
             warnings.warn(message, SDVVersionWarning)
+
+
+def _check_greater_version(current_version, synthesizer_version):
```
frances-h, line 353: minor nitpick: can we swap the argument order here or change the function name to `_check_is_lower_version`? It's counterintuitive to me that `_is_greater_version(a, b)` returns `True` when `a < b` 😅

sdv/_utils.py
pvk-developer: Yes, I will correct the `BaseSingleTableSynthesizer` to `BaseSynthesizer` since that is the main class being inherited.

frances-h (approved):
LGTM!

PR #1874, opened 2024-03-26
Add utility function to simplify multi-table schemas

### LINKED ISSUE #1832: Add utility function to simplify multi-table schemas
### Problem Description
Currently, HMA cannot run on certain multi-table schemas. We issue a warning when a schema will generate too many columns, and we should provide a utility function to easily reduce a multi-table schema so it can successfully run on HMA.

### Expected behavior
Add a new utility function `utils.simplify_schema`:
Parameters:
- `data` - the data dictionary
- `metadata` - the MultiTableMetadata for this dataset

Returns:
- A data dictionary mapping table names to simplified tables
- `MultiTableMetadata` for the simplified data schema

```python
from sdv.utils import simplify_schema

simple_data, simple_metadata = simplify_schema(
  data=my_data,
  metadata=my_metadata
)
```

#### Algorithm overview
For every root table:
  - drop any table that is depth > 2 away from the parent (i.e. keep only direct children and grandchildren) and count the number of tables connected to the root

Select the root with the greatest number of descendant tables
Calculate the number of extended columns we can add to the root (we can reuse the logic used to generate the warning in HMA)
Allocate a # of augmented columns to each child relationship
For each child:
- Determine the number of modelable columns and add the number of child relationships for that child
- If the number of modelable columns will generate more than the allowed number of extended columns, drop modelable columns from the child
  - Try to keep a variety of sdtypes
  - If we cannot drop columns so that we will not exceed the maximum number of extended_columns, drop any grandchild tables until we can

For each grandchild:
- Drop all modelable columns (grandchildren should only generate a `num_rows` column in their parents)

### Additional context
We should also change the warning in HMA to point to this utility function:
```
>>> synthesizer = HMASynthesizer(metadata)
PerformanceAlert: Using the HMASynthesizer on this metadata schema is not recommended because HMA will generate a large number of columns

Table Name   # Columns in Metadata   Est # Columns
users        12                      123123123
transactions   
...

We recommend simplifying your metadata schema using utils.simplify_schema
```

R-Palazzo:
CU-86azkea74
Resolve #1832

Formula to determine how many columns to drop for a child
- $n$: number of columns of the table
- $k$: number of parameter columns per column. For beta's distribution $k=4$
- $m = \frac{1000}{\text{total number of relationship}}$: maximum number of column per relationship
- minimum number of column to drop = $n + k - \sqrt{k^2 + 1 + 2m}$

Some other info:
- I updated some methods inside HMA to be `class` or `static methods` so I can use them outside HMA. I preferred keeping them inside HMA, though, because they are very specific to this model (not used for HSA, for instance.)
- There is an integration test over our biggest demo datasets.
The method `_simplify_data(data, metadata)` makes your data match your metadata.
It can be useful when you're updating your metadata (removing columns and tables...) to make it work along with your data. If we want, it can be public.


sdv-team:
Task linked: [CU-86azkea74 SDV - Add utility function to simplify multi-table schemas #1832](https://app.clickup.com/t/86azkea74)

amontanez24 (changes requested):
This is looking pretty good! I think in a couple of places we can maybe reduce the number of functions and simplify the logic a few steps. It would be helpful to add a docstring for the `_get_num_column_to_drop` so we can discuss it more

Also in general can we fix some of the names:

- childs -> children
- grandchilds -> grandchildren
- simplify -> simplified if it is being used as an adjective

sdv/multi_table/utils.py
```diff
@@ -0,0 +1,421 @@
...
+    Args:
+        relationships (list[dict]):
+            List of relationships between the tables.
+        order (int):
+            Order of the descendants.
+    """
+    root_tables = _get_root_tables(relationships)
+    all_descendants = {}
+    for root in root_tables:
+        order_descendances = _get_n_order_descendants(relationships, root, order)
```
frances-h, line 67: minor: typo

sdv/multi_table/utils.py
```diff
@@ -152,7 +152,27 @@ def _simplify_grandchilds(metadata, grandchilds):
...
+        metadata (MultiTableMetadata):
+            Metadata of the datasets.
+        child_table (str):
+            Name of the child table.
+        max_col_per_relationships (int):
+            Maximum number of columns to model per relationship.
+
+    Returns:
+        int:
+            Number of columns to drop.
```
amontanez24, line 184: it looks like this returns a tuple

tests/unit/utils/test_poc.py
```diff
@@ -409,3 +411,33 @@ def test_simplify_schema(mock_print_summary, mock_get_total_estimated_columns,
...
+        ' data/metadata combination is valid before trying to simplify the schema.'
+    )
+    with pytest.raises(InvalidMetadataError, match=expected_message):
+        simplify_schema(real_data, metadata)
+
+
+def test_simplify_schema_invalid_data():
+    """Test ``simplify_schema`` when the data is not valid."""
+    # Setup
+    real_data, metadata = download_demo('multi_table', 'fake_hotels')
```
amontanez24, line 434: Ideally we shouldn't download data for the unit tests

sdv/multi_table/utils.py
```diff
@@ -97,31 +89,50 @@ def _simplify_relationships(metadata, root_table, descendants_to_keep):
...
         is_valid_parent = is_parent_root or is_parent_in_descendants
         is_valid_child = child_table in descendants_to_keep
-        if not is_valid_parent or not is_valid_child:
-            metadata.remove_relationship(parent_table, child_table)
-        else:
+        if is_valid_parent and is_valid_child:
             if is_parent_root:
                 children.append(child_table)
             else:
                 grandchildren.append(child_table)
```
amontanez24: I think I misunderstood what this was doing before. It's fine to prune the metadata here while you're already looping through the relationships. Can we update the docstring though to reflect all of that?

R-Palazzo:
Thanks for the review @amontanez24, I addressed latest comments in this commit [8ca4158](https://github.com/sdv-dev/SDV/pull/1874/commits/8ca41583f535db84f8d4b14ade75b16d211be72a).

sdv/multi_table/utils.py
```diff
@@ -0,0 +1,431 @@
...
+from sdv.multi_table.hma import MAX_NUMBER_OF_COLUMNS
+
+MODELABLE_SDTYPE = ['categorical', 'numerical', 'datetime', 'boolean']
+
+
+def _get_relationship_for_child(relationships, child_table):
+    return [rel for rel in relationships if rel['child_table_name'] == child_table]
+
+
+def _get_relationship_for_parent(relationships, parent_table):
```
amontanez24, line 20: minor: can we make plural (ie. get_relationships)

sdv/multi_table/utils.py
```diff
@@ -0,0 +1,431 @@
...
+    root_tables = _get_root_tables(relationships)
+    all_descendants = {}
+    for root in root_tables:
+        order_descendants = _get_n_order_descendants(relationships, root, order)
+        all_descendant_root = set()
+        for order_key in order_descendants:
+            all_descendant_root.update(order_descendants[order_key])
+
+        all_descendants[root] = all_descendant_root
+
```
amontanez24, line 68: could we prevent having to recompute children and grandchildren later by having this method return a dict in the following format:
```
root -> {'order_1': ..., 'order_2': ..., 'num_descendants': 6}
```

R-Palazzo: Good idea, done in [209b720](https://github.com/sdv-dev/SDV/pull/1874/commits/209b720e703f9fef6eb4ceb112c5506804198b66)

sdv/multi_table/utils.py
R-Palazzo: Yes, done in [209b720](https://github.com/sdv-dev/SDV/pull/1874/commits/209b720e703f9fef6eb4ceb112c5506804198b66)

sdv/multi_table/hma.py
```diff
@@ -172,51 +245,12 @@ def _print_estimate_warning(self):
...
-        columns_per_table = self._get_num_data_columns()
-
-        # Starting at root tables, recursively estimate the number of columns
-        # each table will model
-        visited = set()
-        for table_name in self._get_root_parents():
-            self._estimate_columns_traversal(table_name, columns_per_table, visited)
-
-        return columns_per_table
+                "We recommend simplifying your metadata schema using 'sdv.utils.simplify_schema'."
```
frances-h, line 251: minor: should be "using 'sdv.utils.poc.simplify_schema'"

sdv/multi_table/utils.py
```diff
@@ -0,0 +1,400 @@
...
+    """Get the descendants of the parent table until a given order.
+
+    Args:
+        relationships (list[dict]):
+            List of relationships between the tables.
+        parent_table (str):
+            Name of the parent table.
+        order (int):
+            Order of the descendants.
+    """
```
frances-h, line 34: minor: add returns to docstring

sdv/multi_table/utils.py
```diff
@@ -0,0 +1,400 @@
...
+
+def _get_all_descendant_per_root_at_order_n(relationships, order):
+    """Get all the descendants of the root tables at a given order.
+
+    Args:
+        relationships (list[dict]):
+            List of relationships between the tables.
+        order (int):
+            Order of the descendants.
+    """
```
frances-h, line 58: minor: add returns to docstring

sdv/multi_table/utils.py
```diff
@@ -0,0 +1,400 @@
...
+        is_parent_to_drop = parent_table in tables_to_drop
+        is_child_to_drop = child_table in tables_to_drop
+        if is_parent_to_drop or is_child_to_drop:
+            metadata.relationships.remove(relationship)
+            if is_parent_to_drop and parent_table in metadata.tables:
+                del metadata.tables[parent_table]
+            elif is_child_to_drop and child_table in metadata.tables:
+                del metadata.tables[child_table]
+
+    return metadata
```
frances-h, line 103: Since this function will actually modify the metadata object we pass in, we should remove the return. The convention is usually to have functions that modify arguments in-place to not return anything.

sdv/multi_table/utils.py
```diff
@@ -0,0 +1,400 @@
...
+    """
+    for grandchild in grandchildren:
+        columns = metadata.tables[grandchild].columns
+        columns_to_drop = [
+            col_name for col_name in columns if columns[col_name]['sdtype'] in MODELABLE_SDTYPE
+        ]
+        for column in columns_to_drop:
+            del columns[column]
+
+    return metadata
```
frances-h, line 126: Same as above, we should remove the return since we're directly modifying the passed in metadata object

sdv/multi_table/utils.py
```diff
@@ -0,0 +1,400 @@
...
+            Name of the child table.
+        max_col_per_relationships (int):
+            Maximum number of columns to model per relationship.
+
+    Returns:
+        tuple:
+            - num_cols_to_drop (int): Number of columns to drop from the child table.
+            - modelable_columns (dict): Dictionary that maps the modelable sdtype to their columns.
+    """
+    num_column_parameter = 4  # for beta distribution
```
frances-h, line 152: Could we pull this from HMA instead of using a magic number? Maybe update the distribution to number of parameters dict to a constant, and pull it and HMA's default distribution (in case it ever changes) directly from HMA?

sdv/multi_table/utils.py
```diff
@@ -0,0 +1,400 @@
...
+    columns_to_sdtypes = {
+        column: columns[column]['sdtype'] for column in columns
+    }
+    sdtypes_to_columns = defaultdict(list)
+    for col, sdtype in columns_to_sdtypes.items():
+        sdtypes_to_columns[sdtype].append(col)
+
+    modelable_columns = {
+        key: value for key, value in sdtypes_to_columns.items() if key in MODELABLE_SDTYPE
+    }
```
frances-h, line 163: I think this can be simplified to something like:
```python
modelable_columns = defaultdict(list)
for col, col_meta in columns.items():
    if col_meta['sdtype'] in MODELABLE_SDTYPE:
        modelable_columns[col_meta['sdtype'].append(col)
```

amontanez24 (approved):
LGTM🍾 🎆

frances-h (approved):
LGTM, thanks for addressing! 🎉

PR #1886, opened 2024-04-03
Raise a `VersionError` when re-fitting a synthesizer that was created on a previous SDV version

### LINKED ISSUE #1838: Show an error when attempting to re-train a synthesizer that was created on a previous SDV version
### Problem Description

1. I am on the latest version of SDV (1.10.0) and I encounter a bug
2. I am able to find a workaround so I move on from the bug. I train and save a synthesizer on SDV 1.10.0
2. In a future SDV version (eg. 1.11.0), I am told that the bug is fixed so I upgrade to the new version
3. I load in my saved synthesizer from (3) and try to re-train it in order to fix the bug. It doesn't work.

The scenario is confusing because I might imagine re-training a synthesizer will work. But because the synthesizer object was created on an older version, this may lead to incompatibilities.

### Expected behavior
Create a new VersionError type and show this if I try to fit on any synthesizer in this scenario.

```python
>>> synthesizer.fit(data)
VersionError: You are currently on SDV version 1.11.0 but this synthesizer was created on version 1.10.0. Fitting 
this synthesizer again is not supported. Please create a new synthesizer.
```

### Additional context

Apply this error to all synthesizers (available to public and enterprise)

Note that _sampling_ the synthesizer is still ok. (This is what we guarantee.) It's just refitting that is not supported. 

Note that the trained version is available under synthesizer.get_info() (the enterprise version may not be -- perhaps we should add this)

pvk-developer:
Resolves #1838 
CU-86azkp7m4

### DIFF
--- a/sdv/_utils.py
+++ b/sdv/_utils.py
@@ -299,8 +299,12 @@ def _compare_versions(current_version, synthesizer_version, compare_operator=ope
 
     Returns:
         bool:
-            True if the synthesizer version is greater than the current version, False otherwise.
+            Depending on the ``operator`` function it will return ``True`` or ``False`` if
+            ``current_version`` is bigger or lower than ``synthesizer_version``.
     """
+    if None in (current_version, synthesizer_version):
+        return False
+
     current_version = current_version.split('.')
     synthesizer_version = synthesizer_version.split('.')
     for current_v, synth_v in zip(current_version, synthesizer_version):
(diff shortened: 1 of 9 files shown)


sdv-team:
Task linked: [CU-86azkp7m4 SDV - Show an error when attempting to re-train a synthesizer that was created on a previous SDV version #1838](https://app.clickup.com/t/86azkp7m4)

R-Palazzo (approved):
LGTM! 👍

@amontanez24, @npatki The current solution will not be applied to DayZ. When updating DayZ, we may want to fix this as well.

pvk-developer:
> LGTM! 👍
> 
> @amontanez24, @npatki The current solution will not be applied to DayZ. When updating DayZ, we may want to fix this as well.

Bear in mind that `DayZ` does not `fit`.

sdv/_utils.py
```diff
@@ -301,63 +307,81 @@ def _check_is_lower_version(current_version, synthesizer_version):
         try:
             current_v = int(current_v)
             synth_v = int(synth_v)
-            if current_v > synth_v:
+            if compare_operator(current_v, synth_v):
                 return False
 
-            if synth_v > current_v:
+            if compare_operator(synth_v, current_v):
```
amontanez24, line 313: what happens when the versions have .dev0 or r1 in them?

sdv/_utils.py
```diff
@@ -301,63 +307,81 @@ def _check_is_lower_version(current_version, synthesizer_version):
...
+    fit_public_version = getattr(synthesizer, '_fitted_sdv_version', None)
+    fit_enterprise_version = getattr(synthesizer, '_fitted_sdv_enterprise_version', None)
 
-        if is_public_lower:
-            raise VersionError(
-                f'You are currently on SDV version {current_public_version} but this '
-                f'synthesizer was created on version {fit_public_version}. {downgrade_message}'
-            )
+    is_public_lower = False
+    if None not in (current_public_version, fit_public_version):
```
amontanez24, line 350: why not have compare_versions return False if either one is None

sdv/_utils.py
pvk-developer: We do not compare that. If we have `dev0` vs`dev1` for example, it will fail and return `False` (because of the `try except`), I thought that we wanted something basic, not going that depth. If thats the case the best approach is to just use `packaging` which compares everything (dev0 to dev1, dev0 to r1 even rc1 and other nomenclatures).

sdv/_utils.py
pvk-developer: I think that by the `try except` it should return false actually.

frances-h (approved):

amontanez24 (approved):

PR #1889, opened 2024-04-04
Metadata isn't validating sdtypes in a column relationship (public SDV only)

### LINKED ISSUE #1781: Metadata isn't validating sdtypes in a column relationship (public SDV only)
### Environment Details

* SDV version: 1.9.0
* Python version: Any
* Operating System: Any

### Error Description
As a public SDV user, I want to be able to note down when a group of columns should be considered together as an address. According to [the documentation](https://docs.sdv.dev/sdv/reference/metadata-spec/single-table-metadata-json#column-relationships), this type of column relationship only supports a specific set of sdtypes:

> An address is defined by 2 or more columns that have the following sdtypes: `country_code`, `administrative_unit`, `state`, `state_abbr`, `city`, `postcode`, `street_address` and `secondary_address`

But in practice, I find that I'm able to provide it any sdtype and the metadata considers it to be valid. This is a problem because if I upgrade to the Enterprise SDK, then the same exact metadata is considered invalid. 

**Metadata is supposed to represent ground source-of-truth, so it cannot valid for the Public SDK and then invalid for the Enterprise SDK.**

### Steps to reproduce

```python

    'columns': {
        'user_city': { 'sdtype': 'city' },
        'user_zip': { 'sdtype': 'postcode' },
        'user_value': { 'sdtype': 'unknown'} # this is not allowed in an address relationship
    },
    'column_relationships': [{
        'type': 'address',
        'column_names': ['user_city', 'user_zip', 'user_value']
    }]
})

metadata.validate()
```

In public SDV it validates. In the enterprise version,  it produces this error:
```
InvalidMetadataError: The following errors were found in the metadata:

Column relationships have following errors:
Column 'user_value' has an unsupported sdtype 'unknown'.
Please provide a column that is compatible with address data.
```

This error should be expected in public SDV as well.

R-Palazzo:
CU-86az9j5en
Resolve #1781


sdv-team:
Task linked: [CU-86az9j5en SDV - Metadata isn't validating sdtypes in a column relationship (public SDV only) #1781](https://app.clickup.com/t/86az9j5en)

tests/test_tasks.py
```diff
@@ -25,12 +25,12 @@ def test_get_minimum_versions():
     # Assert
     expected_versions_39 = [
         'numpy==1.20.0',
-        'pandas @ git+https://github.com/pandas-dev/pandas.git@master#egg=pandas',
+        'git+https://github.com/pandas-dev/pandas.git@master#egg=pandas',
```
amontanez24, line 28: we're gonna need to update this everywhere

amontanez24 (approved):
This came together quite nicely imo

fealho (approved):

PR #1894, opened 2024-04-04
Add future warnings to all public functions of SingleTablePreset

### LINKED ISSUE #1855: Deprecate `SingleTablePreset` (including `FastML` Preset)
### Problem Description
Currently, we recommend the `SingleTablePreset` (with `'FAST_ML'` option) as the first quickstart synthesizer that a user should try. At the time we added this, the intention was to provide the easiest option for users to try creating synthetic data. But since then, there have been changes that make the `SingleTablePreset` more of a liability.

1. We've improved the `GaussianCopulaSynthesizer` so much that its speed is similar to `SingleTablePreset`
2. The `GaussianCopulaSynthesizer` (and other SDV synthesizers) are hooked up to other customizations (constraints, transformers, etc.) that are not possible with `SingleTablePreset`. So users have to move to these synthesizers anyway.
3. `SingleTablePreset` is a special wrapper on top of `GaussianCopulaSynthesizer` and as such, it costs us more to maintain and update it with features.
4. The original intention was to add other options (besides `FAST_ML`) to `SingleTablePreset` but upon further consideration, we would rather expose these as different synthesizers.

### Expected behavior
Deprecate `SingleTablePreset`. To be backwards compatible, we should still allow users to use the functionality without crashing. However, _any_ function within the `SingleTablePreset` class (including `__init__`) should yield a `FutureWarning` asking the user to switch to `GaussianCopulaSynthesizer` instead.

```
>>> synth = SingleTablePreset(metadata, name='FAST_ML')
Future Warning: The 'SingleTablePreset' is deprecated. For equivalent Fast ML functionality, please use the 'GaussianCopulaSynthesizer'.
``` 

### Additional context
We will also need to update our docs and Quickstart tutorial after this is updated.

lajohn4747:
resolve #1855 
CU_86azqyj6k

Add warnings indicating the deprecation of `SingleTablePreset`

### DIFF
--- a/sdv/lite/single_table.py
+++ b/sdv/lite/single_table.py
@@ -3,6 +3,7 @@
 import inspect
 import logging
 import sys
+import warnings
 
 import cloudpickle
 
@@ -14,6 +15,10 @@ FAST_ML_PRESET = 'FAST_ML'
 PRESETS = {
     FAST_ML_PRESET: 'Use this preset to minimize the time needed to create a synthetic data model.'
 }
+DEPRECATION_MSG = (
+    "The 'SingleTablePreset' is deprecated. For equivalent Fast ML "
+    "functionality, please use the 'GaussianCopulaSynthesizer'."
+)
 
 
 class SingleTablePreset:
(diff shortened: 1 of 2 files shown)


amontanez24 (commented):
This looks good! Can we just add integration tests to make sure warnings are showing

fealho (approved):

amontanez24 (approved):
LGTM!

PR #1897, opened 2024-04-05
Add warning for `datetimes` when they are missing `datetime_format` and are represented as `dtype` object

### LINKED ISSUE #1847: Add warning when user tries to train a model using datetime values without a datetime_format set
**Situation**

If you read data into a pandas dataframe, datetime columns often are (by default) set to the object (aka string) dtype. The user then must take some action to make SDV synthesize better datetime values:

- user can manually cast the column to datetime dtype
- user can set a `datetime_format` when creating the metadata for SDV to use

It's easy for someone (especially a new user) to skip this step entirely, causing issues later.

**Suggested Warning**

After discussing with Neha, I'm opening this feature request. Ideally, we can add a **warning** when the user tries to make progress with SDV (e.g. maybe training a synthesizer) that they should add a `datetime_format`.

pvk-developer:
Resolves #1847 
CU-86aznj7ak

### DIFF
--- a/sdv/metadata/multi_table.py
+++ b/sdv/metadata/multi_table.py
@@ -839,6 +839,14 @@ class MultiTableMetadata:
         Args:
             data (pd.DataFrame):
                 The data to validate.
+
+        Raises:
+            InvalidDataError:
+                This error is being raised if the data is not matching its sdtype requirements.
+
+        Warns:
+            A warning is being raised if ``datetime_format`` is missing from a column represented
+            as ``object`` in the dataframe and its sdtype is ``datetime``.
         """
         errors = []
         errors += self._validate_missing_tables(data)
--- a/sdv/metadata/multi_table.py
+++ b/sdv/metadata/multi_table.py
@@ -784,13 +784,14 @@ class MultiTableMetadata:
                     warning_dataframes.append(df)
 
         if warning_dataframes:
-            df = pd.concat(warning_dataframes)
-            warnings.warn(
+            warning_df = pd.concat(warning_dataframes)
+            warning_msg = (
                 "No 'datetime_format' is present in the metadata for the following columns:\n "
-                f'{df.to_string(index=False)}\n\n'
+                f'{warning_df.to_string(index=False)}\n'
                 'Without this specification, SDV may not be able to accurately parse the data. '
                 "We recommend adding datetime formats using 'update_column'."
             )
+            warnings.warn(warning_msg)
 
         return errors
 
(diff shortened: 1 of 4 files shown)


sdv-team:
Task linked: [CU-86aznj7ak SDV - Add warning when user tries to train a model using datetime values without a datetime_format set #1847](https://app.clickup.com/t/86aznj7ak)

sdv/metadata/single_table.py
```diff
@@ -1060,13 +1068,18 @@ def _validate_column_data(self, column):
...
+                sdtype_warnings['datetime_format'].append(datetime_format)
+
         if invalid_values:
             invalid_values = _format_invalid_values_string(invalid_values, 3)
             return [f"Invalid values found for {sdtype} column '{column.name}': {invalid_values}."]
 
         return []
 
-    def validate_data(self, data):
+    def validate_data(self, data, sdtype_warnings=None):
```
amontanez24, line 1082: update docstring

sdv/metadata/single_table.py
```diff
@@ -1030,8 +1031,15 @@ def _get_invalid_column_values(column, validation_function):
 
         return set(column[~valid])
 
-    def _validate_column_data(self, column):
-        """Validate values of the column satisfy its sdtype properties."""
+    def _validate_column_data(self, column, sdtype_warnings):
+        """Validate values of the column satisfy its sdtype properties.
```
amontanez24, line 1035: can we update the return type of this method? I think it's a little confusing because usually our validate methods raise errors but here it returns list with the error message. Similarly, we are appending warnings to a list instead of raising them and bubbling them up. This is fine but I think a description could help

sdv/metadata/multi_table.py
```diff
@@ -753,9 +755,12 @@ def _validate_missing_tables(self, data):
     def _validate_all_tables(self, data):
         """Validate every table of the data has a valid table/metadata pair."""
         errors = []
+        warning_dataframes = []
         for table_name, table_data in data.items():
+            table_sdtype_warnings = defaultdict(list)
             try:
-                self.tables[table_name].validate_data(table_data)
+                with warnings.catch_warnings(record=True):
```
amontanez24, line 762: since we're catching the warnings, why not access them directly instead of adding them to a list?
something like
```python
try:
    with warnings.catch_warnings(record=True) as caught_warnings:
        ...

finally:
    for warning in caught_warnings:
        message = warning.message
        do_what_you_want_with_message()
```
Or was it just easier to add to a list?

gsheni (approved):
lgtm

amontanez24 (approved):
:shipit:

PR #1905, opened 2024-04-09
Create unique id for each synthesizer

### LINKED ISSUE #1902: Create unique id for each synthesizer
### Problem Description

_As anyone debugging logs, it would be useful to know which operations were executed by the same synthesizer. To do this, we need a way of identifying synthesizers._

### Expected behavior

- Store a private attribute, `_synthesizer_id` on each synthesizer during its initialization.
- Make sure this attribute is stored and reloaded when saving and loading

pvk-developer:
Resolves #1902 
CU-86b00mxjz

This is how the `IDS` look :
image

I'm printing both `instance` and `id` just in case we wanted to change the `representation` to be the id of the synthesizer.

### DIFF
--- a/sdv/_utils.py
+++ b/sdv/_utils.py
@@ -1,5 +1,6 @@
 """Miscellaneous utility functions."""
 import operator
+import uuid
 import warnings
 from collections import defaultdict
 from collections.abc import Iterable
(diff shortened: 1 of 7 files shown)


sdv-team:
Task linked: [CU-86b00mxjz SDV - Create unique id for each synthesizer #1902](https://app.clickup.com/t/86b00mxjz)

frances-h (approved):
LGTM!

sdv/_utils.py
```diff
@@ -388,3 +389,23 @@ def _get_root_tables(relationships):
...
+        synthesizer (BaseSynthesizer or BaseMultiTableSynthesizer):
+            An SDV model instance to check versions against.
+
+    Returns:
+        ID:
+            A unique identifier for this synthesizer.
+    """
+    class_name = synthesizer.__class__.__name__
+    synth_version = version.public
+    unique_id = str(uuid.uuid4()).split('-')[-1]
```
amontanez24, line 410: If we take only a portion of it, it's not guaranteed to be unique anymore right?

pvk-developer: Yeah, I think that a portion is not guarateed to be unique and the odds of it repeating are small but there are. Should we use the entire UUID4 ?

amontanez24: Yeah I think we should

pvk-developer: Done in https://github.com/sdv-dev/SDV/pull/1905/commits/0e6f575ee35c9428606ad59ac82fab8ba28425ca

PR #1908, opened 2024-04-10
Automated Latest Dependency Updates

sdv-team:
This is an auto-generated PR with **latest** dependency updates.

### DIFF
--- a/latest_requirements.txt
+++ b/latest_requirements.txt
@@ -1,5 +1,5 @@
-boto3==1.34.79
-botocore==1.34.79
+boto3==1.34.81
+botocore==1.34.81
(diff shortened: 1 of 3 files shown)


latest_requirements.txt
```diff
@@ -1,5 +1,5 @@
-boto3==1.34.80
-botocore==1.34.80
+boto3==1.34.81
+botocore==1.34.81
```
amontanez24, line 2: should we remove boto from the list of libraries we're auto checking for and just cap at the next major version?

gsheni: yah, they have very frequent releases

amontanez24 (approved):

PR #1920, opened 2024-04-17
Add usage logging

### LINKED ISSUE #1903: Add usage logging
### Problem Description

It would be helpful for debugging and overall library management to have a better understanding of SDV usage. We want to log more information about how synthesizers are used.

### Expected behavior

For each of the following events, create one INFO level log statement with what is required

#### Synthesizer Initialization
- Timestamp
- Synthesizer class name (GaussianCopula, CTGAN, HSA, etc.)
- [Synthesizer id](https://github.com/sdv-dev/SDV/issues/1902)

#### Fit
- Timestamp
- Synthesizer class name (GaussianCopula, CTGAN, HSA, etc.)
- Statistics of the fit data:
    - Total # of tables
    - Total # of rows
    - Total # of columns
- Synthesizer Id

#### Sample
- Timestamp
- Synthesizer class name (GaussianCopula, CTGAN, HSA, etc.)
- Statistics of the sample size:
    - Total # of tables
    - Total # of rows
    - Total # of columns
- Synthesizer Id

#### Synthesizer save and load
- Timestamp
- Synthesizer class name (GaussianCopula, CTGAN, HSA, etc.)
- [Synthesizer id](https://github.com/sdv-dev/SDV/issues/1902)

#### Metadata save
- Timestamp
- Metadata type (single or multi table)
- Statistics about the metadata
    - Total # of tables
    - Total # of columns
    - Total # of relationships

### Additional context

We should use a different logger than the loggers we currently use to record this information. 
#### Proposal
- Create a yaml file that defines the configuration for the logger that should be used for usage. Whenever we need the logger, we can load that yaml file and [configure the logger](https://docs.python.org/3/howto/logging.html#configuring-logging) from a [dictionary](https://docs.python.org/3/library/logging.config.html#logging.config.dictConfig).
- Store this yaml in sdv/logging
- Add a variable to the logger for how the logs should be stored (valid values will be None or 'local' for now). If set to None, we should not log this info.
Example yaml
```yaml
log_style: 'local'
version: 1
handlers:
  file:
    class: logging.FileHandler
    filename: sdv_logs.log
loggers:
  BasSynthesizer:
    level: INFO
    handlers: [file]
    propagate: no
```

pvk-developer:
Resolves #1903 
CU-86b00n5d2


sdv/logging/utils.py
```diff
@@ -0,0 +1,91 @@
...
+    """Return a dictionary with the logging configuration."""
+    logging_path = Path(__file__).parent
+    with open(logging_path / 'sdv_logger_config.yml', 'r') as f:
+        logger_conf = yaml.safe_load(f)
+
+    # Logfile to be in this same directory
+    for logger in logger_conf.get('loggers', {}).values():
+        handler = logger.get('handlers', {})
+        if handler.get('filename') == 'sdv_logs.log':
+            handler['filename'] = logging_path / handler['filename']
```
amontanez24, line 22: you had ideas for an alternate place to store right? @pvk-developer

sdv/multi_table/base.py
```diff
@@ -56,13 +59,14 @@ def _set_temp_numpy_seed(self):
...
 
     def _initialize_models(self):
-        for table_name, table_metadata in self.metadata.tables.items():
-            synthesizer_parameters = self._table_parameters.get(table_name, {})
-            self._table_synthesizers[table_name] = self._synthesizer(
-                metadata=table_metadata,
-                locales=self.locales,
-                **synthesizer_parameters
-            )
+        with disable_single_table_logger():
```
amontanez24, line 62: I'm wondering if it would be more straightforward to have the single table synthesizer log conditionally ie.
```python
if synthesizer.is_part_of_multitable: # we can have a better name
    SYNTHESIZER_LOGGER.info(...)
```
To do this we'd need to have some attribute that tracks if the synthesizer was made as part of a multitable synthesizer, but I think that could be useful information. @frances-h @pvk-developer What do you think?

sdv/logging/utils.py
```diff
@@ -0,0 +1,91 @@
...
+    Returns:
+        logging.Logger:
+            A logger instance configured according to the logging configuration
+            and the specific settings for the given logger name.
+    """
+    logger_conf = get_sdv_logger_config()
+    logger = logging.getLogger(logger_name)
+    if logger_name in logger_conf.get('loggers'):
+        formatter = None
+        config = logger_conf.get('loggers').get(logger_name)
```
amontanez24, line 68: if the config sets the log_registry to None, we should disable all the loggers here (maybe just give them no handlers or something)

amontanez24 (commented):
This looks great! The one thing I think we need still is the ability to completely disable the loggers from the config

sdv/multi_table/base.py
pvk-developer: I agree. I can add the `table_name` for the `synthesizer` and add it to the `data_processor` as well. Then the logic would be something like 
```python
if not synthesizer.table_name:
    # log the message

File: sdv/logging/utils.py (hunk 1 above)
Comment (pvk-developer) [reply sdv/logging/utils.py:22]: Yes, my idea was to add `platformdirs` and then store it locally there (it would be similar to where we stored credentials on `SDV-Installer`).
```python
    store_path = user_app_path = Path(platformdirs.user_data_dir('sdv', 'sdv-dev'))
```

amontanez24: I think this would be fine

frances-h (commented):
Could we add an integration test as well? Maybe just checking that the file exists and that logs from the run are present at the end of the file?

pyproject.toml
```diff
@@ -38,6 +38,7 @@ dependencies = [
     'deepecho>=0.6.0',
     'rdt>=1.12.0',
     'sdmetrics>=0.14.0',
+    'platformdirs>=4.0'
```
amontanez24, line 41: should we add this to the dependency checker since you're not capping it?

sdv/single_table/base.py
```diff
@@ -107,6 +112,16 @@ def __init__(self, metadata, enforce_min_max_values=True, enforce_rounding=True,
         self._fitted_sdv_version = None
         self._fitted_sdv_enterprise_version = None
         self._synthesizer_id = generate_synthesizer_id(self)
+        if self.table_name is None:
```
amontanez24, line 115: My only fear is if somehow there's another way the table name ends up none. As long as that's impossible, this seems good

amontanez24 (commented):
looks great! A couple small comments

sdv/single_table/base.py
pvk-developer: Well we need to make sure that multi table synthesizers flag this correctly. As you can see on the changes done to `HMASynthesizer`, we should do the same on `HSASynthesizer` and others.

pyproject.toml
pvk-developer: Yeah, let me add it!

sdv/single_table/base.py
pvk-developer: After discussing adding a `flag` that is not in the `__init__` is to over complicate the current creation of the synthesizers and we are rolling back to the previous solution with the `disable_single_table_logging`.

frances-h (approved):

lajohn4747 (approved):

PR #1928, opened 2024-04-18
Add `get_random_subset` poc utility function

### LINKED ISSUE #1877: Add `get_random_subset` poc utility function
### Problem Description
We should create a utility function that will allow users to coherently subsample a very large dataset so that it can be used with HMA.

### Expected behavior
Add a new function, `get_random_subset` to the `utils.poc` module.
```
>>> from sdv.utils import poc

>>> small_dataset = poc.get_random_subset(
            data,
            metadata,
            main_table_name='transactions',
            num_rows=1000,
            verbose=True
)
Success! Your subset has 90% less rows than the original.

Table Name    # Rows (Original)    # Rows (Subset)
sessions      1200                 120            
transactions  5000                 200             

```

#### Parameters
- `data [dict]` - The data dictionary
- `metadata [MultiTableMetadata]` - The metadata for the data
- `main_table_name [str]` - The main table to consider when subsampling
- `num_rows [int]` - The number of rows to subsample from the main table
- `verbose [bool, optional]` - Whether to print a summary of the results of subsampling. Defaults to True. 

#### Returns
- The data dictionary containing the subsampled tables
- If `verbose` is True, it should also print a summary of what the function did:
  - The total percentage of rows that were dropped (i.e. total number of rows in the subsampled data / total number of rows in the original data)
  - For each table, the original number of rows in the table and the number of rows in the subsampled table

#### Algorithm Overview
- [For disconnected schemas, which we don't currently support but may in the future]
  - Calculate ratio of `num_rows` to the original table size for the main table
  - For every root table that is disconnected root from the main table:
    - subsample the root table using ratio found above
- Randomly sample `num_rows` rows from the main table
- If the main table has any parents, for each parent:
  - If all parent rows were referenced in the original main table, drop all parent rows that are no longer referenced by the subsampled main table
  - If there were parent rows that were not referenced (aka childless parent rows) in the original main table, drop any rows that had a reference and are now no longer referenced. Determine the percentage of referenced rows that were dropped, and randomly drop the same percentage of the originally unreferenced parent rows
  - Repeat this process for grandparents, great-grandparents, etc. Note that if we have e.g. a diamond shaped relationship (main table has 2 parents that each share the same parent), we want to be keep all rows in the grandparent that are referenced by either parent.
- Use drop unknown references to enforce referential integrity and drop rows from the descendant tables. Note that this should not change the size of the main table since we only drop unreferenced rows from the parent tables. 
- Perform validation:
  - If any subsampled table has no rows, raise an error and suggest re-trying or increasing the `num_rows` parameter
      - This could happen if a parent is aggressively subsampled, causing `drop_unknown_references` to wipe out a child
      - Since we are randomly sampling, re-trying may give a better result 
- If verbose, print how the results of data was subsampled:
  - Percentage of rows dropped (total number of subsampled rows / original total number of rows)
  - For each table, print original number of rows vs subsampled number of rows

R-Palazzo:
CU-86azvqpqe
Resolve #1877

A few considerations regarding this PR:
1 - `NaNs handling`: Currently, I don't drop NaN `foreign keys`
2 - `Randomness`: For reproducibility, I set a seed, is it fine?
3 - `Index`: Should we reset the index of the tables at the end after all the dropping is done? This is also a question for `drop_unknown_references`.
4 - `Verbosity`: To be consistent with the other `POC methods`, I added a verbose parameter to `simplify_schema()`
5 - `Disconnected schema`: Subsampling disconnected schema should work with `get_random_subset`. I wrote down a test where I mocked the metadata validation since disconnected schemas are not supported there.

Thanks for your review and your thoughts on this ;)


sdv-team:
Task linked: [CU-86azvqpqe SDV - Add `get_random_subset` poc utility function #1877](https://app.clickup.com/t/86azvqpqe)

sdv/multi_table/utils.py
```diff
@@ -7,10 +7,18 @@
 import pandas as pd
 
 from sdv._utils import _get_root_tables
+from sdv.errors import InvalidDataError, SamplingError
 from sdv.multi_table import HMASynthesizer
 from sdv.multi_table.hma import MAX_NUMBER_OF_COLUMNS
 
 MODELABLE_SDTYPE = ['categorical', 'numerical', 'datetime', 'boolean']
+RANDON_STATE = 42
```
gsheni, line 15:
```suggestion
RANDOM_STATE = 42
```

sdv/multi_table/utils.py
```diff
@@ -7,10 +7,18 @@
...
 from sdv._utils import _get_root_tables
+from sdv.errors import InvalidDataError, SamplingError
 from sdv.multi_table import HMASynthesizer
 from sdv.multi_table.hma import MAX_NUMBER_OF_COLUMNS
 
 MODELABLE_SDTYPE = ['categorical', 'numerical', 'datetime', 'boolean']
+RANDOM_STATE = 42
+
+
+def _get_child_tables(relationships):
```
gsheni, line 18: Since we are modifying this function, can we take this chance to centralize this function? It seems there are similar functions in 3 places:
- https://github.com/sdv-dev/SDV/blob/1be52f6af74670fef2fe93ef047875def574036a/sdv/multi_table/base.py#L117-L122
- https://github.com/sdv-dev/SDV/blob/447d0d0f5b85b3fae3a0fc7a737915817e8742e0/sdv/_utils.py#L363-L366

sdv/multi_table/utils.py
```diff
@@ -392,3 +428,203 @@ def _get_rows_to_drop(metadata, data):
...
+
+        _subsample_ancestors(data, metadata, parent, primary_keys_referenced)
+
+
+def _subsample_data(data, metadata, main_table_name, num_rows):
+    """Subsample multi-table table based on a table and a number of rows.
+
+    The strategy is to:
+    - Subsample the disconnected roots tables by keeping a similar proportion of data
+      than the maint table. Ensure referential integrity.
```
frances-h, line 578: minor typo: maint -> main

sdv/utils/poc.py
```diff
@@ -127,6 +116,71 @@ def simplify_schema(data, metadata):
...
+    - Subsample the main table and its descendants to ensure referential integrity.
+    - Subsample the ancestors of the main table by removing primary key rows that are no longer
+      referenced by the descendants and drop also some unreferenced rows.
+
+    Args:
+        metadata (MultiTableMetadata):
+            Metadata of the datasets.
+        data (dict):
+            Dictionary that maps each table name (string) to the data for that
+            table (pandas.DataFrame).
```
frances-h, line 140: Arg order should be swapped

sdv/multi_table/utils.py
R-Palazzo: Yes, done in [43d4441](https://github.com/sdv-dev/SDV/pull/1928/commits/43d4441ccbd89795a56df13d9b1ff99b2465292f)

frances-h:
@R-Palazzo just to respond to some of your considerations:
> 1 - NaNs handling: Currently, I don't drop NaN foreign keys

I think it might make more sense to treat NaNs similarly to how we do with `drop_unknown_references` and add a flag to indicate if we should drop them or not. As an added consideration, if we _don't_ drop them, maybe should we update the functionality to drop a proportional number of NaN foreign keys to keep the table balanced?

> 2 - Randomness: For reproducibility, I set a seed, is it fine?

I'm actually not sure we want a fixed seed here, since it might be helpful to re-run the function to get a different subsample. Maybe instead we could control the randomness in a similar way to how we control it when we sample from synthesizers?

> 3 - Index: Should we reset the index of the tables at the end after all the dropping is done? This is also a question for drop_unknown_references.

In my opinion, I think we can leave the index as-is. We don't use the index, and users might want to be able to compare the subsampled tables back to their original data.

> 4 - Verbosity: To be consistent with the other POC methods, I added a verbose parameter to simplify_schema()

Nice, I think this works well :)

sdv/multi_table/utils.py
```diff
@@ -392,3 +428,203 @@ def _get_rows_to_drop(metadata, data):
...
+    The strategy here is to:
+    - Drop the rows that are no longer referenced by the descendants.
+    - Drop a proportional amount of never-referenced rows.
+
+    Args:
+        parent_table (pandas.DataFrame):
+            Parent table to subsample.
+        parent_primary_key (str):
+            Name of the primary key of the parent table.
+        pk_referenced_before_parent (set):
```
amontanez24, line 505: minor: could we move parent to the front? ie. `parent_pks_referenced_before`.

sdv/multi_table/utils.py
```diff
@@ -7,10 +7,18 @@
 import pandas as pd
 
 from sdv._utils import _get_root_tables
+from sdv.errors import InvalidDataError, SamplingError
 from sdv.multi_table import HMASynthesizer
 from sdv.multi_table.hma import MAX_NUMBER_OF_COLUMNS
 
 MODELABLE_SDTYPE = ['categorical', 'numerical', 'datetime', 'boolean']
+RANDOM_STATE = 42
```
amontanez24, line 15: Not sure if we should control randomness for this

amontanez24 (commented):
Should we try to align this more with the strategy used in the database subsampling? They seem pretty different at the moment

tests/unit/multi_table/test_utils.py
```diff
@@ -1331,16 +1372,44 @@ def test__subsample_table_and_descendants(mock_drop_rows):
...
+    mock_get_idx_to_drop_nan_foreign_key_table.assert_called_once_with(
+        data, metadata.relationships, 'parent'
+    )
+    mock_drop_rows.assert_called_once_with(data, metadata, drop_missing_values=True)
     assert len(data['parent']) == 3
 
 
+@patch('sdv.multi_table.utils._get_idx_to_drop_nan_foreign_key_table')
+def test__subsample_table_and_descendants_nan_fk(mock_get_idx_to_drop_nan_foreign_key_table):
+    """Test the ``_subsample_table_and_descendants`` when there is too many NaN foreign keys."""
```
amontanez24, line 1392:
```suggestion
    """Test the ``_subsample_table_and_descendants`` when there are too many NaN foreign keys."""
```

sdv/multi_table/utils.py
```diff
@@ -454,13 +468,37 @@ def _subsample_disconnected_roots(data, metadata, table, ratio_to_keep):
...
 
-    _drop_rows(data, metadata, drop_missing_values=False)
+    _drop_rows(data, metadata, drop_missing_values=True)
 
 
 def _subsample_table_and_descendants(data, metadata, table, num_rows):
-    """Subsample the table and its descendants."""
+    """Subsample the table and its descendants.
+
+    The logic is to first subsample all the NaN foreign key of the table.
```
amontanez24, line 477:
```suggestion
    The logic is to first subsample all the NaN foreign keys of the table.
```

sdv/multi_table/utils.py
```diff
@@ -429,16 +430,29 @@ def _get_rows_to_drop(data, metadata):
     return table_to_idx_to_drop
 
 
+def _get_idx_to_drop_nan_foreign_key_table(data, relationships, table):
```
amontanez24, line 433: minor: I'd rename this to something like `_get_nan_fk_indices_table` or something like that. The indexes aren't dropped here and if this method is useful elsewhere it might not be to ddrop the indices so I'd leave that part out

amontanez24 (approved):

gsheni (approved):

frances-h (approved):

PR #1929, opened 2024-04-18
Improve addon detection to include modules

### LINKED ISSUE #1923: Add io module stub
### Problem Description

_As a user, it would be helpful to have tools that aid in the importing of my data from different sources._

### Expected behavior

- Add an io module

### LINKED ISSUE #1924: Add add-on modules to sys.modules
### Problem Description

_As a user, it would be useful if importing for premium features worked the same as for public features._

### Expected behavior

For some premium features, we want to add modules that don't exist and be able to import from them.
```python
from sdv.new_module import NewClass
```
Currently this is not supported. To fix this, we need to add the new modules to sys.modules, since that is the dict python uses for imports.

amontanez24:
resolves #1924 
resolves #1923


sdv/io/__init__.py
```diff
@@ -0,0 +1 @@
+"""I/O module."""
```
amontanez24, line 1: This is needed to add an entry point for the connectors

pvk-developer (approved):

frances-h (approved):
LGTM!

PR #1931, opened 2024-04-18
When generating ids without a regex, create them randomly

### LINKED ISSUE #1922: When generating ids without a regex, create them randomly
### Problem Description
For any column that has sdtype id **without** a user-provided regex, the SDV currently generates generates index values in a sequential manner (eg, `0`, `1`, `2`, ...) The resulting data doesn't look realistic.

### Expected behavior
For any columns of sdtype `id` that **do not** have a user-provided regex, ensure that the synthetic data is created randomly. 

In technical terms: We currently generate sequential values by assigning the `IDGenerator` RDT to these columns. Instead, we should assign the `AnonymizedFaker` to those columns and use the `bothify` function to make random strings. The exact params to `bothify` depend on the dtype (storage type) of the data.

- If the column is numeric (int, float, etc), we need to ensure that the resulting synthetic values can be cast back to this dtype. In this case, assign the following transformer: `AnonymizedFaker(provider_name=None, function_name='bothify', function_kwargs={'text': '##########'})`
  - This allows for 1 billion possible values
  - When cast back to numbers, they will be completely randomized
  - *If* it's a primary key, then also set `cardinality_rule='unique'`
- Otherwise, the synthetic values can remain as a string (object) type. In this case, assign the following transformer: `AnonymizedFaker(provider_name=None, function_name='bothify', function_kwargs={'text': '??????'})
  - This allows for well over 1 billion possible values
  - *If* it's a primary key, then also set `cardinality_rule='unique'`

### Additional context
- This change only applies if a column is sdtype 'id' AND there is *no* 'regex_format' available.

frances-h:
Resolve #1922
CU-86b047n3r

### DIFF
--- a/sdv/data_processing/data_processor.py
+++ b/sdv/data_processing/data_processor.py
@@ -9,7 +9,7 @@ from pathlib import Path
 import pandas as pd
 import rdt
 from pandas.api.types import is_float_dtype, is_integer_dtype
-from rdt.transformers import AnonymizedFaker, IDGenerator, get_default_transformers
+from rdt.transformers import AnonymizedFaker, get_default_transformers
 from rdt.transformers.pii.anonymization import get_anonymized_transformer
 
 from sdv.constraints import Constraint
@@ -567,21 +567,22 @@ class DataProcessor:
                     )
                     sdtypes[column] = 'text'
 
-                elif column in self._keys:
-                    prefix = None
-                    if not is_numeric:
-                        prefix = 'sdv-id-'
-
-                    transformers[column] = IDGenerator(prefix=prefix)
-                    sdtypes[column] = 'text'
-
                 else:
+                    bothify_format = 'sdv-id-??????'
+                    if is_numeric:
+                        bothify_format = '##########'
+
+                    cardinality_rule = None
+                    if column in self._keys:
+                        cardinality_rule = 'unique'
+
                     transformers[column] = AnonymizedFaker(
                         provider_name=None,
                         function_name='bothify',
-                        function_kwargs={'text': '#####'}
+                        function_kwargs={'text': bothify_format},
+                        cardinality_rule=cardinality_rule
                     )
-                    sdtypes[column] = 'pii'
+                    sdtypes[column] = 'text'
 
             elif sdtype == 'unknown':
                 sdtypes[column] = 'pii'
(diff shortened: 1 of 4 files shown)


sdv-team:
Task linked: [CU-86b047n3r SDV - When generating ids without a regex, create them randomly #1922](https://app.clickup.com/t/86b047n3r)

tests/integration/single_table/test_copulas.py
```diff
@@ -365,14 +365,11 @@ def test_numerical_columns_gets_pii():
...
     # Setup
     data = pd.DataFrame(data={
-        'id': [0, 1, 2, 3, 4],
         'city': [0, 0, 0, 0, 0],
         'numerical': [21, 22, 23, 24, 25]
     })
     metadata = SingleTableMetadata.load_from_dict({
-        'primary_key': 'id',
         'columns': {
-            'id': {'sdtype': 'id'},
```
amontanez24, line 375: why are we removing these?

amontanez24 (commented):
Just one question, otherwise looks good!

tests/integration/single_table/test_copulas.py
frances-h: I removed it since it'll just be randomly generated numbers and I couldn't think of a good way to add an `assert` around them. I could change them to non-numerical and check the `sdv-id-` format instead?

amontanez24: Ah I see. Yeah that's fine. I thought we controlled faker's randomness with a seed though

frances-h: Ah yeah, I can add it back then if we want to keep the ids

amontanez24: Maybe we should so that we know if the randomness isn't being controlled

sdv/data_processing/data_processor.py
```diff
@@ -567,21 +567,23 @@ def _create_config(self, data, columns_created_by_constraints):
                     )
                     sdtypes[column] = 'text'
 
-                elif column in self._keys:
-                    prefix = None
-                    if not is_numeric:
-                        prefix = 'sdv-id-'
+                else:
+                    bothify_format = 'sdv-id-??????'
```
pvk-developer, line 571: This will only generate `a-zA-Z` characters. If we want to use `0-9a-zAZ` we could pass `string.ascii_letters + string.digits` as function_kwargs.
Here is an example:
image

amontanez24: After speaking with @npatki I think we can leave as is

amontanez24 (approved):

pvk-developer (approved):

PR #1932, opened 2024-04-18
If using regex to generate values, scramble them

### LINKED ISSUE #1921: If using regex to generate values, scramble them
### Problem Description
For any column that has sdtype `id` with a provided regex, the SDV currently generates the regex in sequential (alphanumeric) order. The resulting data doesn't look realistic.

### Expected behavior
For any `id` columns with a regex, ensure that the synthetic data does not have values in sequential order. A simple way to do this is to scramble them.

In technical terms: We assign RDT's RegexGenerator to accomplish this. By default, we should assign the RegexGenerator with `generation_order='scrambled'` to these columns.

### Additional context
- This change only applies if a column is sdtype `'id'` AND there is a user-provided `'regex_format'`. If there is no regex provided in the metadata, then no changes will be made for this issue
- This issue depends on RDT changes to add the `generation_order` parameter. See [RDT issue #800](https://github.com/sdv-dev/RDT/issues/800)

amontanez24:
resolves #1921 
CU-86b047ax2

### DIFF
--- a/tests/integration/single_table/test_base.py
+++ b/tests/integration/single_table/test_base.py
@@ -226,6 +226,26 @@ def test_sample_remaining_columns_with_some_nans():
         synthesizer.sample_remaining_columns(known_columns=known_columns)
 
 
+def test_sample_keys_are_scrambled():
+    """Test that the keys are scrambled in the sampled data."""
+    # Setup
+    data, metadata = download_demo(
+        modality='single_table',
+        dataset_name='fake_hotel_guests'
+    )
+    metadata.update_column('guest_email', sdtype='id', regex_format='[A-Z]{3}')
+    synthesizer = GaussianCopulaSynthesizer(metadata)
+    synthesizer.fit(data)
+
+    # Run
+    sampled = synthesizer.sample(1000)
+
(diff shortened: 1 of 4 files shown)


sdv-team:
Task linked: [CU-86b047ax2 SDV - If using regex to generate values, scramble them #1921](https://app.clickup.com/t/86b047ax2)

tests/integration/single_table/test_base.py
```diff
@@ -267,9 +287,19 @@ def test_multiple_fits():
 @pytest.mark.parametrize('synthesizer', SYNTHESIZERS)
 def test_sampling(synthesizer):
     """Test that samples are different when ``reset_sampling`` is not called."""
+    # Setup
+    data = pd.DataFrame({
+        'column1': list(range(100)),
+        'column2': list(range(100)),
+        'column3': list(range(100))
+    })
+    synthesizer.fit(data)
```
amontanez24, line 296: this was crashing if you run the test on its own

pvk-developer (approved):

frances-h (approved):

PR #1957, opened 2024-04-23
 Only run unit and integration tests on oldest and latest python versions for macos

### LINKED ISSUE #1948: Only run unit and integration tests on oldest and latest python versions for macos
### Problem Description

_As an engineer, I'd like to see the results of CI workflows quickly so that I cna make changes or merge PRs more efficiently._

Github only allows 5 concurrent mac machines, while it allows 20 total machines. For this reason, our CI builds often get bottlenecked at the mac builds.

### Expected behavior

- Only run the unit and integration workflows on python 3.8 and 3.12 for Macos

frances-h:
CU-86b05rg34
Resolve #1948

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -12,6 +12,13 @@ jobs:
       matrix:
         python-version: [ '3.8', '3.9', '3.10', '3.11', '3.12']
         os: [ubuntu-latest, macos-latest, windows-latest]
+        exclude:
+          - os: macos-latest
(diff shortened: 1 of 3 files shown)


fealho (commented):
Do we want to exclude the minimum unit tests as well?

rwedge:
I think removing `macos` from the matrix and adding the mac configurations via `include` would be easier to follow

fealho (approved):

rwedge (approved):

PR #1962, opened 2024-04-25
Add ExcelHandler

### LINKED ISSUE #1950: Add ExcelHandler
### Problem Description

_As a user, I'd like an streamlined way to load my data and metadata from files so that I can get right to using SDV._

### Expected behavior

- In the `sdv.io` subpackage, add a folder called `local`
- In that folder add a class called `ExcelHandler`

### \_\_init\_\_
**Parameters**
```python
from sdv.io.local import ExcelHandlder

handler = ExcelHandler() 
```

### read
**Functionality** 

Internally, reading should use the [read_excel](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_excel.html) function from pandas. A few things should be hardcoded by default
- Pandas _should not_ detect an index column from the data
- Pandas _should not_ try to infer datetime formats (or cast them to np.datetime objects). Any datetime column should be left as a dtype 'object'
- After reading the data, we should use it to infer a MultiTableMetadata object. (Even if there is only 1 table, we should still create a MultiTableMetadata object.)

**Parameters**
- (required) file_path: A string describing the path of the Excel file to read
- sheet_name: A list of strings denoting which sheets in the Excel file to read from
    - (default) None: Read all the sheets in the file
    - list(str): Read only the sheets listed

**Returns**
- data: A dictionary mapping each table name to a pandas DataFrame with the data. The table name is the same as the sheet name
- metadata: A `MultiTableMetadata` object that describes the data

### write
**Functionality**
Internally, writing should use the [to_excel](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_excel.html) function from pandas. A few things should be hardcoded by default
- Do not write the index column
- Each table of the synthetic data should be written as a new sheet within the file. The name of the sheet should be the same as the name of the table
- If a sheet already exists with the same name, completely override it 

**Parameters**
- (required) synthetic_data: A dictionary that maps each table name to a pandas.DataFrame containing data from it
- (required) file_name: The name of the excel file to write
- sheet_name_suffix: A string with a suffix to add to each sheet name
    - (default) None: The name of the table should be the name of the sheet
    - (str) Append this string as the suffix. Eg. suffix of "_synthetic" will make sheets with "TABLENAME_synthetic"

- mode: A string signaling which mode of writing to use
    - (default) 'w': Write sheets to a new file, clearing any existing file that may exist
    - 'a': Append new sheets within the existing file. Note: You cannot append data to existing sheets.

### Additional context

- We will add a number of local file handlers for different file types. Therefore the implementation of this class should also add a base class. 
- Optionally, the init, read and write functions can include a subset of arguments that the corresponding pandas functions use
    - if both the read and write for pandas are the same for a parameter (eg. [decimal](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_csv.html)), then put it in the init. 
    - We can ignore most of these parameters. Only add ones that seem impactful

pvk-developer:
Resolves #1950
CU-86b05t4z0

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,11 +52,9 @@ dependencies = [
 sdv = { main = 'sdv.cli.__main__:main' }
 
 [project.optional-dependencies]
-excel_handler = [
-    'pandas[excel]'
-]
+excel = ['pandas[excel]']
 test = [
-    'sdv[excel_handler]',
+    'sdv[excel]',
     'pytest>=3.4.2',
     'pytest-cov>=2.6.0',
     'pytest-rerunfailures>=10.3,<15',
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,9 @@ dependencies = [
     'deepecho>=0.6.0',
     'rdt>=1.12.0',
     'sdmetrics>=0.14.0',
-    'platformdirs>=4.0'
+    'platformdirs>=4.0',
+    'xlsxwriter>=3.1.0',
+    'openpyxl>=3.1.0'
 ]
 
 [project.urls]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,8 +39,6 @@ dependencies = [
     'rdt>=1.12.0',
     'sdmetrics>=0.14.0',
     'platformdirs>=4.0',
-    'xlsxwriter>=3.1.0',
-    'openpyxl>=3.1.0'
 ]
 
 [project.urls]
@@ -54,7 +52,11 @@ dependencies = [
 sdv = { main = 'sdv.cli.__main__:main' }
 
 [project.optional-dependencies]
+excel_handler = [
+    'pandas[excel]'
+]
 test = [
+    'sdv[excel_handler]',
     'pytest>=3.4.2',
     'pytest-cov>=2.6.0',
     'pytest-rerunfailures>=10.3,<15',
--- a/sdv/io/local/local.py
+++ b/sdv/io/local/local.py
@@ -264,7 +264,14 @@ class ExcelHandler(BaseLocalHandler):
                     sheet_name = f'{table_name}{sheet_name_suffix}'
                     suffix_added = True
 
-                temp_data[sheet_name] = table
+                if temp_data.get(sheet_name):
+                    temp_data[sheet_name] = pd.concat([
+                        temp_data[sheet_name],
+                        synthetic_data[sheet_name]
+                    ])
+
+                else:
+                    temp_data[sheet_name] = table
 
         writer = pd.ExcelWriter(file_name)
         for table_name, table_data in temp_data.items():
--- a/sdv/io/local/local.py
+++ b/sdv/io/local/local.py
@@ -265,10 +265,10 @@ class ExcelHandler(BaseLocalHandler):
                     suffix_added = True
 
                 if temp_data.get(sheet_name) is not None:
-                    temp_data[sheet_name] = pd.concat([
-                        temp_data[sheet_name],
-                        synthetic_data[sheet_name]
-                    ])
+                    temp_data[sheet_name] = pd.concat(
+                        [temp_data[sheet_name], synthetic_data[sheet_name]],
+                        ignore_index=True
+                    )
 
                 else:
                     temp_data[sheet_name] = table
--- a/sdv/io/local/local.py
+++ b/sdv/io/local/local.py
@@ -192,3 +192,71 @@ class CSVHandler(BaseLocalHandler):
                 quoting=self.quoting,
                 mode=mode,
             )
+
+
+class ExcelHandler(BaseLocalHandler):
+    """A class for handling Excel files."""
+
+    def read(self, file_path, sheet_names=None):
+        """Read data from Excel files and returns it along with metadata.
+
+        Args:
+            file_path (str):
+                The path to the Excel file to read.
+            sheet_names (list of str, optional):
+                The names of sheets to read. If None, all sheets are read.
+
+        Returns:
+            tuple:
+                A tuple containing the data as a dictionary and metadata. The dictionary maps
+                table names to pandas DataFrames. The metadata is an object describing the data.
+        """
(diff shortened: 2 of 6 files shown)


sdv-team:
Task linked: [CU-86b05t4z0 SDV - Add ExcelHandler #1950](https://app.clickup.com/t/86b05t4z0)

tests/unit/io/local/test_local.py
```diff
@@ -210,3 +210,140 @@ def test_write_file_exists_mode_is_w(self, tmpdir):
...
+        instance = ExcelHandler()
+
+        # Run and Assert
+        error_msg = "'sheet_names' must be None or a list of strings."
+        with pytest.raises(ValueError, match=error_msg):
+            instance.read(file_path, sheet_names)
+
+    @patch('sdv.io.local.local.pd')
+    def test_write(self, mock_pd):
+        # Setup
```
R-Palazzo, line 324: Add test docstring

sdv/io/local/local.py
```diff
@@ -192,3 +192,71 @@ def write(self, synthetic_data, folder_name, file_name_suffix=None, mode='x'):
...
+                parse_dates=False,
+                decimal=self.decimal,
+                index_col=None
+            )
+
+        metadata = self._infer_metadata(data)
+        return data, metadata
+
+    def write(self, synthetic_data, file_name, sheet_name_suffix=None, mode='w'):
+        """Write synthetic data to an Excel File.
```
R-Palazzo, line 236: So in this case the `mode` parameter is useless?

sdv/io/local/local.py
```diff
@@ -192,3 +192,71 @@ def write(self, synthetic_data, folder_name, file_name_suffix=None, mode='x'):
                 quoting=self.quoting,
                 mode=mode,
             )
+
+
+class ExcelHandler(BaseLocalHandler):
+    """A class for handling Excel files."""
+
+    def read(self, file_path, sheet_names=None):
+        """Read data from Excel files and returns it along with metadata.
```
R-Palazzo, line 201: return

sdv/io/local/local.py
```diff
@@ -192,3 +192,71 @@ def write(self, synthetic_data, folder_name, file_name_suffix=None, mode='x'):
...
+        metadata = MultiTableMetadata()
+        if sheet_names is not None and not isinstance(sheet_names, list):
+            raise ValueError("'sheet_names' must be None or a list of strings.")
+
+        if sheet_names is None:
+            xl_file = pd.ExcelFile(file_path)
+            sheet_names = xl_file.sheet_names
+
+        for sheet_name in sheet_names:
+            data[sheet_name] = pd.read_excel(
```
R-Palazzo, line 224: Should we add a check that the `sheet_name` exists in the `file_path` here? Or it's ok to let pandas raise an error

R-Palazzo (commented):
Looks good!

Just 1 or 2 questions.

pyproject.toml
```diff
@@ -38,7 +38,9 @@ dependencies = [
     'deepecho>=0.6.0',
     'rdt>=1.12.0',
     'sdmetrics>=0.14.0',
-    'platformdirs>=4.0'
+    'platformdirs>=4.0',
+    'xlsxwriter>=3.1.0',
+    'openpyxl>=3.1.0'
```
amontanez24, line 43: After speaking with @npatki I think we should make these optional dependencies. I would also just refer to pandas' dependencies instead of maintaining this ourselves. Something like
```toml
[project.optional-dependencies]
excel_handler = [
    'pandas[excel]'
]
```

sdv/io/local/local.py
amontanez24: Mode should be used. If set to 'w' we should write the sheets to a new file. If set to 'a' we should write the sheets to an existing file. What is currently happening if the file already exists?

amontanez24 (changes requested):

R-Palazzo (approved):

tests/unit/io/local/test_local.py
```diff
@@ -347,3 +348,48 @@ def test_write(self, mock_pd):
...
+            float_format=None,
+            index=False
+        )
+        synth_sheet_two.to_excel.assert_called_once_with(
+            mock_pd.ExcelWriter.return_value,
+            sheet_name='Sheet2_synthetic',
+            float_format=None,
+            index=False
+        )
+        mock_pd.ExcelWriter.return_value.close.assert_called_once_with()
```
amontanez24, line 395: if there is no parameter we can just say `assert_called_once` right?

sdv/io/local/local.py
```diff
@@ -192,3 +192,90 @@ def write(self, synthetic_data, folder_name, file_name_suffix=None, mode='x'):
...
+        for table_name, table_data in temp_data.items():
+            if sheet_name_suffix and not suffix_added:
+                table_name += sheet_name_suffix
+
+            table_data.to_excel(
+                writer,
+                sheet_name=table_name,
+                float_format=self.float_format,
+                index=False
+            )
```
amontanez24, line 279: does this overwrite the file if it already existed?

pvk-developer: This overwrites the sheet. The way excel writer works is that always overwrites the document. There is no 'append' like in `csv`.
That is why in order to 'append' we have to:
* Read the file.
* Extract the sheets that are already in the file.
* Update them.
* Save the file again.

Now that I'm writing this I realize, that I overwrite the sheet name, and I should `append` to it. I'm addressing that in the following commit.

amontanez24: I meant if mode is 'w'. The spec says if mode is 'w' and there was anything in the file before, it should be completely overwritten and any other sheets deleted. Is that what happens?

pvk-developer: Yes, that does happen. It always overwrites even if the file exists.

amontanez24 (approved):

PR #1963, opened 2024-04-26
Include `sdv_logger_config.yml` with the package

pvk-developer:
Include the `sdv_logger_config.yml` with the package.
Update the `github workflow` for `macos` since `latest` is using `arm` now instead of `x86` (intel) and `numpy` does not have a compiled distribution for it.
Also fixes logger backwards compatibility in the `SingleTableSynthesizer` and `Par`.


gsheni (approved):

sdv/sequential/par.py
```diff
@@ -87,12 +87,13 @@ def _get_context_metadata(self):
 
     def __init__(self, metadata, enforce_min_max_values=True, enforce_rounding=False,
                  locales=['en_US'], context_columns=None, segment_size=None, epochs=128,
-                 sample_size=1, cuda=True, verbose=False):
+                 sample_size=1, cuda=True, verbose=False, table_name=None):
```
amontanez24, line 90: I just realized that this parameter didn't previously exist. I don't think we should add any parameters to the init. We should probably remove it from the single table synthesizers as well or at least consult @npatki 

Instead maybe we set the name on the metadata directly here
https://github.com/sdv-dev/SDV/blob/275955dcaaa81dcf3385b03965839843b80c255b/sdv/multi_table/base.py#L63-L70

.github/workflows/minimum.yml
```diff
@@ -13,7 +13,7 @@ jobs:
         python-version: [ '3.8', '3.9', '3.10', '3.11', '3.12']
         os: [ubuntu-latest, windows-latest]
         include:
-          - os: macos-latest
+          - os: macos-11
```
amontanez24, line 16: I don't think we need to change this. It doesn't really matter what kind of a mac these tests run on

amontanez24 (changes requested):

.github/workflows/minimum.yml
pvk-developer: Not randomly, the minimum `numpy` which is 1.20, doesn't have the `arm` builds or `universal` that is why.

amontanez24: Oh ok

amontanez24 (approved):

amontanez24 (approved):

frances-h (approved):

PR #1976, opened 2024-05-01
Convert integer column names to strings to allow for default column names

### LINKED ISSUE #1935: Synthesizers crash when column names are integers (`TypeError: unsupported operand`)
### Environment Details
* SDV version: 1.12.0

### Error Description
If my datafame's column names are represented with integers (instead of with strings), then the synthesizer's fit crashes with a `TypeError`

### Steps to reproduce
```python
import pandas as pd
import numpy as np

# generate a dataframe with integer column names
num_rows = 100
num_cols = 20
values = {i+1: np.random.randint(0, 100, size=num_rows) for i in range(num_cols)}
data = pd.DataFrame(values)

# create metadata
metadata_dict = {'columns': {}}
for i in range(num_cols):
  metadata_dict['columns'][i+1] = { 'sdtype': 'numerical' }
metadata = SingleTableMetadata.load_from_dict(metadata_dict)

# try to model it
synth.fit(data)
```
```
TypeError: unsupported operand type(s) for +=: 'int' and 'str'
```

[stack_trace.txt](https://github.com/sdv-dev/SDV/files/15031009/stack_trace.txt)

### Additional Context
- This is replicable for any synthesizer with a fit function (CTGAN, etc.)
- The metadata auto detection doesn't work for these dataframes either -- see #1933
- When would you have a DataFrame with integer column names? One use case is if I have a CSV that doesn't have any column names listed. If I tell pandas that there is no header in my data, then it will create column names with index values 0, 1, ...
```python
df = pd.read_csv('my_data.csv', header=None)
```

### Workaround
A simple workaround would be to rename the columns using strings instead of integers.

lajohn4747:
resolves #1935 
CU-86b04ut1e

Handle columns that are integers
This PR allows the base synthesizer to ingest data and convert the columns into strings. When sampling, it reassigns the correct columns so the user does have different columns for synthetic data.

Please note metadata dictionaries `columns` need to be a string.


sdv-team:
Task linked: [CU-86b04ut1e SDV - Synthesizers crash when column names are integers (`TypeError: unsupported operand`) #1935](https://app.clickup.com/t/86b04ut1e)

tests/integration/multi_table/test_hma.py
```diff
@@ -1738,3 +1738,48 @@ def test_synthesizer_logger(mock_datetime, mock_generate_id):
         '    Total number of columns: 15\n'
         '  Synthesizer id: HMASynthesizer_1.0.0_92aff11e9a5649d1a280990d1231a5f5\n'
     )
+
+
+def test_fit_and_sample_numerical_col_names():
+    """Test fit and sampling when column names are integers"""
```
R-Palazzo, line 1744: fit/sample or fitting/sampling

tests/integration/single_table/test_base.py
```diff
@@ -855,3 +855,37 @@ def test_synthesizer_logger(mock_datetime, mock_generate_id):
         '    Total number of columns: 3\n'
         '  Synthesizer id: GaussianCopulaSynthesizer_1.0.0_92aff11e9a5649d1a280990d1231a5f5\n'
     )
+
+
+def test_fit_and_sample_numerical_col_names():
+    """Test fit and sampling when column names are integers"""
```
R-Palazzo, line 861: fit/sample or fitting/sampling

tests/integration/single_table/test_base.py
```diff
@@ -855,3 +855,37 @@ def test_synthesizer_logger(mock_datetime, mock_generate_id):
...
+
+    # Run
+
+    synthesizers = [
+        CTGANSynthesizer,
+        TVAESynthesizer,
+        GaussianCopulaSynthesizer,
+        CopulaGANSynthesizer
+    ]
+    for synthesizer_class in synthesizers:
```
R-Palazzo, line 883: You can use `@pytest.mark.parametrize` instead of the `for` boucle

tests/integration/multi_table/test_hma.py
```diff
@@ -1738,3 +1738,48 @@ def test_synthesizer_logger(mock_datetime, mock_generate_id):
...
+    metadata = MultiTableMetadata.load_from_dict(metadata_dict)
+    metadata.set_primary_key('0', '1')
+
+    # Run
+    synth = HMASynthesizer(metadata)
+    synth.fit(data)
+    first_sample = synth.sample()
+    second_sample = synth.sample()
+
+    with pytest.raises(AssertionError):
```
R-Palazzo, line 1784: Add `#Assert`

tests/integration/single_table/test_base.py
```diff
@@ -855,3 +855,37 @@ def test_synthesizer_logger(mock_datetime, mock_generate_id):
...
+        synth = synthesizer_class(metadata)
+        synth.fit(data)
+        sample_1 = synth.sample(10)
+        sample_2 = synth.sample(10)
+
+        # Assert
```
R-Palazzo, line 889: Can we check that the string version of the column names matches the real data here?

R-Palazzo (commented):
Looks good!

Can you add some unit tests?

sdv/single_table/base.py
```diff
@@ -383,7 +384,18 @@ def preprocess(self, data):
                 "please refit the model using 'fit' or 'fit_processed_data'."
             )
 
-        return self._preprocess(data)
+        for column in data.columns:
+            if isinstance(column, int):
```
R-Palazzo, line 388: Do we need this check? Or can we do the same as done for the multi-table case?

tests/integration/multi_table/test_hma.py
```diff
@@ -1738,3 +1738,52 @@ def test_synthesizer_logger(mock_datetime, mock_generate_id):
         '    Total number of columns: 15\n'
         '  Synthesizer id: HMASynthesizer_1.0.0_92aff11e9a5649d1a280990d1231a5f5\n'
     )
+
+
+def test_fit_and_sample_numerical_col_names():
+    """Test fitting/sampling when column names are integers"""
+    # Setup data
```
R-Palazzo, line 1745: `# Setup`

tests/integration/multi_table/test_hma.py
```diff
@@ -1738,3 +1738,52 @@ def test_synthesizer_logger(mock_datetime, mock_generate_id):
...
+    # Run
+    synth = HMASynthesizer(metadata)
+    synth.fit(data)
+    first_sample = synth.sample()
+    second_sample = synth.sample()
+    assert first_sample['0'].columns.tolist() == data['0'].columns.tolist()
+    assert first_sample['1'].columns.tolist() == data['1'].columns.tolist()
+    assert second_sample['0'].columns.tolist() == data['0'].columns.tolist()
+    assert second_sample['1'].columns.tolist() == data['1'].columns.tolist()
+    # Assert
```
R-Palazzo, line 1787: Add a blank line before `# Assert`

sdv/single_table/base.py
```diff
@@ -383,7 +384,18 @@ def preprocess(self, data):
...
+                self._original_columns = data.columns
+                data.columns = data.columns.astype(str)
+                break
+
+        preprocess_data = self._preprocess(data)
+
+        if not self._original_columns.empty:
+            data.columns = self._original_columns
```
R-Palazzo, line 396: Can we split the logic of the `_original_columns` and the `preprocessing`? By having a private method like `_set_original_column_names(data)` or similar

sdv/single_table/utils.py
```diff
@@ -303,7 +303,7 @@ def unflatten_dict(flat):
 
             else:
                 subdict = unflattened.setdefault(key, {})
-                if subkey.isdigit():
+                if subkey.isdigit() and key != 'univariates':
```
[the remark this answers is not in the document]
lajohn4747: Univariate rows are just column names and don't have nested elements like `correlation` and `distribution`. If the column name is an integer it gets picked up by this line but the `univariate` column doesn't encode any values other than the name of the original column so it shouldn't matter if it is converted to an `int`. This prevents later methods from crashing due to a column name being an `int`

sdv/multi_table/base.py
```diff
@@ -337,10 +350,7 @@ def preprocess(self, data):
             dict:
                 A dictionary with the preprocessed data.
         """
-        for table, dataframe in data.items():
-            self._original_table_columns[table] = dataframe.columns
-            dataframe.columns = dataframe.columns.astype(str)
-            data[table] = dataframe
+        list_of_chnaged_tables = self._store_and_convert_original_cols(data)
```
R-Palazzo, line 353: chnaged -> changed

sdv/single_table/utils.py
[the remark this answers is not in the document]
R-Palazzo: I see, thanks

R-Palazzo (approved):
Thanks for addressing the comments :)

sdv/multi_table/base.py
```diff
@@ -480,6 +499,10 @@ def sample(self, scale=1.0):
             total_rows += len(table)
             total_columns += len(table.columns)
 
+        for table in sampled_data:
+            if table in self._original_table_columns:
```
frances-h, line 503: For backwards compatibility, we should use `getattr` and default to an empty dict

sdv/single_table/base.py
```diff
@@ -891,6 +909,9 @@ def sample(self, num_rows, max_tries_per_batch=100, batch_size=None, output_file
             show_progress_bar=show_progress_bar
         )
 
+        if not self._original_columns.empty:
```
frances-h, line 912: We should update this to use `hasattr` or `getattr` for backwards compatability

frances-h (approved):
LGTM 👍

PR #1981, opened 2024-05-02
Update the logger settings

pvk-developer:
After discussing with @amontanez24 , we would like to set the default config to None and not log the public usage.
This will be changed in enterprise.

CU-86b0ah0kh


amontanez24 (approved):

gsheni:
Why are the unit tests failing?

amontanez24:
> Why are the unit/integration tests failing?

It's unrelated and should be fixed by #1982

lajohn4747 (approved):

PR #1988, opened 2024-05-06
Split out metadata creation from data

### LINKED ISSUE #1975: Split out metadata creation from data import in the local files handlers
### Problem Description
The local file handlers (for CSV and Excel) currently return both the metadata and data in one go.

```python
data, metadata = handler.read(folder_name='project/data')
```

While this approach is ok, it is not consistent with other parts of our product and may not be future-proof.

### Expected behavior

1. Update the `handler.read` function (for both CSV and Excel) to _just_ read and return the data. Do _not_ create or return metadata in this function.
2. Create a new function called `create_metadata` (for both CSV and Excel). This function should take as input a dictionary of DataFrames. It should output a `MultiTableMetadata` object.

Below is the updated user journey:
```python
from sdv.io.local import CSVHandler

handler = CSVHandler(sep='\t', encoding='UTF') 
data = handler.read(folder_name='project/data')
metadata = handler.create_metadata(data)
```

### Additional context
In the future, we would like to allow flags in create_metadata in order to control whether inference should be done. But this is **out of scope** for the current issue.

```python
# EXAMPLE only: out of scope
metadata = handler.create_metadata(
  infer_sdtypes=True,
  infer_primary_keys=False,
  infer_foreign_keys=True)
```

pvk-developer:
Resolves #1975 
CU-86b09kwrd

### DIFF
--- a/sdv/io/local/local.py
+++ b/sdv/io/local/local.py
@@ -16,7 +16,7 @@ class BaseLocalHandler:
         self.decimal = decimal
         self.float_format = float_format
 
-    def _infer_metadata(self, data):
+    def create_metadata(self, data):
         """Detect the metadata for all tables in a dictionary of dataframes.
 
         Args:
@@ -110,8 +110,6 @@ class CSVHandler(BaseLocalHandler):
                 If the specified files do not exist in the folder.
         """
         data = {}
-        metadata = MultiTableMetadata()
-
         folder_path = Path(folder_name)
 
         if file_names is None:
@@ -156,8 +154,7 @@ class CSVHandler(BaseLocalHandler):
                 **kwargs
             )
 
-        metadata = self._infer_metadata(data)
-        return data, metadata
+        return data
 
     def write(self, synthetic_data, folder_name, file_name_suffix=None, mode='x'):
         """Write synthetic data to CSV files.
@@ -229,13 +226,10 @@ class ExcelHandler(BaseLocalHandler):
                 A tuple containing the data as a dictionary and metadata. The dictionary maps
                 table names to pandas DataFrames. The metadata is an object describing the data.
         """
-        metadata = MultiTableMetadata()
         if sheet_names is not None and not isinstance(sheet_names, list):
             raise ValueError("'sheet_names' must be None or a list of strings.")
 
-        data = self._read_excel(file_path, sheet_names)
-        metadata = self._infer_metadata(data)
-        return data, metadata
+        return self._read_excel(file_path, sheet_names)
 
     def write(self, synthetic_data, file_name, sheet_name_suffix=None, mode='w'):
         """Write synthetic data to an Excel File.
--- a/sdv/io/local/local.py
+++ b/sdv/io/local/local.py
@@ -38,9 +38,8 @@ class BaseLocalHandler:
         This method must be implemented by subclasses.
 
         Returns:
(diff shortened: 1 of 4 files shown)


sdv-team:
Task linked: [CU-86b09kwrd SDV - Split out metadata creation from data import in the local files handlers #1975](https://app.clickup.com/t/86b09kwrd)

tests/unit/io/local/test_local.py
```diff
@@ -6,10 +6,61 @@
...
 
-from sdv.io.local.local import CSVHandler, ExcelHandler
+from sdv.io.local.local import BaseLocalHandler, CSVHandler, ExcelHandler
 from sdv.metadata.multi_table import MultiTableMetadata
 
 
+class TestBaseLocalHandler:
+
+    def test___init__(self):
+        """Test the dafault initialization of the class."""
```
R-Palazzo, line 16: dafault -> default

setup.cfg
```diff
@@ -14,7 +14,9 @@ extend-ignore =
 	# TokenError: unterminated string literal
 	E902,
 	# Mutable default arg of type List
-	M511
+	M511,
+    # Logging and IO shadowing python's builtins
+    A005
```
R-Palazzo, line 19: Out of curiosity, did the Lint GitHub workflow crash without it? I made this change locally because it was, but I didn't push it since it was working on github.

tests/unit/io/local/test_local.py
```diff
@@ -6,10 +6,61 @@
...
+                'hotel_id': [1, 1, 3, 2, 3]
+            })
+        }
+        instance = BaseLocalHandler()
+
+        # Run
+        metadata = instance.create_metadata(data)
+
+        # Assert
+        assert isinstance(metadata, MultiTableMetadata) is True
```
R-Palazzo, line 43: You don't need `is True` here I think

R-Palazzo (approved):
Looks good!

I think there is some docs to update, for instance here:
https://github.com/sdv-dev/SDV/blob/248bc1a4c96c11edbdac97105e082ebda820ae39/sdv/io/local/local.py#L105

sdv/io/local/local.py
```diff
@@ -229,13 +226,10 @@ def read(self, file_path, sheet_names=None):
                 A tuple containing the data as a dictionary and metadata. The dictionary maps
```
fealho, line 226: Update this

sdv/io/local/local.py
```diff
@@ -110,8 +110,6 @@ def read(self, folder_name, file_names=None):
                 If the specified files do not exist in the folder.
```
fealho, line 110: Update Returns.

setup.cfg
fealho: Did you run using Python 3.9?

fealho (approved):

setup.cfg
pvk-developer: Yes. I run on python 3.9. It is okay to have this here I believe.

tests/unit/io/local/test_local.py
pvk-developer: I think that I started doing this after `assert not False` or `assert None` started asking me to do `assert is None` or `assert is False`. But I think its fine to be just `assert`.

PR #1989, opened 2024-05-06
Convert metadata columns from integer to columns to ensure that SDV works properly

### LINKED ISSUE #1933: Metadata detection crashes when the column names are integers (`AttributeError: 'int' object has no attribute 'lower'`)
### Environment Details
* SDV version: 1.12.0 (latest)

### Error Description
If my datafame's column names are represented with integers (instead of with strings), then the metadata detection crashes with a `TypeError`. 

### Steps to reproduce

```python
import pandas as pd
import numpy as np

# generate a dataframe with integer column names
num_rows = 100
num_cols = 20
values = {i+1: np.random.randint(0, 100, size=num_rows) for i in range(num_cols)}
data = pd.DataFrame(values)

metadata.detect_from_dataframe(data=df)
```
```
AttributeError: 'int' object has no attribute 'lower'
```
[stack_trace.txt](https://github.com/sdv-dev/SDV/files/15030588/stack_trace.txt)

### Additional Context
**When would you have a DataFrame with integer column names?** One use case is if I have a CSV that doesn't have any column names listed. If I tell pandas that there is no header in my data, then it will create column names with index values `0`, `1`, ...

```python
df = pd.read_csv('my_data.csv', header=None)
```

### Workaround
A simple workaround would be to rename the columns using strings instead of integers.

lajohn4747:
resolves #1933 
CU-86b04u2ae

Metadata columns need to be strings as much of SDV metadata detection assumes  all columns are of `string` type

### DIFF
--- a/sdv/metadata/multi_table.py
+++ b/sdv/metadata/multi_table.py
@@ -389,7 +389,7 @@ class MultiTableMetadata:
         """
         self._validate_table_exists(table_name)
         table = self.tables.get(table_name)
-        table.update_column(column_name, **kwargs)
+        table.update_column(str(column_name), **kwargs)
 
     def update_columns(self, table_name, column_names, **kwargs):
         """Update multiple columns with the same metadata kwargs.
@@ -523,7 +523,9 @@ class MultiTableMetadata:
 
     def detect_table_from_dataframe(self, table_name, data):
         """Detect the metadata for a table from a dataframe.
-        All data columns are converted to strings
+
+        This method automatically detects the ``sdtypes`` for the given ``pandas.DataFrame``,
+        for a specified table. All data columns are converted to strings.
 
         Args:
             table_name (str):
@@ -539,7 +541,9 @@ class MultiTableMetadata:
 
     def detect_from_dataframes(self, data):
         """Detect the metadata for all tables in a dictionary of dataframes.
-        All data columns are converted to strings
+
+        This method automatically detects the ``sdtypes`` for the given ``pandas.DataFrame``.
+        All data columns are converted to strings.
 
         Args:
             data (dict):
--- a/sdv/metadata/multi_table.py
+++ b/sdv/metadata/multi_table.py
@@ -389,7 +389,7 @@ class MultiTableMetadata:
         """
         self._validate_table_exists(table_name)
         table = self.tables.get(table_name)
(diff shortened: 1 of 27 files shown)


sdv-team:
Task linked: [CU-86b04u2ae SDV - Metadata detection crashes when the column names are integers (`TypeError: unsupported operand type(s)`) #1933](https://app.clickup.com/t/86b04u2ae)

sdv/multi_table/hma.py
```diff
@@ -114,7 +114,7 @@ def _estimate_columns_traversal(cls, metadata, table_name,
             columns_per_table[table_name] += \
                 cls._get_num_extended_columns(
                     metadata, child_name, table_name, columns_per_table, distributions
-                )
+            )
```
amontanez24, line 117: why move this? The bracket should line up with the statement it closes right?

amontanez24 (commented):
Can we add an integration test for both the single and multi table case?

sdv/metadata/multi_table.py
```diff
@@ -524,6 +524,9 @@ def _detect_relationships(self):
     def detect_table_from_dataframe(self, table_name, data):
         """Detect the metadata for a table from a dataframe.
 
+        This method automatically detects the ``sdtypes`` for the given ``pandas.DataFrame``,
+        for a specified table. All data columns are converted to strings.
```
amontanez24, line 528: we should change this to say column names so they don't think we're converting the columns themselves

amontanez24 (approved):

tests/integration/single_table/test_base.py
```diff
@@ -831,3 +831,20 @@ def test_sample_not_fitted(synthesizer):
...
+        1: [1, 2, 3],
+        2: [4, 5, 6],
+        3: ['a', 'b', 'c'],
+    })
+    metadata = SingleTableMetadata()
+    metadata.detect_from_dataframe(data)
+    instance = synthesizer_class(metadata)
+    instance.fit(data)
+    sample = instance.sample(5)
+    assert sample.columns.tolist() == data.columns.tolist()
```
R-Palazzo, line 850: and `# Assert` and `# Run` before

R-Palazzo (approved):
LGTM!

PR #2003, opened 2024-05-14
Remove unexpected NaN values in sequence_index

### LINKED ISSUE #1973: Unexpected NaN values in `sequence_index` when dataframe isn't reset
### Environment Details
* SDV version: 1.12.1 (latest)

### Error Description
Like any SDV synthesizer, the PARSynthesizer is supposed to synthesize NaN values if (and only if) the real data also had NaN values for the column. However, we see that sometimes, PAR creates NaN values for the _sequence index_ column even though the real data does not have missing values.

This is replicable for the particular case where the index (aka row count) of the pandas dataframe hasn't been reset. This can happen, for instance, when you subset the rows. _Note that the pandas dataframe index is a separate concept than the PAR's sequence index! These are two separate columns._

### Steps to reproduce
As described [in this comment](https://github.com/sdv-dev/SDV/issues/1808#issuecomment-2081551310), we can replicate the issue by taking the demo dataset and dropping a few of the sequences.

```python
import numpy as np
import pandas as pd

data, metadata = download_demo(modality='sequential', dataset_name='nasdaq100_2019',)

# modify the data by choosing a subset of it
data_subset = data.copy()
np.random.seed(1234)
symbols = data['Symbol'].unique()

# only select a subset of data in each sequence
for i, symbol in enumerate(symbols):
    symbol_mask = data_subset['Symbol'] == symbol
    data_subset = data_subset.drop(data_subset[symbol_mask].sample(frac = i / (2 * len(symbols))).index)

# now run PAR
# for performance's sake, we'll running with fewer epochs and create fewer samples
synthesizer = PARSynthesizer(metadata, epochs=5, verbose=True)
synthesizer.fit(data_subset)
synthetic_data = synthesizer.sample(num_sequences=5)

# observe that the synthetic data contains NaN values in sequence index column 'Date'
# even though the real data does not have NaN values for it
print('Real data has NaNs?:', pd.isnull(data_subset['Date']).any())
print('Synthetic data has NaNs?:', pd.isnull(synthetic_data['Date']).any())
```
```
Real data has NaNs?: False
Synthetic data has NaNs?: True
```

### Root Cause
This is happening because the _pandas DataFrame_ index is not a continuously increasing index (due to the subsetting). For example, observe how row 318 does not exist (because the subset didn't include it):

image

### Workaround
The workaround would be to reset the index of the subsetted data _before_ fitting it with PAR

```python
# reset the index column so it doesn't randomly skip rows
data_subset_reset = data_subset.reset_index().drop(columns='index')

# now fit PR on the reset data
synthesizer = PARSynthesizer(metadata, epochs=5, verbose=True)
synthesizer.fit(data_subset_reset)
synthetic_data2 = synthesizer.sample(num_sequences=5)
```

The problem then goes away.
```python
print('Real data has NaNs?:', pd.isnull(data_subset_reset['Date']).any())
print('Synthetic data has NaNs?:', pd.isnull(synthetic_data2['Date']).any())
```
```
Real data has NaNs?: False
Synthetic data has NaNs?: False
```

fealho:
CU-86b090qgm, Resolve #1973.

### DIFF
--- a/sdv/multi_table/hma.py
+++ b/sdv/multi_table/hma.py
@@ -114,7 +114,7 @@ class HMASynthesizer(BaseHierarchicalSampler, BaseMultiTableSynthesizer):
             columns_per_table[table_name] += \
                 cls._get_num_extended_columns(
                     metadata, child_name, table_name, columns_per_table, distributions
-                )
+            )
 
         visited.add(table_name)
 
--- a/sdv/multi_table/utils.py
+++ b/sdv/multi_table/utils.py
@@ -47,10 +47,10 @@ def _get_n_order_descendants(relationships, parent_table, order):
     descendants = {}
     order_1_descendants = _get_relationships_for_parent(relationships, parent_table)
     descendants['order_1'] = [rel['child_table_name'] for rel in order_1_descendants]
-    for i in range(2, order+1):
+    for i in range(2, order + 1):
         descendants[f'order_{i}'] = []
         prov_descendants = []
-        for child_table in descendants[f'order_{i-1}']:
+        for child_table in descendants[f'order_{i - 1}']:
             order_i_descendants = _get_relationships_for_parent(relationships, child_table)
             prov_descendants.extend([rel['child_table_name'] for rel in order_i_descendants])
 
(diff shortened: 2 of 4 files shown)


sdv-team:
Task linked: [CU-86b090qgm SDV - Unexpected NaN values in `sequence_index` when dataframe isn't reset #1973](https://app.clickup.com/t/86b090qgm)

pvk-developer (approved):

frances-h (approved):

PR #2005, opened 2024-05-14
Update logs to write to CSV

### LINKED ISSUE #1974: Save usage log file as a csv
### Problem Description

_As an engineer debugging, it would be nice to store the logs as a csv so that they can be loaded as a DataFrame and analyzed._

### Expected behavior

- Store the [log file](https://github.com/sdv-dev/SDV/blob/25388141122ab3b0a2a4948cbc3d4e9a7214723b/sdv/logging/sdv_logger_config.yml#L9) for usage as a CSV 

### Additional context

- It seems like we can implement a CSVFormatter that gets used by the logger. See [this post
](https://stackoverflow.com/questions/19765139/what-is-the-proper-way-to-do-logging-to-a-csv-file)

frances-h:
Resolve #1974 
CU-86b094e79

### DIFF
--- a/sdv/logging/logger.py
+++ b/sdv/logging/logger.py
@@ -7,7 +7,7 @@ from io import StringIO
 from sdv.logging.utils import get_sdv_logger_config
 
 
-class CSVHandler(logging.Formatter):
+class CSVFormatter(logging.Formatter):
     """Logging formatter to convert to CSV."""
 
     def __init__(self):
@@ -59,11 +59,14 @@ def get_sdv_logger(logger_name):
             logger.removeHandler(handler)
 
         if logger_name in logger_conf.get('loggers'):
-            formatter = CSVHandler()
+            formatter = None
             config = logger_conf.get('loggers').get(logger_name)
             log_level = getattr(logging, config.get('level', 'INFO'))
             if config.get('format'):
-                formatter = logging.Formatter(config.get('format'))
+                if config.get('format') == 'CSV':
+                    formatter = CSVFormatter()
+                else:
+                    formatter = logging.Formatter(config.get('format'))
 
             logger.setLevel(log_level)
             logger.propagate = config.get('propagate', False)
--- a/sdv/logging/logger.py
+++ b/sdv/logging/logger.py
@@ -1,9 +1,33 @@
 """SDV Logger."""
-
+import csv
 import logging
 from functools import lru_cache
+from io import StringIO
 
 from sdv.logging.utils import get_sdv_logger_config
+
+
+class CSVHandler(logging.Formatter):
+    """Logging formatter to convert to CSV."""
+
+    def __init__(self):
+        super().__init__()
+        self.output = StringIO()
+        headers = [
+            'LEVEL', 'EVENT', 'TIMESTAMP', 'SYNTHESIZER CLASS NAME', 'SYNTHESIZER ID',
+            'TOTAL NUMBER OF TABLES', 'TOTAL NUMBER OF ROWS', 'TOTAL NUMBER OF COLUMNS'
+        ]
+        self.writer = csv.DictWriter(self.output, headers)
+
+    def format(self, record):  # noqa: A003
+        """Format the record and write to CSV."""
+        row = record.msg
+        row['LEVEL'] = record.levelname
+        self.writer.writerow(row)
+        data = self.output.getvalue()
+        self.output.truncate(0)
+        self.output.seek(0)
+        return data.strip()
 
 
 @lru_cache()
@@ -35,7 +59,7 @@ def get_sdv_logger(logger_name):
             logger.removeHandler(handler)
 
         if logger_name in logger_conf.get('loggers'):
-            formatter = None
+            formatter = CSVHandler()
             config = logger_conf.get('loggers').get(logger_name)
             log_level = getattr(logging, config.get('level', 'INFO'))
             if config.get('format'):
--- a/sdv/logging/logger.py
+++ b/sdv/logging/logger.py
@@ -59,14 +59,14 @@ def get_sdv_logger(logger_name):
             logger.removeHandler(handler)
 
         if logger_name in logger_conf.get('loggers'):
-            formatter = None
+            formatter = None            
             config = logger_conf.get('loggers').get(logger_name)
             log_level = getattr(logging, config.get('level', 'INFO'))
-            if config.get('format'):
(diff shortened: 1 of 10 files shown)


sdv-team:
Task linked: [CU-86b094e79 SDV - Save usage log file as a csv #1974](https://app.clickup.com/t/86b094e79)

sdv/logging/logger.py
```diff
@@ -1,11 +1,35 @@
...
-
+import csv
 import logging
 from functools import lru_cache
+from io import StringIO
 
 from sdv.logging.utils import get_sdv_logger_config
 
 
+class CSVHandler(logging.Formatter):
```
pvk-developer, line 10: Could we use a different name for this ? I'm a bit concerned about having another class already with this name:
https://github.com/sdv-dev/SDV/blob/08a464c10d82f054e00d5807f6574579d7204329/sdv/io/local/local.py#L54

sdv/logging/logger.py
```diff
@@ -35,7 +59,7 @@ def get_sdv_logger(logger_name):
             logger.removeHandler(handler)
 
         if logger_name in logger_conf.get('loggers'):
-            formatter = None
+            formatter = CSVHandler()
```
pvk-developer, line 62: I think that this is not a `formatter` but a `handler`.
I believe that we could do something like this:
```python
if handler_class == 'CSVHandler':
    # Do the logic in here
```

sdv/logging/sdv_logger_config.yml
```diff
@@ -6,22 +6,22 @@ loggers:
     propagate: false
     handlers:
       class: logging.FileHandler
```
pvk-developer, line 8: instead of `logging.FileHandler` this should be `sdv.logging.logger.CSVHandler`

sdv/multi_table/base.py
```diff
@@ -119,15 +119,12 @@ def __init__(self, metadata, locales=['en_US'], synthesizer_kwargs=None):
...
-            datetime.datetime.now(),
-            self.__class__.__name__,
-            self._synthesizer_id
-        )
+        SYNTHESIZER_LOGGER.info({
+            'EVENT': 'Instance',
+            'TIMESTAMP': datetime.datetime.now(),
+            'SYNTHESIZER CLASS NAME': self.__class__.__name__,
+            'SYNTHESIZER ID': self._synthesizer_id
+        })
```
pvk-developer, line 127: 🤔 what if we use other handler that is not `CSV`?  

CC: @amontanez24 
Are we going to support multiple handlers or just `csv` would be the end goal of this ? If so I think we can simplify things a bit regarding the `logger`.

pvk-developer (commented):
Left some comments but open to discuss them.

sdv/multi_table/base.py
frances-h: We can always write other `logging.Formatter` classes that take in this input dict and output it however we want it formatted.

sdv/logging/logger.py
frances-h: Ah good catch, it should probably be `CSVFormatter` to be clearer.

sdv/logging/logger.py
frances-h: We're still using the FileHandler as our handler since we're writing to the file. The `CSVHandler` (to be renamed CSVFormatter) just formats the log message before passing it along to the handler, in this case converting it to a CSV row. We can add it to the config though, or case it on the file target extension.

sdv/logging/sdv_logger_config.yml
amontanez24: I think here the CSV part is being handled at the Formatter level, not the handler level.

sdv/logging/sdv_logger_config.yml
```diff
@@ -4,24 +4,28 @@ loggers:
   SingleTableSynthesizer:
     level: INFO
     propagate: false
+    format: CSV
```
amontanez24, line 7: the standard config for logging has the `formatter` key word which can be specified under the handler section. https://docs.python.org/3/library/logging.config.html#dictionary-schema-details

sdv/multi_table/base.py
amontanez24: I think csv will be it

sdv/logging/sdv_logger_config.yml
pvk-developer: Yeah sorry the `CSVHandler` threw me off thinking as a `CSVHandler` like inheriting from `FileHandler`

pvk-developer (approved):
LGTM!

amontanez24 (approved):

frances-h:
Sample logfile:

INFO,Instance,2024-05-15 14:35:34.602861,HMASynthesizer,HMASynthesizer_1.12.2.dev0_6dc42d45ad6044e7a46308596db128ed,,,
INFO,Fit,2024-05-15 14:35:38.887712,HMASynthesizer,HMASynthesizer_1.12.2.dev0_6dc42d45ad6044e7a46308596db128ed,2,668,15
INFO,Fit processed data,2024-05-15 14:35:39.201517,HMASynthesizer,HMASynthesizer_1.12.2.dev0_6dc42d45ad6044e7a46308596db128ed,2,668,11
INFO,Sample,2024-05-15 14:35:45.569641,HMASynthesizer,HMASynthesizer_1.12.2.dev0_6dc42d45ad6044e7a46308596db128ed,2,668,15
INFO,Sample,2024-05-15 14:35:48.678179,HMASynthesizer,HMASynthesizer_1.12.2.dev0_6dc42d45ad6044e7a46308596db128ed,2,1336,15
INFO,Instance,2024-05-15 14:36:01.630464,HSASynthesizer,HSASynthesizer_1.12.2.dev0_b1e1003996bf400b8e9f4c4c95b28d0b,,,
INFO,Fit,2024-05-15 14:36:01.738376,HSASynthesizer,HSASynthesizer_1.12.2.dev0_b1e1003996bf400b8e9f4c4c95b28d0b,2,668,15
INFO,Fit processed data,2024-05-15 14:36:01.944802,HSASynthesizer,HSASynthesizer_1.12.2.dev0_b1e1003996bf400b8e9f4c4c95b28d0b,2,668,11
INFO,Sample,2024-05-15 14:36:02.827084,HSASynthesizer,HSASynthesizer_1.12.2.dev0_b1e1003996bf400b8e9f4c4c95b28d0b,2,2004,15

PR #2007, opened 2024-05-15
Add pyyaml dependency and add install tests workflow

gsheni:
- Add a install test workflow which install core sdv requirements, imports sdv, and gets the version
- Add pyyaml dependency

### DIFF
--- a/.github/workflows/install.yaml
+++ b/.github/workflows/install.yaml
@@ -1,0 +1,35 @@
+name: Install Tests
+on:
+  pull_request:
+    types: [opened, synchronize]
+  push:
+    branches:
+      - main
+jobs:
+  install:
+    name: ${{ matrix.os }} - ${{ matrix.python_version }} install
+    strategy:
+      fail-fast: true
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+        python_version: ["3.8", "3.12"]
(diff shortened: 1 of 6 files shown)


.github/workflows/install.yaml
```diff
@@ -0,0 +1,35 @@
...
+  push:
+    branches:
+      - main
+jobs:
+  install:
+    name: ${{ matrix.os }} - ${{ matrix.python_version }} install
+    strategy:
+      fail-fast: true
+      matrix:
+        os: [ubuntu-latest, macos-latest]
```
amontanez24, line 14: let's just run on ubuntu

frances-h, line 14: should we add windows as well?

amontanez24 (approved):

.github/workflows/install.yaml
gsheni: Fixed:
https://github.com/sdv-dev/SDV/pull/2007/commits/e7b01d98d0302fdb0c6eb7f4953deddc6242769f

gsheni: Doing just ubuntu for now to keep testing to a minimal

frances-h (approved):

Issue #1778, opened 2024-02-06
RuntimeError when CPAR model is trained on CUDA-enabled device, then trying to sample from the model on cpu-only device


Faiga91:
### Environment Details

Please indicate the following details about the environment in which you found the bug:

* SDV version: **1.9.0**
* Python version:   **3.10.13**
* Operating System: **macOS 14.2.1**

### Error Description

When I train the PAR synthesiser code on CUDA enabled device, then I try to generate new samples from the trained synthesizer on a CPU-only device, I get the following error: 

![image](https://github.com/sdv-dev/SDV/assets/16253252/9513c4bf-dd6b-433b-bea1-0920b4a5a917)

npatki:
Hi @Faiga91 thanks for reporting this issue.

Can you confirm that sampling works on the _same_ CUDA-enabled device that you fitted your data on? If this doesn't work then it is indeed a bug.

As for switching to a CPU-only device for sampling: We currently do not support the use case fitting on a GPU and then sampling from CPU. I will turn this into a feature request for us to track. We prioritize feature requests based on demand and importance to your project, so any other information you have about your use case will definitely help us prioritize.

Faiga91:
Thanks a lot for following up. It did indeed work on the same CUDA-enabled device. 
I managed to work around this issue, and my project is running with no issue. However, it will be nice to have a more customised error message when a user is trying to sample on a non-CUDA-enabled device.

npatki:
Hi @Faiga91, no problem.

Let's keep this issue open as a feature request for allowing sampling from a CPU-only device. In the meantime, I agree that it would at least be nice to have a better error message. I will file a new issue the error message, which we can probably address even sooner.

PR #2016, opened 2024-05-16
Improve error message when sampling on a non-CPU device

### LINKED ISSUE #1819: Improve error message when sampling on a non-CPU device
### Problem Description
When using the `PARSynthesizer`, SDV currently expects that the fitting and sampling are both done on machines with the same processors -- i.e. it is _not_ possible for you to fit on a GPU-enabled device, but then sample from a CPU-only device. The feature request for this is in #1778.

Until we are able to accommodate this feature, we should at least surface a better error message for users explaining this limitation. The current error message looks like this:

```
RuntimeError: Attempting to deserialize object on a CUDA device but torch.cuda.is_available() is False. 
If you are running on a CPU-only machine, please use torch.load with map_location=torch.device('cpu') 
to map your storages to the CPU.
```

### Expected behavior
Create a more generic error explaining that this feature isn't yet possible.

```
SamplingError: This synthesizer was created on a machine with GPU but the current machine is 
CPU-only. This feature is currently unsupported. We recommend sampling on the same GPU-enabled machine.
```

fealho:
CU-86azg7w2c, Resolve #1819.

### DIFF
--- a/sdv/multi_table/base.py
+++ b/sdv/multi_table/base.py
@@ -696,12 +696,19 @@ class BaseMultiTableSynthesizer:
         with open(filepath, 'rb') as f:
             try:
                 synthesizer = cloudpickle.load(f)
-            except RuntimeError:
-                raise SamplingError(
-                    'This synthesizer was created on a machine with GPU but the current machine is'
-                    ' CPU-only. This feature is currently unsupported. We recommend sampling on '
-                    'the same GPU-enabled machine.'
+            except RuntimeError as e:
+                err_msg = (
+                    'Attempting to deserialize object on a CUDA device but '
+                    'torch.cuda.is_available() is False. If you are running on a CPU-only machine,'
+                    " please use torch.load with map_location=torch.device('cpu') "
+                    'to map your storages to the CPU.'
                 )
+                if str(e) == err_msg:
+                    raise SamplingError(
+                        'This synthesizer was created on a machine with GPU but the current machine is'
+                        ' CPU-only. This feature is currently unsupported. We recommend sampling on '
+                        'the same GPU-enabled machine.'
+                    )
 
         check_synthesizer_version(synthesizer)
         check_sdv_versions_and_warn(synthesizer)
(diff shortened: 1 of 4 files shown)


sdv-team:
Task linked: [CU-86azg7w2c SDV - Improve error message when sampling on a non-CPU device #1819](https://app.clickup.com/t/86azg7w2c)

sdv/single_table/base.py
```diff
@@ -492,7 +492,14 @@ def load(cls, filepath):
...
         with open(filepath, 'rb') as f:
-            synthesizer = cloudpickle.load(f)
+            try:
+                synthesizer = cloudpickle.load(f)
+            except RuntimeError:
+                raise SamplingError(
+                    'This synthesizer was created on a machine with GPU but the current machine is'
+                    ' CPU-only. This feature is currently unsupported. We recommend sampling on '
+                    'the same GPU-enabled machine.'
+                )
```
pvk-developer, line 502: Shouldn't this occur on the `sampling` instead ? or the error gets raised while you load the synthesizer?

fealho: From the [user issue](https://github.com/sdv-dev/SDV/issues/1778), they ran into this problem when loading the synthesizer, so that's where I added the warning.

pvk-developer (approved):

sdv/multi_table/base.py
```diff
@@ -694,7 +694,14 @@ def load(cls, filepath):
                 The loaded synthesizer.
         """
         with open(filepath, 'rb') as f:
-            synthesizer = cloudpickle.load(f)
+            try:
+                synthesizer = cloudpickle.load(f)
+            except RuntimeError:
+                raise SamplingError(
```
rwedge, line 700: Could we only switch to the new error message after confirming the error that surfaced has that specific message about the current machine being CPU-only?

rwedge (commented):
If the error message doesn't match the specific case we should re-raise it

rwedge (approved):

PR #2021, opened 2024-05-21
Include validation check for single table auto_assign_transformers

### LINKED ISSUE #1509: Better data validation message for `auto_assign_transformers`
### Problem Description
If I use the `auto_assign_transformers` functionality with invalid data*, then I receive an error that doesn't really make sense.

*Invalid data is any data that does not match the metadata

```python
import numpy as np
import pandas as pd

    'columns': {
        'a': { 'sdtype': 'categorical' },
    }
})

# input data that does not match the metadata
data = pd.DataFrame({'b': list(np.random.choice(['M', 'F'], size=10)) })
```

Output:
```
AttributeError: 'NoneType' object has no attribute 'get'
```

### Expected behavior
I expect an error that is more descriptive to the problem. We should re-use the error message from using `fit` on invalid data.

```python
synthesizer.fit(data)
```

```
The columns ['b'] are not present in the metadata.

The metadata columns ['a'] are not present in the data.
```

### Additional context
It appears that `fit` (and `fit_processed_data`) are actually running a validation check between the data and metadata. It seems that the `auto_assign_transformers` method is NOT running the check.

Should we run the check in this method? If so, maybe the fit functions don't need it (since they internally call this method first).

lajohn4747:
resolves #1509 
CU-85ztjjzrw

Include better messsaging for auto_assign_transformers by checking metadata and data before going through assigning the transformers.

### DIFF
--- a/sdv/single_table/base.py
+++ b/sdv/single_table/base.py
@@ -305,6 +305,7 @@ class BaseSynthesizer:
             data (pandas.DataFrame):
                 The raw data (before any transformations) that will be used to fit the model.
         """
+        self.metadata._validate_metadata_matches_data(data)
         self._data_processor.prepare_for_fitting(data)
 
     def get_transformers(self):
--- a/tests/unit/single_table/test_base.py
+++ b/tests/unit/single_table/test_base.py
@@ -12,7 +12,8 @@ from rdt.transformers import (
 
 from sdv import version
 from sdv.constraints.errors import AggregateConstraintsError
-from sdv.errors import ConstraintsNotMetError, SamplingError, SynthesizerInputError, VersionError
+from sdv.errors import (
+    ConstraintsNotMetError, InvalidDataError, SamplingError, SynthesizerInputError, VersionError)
 from sdv.metadata.single_table import SingleTableMetadata
 from sdv.sampling.tabular import Condition
 from sdv.single_table import (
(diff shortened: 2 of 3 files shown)


sdv-team:
Task linked: [CU-85ztjjzrw SDV - Better data validation message for `auto_assign_transformers` #1509](https://app.clickup.com/t/85ztjjzrw)

pvk-developer (approved):

lajohn4747:
> Does this work for multi table?

Yes it does, I added a test to multitable to confirm it as well. Multitable synthesizer handles `auto_assign_transformer` at the table level which happens to call the single table `auto_assign_transformer` function

tests/unit/single_table/test_base.py
```diff
@@ -218,6 +219,27 @@ def test_auto_assign_transformers(self):
         # Assert
         instance._data_processor.prepare_for_fitting.assert_called_once_with(data)
 
+    def test_auto_assign_transformers_with_invalid_data(self):
+        """Test that the ``DataProcessor.prepare_for_fitting`` is being called."""
```
R-Palazzo, line 223: The description doesn't seem to describe the test

tests/unit/single_table/test_base.py
```diff
@@ -218,6 +219,27 @@ def test_auto_assign_transformers(self):
...
+        synthesizer = GaussianCopulaSynthesizer(metadata)
+
+        # input data that does not match the metadata
+        data = pd.DataFrame({'b': list(np.random.choice(['M', 'F'], size=10))})
+        error_msg = re.escape(
+            'The provided data does not match the metadata:\n'
+            "The columns ['b'] are not present in the metadata.\n\n"
+            "The metadata columns ['a'] are not present in the data."
+        )
+        with pytest.raises(InvalidDataError, match=error_msg):
```
R-Palazzo, line 240: Add `# Run and Assert`

R-Palazzo (commented):
Looks good!

I just have one question

sdv/single_table/base.py
```diff
@@ -305,6 +305,7 @@ def auto_assign_transformers(self, data):
             data (pandas.DataFrame):
                 The raw data (before any transformations) that will be used to fit the model.
         """
+        self.metadata._validate_metadata_matches_data(data)
```
R-Palazzo, line 308: Should we use `validate()` here? I think it's the method called by `fit()`:
https://github.com/sdv-dev/SDV/blob/9421b160b8a2bd04950c305eee7244bc752b6806/sdv/single_table/base.py#L158

tests/unit/sequential/test_par.py
```diff
@@ -32,7 +32,7 @@ def get_metadata(self, add_sequence_key=True, add_sequence_index=False):
 
     def get_data(self):
         data = pd.DataFrame({
-            'time': [1, 2, 3],
```
lajohn4747, line 35: I actually use datetimes as now `auto_assign_transformer` will validate this data which would error out. It should be date-time anyway

R-Palazzo (approved):

amontanez24 (approved):

PR #2043, opened 2024-05-31
Do not enforce min/max on sequence index column

### LINKED ISSUE #2031: PARSynthesizer: Duplicate sequence index values when `sequence_length` is higher than real data
### Environment Details

* SDV version: SDV 1.12 and SDV 1.13.1

### Error Description

When the desired sequence length is higher than the real data's sequence length and min-max enforcement is enabled, PARSynthesizer can generate duplicate values. This seems to happen especially when the sequence index column is a datetime column. When synthesizing values for the sequence key column, PARSynthesizer runs into the max value and repeats it.

### Steps to reproduce

**Original Data**

2 sequences, each with 5 unique values for the `visits` column (the sequence index)

Screenshot 2024-05-28 at 6 38 57 PM

**Synthetic Data**

Synthetic Data example when you set `sequence_length` parameter to 25:

Screenshot 2024-05-28 at 5 14 50 PM

Synthetic Data example when you set `sequence_length` parameter to 7:

Screenshot 2024-05-28 at 5 15 32 PM

[Full code in Internal Colab Notebook here](https://colab.research.google.com/drive/1DwOOWZb8QnQk3y14iE06Cidv76YjBYIj?authuser=1#scrollTo=X0CahNzH7puz)

### Workarounds

- You can keep `enforce_min_max_values` as `False` and this will remove the max value ceiling for the datetime sequence_key column. But this will mean that the synthesized data will be less representative of your real data so this is a big tradeoff until this bug is fixed.
- You can set `num_sequences` to be identical to the number of rows in your smallest, least unique (when it comes to the sequence key column) sequence from your real data. E.g. if you have a small sequence with 5 unique values for the sequence key, don't generate more than 5 rows per sequence. But this is also a limitation of SDV until this bug is fixed.

Original Discussion here: https://github.com/sdv-dev/SDV/issues/2004

lajohn4747:
resolves #2031 
CU-86b0me61g

Do not enforce min/max on sequence index column so it does not cause duplication of the seq_index

### DIFF
--- a/sdv/sequential/par.py
+++ b/sdv/sequential/par.py
@@ -194,6 +194,28 @@ class PARSynthesizer(LossValuesMixin, BaseSynthesizer):
 
         return data
 
+    def auto_assign_transformers(self, data):
+        """Automatically assign the required transformers for the given data and constraints.
+
+        This method will automatically set a configuration to the ``rdt.HyperTransformer``
+        with the required transformers for the current data.
+
+        Args:
+            data (dict):
+                Mapping of table name to pandas.DataFrame.
+
+        Raises:
+            InvalidDataError:
+                If a table of the data is not present in the metadata.
+        """
+        super().auto_assign_transformers(data)
+
+        # Ensure that sequence index does not get auto assigned with enforce_min_max_values
+        if self._sequence_index:
+            sequence_index_transformer = self.get_transformers()[self._sequence_index]
+            if sequence_index_transformer.enforce_min_max_values:
+                sequence_index_transformer.enforce_min_max_values = False
+
     def _preprocess(self, data):
         """Transform the raw data to numerical space.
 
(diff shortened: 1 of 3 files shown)


sdv-team:
Task linked: [CU-86b0me61g SDV - PARSynthesizer: Duplicate sequence index values when sequence_length is higher than real data #2031](https://app.clickup.com/t/86b0me61g)

sdv/sequential/par.py
```diff
@@ -216,6 +216,9 @@ def _preprocess(self, data):
         preprocessed = super()._preprocess(data)
 
         if self._sequence_index:
+            sequence_index_transformer = self.get_transformers()[self._sequence_index]
+            if sequence_index_transformer.enforce_min_max_values:
+                sequence_index_transformer.enforce_min_max_values = False
```
frances-h, line 221: If a user updated the sequence index transformer and set `enforce_min_max_values` to `True`, would this ignore that? It might be better to only change the transformer when it's been automatically assigned, so that way if a user modifies it we don't overwrite their change.

lajohn4747: Sounds good! I moved it into a overridden `auto_assign_method` so it's always after being auto assigned.

frances-h (approved):
LGTM! 👍

amontanez24 (approved):

PR #2044, opened 2024-06-03
Allow for simple PARSynthesizer constraints

### LINKED ISSUE #1936: Allow constraints in PARSynthesizer (for all context cols, or all non-context columns)
### Problem Description
There is an increasing user demand for applying constraints to the PARSynthesizer -- see #570. One reason why we have been unable to support constraints is that the PARSynthesizer has unchanging, `context_columns` while other synthesizers do not.

To at least address _some_ constraint uses, it would be good to allow constraints if:
- All of the involved columns are NOT context columns or
- All of the involved columns ARE context columns

i.e. mixing-and-matching columns from context and non-context are not allowed within a constraint

Another reason is that PAR (and sequence-based synthesizers) cannot easily accommodate reject sampling based approaches. So it would only be possible to add non-overlapping constraints.

### Expected behavior
Enable the PARSynthesizer to be used with constraints, in the same manner that any other synthesizer can be used. But only for select cases:

1. You can only use a constraint for columns that are all contextual, or all non-contextual
2. You cannot supply multiple overlapping constraints (that is 2 constraints that each independently act on the same column)
3. You cannot use it for custom constraints (for now, as it is complicated with a custom constraint to determine whether points 1 and 2 can be met).

If a user passes in constraints that violate these assumptions, we should throw a `SynthesizerInputError`.

```python
my_constraint = {
  'constraint_parameters': {
    'high_column_name': 'A',
    'low_column_name': 'B'
  }
}

synthesizer = PARSynthesizer(metadata, context_columns=['A'])
synthesizer.add_constraints([my_constraint])
```
```
SynthesizerInputError: The PARSynthesizer cannot accommodate constraints with a mix of context
and non-context columns.
```

For overlapping constraints:
```
SynthesizerInputError: The PARSynthesizer cannot accommodate multiple constraints that overlap on 
the same columns.
```
For custom constraints:
```
SynthesizerInputError: The PARSynthesizer cannot accommodate custom constraints.
```

lajohn4747:
resolves #1936 
CU-86b04v6vb

Allow for constraints for PARSynthesize if they but only for select cases:

- You can only use a constraint for columns that are all contextual, or all non-contextual
- You cannot supply multiple overlapping constraints (that is 2 constraints that each independently act on the same column)
- You cannot use it for custom constraints (for now, as it is complicated with a custom constraint to determine whether points 1 and 2 can be met).

### DIFF
--- a/sdv/sequential/par.py
+++ b/sdv/sequential/par.py
@@ -3,7 +3,6 @@
 import inspect
 import logging
 import uuid
-import warnings
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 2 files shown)


sdv-team:
Task linked: [CU-86b04v6vb SDV - Allow constraints in PARSynthesizer (for all context cols, or all non-context columns) #1936](https://app.clickup.com/t/86b04v6vb)

frances-h (commented):
Can we add an integration test as well?

sdv/sequential/par.py
```diff
@@ -135,13 +134,53 @@ def get_parameters(self):
...
+                        'that overlap on the same columns.')
+                constraint_cols.append(col)
+
+        all_context = all(col in context_set for col in constraint_cols)
+        no_context = all(col not in context_set for col in constraint_cols)
+
+        if all_context or no_context:
+            super().add_constraints(constraints)
+        else:
+            raise SynthesizerInputError(
```
amontanez24, line 177: I think we need to add a check here as well that raises an error if someone tries to add a custom constraint dict. Even though they won't be able to load it, if they skip that step and add the dict then this method won't give them any feedback.

sdv/sequential/par.py
```diff
@@ -135,13 +134,53 @@ def get_parameters(self):
...
+                'The PARSynthesizer cannot accommodate constraints '
+                'with a mix of context and non-context columns.')
+
+    def load_custom_constraint_classes(self, filepath, class_names):
```
amontanez24, line 181: The following function should probably also be overridden even though it isn't public (in the docs)
https://github.com/sdv-dev/SDV/blob/70c2bf7060bac864757c1178123a3c68cf0fdf23/sdv/single_table/base.py#L261

sdv/sequential/par.py
lajohn4747: If we try to run a custom constraint it won't be found due to us not being able to load it without `add_custom_constraint_class` and `load_custom_constraint_classes`. We run into the default `InvalidConstraintsError` for a custom constraint that is not predefined. I think it is suitable and an extra check is not needed. What do you think? @amontanez24 

I also added a test to confirm the error is thrown for adding a custom constraint.

amontanez24: makes sense

amontanez24 (approved):

frances-h (approved):

PR #2053, opened 2024-06-11
Removing the saving of a temp file unless specified

### LINKED ISSUE #2042: If no filepath is provided, do not create a file during `sample`
### Problem Description
As described in issue #1310, the single table `sample` method has the following functionality:

- If you specify an `output_filepath`, the sampling function will write to the filepath after every batch, ensuring that you do not lose previously sampled data if there is a crash
- If you do not specify an `output_filepath`, then the sampling function will still save batched results to `.sample.csv.temp` (still want to ensure that you have access to some data in the event of a crash). Once the sampling completes, we delete this file.

Problems:
1. When you specify that `output_filepath=None`, SDV creates a temporary file anyway. This is unintuitive for users who assume that `None` means that no output filepath will be provided. It is also unintuitive that SDV deletes the file after-the-fact. 
2. In the event of a crash, there is a message asking users to check `.sample.csv.temp`, which is unexpected. This was meant to give the user *some data* rather than nothing. But practice there are always 0 rows in this file, because the default `batch_size` is kept equal to the sample size.
3. There is no way to turn _off_ the functionality for saving to a file. Users in both #1310 and #2029 mention that they'd like to turn it off because it does not work well within their filesystem (they are not worried about crashes during sampling). 

### Expected behavior

- If the `output_filepath=None`, we should no longer create or write to a temporary file. This will align with the user expectations and it will enable more forms of usage. In the event of a crash, update the error message. There will no longer be a file to reference.
- If `output_filepath` is provided, we should create and periodically write to the file as we do today. In the event of a crash, the error message should continue to ask the user to check the file.

**Default**: We can make the default `None`. In the future, we should figure out a better default for `batch_size` for very large samples. In such a case, it would make sense to have a output filepath supplied as a default.

lajohn4747:
resolves #2042 
CU-86b0p02f3

Remove the saving of a temp file during a synthesizer sample unless the user provides an output filepath.

### DIFF
--- a/sdv/lite/single_table.py
+++ b/sdv/lite/single_table.py
@@ -136,8 +136,7 @@ class SingleTablePreset:
                 The batch size to use per attempt at sampling. Defaults to 10 times
                 the number of rows.
             output_file_path (str or None):
-                The file to periodically write sampled rows to. Defaults to
-                a temporary file, if None.
+                The file to periodically write sampled rows to.
 
         Returns:
             pandas.DataFrame:
@@ -168,8 +167,7 @@ class SingleTablePreset:
                 The batch size to use per attempt at sampling. Defaults to 10 times
                 the number of rows.
             output_file_path (str or None):
-                The file to periodically write sampled rows to. Defaults to
-                a temporary file, if None.
+                The file to periodically write sampled rows to.
 
         Returns:
             pandas.DataFrame:
--- a/sdv/lite/single_table.py
+++ b/sdv/lite/single_table.py
@@ -136,7 +136,7 @@ class SingleTablePreset:
                 The batch size to use per attempt at sampling. Defaults to 10 times
                 the number of rows.
             output_file_path (str or None):
-                The file to periodically write sampled rows to.
+                The file to periodically write sampled rows to. Defaults to None.
 
         Returns:
             pandas.DataFrame:
@@ -167,7 +167,7 @@ class SingleTablePreset:
                 The batch size to use per attempt at sampling. Defaults to 10 times
                 the number of rows.
             output_file_path (str or None):
-                The file to periodically write sampled rows to.
+                The file to periodically write sampled rows to. Defaults to None.
 
         Returns:
             pandas.DataFrame:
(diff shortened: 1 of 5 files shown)


sdv-team:
Task linked: [CU-86b0p02f3 SDV - If no filepath is provided, do not create a file during `sample` #2042](https://app.clickup.com/t/86b0p02f3)

sdv/lite/single_table.py
```diff
@@ -168,8 +167,7 @@ def sample_remaining_columns(self, known_columns, max_tries_per_batch=100,
                 The batch size to use per attempt at sampling. Defaults to 10 times
                 the number of rows.
             output_file_path (str or None):
-                The file to periodically write sampled rows to. Defaults to
-                a temporary file, if None.
+                The file to periodically write sampled rows to.
```
amontanez24, line 170:
```suggestion
                The file to periodically write sampled rows to. Defaults to None.
```

sdv/lite/single_table.py
```diff
@@ -136,8 +136,7 @@ def sample_from_conditions(self, conditions, max_tries_per_batch=100,
...
```
amontanez24, line 139:
```suggestion
                The file to periodically write sampled rows to. Defaults to None.
```

sdv/single_table/base.py
```diff
@@ -935,8 +929,7 @@ def _sample_with_conditions(self, conditions, max_tries_per_batch, batch_size,
             progress_bar (tqdm.tqdm or None):
                 The progress bar to update.
             output_file_path (str or None):
-                The file to periodically write sampled rows to. Defaults to
-                a temporary file, if None.
+                The file to periodically write sampled rows to.
```
amontanez24, line 932:
```suggestion
                The file to periodically write sampled rows to. Defaults to None.
```

sdv/single_table/base.py
```diff
@@ -1059,8 +1052,7 @@ def sample_from_conditions(self, conditions, max_tries_per_batch=100,
             batch_size (int):
                 The batch size to use per sampling call.
             output_file_path (str or None):
-                The file to periodically write sampled rows to. Defaults to
-                a temporary file, if None.
+                The file to periodically write sampled rows to.
```
amontanez24, line 1055:
```suggestion
                The file to periodically write sampled rows to. Defaults to None.
```

amontanez24 (commented):
Some small doc comments but after that looks good!

amontanez24 (approved):

PR #2068, opened 2024-06-13
Set numerical distributions to truncnorm for extension columns

### LINKED ISSUE #1895: If a parent has 0/1 children, HMASynthesizer may create constant data
### Environment Details
* SDV version: 1.11.0

### Error Description
If many rows of a parent table have 0/1 children, then the HMASynthesizer is not well-suited to handle the data. Since data from 0/1 children cannot be aggregated (eg. a correlation cannot be computed), the HMA end up with a lot of NaN values which are then filled with a mean.

This resulting data is hard to model. As a result, the synthetic data is often constant at a min/max value -- the result of clipping out-of-bounds synthetic data.

### Steps to reproduce

```python
import pandas as pd
import numpy as np

parent_table = pd.DataFrame(data={
    'id': [i for i in range(1000)],
    'col_A': list(np.random.choice(['A', 'B', 'C', 'D', 'E'], size=1000))
})

child_table_data = {
    'parent_id': [],
    'col_B': [],
    'col_C': []
}

for i in range(1000):
  num_children = np.random.choice([0, 1, 10, 15], p=[0.4, 0.5, 0.05, 0.05])
  if num_children == 0:
    continue

  child_table_data['parent_id'].extend([i]*num_children)
  child_table_data['col_B'].extend([round(i, 2) for i in np.random.uniform(low=0, high=10, size=num_children)])
  child_table_data['col_C'].extend(list(np.random.choice(['A', 'B', 'C', 'D', 'E'], size=num_children)))

child_table = pd.DataFrame(data=child_table_data)

data = {
    'parent': parent_table,
    'child': child_table
}

metadata = MultiTableMetadata.load_from_dict({
    'tables': {
        'parent': {
            'primary_key': 'id',
            'columns': {
                'id': { 'sdtype': 'id' },
                'col_A': { 'sdtype': 'categorical' }
            }
        },
        'child': {
            'columns': {
                'parent_id': { 'sdtype': 'id' },
                'col_B': { 'sdtype': 'numerical' },
                'col_C': { 'sdtype': 'categorical' }
            }
        }

    },
    'relationships': [{
        'parent_table_name': 'parent',
        'child_table_name': 'child',
        'parent_primary_key': 'id',
        'child_foreign_key': 'parent_id'
    }]
})

synth = HMASynthesizer(metadata)
synth.fit(data)
synthetic_data = synth.sample()
```

Observe how many values in the synthetic child are constant:

image

gsheni:
- Fixes #1895

### DIFF
--- a/sdv/multi_table/hma.py
+++ b/sdv/multi_table/hma.py
@@ -273,7 +273,7 @@ class HMASynthesizer(BaseHierarchicalSampler, BaseMultiTableSynthesizer):
 
     def _set_extended_columns_distributions(self, synthesizer, table_name, valid_columns):
         numerical_distributions = {}
-        for extended_column in self._parent_extended_columns.get(table_name, []):
+        for extended_column in self._parent_extended_columns[table_name]:
             if extended_column in valid_columns:
                 numerical_distributions[extended_column] = DEFAULT_EXTENDED_COLUMNS_DISTRIBUTION
         synthesizer._set_numerical_distributions(numerical_distributions)
--- a/sdv/multi_table/hma.py
+++ b/sdv/multi_table/hma.py
@@ -16,6 +16,7 @@ from sdv.sampling import BaseHierarchicalSampler
 
 LOGGER = logging.getLogger(__name__)
 MAX_NUMBER_OF_COLUMNS = 1000
+DEFAULT_EXTENDED_COLUMNS_DISTRIBUTION = 'truncnorm'
 
 
 class HMASynthesizer(BaseHierarchicalSampler, BaseMultiTableSynthesizer):
@@ -272,14 +273,10 @@ class HMASynthesizer(BaseHierarchicalSampler, BaseMultiTableSynthesizer):
 
     def _set_extended_columns_distributions(self, synthesizer, table_name, valid_columns):
         numerical_distributions = {}
-        if (
-            table_name in self._parent_extended_columns
-            and len(self._parent_extended_columns[table_name]) > 0
-        ):
+        if table_name in self._parent_extended_columns:
             for extended_column in self._parent_extended_columns[table_name]:
                 if extended_column in valid_columns:
-                    numerical_distributions[extended_column] = 'truncnorm'
-
+                    numerical_distributions[extended_column] = DEFAULT_EXTENDED_COLUMNS_DISTRIBUTION
             synthesizer._set_numerical_distributions(numerical_distributions)
 
     def _get_extension(self, child_name, child_table, foreign_key, progress_bar_desc):
--- a/sdv/multi_table/hma.py
+++ b/sdv/multi_table/hma.py
@@ -269,6 +269,18 @@ class HMASynthesizer(BaseHierarchicalSampler, BaseMultiTableSynthesizer):
 
         return processed_data
 
+    def _set_extended_columns_distributions(self, synthesizer,
+                                            table_name,
+                                            valid_columns):
+        numerical_distributions = {}
+        if table_name in self.parent_extended_columns and \
+                len(self.parent_extended_columns[table_name]) > 0:
+            for extended_column in self.parent_extended_columns[table_name]:
+                if extended_column in valid_columns:
+                    numerical_distributions[extended_column] = 'truncnorm'
+
+            synthesizer._set_numerical_distributions(numerical_distributions)
+
     def _get_extension(self, child_name, child_table, foreign_key, progress_bar_desc):
         """Generate the extension columns for this child table.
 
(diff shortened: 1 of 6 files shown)


sdv/multi_table/hma.py
```diff
@@ -398,6 +399,10 @@ def _augment_table(self, table, tables, table_name):
                     self.extended_columns[child_name][column] = FloatFormatter(
                         enforce_min_max_values=True)
                     self.extended_columns[child_name][column].fit(extension, column)
+                if len(extension.columns) > 0:
+                    if table_name not in self.parent_extended_columns:
+                        self.parent_extended_columns[table_name] = []
+                    self.parent_extended_columns[table_name] = list(extension.columns)
```
frances-h, line 405: I think this should be an append, otherwise if you have more than 1 child the 2nd child will overwrite the extended columns from the first

gsheni: Fixed with using extend:
https://github.com/sdv-dev/SDV/pull/2068/commits/e38bc2a2c9eabcd91329166617dea2d3266696f7#diff-c18ffdc4352bb1e367eaefa2e851a0da762907b7bf727f00bc6094446c889beeR419

sdv/multi_table/hma.py
```diff
@@ -159,6 +160,7 @@ def __init__(self, metadata, locales=['en_US'], verbose=True):
         self._augmented_tables = []
         self._learned_relationships = 0
         self._default_parameters = {}
+        self.parent_extended_columns = defaultdict(list)
```
amontanez24, line 163: we can probably make this private

sdv/multi_table/hma.py
```diff
@@ -215,8 +217,8 @@ def _get_distributions(self):
         distributions = {}
         for table in self.metadata.tables:
             parameters = self.get_table_parameters(table)
-            sythesizer_parameter = parameters.get('synthesizer_parameters', {})
-            distributions[table] = sythesizer_parameter.get('default_distribution', None)
+            synthesizer_parameter = parameters.get('synthesizer_parameters', {})
+            distributions[table] = synthesizer_parameter.get('default_distribution', None)
```
amontanez24, line 221:
```suggestion
            synthesizer_parameters = parameters.get('synthesizer_parameters', {})
            distributions[table] = synthesizer_parameters.get('default_distribution', None)
```

sdv/multi_table/hma.py
```diff
@@ -462,6 +481,9 @@ def _model_tables(self, augmented_data):
             )
 
             if not table.empty:
+                self._set_extended_columns_distributions(
+                    self._table_synthesizers[table_name], table_name, table.columns
+                )
```
amontanez24, line 486: why do you need this both here and in `_get_extension`?

sdv/multi_table/hma.py
gsheni: Done:
https://github.com/sdv-dev/SDV/pull/2068/commits/0490d81af6463be113af32a8b7f6cfeb80bd3fcd#diff-c18ffdc4352bb1e367eaefa2e851a0da762907b7bf727f00bc6094446c889beeR163

sdv/multi_table/hma.py
gsheni: Plural added:
https://github.com/sdv-dev/SDV/pull/2068/commits/0490d81af6463be113af32a8b7f6cfeb80bd3fcd#diff-c18ffdc4352bb1e367eaefa2e851a0da762907b7bf727f00bc6094446c889beeR220

sdv/multi_table/hma.py
gsheni: - There is 2 synthesizers that are created. One at the table level (in `_model_tables`) and one per each FK grouping (the grouped child rows in `_get_extension`). 
- We need to call it twice to make sure that in both situations, the extension columns use `truncnorm`. 
- If I called it at just the table level, the synthesizer for the grouped child rows will just be a default GaussianCopula (all distributions set to beta). This would result in the constant values with 0/1 child rows (current behavior).

sdv/multi_table/hma.py
```diff
@@ -268,6 +270,18 @@ def preprocess(self, data):
 
         return processed_data
 
+    def _set_extended_columns_distributions(self, synthesizer, table_name, valid_columns):
+        numerical_distributions = {}
+        if (
+            table_name in self._parent_extended_columns
+            and len(self._parent_extended_columns[table_name]) > 0
+        ):
```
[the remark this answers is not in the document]
frances-h: I don't think we even need the get, if we directly access a key that's not present we'll get back an empty list anyways.

Issue #1803, opened 2024-02-20
Switch to using ruff for Python linting and code formatting


gsheni:
### Problem Description
- We currently use `isort`, `flake8`, and `pydocstyle` for managing out python linting and code formatting.
- However, many libraries have switched to using `ruff`. It is a much faster linter/formatter and can replace the `isort`/`flake8`/`pydocstyle` libraries (among others)

### Expected behavior
- Use `ruff` as the standard linter/formatter

PR #2069, opened 2024-06-14
Add ruff for linting, remove flake8, remove isort, remove pylint

gsheni:
CU-86azdvty7

- Closes https://github.com/sdv-dev/SDV/issues/1803
- Added [ruff](https://github.com/charliermarsh/ruff) for running linting
- Added **ruff** commands to Makefile
- Removed **flake8** as dev dependency, since **ruff** does this, but faster
- Remove **isort** as dev dependency, since **ruff** has the [same function](https://github.com/charliermarsh/ruff#how-does-ruffs-import-sorting-compare-to-isort), but faster
- Removed **pylint** as dev dependency, since **ruff** has [more implemented rules](https://docs.astral.sh/ruff/faq/#how-does-ruffs-linter-compare-to-pylint)
- Removed setup.cfg


sdv-team:
Task linked: [CU-86azdvty7 SDV - Switch to using ruff for Python linting and code formatting #1803](https://app.clickup.com/t/86azdvty7)

pyproject.toml
```diff
@@ -190,19 +165,55 @@ filename = "sdv/__init__.py"
...
 [tool.pytest.ini_options]
 addopts = "--ignore=pyproject.toml"
 
 [build-system]
 requires = ['setuptools', 'wheel']
 build-backend = 'setuptools.build_meta'
+
+[tool.ruff]
+preview = true
+line-length = 100
```
pvk-developer, line 177: I noticed that our other libraries with `Ruff` use this line-length, so I'm fine with increasing it. I'm just curious if there's a new industry standard. 

Fun fact: The [pep-8 standard](https://peps.python.org/pep-0008/#maximum-line-length) was 72-79. However this has been the most 'overwritten' rule.

pvk-developer (approved):
LGTM! I really like this Ruff addition, and as you mentioned and we have seen it, its very fast! 💯

PR #2071, opened 2024-06-14
Added constant for float relative tolerance

rgerkin:
There is currently no user-accessible way to control the float tolerance for conditionally sampling.  While exposing this in the API would be best, this PR at least allows the user to modify this value directly in the module.

N.B. There are actually 2 different values of float_rtol used in the module, one was 0.1 by default and the rest were 0.01.  I set these all to the same constant, since I saw no clear reason why they should be different.

### DIFF
--- a/sdv/single_table/base.py
+++ b/sdv/single_table/base.py
@@ -33,6 +33,7 @@ SYNTHESIZER_LOGGER = get_sdv_logger('SingleTableSynthesizer')
 
 COND_IDX = str(uuid.uuid4())
 FIXED_RNG_SEED = 73251
+FLOAT_RTOL = 0.01
 
 
 class BaseSynthesizer:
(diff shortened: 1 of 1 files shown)


sdv-team:
Hi @rgerkin , it’s great to see your interest in the SDV ecosystem. The SDV software (and its related libraries) is owned and maintained by DataCebo. It is available under the Business Source License for you to browse.

We support a large set of users with enterprise-specific intricacies and reliability needs. This has required us to be deliberate about setting the roadmap for SDV libraries. As a result, we are unable to prioritize reviewing and accepting external pull requests.

Is your PR is meant to address a bug or feature request? If you haven't already, we would greatly appreciate it if you could file an issue instead with the overall description of your problem. We can determine whether it’s aligned with our framework. Once discussed, our team typically resolves smaller issues within a few release cycles. We appreciate your understanding.

PR #2083, opened 2024-06-20
Release notes should not include PRs

### LINKED ISSUE #2074: Release notes should not include PRs
### Problem Description
We currently have an automated script that creates the release notes for each SDV release. This script is meant to pull in the issues that we've included in the release (based off the milestone), and sort them as feature request, bug, or miscellaneous based on the tag.

But in practice, we see that the "Miscellaneous" section actually contains PRs, not issues. This is confusing because each PR should already be attached to an issue -- meaning that the bug or feature it represents is already included in the release notes. It leads to doubly mentioning the bug or feature.

Eg. In [the 1.14.0 release notes](https://github.com/sdv-dev/SDV/releases/tag/v1.14.0) we see PR #2043 linked under Miscellaneous. But this PR is fixing issue #2031 which is already listed under bugs.

### Expected behavior
- Only issues should be tagged in the release notes, not PRs
- The Miscellaneous section should only include issues that are NOT tagged as feature request or bug

amontanez24:
resolves #2074 

This PR also changes the order that each category appears in for the release notes to put features and customer issues at the top

### DIFF
--- a/scripts/release_notes_generator.py
+++ b/scripts/release_notes_generator.py
@@ -25,7 +25,8 @@ ISSUE_LABELS = [
 ]
 ISSUE_LABELS_ORDERED_BY_IMPORTANCE = [
     'feature request',
-    'customer success' 'bug',
+    'customer success',
+    'bug',
     'documentation',
     'internal',
     'maintenance',
--- a/scripts/release_notes_generator.py
+++ b/scripts/release_notes_generator.py
@@ -13,7 +13,7 @@ LABEL_TO_HEADER = {
     'maintenance': 'Maintenance',
     'customer success': 'Customer Success',
     'documentation': 'Documentation',
-    'misc': 'Miscellaneous',
+    'misc': 'Miscellaneous'
 }
 ISSUE_LABELS = [
     'documentation',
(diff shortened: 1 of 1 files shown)


scripts/release_notes_generator.py
```diff
@@ -23,6 +23,13 @@
     'feature request',
     'customer success',
 ]
+ISSUE_LABELS_ORDERED_BY_IMPORTANCE = [
+    'feature request',
+    'customer success' 'bug',
```
gsheni, line 28:
```suggestion
    'customer success',
    'bug',
```

scripts/release_notes_generator.py
```diff
@@ -23,6 +23,14 @@
     'feature request',
     'customer success',
 ]
+ISSUE_LABELS_ORDERED_BY_IMPORTANCE = [
```
gsheni, line 26: Would it be better to just change the order of `ISSUE_LABELS` to match what you want, rather than creating a new global variable?

amontanez24: The other one is a different order because when picking which category to assign a label to, I want the order to be different. For example, if an issue is labeled both a feature request and internal, I want it to be only put in the internal category of the release notes. So the order for `ISSUE_LABELS` is intentionally different

gsheni (approved):

rwedge (approved):

PR #2084, opened 2024-06-20
Add diff printout for ruff format failure

gsheni:


amontanez24 (approved):

tasks.py
```diff
@@ -123,8 +123,13 @@ def tutorials(c):
 def lint(c):
     check_dependencies(c)
     c.run('ruff check .')
-    c.run('ruff format . --check')
+    c.run('ruff format --check --diff .')
 
+@task
+def fix_lint(c):
```
amontanez24, line 129: 🎉

R-Palazzo:
Hi @gsheni, I noticed that `print()` statements are no longer raised when running `invoke lint.` Also running `make fix-lint` will not remove print statements.

amontanez24:
> Hi @gsheni, I noticed that `print()` statements are no longer raised when running `invoke lint.` Also running `make fix-lint` will not remove print statements.

@gsheni If we can we should fix this to match the behavior of the previous lint rules

R-Palazzo (approved):

amontanez24 (approved):

PR #2089, opened 2024-06-25
Numerical unknowns should not be converted to sdv-pii-????

### LINKED ISSUE #2064: HMA sampling crashes when unknown sdtype detected for numerical column
### Environment Details

* SDV version: 1.13.1 and on (maybe before that)
* Python version: Any
* Operating System: Any

### Error Description

If a numerical column is detected as `unknown` in the metadata, then sampling seems to fail because it can't cast the generated value which is a string with the prefix `sdv-pii-` to an int.

### Steps to reproduce

```python
import pandas as pd
import faker
import numpy as np

fake = faker.Faker()

table1 = pd.DataFrame({
    'name': [fake.name() for i in range(20)],
    'salary': np.random.randint(20_000, 250_000, 20),
    'age': np.random.randint(18, 70, 20),
    'address': [fake.address() for i in range(20)]
})
table2 = pd.DataFrame({
    'company': [fake.company() for i in range(20)],
    'employee_count': np.random.randint(15, 4000, 20),
    'revenue': np.random.randint(100_000, 4_000_000_000)
})

tables_dict = {'people': table1, 'company': table2}

metadata.detect_from_dataframes(tables_dict)

synth = HMASynthesizer(metadata)
synth.fit(tables_dict)
synth.sample(1)
```
The code above produces
```
[<ipython-input-4-f0bab0e0379c>](https://localhost:8080/#) in <cell line: 5>()
      3 synth = HMASynthesizer(metadata)
      4 synth.fit(tables_dict)
----> 5 synth.sample(1)

9 frames
[/usr/local/lib/python3.10/dist-packages/pandas/core/dtypes/astype.py](https://localhost:8080/#) in _astype_nansafe(arr, dtype, copy, skipna)
    136     if copy or is_object_dtype(arr.dtype) or is_object_dtype(dtype):
    137         # Explicit copy, or required since NumPy can't view from / to object.
--> 138         return arr.astype(dtype, copy=True)
    139 
    140     return arr.astype(dtype, copy=copy)

ValueError: invalid literal for int() with base 10: 'sdv-pii-cs7y0'
```

lajohn4747:
resolves #2064 
CU-86b0wh849

PII Type adds the prefix `sdv-pii`. If the sdtype is unknown, the transformers will auto-assign but the reverse transform for a `pii` sdtype should always have the dtype of `object` as it will contain a prefix. If a numerical column is detected we will add change the `Faker` function to use `numerify` and cap at the max amount of digits.

Adjusted some unit test to avoid test failures due to mocking

### DIFF
--- a/sdv/data_processing/data_processor.py
+++ b/sdv/data_processing/data_processor.py
@@ -587,13 +587,23 @@ class DataProcessor:
 
             elif sdtype == 'unknown':
                 sdtypes[column] = 'pii'
-                transformers[column] = AnonymizedFaker(
-                    function_name='bothify',
-                )
-                transformers[column].function_kwargs = {
+                function_name = 'bothify'
+                function_kwargs = {
                     'text': 'sdv-pii-?????',
                     'letters': '0123456789abcdefghijklmnopqrstuvwxyz',
                 }
+                if pd.api.types.is_numeric_dtype(data[column]):
+                    digits = data[column].apply(lambda x: len(str(abs(x))))
+                    max_digits = max(digits)
+                    text = '!' * max_digits
+                    function_name = 'numerify'
+                    function_kwargs = {
+                        'text': text,
+                    }
+                transformers[column] = AnonymizedFaker(
+                    function_name=function_name,
+                )
+                transformers[column].function_kwargs = function_kwargs
 
             elif pii:
                 sdtypes[column] = 'pii'
--- a/sdv/data_processing/data_processor.py
+++ b/sdv/data_processing/data_processor.py
@@ -595,7 +595,8 @@ class DataProcessor:
                 if pd.api.types.is_numeric_dtype(data[column]):
                     digits = data[column].apply(lambda x: len(str(abs(x))))
                     max_digits = max(digits)
-                    text = '!' * max_digits
+                    min_digits = min(digits)
+                    text = ('!' * (max_digits - min_digits)) + '%' + ('#' * (min_digits - 1))
                     function_name = 'numerify'
                     function_kwargs = {
                         'text': text,
--- a/sdv/data_processing/data_processor.py
+++ b/sdv/data_processing/data_processor.py
@@ -593,9 +593,8 @@ class DataProcessor:
                     'letters': '0123456789abcdefghijklmnopqrstuvwxyz',
                 }
                 if pd.api.types.is_numeric_dtype(data[column]):
-                    digits = data[column].apply(lambda x: len(str(abs(x))))
-                    max_digits = max(digits)
-                    min_digits = min(digits)
+                    max_digits = len(str(abs(max(data[column]))))
+                    min_digits = len(str(abs(min(data[column]))))
                     text = ('!' * (max_digits - min_digits)) + '%' + ('#' * (min_digits - 1))
                     function_name = 'numerify'
                     function_kwargs = {
--- a/sdv/sampling/hierarchical_sampler.py
+++ b/sdv/sampling/hierarchical_sampler.py
@@ -244,6 +244,13 @@ class BaseHierarchicalSampler:
             synthesizer = self._table_synthesizers.get(table_name)
             dtypes = synthesizer._data_processor._dtypes
             for name, dtype in dtypes.items():
+                is_pii = (
+                    self.metadata.tables.get(table_name, {}).columns.get(name, {}).get('pii', False)
+                )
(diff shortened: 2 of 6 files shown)


sdv-team:
Task linked: [CU-86b0wh849 SDV - HMA sampling crashes when unknown sdtype detected for numerical column #2064](https://app.clickup.com/t/86b0wh849)

sdv/sampling/hierarchical_sampler.py
```diff
@@ -244,6 +244,13 @@ def _finalize(self, sampled_data):
             synthesizer = self._table_synthesizers.get(table_name)
             dtypes = synthesizer._data_processor._dtypes
             for name, dtype in dtypes.items():
+                is_pii = (
+                    self.metadata.tables.get(table_name, {}).columns.get(name, {}).get('pii', False)
+                )
+                # If pii column, the original dtype should be converted to object and not the transformer dtype
+                # as the reverse transform will remain in a pii format
+                if is_pii:
+                    dtype = 'object'
```
pvk-developer, line 253: just a nitpick, add a breakline here

pvk-developer (approved):

R-Palazzo (approved):
LGTM! Just checking, it it an issue for HMA only? @lajohn4747, @amontanez24

lajohn4747:
> LGTM! Just checking, it it an issue for HMA only? @lajohn4747, @amontanez24

It appears to be that way after some local testing

amontanez24 (changes requested):
I don't think we should force the reversed type to be object. The reversed type should match whatever the input was. In the case where the original type was numerical, I don't think we should be adding that prefix

sdv/sampling/hierarchical_sampler.py
```diff
@@ -241,7 +241,13 @@ def _finalize(self, sampled_data):
             synthesizer = self._table_synthesizers.get(table_name)
             dtypes = synthesizer._data_processor._dtypes
             for name, dtype in dtypes.items():
-                table_rows[name] = table_rows[name].dropna().astype(dtype)
+                try:
+                    table_rows[name] = table_rows[name].dropna().astype(dtype)
+                except Exception:
+                    LOGGER.info(
+                        'Could not cast back to transformer dtype, keeping original typing.'
```
amontanez24, line 248:
```suggestion
                        'Could not cast back to column's original dtype, keeping original typing.'
```

sdv/data_processing/data_processor.py
```diff
@@ -587,13 +587,24 @@ def _create_config(self, data, columns_created_by_constraints):
...
-                transformers[column].function_kwargs = {
+                function_name = 'bothify'
+                function_kwargs = {
                     'text': 'sdv-pii-?????',
                     'letters': '0123456789abcdefghijklmnopqrstuvwxyz',
                 }
+                if pd.api.types.is_numeric_dtype(data[column]):
+                    digits = data[column].apply(lambda x: len(str(abs(x))))
+                    max_digits = max(digits)
+                    min_digits = min(digits)
```
amontanez24, line 598: instead of applying to the whole column, why don't we just get the min and max of the column and only compute the digits for those? I think the apply function is kind of slow especially when it has to convert everything to strings

amontanez24, line 598: can we add some unit tests for the digit logic?

amontanez24 (changes requested):

sdv/sampling/hierarchical_sampler.py
```diff
@@ -241,7 +241,13 @@ def _finalize(self, sampled_data):
...
             dtypes = synthesizer._data_processor._dtypes
             for name, dtype in dtypes.items():
-                table_rows[name] = table_rows[name].dropna().astype(dtype)
+                try:
+                    table_rows[name] = table_rows[name].dropna().astype(dtype)
+                except Exception:
+                    LOGGER.info(
+                        "Could not cast back to column's original dtype, keeping original typing."
+                    )
+                    table_rows[name] = table_rows[name].dropna()
```
amontanez24, line 250: let's add unit tests for this as well

amontanez24 (commented):
Can we also add an integration test with the example in the issue

sdv/sampling/hierarchical_sampler.py
lajohn4747: Sounds good, added both unit test for this try and catch and an integration test.

amontanez24 (approved):

PR #2095, opened 2024-06-27
Set max string length for ruff

gsheni:


pyproject.toml
```diff
@@ -225,7 +225,11 @@ lines-between-types = 0
 [tool.ruff.lint.per-file-ignores]
 "__init__.py" = ["F401", "E402", "F403", "F405", "E501", "I001"]
 "errors.py" = ["D105"]
-"tests/**.py" = ["D"]
+"tests/**.py" = ["D", "W505"]
```
amontanez24, line 228: why are we ignoring for the tests?

gsheni: - When we enabled the length checking, I got error from the test directory.
- I figured that checking docstring length check (`W505`) was not important for the tests directory.

amontanez24: I think we should apply it everywhere. The main use case as I see it is to make it easier to read and review, so I think tests would still have that problem

gsheni: Fixed:
https://github.com/sdv-dev/SDV/pull/2095/commits/02b804a5f2a397e4ca7179a5dc58f56980a193d6

amontanez24 (approved):

pvk-developer (approved):

tests/unit/sequential/test_par.py
```diff
@@ -469,8 +469,8 @@ def test__fit_context_model_with_context_columns(self, gaussian_copula_mock):
 
     @patch('sdv.sequential.par.PARSynthesizer.get_transformers')
     def test_auto_assign_transformers_without_enforce_min_max(self, mock_get_transfomers):
-        """Test to see if auto_assign_transformers does not add enforce_min_max_values if the transformer
-        does not contain it already
+        """Test to see if auto_assign_transformers does not add enforce_min_max_values
+        if the transformer does not contain it already
```
pvk-developer, line 473: I think that this should be failing 🤔

pyproject.toml
```diff
@@ -229,3 +229,7 @@ lines-between-types = 0
 
 [tool.ruff.lint.pydocstyle]
 convention = "google"
+
+[tool.ruff.lint.pycodestyle]
+max-doc-length = 100
+max-line-length = 100
```
R-Palazzo, line 235: Should it be 100 or 99? On Enterprise it's 99

PR #2100, opened 2024-06-28
Support null foreign key values in HMA

### LINKED ISSUE #2063: Support nullable foreign keys in HMA
### Problem Description

HMASynthesizer does not currently support null values in foreign key columns.  Adding the ability to handle null values for foreign keys would expand the range of datasets HMA can model. 

### Expected behavior

HMASynthesizer is able to fit on data that contains null values in foreign key columns and the presence of nulls is reflected in the sampled data.

### Additional context

Changes to the fit process:
When generating extension columns for a child table, treat null as a valid foreign key value and calculate the extension row values for a null parent.  Store these null parent extension values separately from the parent table but still have them retrievable for sampling.

Changes to sampling:
When creating a child table, HMA should leave some rows to be generated using the null parent's child synthesizer based on the percentage of null foreign keys in the relationship being used to create the child table (perhaps this could be handled in `_enforce_table_size`).
When finding parent_ids of other foreign key columns on a child table, treat the stored null parent extension row as another parent candidate and create a corresponding synthesizer to get likelihoods from.

rwedge:
Fixes #2063
CU-86b0wdby9

### DIFF
--- a/tests/integration/multi_table/test_hma.py
+++ b/tests/integration/multi_table/test_hma.py
@@ -1314,7 +1314,7 @@ class TestHMASynthesizer:
         assert len(record) == 1
 
     def test_null_foreign_keys(self):
-        """Test that the synthesizer crashes when there are null foreign keys."""
+        """Test that the synthesizer does not crash when there are null foreign keys."""
         # Setup
         metadata = MultiTableMetadata()
         metadata.add_table('parent_table')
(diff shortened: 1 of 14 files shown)


sdv-team:
Task linked: [CU-86b0wdby9 SDV - Support nullable foreign keys in HMA #2063](https://app.clickup.com/t/86b0wdby9)

tests/integration/multi_table/test_hma.py
```diff
@@ -1355,11 +1356,12 @@ def test_null_foreign_keys(self):
 
         data = {
             'parent_table': pd.DataFrame({'id': [1, 2, 3]}),
-            'child_table1': pd.DataFrame({'id': [1, 2, 3], 'fk': [1, 2, np.nan]}),
+            # 'child_table1': pd.DataFrame({'id': [1, 2, 3], 'fk': [1, 2, np.nan]}),
             'child_table2': pd.DataFrame({
                 'id': [1, 2, 3],
                 'fk1': [1, 2, np.nan],
                 'fk2': [1, 2, np.nan],
+                'cat_type': ['siamese','persian', 'american shorthair'],
```
gsheni, line 1364:
```suggestion
                'cat_type': pd.Series(['siamese','persian', 'american shorthair'], dtype="object")
```

Issue #2015, opened 2024-05-16
Use Github Action Caching to Speed Up Unit and Integration Tests


gsheni:
### Problem Description
- Our unit, integration, and minimum tests currently take a long time to complete. 
  - Longest integration test - `16m 22s` (3.12, windows-latest) 
  - Longest unit test - `8m 31s` (3.10, windows-latest)
  - Longest minimum test - `17m 10s` (3.11, windows-latest)

### Expected behavior
- We can cache our python dependencies to speed up our workflow. 
- Specifically, we can cache for integration, unit, minimum, and lint workflows.
- The change would look something like this:
```
      - uses: actions/checkout@v4
      - name: Set up python ${{ matrix.python-version }}
        uses: actions/setup-python@v5
        with:
          python-version: ${{ matrix.python-version }}
          cache: 'pip'
          cache-dependency-path: "pyproject.toml"
      - uses: actions/cache@v3
        id: cache
        with:
          path: ${{ env.pythonLocation }}
          key: py-${{ matrix.python-version }}-tests-${{ hashFiles('pyproject.toml') }}-v02
      - name: Install all deps (if cache has changed)
        if: (steps.cache.outputs.cache-hit != 'true')
        run: |
          make install-all
      - name: Install all deps (if cache has not changed)
        if: (steps.cache.outputs.cache-hit == 'true')
        run:  |
          python -m pip install --no-dependencies .
```

### Additional context
- https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows

lajohn4747:
Appears to be minimal gain with caching and we have a limit to how much we can cache for tests. Closing without merging since the investigation results do not warrant the addition.

PR #2107, opened 2024-07-03
Try out caching pip

lajohn4747:
part of #2015 
Longest minimum test (3.11, windows-latest)- **17m 10s** went down to **14m 41s**

### DIFF
--- a/.github/workflows/minimum.yml
+++ b/.github/workflows/minimum.yml
@@ -22,12 +22,9 @@ jobs:
     - uses: actions/checkout@v4
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v5
-      with:
-        python-version: ${{ matrix.python-version }}
-        cache: 'pip'
-        cache-dependency-path: "pyproject.toml"
-    - name: Cache pip
+    - name: Cache pip on windows
       id: cache-pip
+      if: runner.os == 'Windows'
       uses: actions/cache@v3
       with:
         path: ~/.cache/pip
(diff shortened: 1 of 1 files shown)


.github/workflows/minimum.yml
```diff
@@ -24,6 +24,19 @@ jobs:
...
+        path: C:\Users\runneradmin\AppData\Local\pip\Cache
+        key: ${{ runner.os }}-pip-${{ matrix.python-version }}-${{ hashFiles('**/pyproject.toml') }}
+        restore-keys: |
+            ${{ runner.os }}-pip-${{ matrix.python-version }}-
+    - name: Check cache hit
+      if: runner.os == 'Windows'
+      run: |
+        echo "Cache hit status: ${{ steps.cache-pip.outputs.cache-hit }}"
     - name: Install dependencies
       run: |
```
gsheni, line 41: You can use the cache to determine how to install dependencies:
```suggestion
    - name: Install SDV (cache has not changed)
      if: steps.cache-pip.outputs.cache-hit == 'true'
      run:  |
        python -m pip install --no-dependencies .
    - name: Install SDV and dependencies (cache has changed)
      if: steps.cache-pip.outputs.cache-hit != 'true'
      run: |
```

lajohn4747: Cannot call the force command `--no-dependencies`, as that command ignores installing dependencies altogether ([Job example](https://github.com/sdv-dev/SDV/actions/runs/9878701812/job/27283222021)). 

Pip looks into the cache location path and will directly install those files rather than re-download them so we don't need to have separate steps. Example [here](https://github.com/sdv-dev/SDV/actions/runs/9876951679/job/27277353093): we can see it using the cache and cut down the time for windows installation by not downloading. (~2-3 mins saved)

lajohn4747:
Appears to be minimal gain with caching and we have a limit to how much we can cache for tests. Closing without merging since the investigation results do not warrant the addition.

PR #2108, opened 2024-07-03
Add error message if sequence key present in context columns

### LINKED ISSUE #2097: Context column cannot be a sequence key: Need better error message for this case
### Environment Details

* SDV version: 1.14.0 (latest)

### Error Description
For sequential data, it should not be possible for the sequence key to be the same as a context column. This is because the sequence key is an identifier for each sequence, whereas a context column is just another column that _happens to never vary within a sequence_. There is no need to declare a sequence key as a context column because a sequence key is already guaranteed not to vary within a sequence -- rather, it is _defining_ what a sequence is.

Yet, the code somehow allows me to instantiate a `PARSynthesizer` with a context column the same as the sequence key. When I try to fit it, I get an error that is not really relevant to the issue. 

```python
import pandas as pd

    'columns': {
        'A': { 'sdtype': 'id' },
        'B': { 'sdtype': 'datetime', 'datetime_format': '%Y-%m-%d' },
        'C': { 'sdtype': 'numerical' },
        'D': { 'sdtype': 'categorical' }
    },
    'sequence_key': 'A'
})

data = pd.DataFrame(data={
    'A': [0, 0, 0, 1, 1, 1],
    'B': ['2020-03-02', '2020-03-04', '2020-03-05', '2020-03-01', '2020-03-03', '2020-03-06'],
    'C': [12, 13, 34, 10, 45, 21],
    'D': ['Yes', 'Yes', 'Yes', 'No', 'No', 'No']
})

synth = PARSynthesizer(metadata, context_columns=['A'])
synth.fit(data)
```

Error:
```
[/usr/local/lib/python3.10/dist-packages/sdv/sequential/par.py](https://localhost:8080/#) in update_transformers(self, column_name_to_transformer)
    298         """
    299         if set(column_name_to_transformer).intersection(set(self.context_columns)):
--> 300             raise SynthesizerInputError(
    301                 'Transformers for context columns are not allowed to be updated.')
    302 

SynthesizerInputError: Transformers for context columns are not allowed to be updated.
```

### Expected Behavior
I should not be allowed to even instantiate a `PARSynthesizer` if any of the context columns are sequence keys. This should immediately throw an error explaining that it is not allowed.

```python
synth = PARSynthesizer(metadata, context_columns=['A'])
```
```
SynthesizerInputError: The sequence key ('A') cannot be a context column. To proceed, please remove the sequence key from the 'context_columns' parameter.
```

gsheni:
CU-86b11ypbf
- Fixes #2097

### DIFF
--- a/sdv/sequential/par.py
+++ b/sdv/sequential/par.py
@@ -195,9 +195,15 @@ class PARSynthesizer(LossValuesMixin, BaseSynthesizer):
         """Error that tells the user custom constraints can't be used in the ``PARSynthesizer``."""
         raise SynthesizerInputError('The PARSynthesizer cannot accommodate custom constraints.')
 
-    def _validate_sequence_key_and_context_columns(
-        self, sequence_key: list[str], context_columns: list[str]
-    ):
+    def _validate_sequence_key_and_context_columns(self, sequence_key, context_columns):
+        """Check that the sequence key is not present in the context colums.
+
+        Args:
+            sequence_key (list[str]):
+                A list of column that identify which row(s) belong to which sequences.
+            context_columns (list[str]):
+                A list of strings, representing the columns that do not vary in a sequence.
+        """
         if set(sequence_key).intersection(set(context_columns)):
             raise SynthesizerInputError(
                 f'The sequence key {sequence_key} cannot be a context column. '
--- a/sdv/sequential/par.py
+++ b/sdv/sequential/par.py
@@ -195,7 +195,9 @@ class PARSynthesizer(LossValuesMixin, BaseSynthesizer):
         """Error that tells the user custom constraints can't be used in the ``PARSynthesizer``."""
         raise SynthesizerInputError('The PARSynthesizer cannot accommodate custom constraints.')
 
-    def _validate_sequence_key_and_context_columns(self, sequence_key, context_columns):
+    def _validate_sequence_key_and_context_columns(
+        self, sequence_key: list[str], context_columns: list[str]
+    ):
         if set(sequence_key).intersection(set(context_columns)):
             raise SynthesizerInputError(
                 f'The sequence key {sequence_key} cannot be a context column. '
--- a/sdv/sequential/par.py
+++ b/sdv/sequential/par.py
@@ -110,6 +110,7 @@ class PARSynthesizer(LossValuesMixin, BaseSynthesizer):
 
         self._sequence_index = self.metadata.sequence_index
         self.context_columns = context_columns or []
+        self._validate_sequence_key_and_context_columns(self._sequence_key, self.context_columns)
         self._extra_context_columns = {}
         self.extended_columns = {}
         self.segment_size = segment_size
@@ -193,6 +194,13 @@ class PARSynthesizer(LossValuesMixin, BaseSynthesizer):
     def add_custom_constraint_class(self, class_object, class_name):
         """Error that tells the user custom constraints can't be used in the ``PARSynthesizer``."""
         raise SynthesizerInputError('The PARSynthesizer cannot accommodate custom constraints.')
+
+    def _validate_sequence_key_and_context_columns(self, sequence_key, context_columns):
+        if set(sequence_key).intersection(set(context_columns)):
+            raise SynthesizerInputError(
+                f'The sequence key {self._sequence_key} cannot be a context column. '
+                'To proceed, please remove the sequence key from the context_columns parameter.'
+            )
 
     def _validate_context_columns(self, data):
         errors = []
--- a/sdv/sequential/par.py
+++ b/sdv/sequential/par.py
@@ -198,7 +198,7 @@ class PARSynthesizer(LossValuesMixin, BaseSynthesizer):
     def _validate_sequence_key_and_context_columns(self, sequence_key, context_columns):
         if set(sequence_key).intersection(set(context_columns)):
             raise SynthesizerInputError(
-                f'The sequence key {self._sequence_key} cannot be a context column. '
+                f'The sequence key {sequence_key} cannot be a context column. '
                 'To proceed, please remove the sequence key from the context_columns parameter.'
             )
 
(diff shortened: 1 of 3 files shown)


sdv-team:
Task linked: [CU-86b11ypbf SDV - Context column cannot be a sequence key: Need better error message for this case #2097](https://app.clickup.com/t/86b11ypbf)

tests/integration/sequential/test_par.py
```diff
@@ -385,3 +385,20 @@ def test_par_sequence_index_is_numerical():
     s1.fit(data)
     sample = s1.sample(2, 5)
     assert sample.columns.to_list() == data.columns.to_list()
+
+
+def test_par_error_on_context_columns():
+    metadata_dict = {
```
lajohn4747, line 391: Can you add docstring for this test and the unit test along with sectioning the test code with:
`#Setup`, `#Run`, `#Assert`

sdv/sequential/par.py
```diff
@@ -194,6 +195,13 @@ def add_custom_constraint_class(self, class_object, class_name):
         """Error that tells the user custom constraints can't be used in the ``PARSynthesizer``."""
         raise SynthesizerInputError('The PARSynthesizer cannot accommodate custom constraints.')
 
+    def _validate_sequence_key_and_context_columns(self, sequence_key, context_columns):
+        if set(sequence_key).intersection(set(context_columns)):
```
pvk-developer, line 199: I think we need to use `set([sequence_key])`. If you pass a `string` to the `set()` you will get the following output:
```python
In [1]: set('abc')
Out[1]: {'a', 'b', 'c'}
```

tests/integration/sequential/test_par.py
```diff
@@ -385,3 +385,20 @@ def test_par_sequence_index_is_numerical():
...
+        'columns': {
+            'A': {'sdtype': 'id'},
+            'B': {'sdtype': 'datetime', 'datetime_format': '%Y-%m-%d'},
+        },
+        'sequence_key': 'A',
```
pvk-developer, line 396: can we change `sequence_key` to something with more letters, if you do `set('A')` it will return what you are expecting but if you do `set('AB')` it will return `{'A', 'B'}`.

pvk-developer (changes requested):

sdv/sequential/par.py
gsheni: This function actually gets a list, the sequence_key is cast to an iterate in the `init`:
https://github.com/sdv-dev/SDV/blob/2097-context-column-cannot-be-a-sequence-key-need-better-error-message-for-this-case/sdv/sequential/par.py#L104

tests/integration/sequential/test_par.py
gsheni: Currently, sdv can only be a single column:
> The sequence key is a column that identify which row(s) belong to which sequences... At this time, SDV does not support composite keys.
- https://docs.sdv.dev/sdv/sequential-data/data-preparation/sequential-metadata-api#set_sequence_key
- Given this, I don't think it makes sense to make sequence_key multiple column names (which would imply composite keys)

tests/integration/sequential/test_par.py
gsheni: Added:
https://github.com/sdv-dev/SDV/pull/2108/commits/ab3c87b279dc66bbd177cb6803c5881f748dd745

sdv/sequential/par.py
gsheni: I added typing to the function to make this clear:
https://github.com/sdv-dev/SDV/pull/2108/commits/ab3c87b279dc66bbd177cb6803c5881f748dd745#diff-f97bec45b00fd2731a7c766f7d860080f74d4355c407317cd4cea7ec7d38274fR199

gsheni: I added a docstring with Args to make this clear:
https://github.com/sdv-dev/SDV/pull/2108/commits/fd50434b8d1f0ff423cc9e58495c53f9ad969bf1

sdv/sequential/par.py
```diff
@@ -110,6 +110,7 @@ def __init__(
 
         self._sequence_index = self.metadata.sequence_index
         self.context_columns = context_columns or []
+        self._validate_sequence_key_and_context_columns(self._sequence_key, self.context_columns)
```
amontanez24, line 113: You don't really need to pass these parameters since they're both on self right?

tests/unit/sequential/test_par.py
```diff
@@ -935,3 +931,18 @@ def test_load(self, mock_file, cloudpickle_mock):
         mock_file.assert_called_once_with('synth.pkl', 'rb')
         cloudpickle_mock.load.assert_called_once_with(mock_file.return_value)
         assert loaded_instance == synthesizer_mock
+
+    def test__par_error_on_context_columns(self):
```
amontanez24, line 935: nitpick: our test names usually go `test_{function_name}_{description}` so this one would be like
`test___init___error_sequence_key_in_context` or something like that

tests/integration/sequential/test_par.py
pvk-developer: Yeah I was referencing my previous comment about the `set(str)` but since this is being automatically casted to iter by `ParSynthesizer` I think is good to go!

pvk-developer (approved):

sdv/sequential/par.py
gsheni: Fixed:
https://github.com/sdv-dev/SDV/pull/2108/commits/4c11961d057bf6de3e4b6aca6bd2fe8bcbabe173#diff-f97bec45b00fd2731a7c766f7d860080f74d4355c407317cd4cea7ec7d38274fL113

tests/unit/sequential/test_par.py
gsheni: Fixed:
https://github.com/sdv-dev/SDV/pull/2108/commits/4c11961d057bf6de3e4b6aca6bd2fe8bcbabe173#diff-8f23874079228dbdef42b9e937b8c44ac58a135681b007cabe5d08b187388d69R390

amontanez24 (approved):

PR #2132, opened 2024-07-15
Separate primary key detection functionality

### LINKED ISSUE #2101: Separate primary key detection functionality
### Problem Description

_As a user, I'd like to be able to detect a subset of the metadata based on my needs. For example, I may only want to detect sdtypes, or only detect primary keys._

### Expected behavior

 We want to split out the functionality for detecting the primary key so that it can be called in its own function.

#### Acceptance criteria

- There should be a private function `_detect_primary_key(self, data)` that will return the guessed primary key

### Additional context

- Currently, our logic for primary key detection is wrapped up in sdtype detection. All of this logic happens in the [_detect_columns ](https://github.com/sdv-dev/SDV/blob/cb49b38693f70e539b4303e919bdb14cc8c71e0f/sdv/metadata/single_table.py#L541) function. We originally did these together to be more efficient since we already looped over all the columns to determine the sdtype and could figure out if a column was a primary key candidate at the same time. Now we should separate this so that it is more modular and also since we may change the logic to determine the primary key anyway.

amontanez24:
resolves #2101

### DIFF
--- a/sdv/metadata/single_table.py
+++ b/sdv/metadata/single_table.py
@@ -209,6 +209,7 @@ class SingleTableMetadata:
         self.column_relationships = []
         self._version = self.METADATA_SPEC_VERSION
         self._updated = False
+        self._primary_key_candidates = None
 
     def _get_unexpected_kwargs(self, sdtype, **kwargs):
         expected_kwargs = self._SDTYPE_KWARGS.get(sdtype, ['pii'])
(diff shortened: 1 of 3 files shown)


sdv/metadata/single_table.py
```diff
@@ -538,20 +538,72 @@ def _determine_sdtype_for_objects(self, data):
...
+            str:
+                The column name of the selected primary key.
+
+        Raises:
+            RuntimeError:
+                If the sdtypes for all columns haven't been detected or set yet.
+        """
+        original_columns = data.columns
+        stringified_columns = data.columns.astype(str)
+        data.columns = stringified_columns
```
frances-h, line 563: Do we need to do this here and in `_detect_columns`?

amontanez24: I think so because the column names that get added to `self.columns` seem to be the [stringified ones](https://github.com/sdv-dev/SDV/blob/2abacb192192182cbe9a712f8e5df3d75db3c480/sdv/metadata/single_table.py#L592). So if we don't do it here then we might get a `KeyError`

amontanez24: Actually now that I think about it, this could be a problem for every other function that takes data and looks in the columns. Maybe we should change the dict to store the original column names instead

frances-h: But we modified the data so the column names are the stringified ones, no? Then at the end we convert it back to the original column names.

frances-h: We also do it at the beginning of `_detect_columns` and then restore them at the end

amontanez24: Yeah but the idea is eventually people will call this method on their own

pvk-developer (approved):

frances-h (approved):

PR #2135, opened 2024-07-16
Add Metadata Class

### LINKED ISSUE #2104: Create unified Metadata class
### Problem Description

_As a user, it can be confusing to figure out which metadata to provide for different scenarios and how to format it. Since majority of the metadata is the same for all cases, it would be nice to only have one object and format to learn._

In an effort to streamline SDV workflows, we want to unify the metadata. This means we will just have one `Metadata` class that all synthesizers will take, whether single table, multi table or sequential.

### Expected behavior

- Do this on new `feature/metadata` branch
- Create a new class called `Metadata`
    - This class should be pretty much the same as the `MultiTableMetadata` class with a couple adjustments
        1. The METADATA_SPEC_VERSION should now be `V1` instead of `MULTI_TABLE_V1`
        2. The `load_from_json` method should be able to load both single and multi table jsons into this new class.
        3. The validate function should not crash if there are no relationships
        4. The validate_data function should not crash if there are no relationships

```python

single_table_metadata = Metadata.load_from_json('single_table_metadata.json')
multi_table_metadata = Metadata.load_from_json('multi_table_metadata.json')
```

### Additional context

- We should try to be DRY and reuse code. One option is to have the new synthesizer inherit from [MultiTableMetadata](https://github.com/sdv-dev/SDV/blob/main/sdv/metadata/multi_table.py) and just override the appropriate fields and methods.
    - We need to support backwards compatibility with the old metadata classes, meaning if someone has a pickled synthesizer with the old metadata, it shouldn't crash. An alternate approach might be to move the code to the new class and have `MultiTableMetadata` inherit from it. If we do this, make sure that pickled `MultiTableMetadata` objects still load and work properly.
- There will be follow up issues to make the synthesizers work with this new object.

lajohn4747:
resolves #2104 
CU-86b147dh4

Adds metadata class that inherits `MultiTableMetadata` for backward compatibility and adds tests to confirm it works fine.

### DIFF
--- a/sdv/metadata/metadata.py
+++ b/sdv/metadata/metadata.py
@@ -2,7 +2,6 @@
 
 from pathlib import Path
 
-from sdv.metadata.errors import InvalidMetadataError
 from sdv.metadata.multi_table import MultiTableMetadata
 from sdv.metadata.single_table import SingleTableMetadata
 from sdv.metadata.utils import read_json
@@ -12,10 +11,6 @@ class Metadata(MultiTableMetadata):
     """Metadata class that handles all metadata."""
 
     METADATA_SPEC_VERSION = 'V1'
-
-    def __init__(self):
-        super().__init__()
-        self._is_multi_table = None
 
     @classmethod
     def load_from_json(cls, filepath):
@@ -30,7 +25,7 @@ class Metadata(MultiTableMetadata):
             - An ``Error`` if the ``json`` file does not contain the ``METADATA_SPEC_VERSION``.
 
         Returns:
-            A ``MultiTableMetadata`` instance.
+            A ``Metadata`` instance.
         """
         filename = Path(filepath).stem
         metadata = read_json(filepath)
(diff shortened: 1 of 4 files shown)


sdv-team:
Task linked: [CU-86b147dh4 SDV - Create unified Metadata class #2104](https://app.clickup.com/t/86b147dh4)

sdv/metadata/metadata.py
```diff
@@ -0,0 +1,89 @@
...
+        if self._is_multi_table:
+            super()._set_metadata_dict(metadata)
+        else:
+            if single_table_name is None:
+                single_table_name = 'default_table_name'
+
+            self.tables[single_table_name] = SingleTableMetadata.load_from_dict(metadata)
+
+    def is_multi_table(self):
+        """Return of this is metadata represents multi-table.
```
amontanez24, line 81:
```suggestion
        """Check if metadata is for multi_table data.
```

sdv/metadata/metadata.py
```diff
@@ -0,0 +1,89 @@
...
+        Args:
+            filepath (str):
+                String that represents the ``path`` to the ``json`` file.
+
+        Raises:
+            - An ``Error`` if the path does not exist.
+            - An ``Error`` if the ``json`` file does not contain the ``METADATA_SPEC_VERSION``.
+
+        Returns:
+            A ``MultiTableMetadata`` instance.
```
amontanez24, line 33: This should be ``Metadata`` now

sdv/metadata/metadata.py
```diff
@@ -0,0 +1,89 @@
...
+
+        if self._is_multi_table:
+            super()._set_metadata_dict(metadata)
+        else:
+            if single_table_name is None:
+                single_table_name = 'default_table_name'
+
+            self.tables[single_table_name] = SingleTableMetadata.load_from_dict(metadata)
+
+    def is_multi_table(self):
```
amontanez24, line 80: I think you can get rid of this method.

sdv/metadata/metadata.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+class Metadata(MultiTableMetadata):
+    """Metadata class that handles all metadata."""
+
+    METADATA_SPEC_VERSION = 'V1'
+
+    def __init__(self):
+        super().__init__()
+        self._is_multi_table = None
```
amontanez24, line 18: I don't think we need the attribute either

amontanez24 (approved):
:shipit:

PR #2140, opened 2024-07-18
Make `Metadata` work with single table synthesizers

### LINKED ISSUE #2128: Enable single table synthesizers to use new Metadata
### Problem Description

_As a user, I'd like to easily create synthetic data for tables in my data once I have metadata independent of modality._

Currently, we have two different types of metadata for single and multi table use cases. This can be confusing and cumbersome for users. Once metadata is created, they should be able to use it with whichever synthesizer will work for their scenario.

### Expected behavior
 Once #2104 is completed, we need to enable all single table and sequential synthesizers to work with the new metadata class.

- All single table and sequential synthesizers should run when passed the new metadata object in the init.
- When a SingleTableSynthesizer is initialized with the new Metadata, check to make sure there is only one table. Otherwise, crash and ask user to use a MultiTableSynthesizer
- Update all references to support both the SingleTableMetadata object and the new Metadata object. Run all integration tests using the new Metadata class and see what failures occur.
    - Update any parts of the code where the attribute might be different to support both the new class and the old SingleTableMetadata.
    - For example, in the DataProcessor, we access the metadata.columns directly. The new object will require you to go into the table first.
- Update unit tests to continue to have 100% coverage
- Deprecate SingleTableMetadata
    - Raise a future warning if users pass old metadata
- The synthesizer should be backwards compatible (Still work with the old metadata)

### Additional context

There will be a lot of references that might change between the new class and the SingleTableMetadata class since many attributes will now be nested. I propose the following strategy for handling these changes
- For constraints and the DataProcessor, use the underlying SingleTableMetadata object. Since the MultiTableMetadata class is made up of a dictionary of SingleTableMetadata classes, you can extract that class for the one table and pass it to the DataProcessor and constraints.
- In the SingleTableSynthesizer classes themselves (Base, GaussianCopula etc.), update the references to work with both the new and old classes. There aren’t as many references in these files so it should be more straightforward.
- **Note:** There is a separate issue for getting the evaluation methods to work

lajohn4747:
resolves #2128 
CU-86b19amxa

Made metadata work with single table synthesizers by converting it to a `SingleTableMetadata` before inserting into the synthesizers.

Added a variable called `.real_metadata` that is used to hold the `Metadata` version. This will help with phasing out `SIngleTableMetadata` in the future. `get_metadata` is updated to always return the `real_metadata` which should be of type class `Metadata` for interface consistency. Altered and fixed some tests so they ensure `Metadata` is the type that is produced.

This approach allows us to keep backward compatibility with `SingleTableMetadata` but still force the move over to `Metadata`. We can converge methods later without the need of creating branch of metadata handling in all the our synthesizers.

### DIFF
--- a/sdv/metadata/metadata.py
+++ b/sdv/metadata/metadata.py
@@ -51,25 +51,6 @@ class Metadata(MultiTableMetadata):
         instance._set_metadata_dict(metadata_dict, single_table_name)
         return instance
 
-    @classmethod
-    def load_from_single_table_metadata(cls, single_metadata_table, table_name=None):
-        """Return a unified Metadata object from a legacy SingleTableMetadata object.
-
-        Args:
-            single_metadata_table (SingleTableMetadata):
-                ``SingleTableMetadata`` object to be converted to a ``Metadata`` object.
-            table_name (string):
-                The name of the table that will be stored in the ``Metadata object.
-
-        Returns:
-            Instance of ``Metadata``.
-        """
-        if not isinstance(single_metadata_table, SingleTableMetadata):
-            raise InvalidMetadataError('Cannot convert given legacy metadata')
-        instance = cls()
-        instance._set_metadata_dict(single_metadata_table.to_dict())
-        return instance
-
     def _set_metadata_dict(self, metadata, single_table_name=None):
         """Set a ``metadata`` dictionary to the current instance.
 
@@ -89,18 +70,12 @@ class Metadata(MultiTableMetadata):
         else:
             if single_table_name is None:
                 single_table_name = 'default_table_name'
-
             self.tables[single_table_name] = SingleTableMetadata.load_from_dict(metadata)
 
     def _convert_to_single_table(self):
-        is_multi_table = len(self.tables) > 1
-        if is_multi_table:
+        if len(self.tables) > 1:
             raise InvalidMetadataError(
                 'Metadata contains more than one table, use a MultiTableSynthesizer instead.'
             )
 
-        if len(self.tables) == 0:
-            return SingleTableMetadata()
-
-        single_table_metadata = next(iter(self.tables.values()))
-        return single_table_metadata
+        return next(iter(self.tables.values()), SingleTableMetadata())
--- a/sdv/metadata/metadata.py
+++ b/sdv/metadata/metadata.py
@@ -2,7 +2,6 @@
 
 from pathlib import Path
 
-from sdv.metadata.errors import InvalidMetadataError
 from sdv.metadata.multi_table import MultiTableMetadata
 from sdv.metadata.single_table import SingleTableMetadata
 from sdv.metadata.utils import read_json
@@ -12,10 +11,6 @@ class Metadata(MultiTableMetadata):
     """Metadata class that handles all metadata."""
 
     METADATA_SPEC_VERSION = 'V1'
-
-    def __init__(self):
-        super().__init__()
-        self._is_multi_table = None
 
     @classmethod
     def load_from_json(cls, filepath):
@@ -30,7 +25,7 @@ class Metadata(MultiTableMetadata):
             - An ``Error`` if the ``json`` file does not contain the ``METADATA_SPEC_VERSION``.
 
         Returns:
-            A ``MultiTableMetadata`` instance.
+            A ``Metadata`` instance.
         """
         filename = Path(filepath).stem
         metadata = read_json(filepath)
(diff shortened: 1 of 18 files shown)


sdv/single_table/base.py
```diff
@@ -97,8 +98,12 @@ def _check_metadata_updated(self):
     def __init__(
         self, metadata, enforce_min_max_values=True, enforce_rounding=True, locales=['en_US']
     ):
+        single_metadata = metadata
+        if isinstance(single_metadata, Metadata):
+            single_metadata = single_metadata._convert_to_single_table()
+
         self._validate_inputs(enforce_min_max_values, enforce_rounding)
-        self.metadata = metadata
+        self.metadata = single_metadata
```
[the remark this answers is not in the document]
lajohn4747: I could update `get_metadata()` to return a `Metadata` object instead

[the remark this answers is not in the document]
amontanez24: Yeah. Another question I have for @npatki is if that method should return a SingleTableMetadata in the case where that's what the user provided, or just always return the new object? I'm thinking of the case where the person saved the synthesizer on an older version, loads it up and does
```python
synthesizer.get_metadata()
```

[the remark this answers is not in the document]
npatki: @amontanez24 @lajohn4747 I imagined that `get_metadata()` would always return the new, streamlined `Metadata` object no matter when the synthesizer was first created.

From this point onwards, the streamlined Metadata object is the only external-facing metadata object in play. All references to `SingleTableMetadata` and `MultiTableMetadata` would either be deprecated or deleted in the docs.

sdv-team:
Task linked: [CU-86b19amxa SDV - Enable single table synthesizers to use new Metadata #2128](https://app.clickup.com/t/86b19amxa)

sdv/metadata/metadata.py
```diff
@@ -71,3 +91,16 @@ def _set_metadata_dict(self, metadata, single_table_name=None):
...
+        if is_multi_table:
+            raise InvalidMetadataError(
+                'Metadata contains more than one table, use a MultiTableSynthesizer instead.'
+            )
+
+        if len(self.tables) == 0:
+            return SingleTableMetadata()
+
+        single_table_metadata = next(iter(self.tables.values()))
+        return single_table_metadata
```
pvk-developer, line 106: I think we can simplify this a bit:
```python
def _convert_to_single_table(self):
    if len(self.tables) > 1:
        raise InvalidMetadataError(
            'Metadata contains more than one table, use a MultiTableSynthesizer instead.'
        )

    return next(iter(self.tables.values()), SingleTableMetadata())
```

Here is an example:
```python

In [9]: next(iter({}.values()), 'default')
Out[9]: 'default'
```

sdv/single_table/base.py
```diff
@@ -97,8 +100,15 @@ def _check_metadata_updated(self):
     def __init__(
         self, metadata, enforce_min_max_values=True, enforce_rounding=True, locales=['en_US']
     ):
+        single_metadata = metadata
```
pvk-developer, line 103: No need to use this variable, you could work directly with the `metadata`.

sdv/single_table/base.py
```diff
@@ -97,8 +100,15 @@ def _check_metadata_updated(self):
...
+        if isinstance(single_metadata, Metadata):
+            single_metadata = single_metadata._convert_to_single_table()
+            self.real_metadata = metadata
+        elif isinstance(single_metadata, SingleTableMetadata):
```
pvk-developer, line 107: Don't forget to raise a `warning` As per the issue:

> * Deprecate SingleTableMetadata
    * Raise a future warning if users pass old metadata

lajohn4747:
I will take a look into converting everything over now instead of trying to make Metadata produce a SingleTableMetadata

pvk-developer (approved):

sdv/single_table/base.py
```diff
@@ -99,6 +107,14 @@ def __init__(
     ):
         self._validate_inputs(enforce_min_max_values, enforce_rounding)
         self.metadata = metadata
+        if isinstance(metadata, Metadata):
+            self.metadata = metadata._convert_to_single_table()
+            self.real_metadata = metadata
```
amontanez24, line 112: I think I'll remove this attribute and just convert it to the new metadata when calling get_metadata

lajohn4747:
@amontanez24 why not take over https://github.com/sdv-dev/SDV/pull/2144 instead?

amontanez24:
> @amontanez24 why not take over #2144 instead?

I think rewriting the code to support the new Metadata class in internal classes is overkill. To maintain backwards compatibility, we have to make sure to convert the class over everywhere that uses it like you were doing in that PR. I think it ends up being harder to read and riskier when the change is really only required for user facing purposes. If we were to do another major release, then we could just drop these old classes and write everything with the new class which would be cleaner, but until then I don't think it's worth it.

I'm also not sure we'll remove the concept of "SingleTableMetadata". The internal classes like constraints and the DataProcessor only work with one table at a time, so that construct still seems to make sense for them

sdv/single_table/base.py
```diff
@@ -269,7 +267,7 @@ def get_parameters(self):
 
     def get_metadata(self):
         """Return the ``Metadata`` for this synthesizer."""
-        return self.real_metadata
+        return Metadata.load_from_dict(self.metadata.to_dict())
```
pvk-developer, line 270: With this change, we always get a new instance of the `metadata` (it's okay, I just want to confirm that this is the expected return).

pvk-developer (approved):

sdv/single_table/base.py
amontanez24: Yeah the docs say we [return a copy](https://docs.sdv.dev/sdv/single-table-data/modeling/synthesizers/gaussiancopulasynthesizer#get_metadata) so it should be fine

frances-h (approved):

PR #2144, opened 2024-07-23
Use metadata over singletablemetadata

### LINKED ISSUE #2128: Enable single table synthesizers to use new Metadata

lajohn4747:
resolves #2128 
CU-86b19amxa

Made metadata work with single table synthesizers by converting  `SingleTableMetadata` to `Metadata` before inserting into the synthesizers. Write a wrapper method so that all single table synthesizers can use Metadata when there is only one table.

Wrote unit tests and update tests to use the methods instead of directly accessing `SingleTableMetadata` attributes.

### DIFF
--- a/sdv/constraints/tabular.py
+++ b/sdv/constraints/tabular.py
@@ -53,6 +53,7 @@ from sdv.constraints.utils import (
     revert_nans_columns,
     sigmoid,
 )
+from sdv.metadata.metadata import Metadata
 
 INEQUALITY_TO_OPERATION = {
     '>': np.greater,
@@ -251,6 +252,7 @@ class FixedCombinations(Constraint):
 
     @staticmethod
     def _validate_metadata_specific_to_constraint(metadata, **kwargs):
+        metadata = Metadata._convert_to_unified_metadata(metadata)
         invalid_columns = []
         column_names = kwargs.get('column_names')
         for column in column_names:
@@ -390,11 +392,13 @@ class Inequality(Constraint):
 
     @classmethod
     def _validate_metadata_columns(cls, metadata, **kwargs):
+        metadata = Metadata._convert_to_unified_metadata(metadata)
         kwargs['column_names'] = [kwargs.get('high_column_name'), kwargs.get('low_column_name')]
         super()._validate_metadata_columns(metadata, **kwargs)
 
     @staticmethod
     def _validate_metadata_specific_to_constraint(metadata, **kwargs):
+        metadata = Metadata._convert_to_unified_metadata(metadata)
         high = kwargs.get('high_column_name')
         low = kwargs.get('low_column_name')
         high_sdtype = metadata.get_columns().get(high, {}).get('sdtype')
@@ -599,6 +603,7 @@ class ScalarInequality(Constraint):
 
     @staticmethod
     def _validate_metadata_specific_to_constraint(metadata, **kwargs):
+        metadata = Metadata._convert_to_unified_metadata(metadata)
         column_name = kwargs.get('column_name')
         sdtype = metadata.get_columns().get(column_name, {}).get('sdtype')
         value = kwargs.get('value')
@@ -772,6 +777,7 @@ class Positive(ScalarInequality):
 
     @staticmethod
     def _validate_metadata_specific_to_constraint(metadata, **kwargs):
+        metadata = Metadata._convert_to_unified_metadata(metadata)
         column_name = kwargs.get('column_name')
         sdtype = metadata.get_columns().get(column_name, {}).get('sdtype')
         if sdtype != 'numerical':
@@ -801,6 +807,7 @@ class Negative(ScalarInequality):
 
     @staticmethod
     def _validate_metadata_specific_to_constraint(metadata, **kwargs):
+        metadata = Metadata._convert_to_unified_metadata(metadata)
         column_name = kwargs.get('column_name')
         sdtype = metadata.get_columns().get(column_name, {}).get('sdtype')
         if sdtype != 'numerical':
@@ -839,6 +846,7 @@ class Range(Constraint):
 
     @classmethod
     def _validate_metadata_columns(cls, metadata, **kwargs):
+        metadata = Metadata._convert_to_unified_metadata(metadata)
         high = kwargs.get('high_column_name')
         low = kwargs.get('low_column_name')
         middle = kwargs.get('middle_column_name')
@@ -847,6 +855,7 @@ class Range(Constraint):
 
     @staticmethod
     def _validate_metadata_specific_to_constraint(metadata, **kwargs):
+        metadata = Metadata._convert_to_unified_metadata(metadata)
         high = kwargs.get('high_column_name')
         low = kwargs.get('low_column_name')
         middle = kwargs.get('middle_column_name')
@@ -1087,6 +1096,7 @@ class ScalarRange(Constraint):
 
     @staticmethod
     def _validate_metadata_specific_to_constraint(metadata, **kwargs):
+        metadata = Metadata._convert_to_unified_metadata(metadata)
         column_name = kwargs.get('column_name')
         if column_name not in metadata.get_columns():
             raise ConstraintMetadataError(
@@ -1442,6 +1452,7 @@ class Unique(Constraint):
 
     @staticmethod
     def _validate_metadata_specific_to_constraint(metadata, **kwargs):
+        metadata = Metadata._convert_to_unified_metadata(metadata)
         column_names = kwargs.get('column_names')
         keys = set()
         if isinstance(metadata.get_primary_key(), tuple):
--- a/sdv/constraints/tabular.py
+++ b/sdv/constraints/tabular.py
@@ -254,7 +254,7 @@ class FixedCombinations(Constraint):
         invalid_columns = []
         column_names = kwargs.get('column_names')
         for column in column_names:
-            if metadata.columns[column]['sdtype'] not in ['boolean', 'categorical']:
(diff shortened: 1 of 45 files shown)


sdv-team:
Task linked: [CU-86b19amxa SDV - Enable single table synthesizers to use new Metadata #2128](https://app.clickup.com/t/86b19amxa)

sdv/lite/single_table.py
```diff
@@ -7,6 +7,9 @@
 
 import cloudpickle
 
+from sdv.metadata.metadata import Metadata
+from sdv.metadata.single_table import DEPRECATION_MSG as sdv_DEPRECATION_MSG
```
pvk-developer, line 11: `sdv_DEPRECATION_MSG` -> `SDV_DEPRECATION_MSG`

sdv/metadata/metadata.py
```diff
@@ -68,6 +72,350 @@ def _set_metadata_dict(self, metadata, single_table_name=None):
...
+            - ``InvalidMetadataError`` if the column already exists.
+            - ``InvalidMetadataError`` if the ``kwargs`` do not contain ``sdtype``.
+            - ``InvalidMetadataError`` if the column has unexpected values or ``kwargs`` for the
+              given ``sdtype``.
+            - ``InvalidMetadataError`` if the table doesn't exist in the ``MultiTableMetadata``.
+        """
+        table_name = self._get_table_name(table_name)
+        if table_name is None:
+            self.tables[DEFAULT_TABLE_NAME] = SingleTableMetadata()
+            table_name = DEFAULT_TABLE_NAME
```
pvk-developer, line 234: Add breakline here

sdv/metadata/metadata.py
```diff
@@ -68,6 +72,350 @@ def _set_metadata_dict(self, metadata, single_table_name=None):
...
+                table only if there is a single table.
+
+        Returns:
+            list:
+                List of alternate keys found in the table.
+        """
+        table = self._get_table_or_default(table_name)
+        if table is None:
+            return []
+        return table.alternate_keys
```
pvk-developer, line 160:
```sugest
return getattr(table, 'alternate_keys', [])
```

sdv/metadata/metadata.py
```diff
@@ -68,6 +72,350 @@ def _set_metadata_dict(self, metadata, single_table_name=None):
...
+            if 'columns' in metadata:
+                self.tables[single_table_name] = SingleTableMetadata.load_from_dict(metadata)
+
+    def _get_table_or_default(self, table_name=None):
+        if table_name is None:
+            if len(self.tables) > 1:
+                raise ValueError(
+                    'This metadata contains more than one table. '
+                    'Please provide a table name in the method.'
+                )
```
pvk-developer, line 86: add breakline

sdv/metadata/metadata.py
```diff
@@ -68,6 +72,350 @@ def _set_metadata_dict(self, metadata, single_table_name=None):
...
+            single_table = next(iter(self.tables.values()), None)
+            return single_table
+        else:
```
pvk-developer, line 89: This can be removed (replace with breakline)

sdv/metadata/metadata.py
```diff
@@ -68,6 +72,350 @@ def _set_metadata_dict(self, metadata, single_table_name=None):
...
+            return self.tables[table_name]
```
pvk-developer, line 90: This can be on the function level
```python
if ...

return self.tables[table_name]
```

sdv/metadata/metadata.py
```diff
@@ -68,6 +72,350 @@ def _set_metadata_dict(self, metadata, single_table_name=None):
...
+                table only if there is a single table.
+
+        Returns:
+            str:
+                Sequence key name
+        """
+        table = self._get_table_or_default(table_name)
+        if table is None:
+            return None
+        return table.sequence_key
```
pvk-developer, line 350: use getattr here as well

sdv/metadata/metadata.py
```diff
@@ -68,6 +72,350 @@ def _set_metadata_dict(self, metadata, single_table_name=None):
...
+                table only if there is a single table.
+
+        Returns:
+            str:
+                Sequence Index name for the given table.
+        """
+        table = self._get_table_or_default(table_name)
+        if table is None:
+            return None
+        return table.sequence_index
```
pvk-developer, line 370: same as above

pvk-developer (changes requested):

sdv/constraints/tabular.py
```diff
@@ -254,7 +254,7 @@ def _validate_metadata_specific_to_constraint(metadata, **kwargs):
         invalid_columns = []
         column_names = kwargs.get('column_names')
         for column in column_names:
-            if metadata.columns[column]['sdtype'] not in ['boolean', 'categorical']:
+            if metadata.get_columns()[column]['sdtype'] not in ['boolean', 'categorical']:
```
pvk-developer, line 257: This won't be backwards compatible. `SingleTableMetadata` does not have `get_columns()`.

lajohn4747: Made it backwards compatible by making sure all metadata is converted and also convert it the synthesizer metadata on `load`.

Also added an integration test for backward compatibility.

tests/integration/metadata/test_metadata.py
```diff
@@ -216,3 +220,38 @@ def test_detect_table_from_csv(tmp_path):
...
+    with pytest.warns(FutureWarning, match=warn_msg):
+        synthesizer = GaussianCopulaSynthesizer(metadata)
+    synthesizer.fit(data)
+    model_path = tmp_path / 'synthesizer.pkl'
+    synthesizer.save(model_path)
+
+    # Assert
+    assert model_path.exists()
+    assert model_path.is_file()
+    loaded_synthesizer = GaussianCopulaSynthesizer.load(model_path)
```
frances-h, line 244: Can we add a test loading a synthesizer with `SingleTableMetadata` to check the deprecation warning is raised and that it gets upgraded to the new `Metadata`?

tests/integration/metadata/test_metadata.py
```diff
@@ -216,3 +220,38 @@ def test_detect_table_from_csv(tmp_path):
     }
 
     assert metadata.to_dict() == expected_metadata
+
+
+def test_single_table_compatibility(tmp_path):
+    """Test if SingleMetadataTable still has compatibility with single table synthesizers."""
+    # Setup
+    data, metadata = download_demo('single_table', 'fake_hotel_guests')
```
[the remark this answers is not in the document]
lajohn4747: Yes will change in a following PR, just adding the test now so that we can confirm nothing breaks in this feature branch in between now and then.

tests/integration/sequential/test_par.py
```diff
@@ -406,6 +410,40 @@ def test_init_error_sequence_key_in_context():
...
+    data = pd.DataFrame(
+        data={
+            'user_id': ['ID_00'] * 5 + ['ID_01'] * 5,
+            'birthdate': ['1995-05-06'] * 5 + ['1982-01-21'] * 5,
+            'timestamp': ['2023-06-21', '2023-06-22', '2023-06-23', '2023-06-24', '2023-06-25'] * 2,
+            'heartrate': [67, 66, 68, 65, 64, 80, 82, 91, 88, 84],
+        }
+    )
+
+    metadata = SingleTableMetadata.load_from_dict({
```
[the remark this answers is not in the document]
lajohn4747: This test is from a different PR so should not be affected.

tests/integration/metadata/test_metadata.py
lajohn4747: Download demo returns `SingleTableMetadata` it will all get changed in the future but that will be addressed in a future PR.

Will add an assert that metadata is `SingleTableMetadata` so that the test clearly covers this use case.

frances-h (approved):
lgtm 👍

lajohn4747:
Closing as we landed #2140 instead

PR #2150, opened 2024-07-26
Changing conditions in ScalarRange for is_datetime

Deathn0t:
I suggest updating the condition that checks for `datetime` types in ScalarRange.

This resulted in an exception when the range values could be interpreted as dates (e.g., 1900, 2000) even though the column type is `int`.

Therefore I suggest raisin the exception only when the column is a `datetime` and not both low/high values of the range are datetime.


Deathn0t:
Hello, would you have any update or comment on this PR? the changes are relatively small.

PR #2155, opened 2024-07-30
Suppress warning when using GANs that suggests that CUDA isn't being used

### LINKED ISSUE #2052: Confusing warning when using GANs that suggests that CUDA isn't being used
### Environment Details

* SDV version: 1.13
* Python version: 3.10.12
* Operating System: Colab (using T4 GPU)

### Warning

If you try to fit CTGANSynthesizer or CopulaGANSynthesizer in Google Colab with a T4 (GPU) runtime, the following warning is displayed the first time only:

```
/usr/local/lib/python3.10/dist-packages/torch/autograd/graph.py:744: UserWarning: Attempting to run cuBLAS, but there was no current CUDA context! Attempting to set the primary context... (Triggered internally at ../aten/src/ATen/cuda/CublasHandlePool.cpp:135.)
  return Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
```

- We've ruled out CUDA not being available by running `torch.cuda.is_available()`, which returns `True`. 
- This warning might be displayed because the CUDA state / context isn't initialized. **But we do end up initializing the CUDA context in our code somewhere (which is why running it twice doesn't throw this warning)**. 

So the warning is technically true, but is confusing to users because they may think that the GPU isn't being used.

### Steps to reproduce

To reproduce warning, change the Colab runtime to T4 (Runtime > Change runtime type, from Colab menu bar) then run this code:

```

    modality='single_table',
)

c = CTGANSynthesizer(metadata, cuda=True)
c.fit(data)
```

To reproduce CopulaGANSynthesizer, delete / restart your runtime:

```
from sdv.single_table import CopulaGANynthesizer

    modality='single_table',
)

c = CopulaGANynthesizer(metadata, cuda=True)
c.fit(data)
```

[Internal Colab Notebook](https://colab.research.google.com/drive/1r7yc1Et7Pjr3i_DAzLS5kGfl_t4kEcBp?authuser=1#scrollTo=6tQZDh39j3Vk)

fealho:
CU-86b0vehd0, Resolve #2052.

This PR suppresses the warning generated in CTGAN: https://github.com/sdv-dev/CTGAN/blob/d132c23a7bf5883cba0d1cb09f6d54aa400c39fb/ctgan/synthesizers/ctgan.py#L43

The warning originates from torch: https://github.com/pytorch/pytorch/blob/f72266eceaf833cd1a44d8e4ac611bc36102539c/aten/src/ATen/cuda/CublasHandlePool.cpp#L135C5-L135C20

### DIFF
--- a/sdv/single_table/ctgan.py
+++ b/sdv/single_table/ctgan.py
@@ -290,7 +290,7 @@ class CTGANSynthesizer(LossValuesMixin, BaseSingleTableSynthesizer):
         )
         self._model = CTGAN(**self._model_kwargs)
         with warnings.catch_warnings():
-            warnings.simplefilter('ignore', category=UserWarning)
+            warnings.filterwarnings('ignore', message='.*Attempting to run cuBLAS.*')
             self._model.fit(processed_data, discrete_columns=discrete_columns)
 
     def _sample(self, num_rows, conditions=None):
--- a/sdv/single_table/ctgan.py
+++ b/sdv/single_table/ctgan.py
@@ -1,5 +1,7 @@
 """Wrapper around CTGAN model."""
 
+import warnings
+
 import numpy as np
 import pandas as pd
 import plotly.express as px
@@ -287,7 +289,9 @@ class CTGANSynthesizer(LossValuesMixin, BaseSingleTableSynthesizer):
             self.get_metadata(), processed_data, transformers
         )
         self._model = CTGAN(**self._model_kwargs)
-        self._model.fit(processed_data, discrete_columns=discrete_columns)
+        with warnings.catch_warnings():
+            warnings.simplefilter('ignore', category=UserWarning)
+            self._model.fit(processed_data, discrete_columns=discrete_columns)
 
     def _sample(self, num_rows, conditions=None):
         """Sample the indicated number of rows from the model.


sdv-team:
Task linked: [CU-86b0vehd0 SDV - UserWarning when using CTGAN caused by potentially uninitialized CUDA context #2052](https://app.clickup.com/t/86b0vehd0)

frances-h (approved):

frances-h:
Are there cases when this warning is valid? If it's only popping up in error then it's probably ok to silence it, but maybe not if it does provide a useful warning in some cases.

frances-h (approved):

sdv/single_table/ctgan.py
```diff
@@ -287,7 +289,9 @@ def _fit(self, processed_data):
             self.get_metadata(), processed_data, transformers
         )
         self._model = CTGAN(**self._model_kwargs)
-        self._model.fit(processed_data, discrete_columns=discrete_columns)
+        with warnings.catch_warnings():
+            warnings.simplefilter('ignore', category=UserWarning)
```
amontanez24, line 293: can we filter only the warnings from torch?

amontanez24 (approved):

PR #2158, opened 2024-07-31
Mock every usage of `is_faker_function` to speed up the unit tests

### LINKED ISSUE #2125: Metadata.add_column can be slow
### Problem Description

The [metadata.add_column](https://github.com/sdv-dev/SDV/blob/8cec1f7c49c4e4242c260653db8bea042ad9f209/sdv/metadata/single_table.py#L259) method can sometimes be unreasonably slow (a few seconds). This seems to be because during its validation, we check if it isn't a faker function, which seems to be a slow method in RDT.
https://github.com/sdv-dev/SDV/blob/8cec1f7c49c4e4242c260653db8bea042ad9f209/sdv/metadata/single_table.py#L229-L243

### Expected behavior

- See if the validation can be sped up. This may require speeding up the underlying [function in RDT](https://github.com/sdv-dev/RDT/blob/6e963a8a01efe2d92a5cefff7d0a77169a8e7574/rdt/transformers/pii/utils.py#L8)

### LINKED ISSUE #2163: Mock every usage of `is_faker_function` to speed up the unit tests
### Problem Description

The initial call of `is_faker_function` from RDT is consistently slow, taking a few seconds. When running the unit tests, the test that makes this first call always takes longer, causing the unit test to run slowly.

### Expected behavior

Running the unit test should take almost no time.
Whenever the `is_faker_function` method is utilized, tests should mock it with the correct return values, whether used directly or indirectly.

### Additional context

To detect which test is doing the first `is_faker_function` call, one can run:
`pytest tests/unit --duration=0`
It will be the test that takes more time.

R-Palazzo:
CU-86b18qaaz
Resolve #2163


sdv-team:
Task linked: [CU-86b18qaaz SDV - Metadata.add_column can be slow #2125](https://app.clickup.com/t/86b18qaaz)

amontanez24 (commented):
Nice job, this is much faster! While we're at it can you fix this [unit test](https://github.com/sdv-dev/SDV/blob/a0e0a76c23c644145b223c7b5ede4a21b31238a0/tests/unit/multi_table/test_hma.py#L1123) to not actually fit and sample from the model. If we can just mock the _model part or something. This is the only test that still takes over a second

R-Palazzo:
> Nice job, this is much faster! While we're at it can you fix this [unit test](https://github.com/sdv-dev/SDV/blob/a0e0a76c23c644145b223c7b5ede4a21b31238a0/tests/unit/multi_table/test_hma.py#L1123) to not actually fit and sample from the model. If we can just mock the _model part or something. This is the only test that still takes over a second

Hi @amontanez24, I didn’t find a good way to mock some methods while preserving the test logic, so I moved it to the `integration` folder. It also makes sense to have it in the `integration tests` because it validates end-to-end whether the estimated number of columns matches the actual one when we fit and sample HMA. There are still other tests in the unit folder for the `_estimate_num_columns` method. Let me know if it works. The change is in [75387aa](https://github.com/sdv-dev/SDV/pull/2158/commits/75387aa4fbdd29f926ade81279b6b5010bb5e196)

amontanez24 (approved):

tests/integration/multi_table/test_hma.py
```diff
@@ -1988,3 +1989,112 @@ def test_hma_synthesizer_with_fixed_combinations():
...
+            {
+                'parent_table_name': 'grandparent',
+                'parent_primary_key': 'GP',
+                'child_table_name': 'parent',
+                'child_foreign_key': 'GP',
+            },
+        ],
+    })
+    synthesizer = HMASynthesizer(metadata)
+    synthesizer._finalize = Mock(return_value=data)
```
fealho, line 2081: @amontanez24 Are we allowed to use mocks in integration tests?

amontanez24: Yeah that's a good question. Ideally you wouldn't have to. This test is a little strange in that it is mocking the internals of another method (_finalize), to assert that the method under test is returning the right values. I'm not really sure of the context, so I don't know why this is the case.

Ideally, this test would just run _estimate_num_columns, and we'd know what the result should be and could just assert that outcome.

fealho (approved):

PR #2161, opened 2024-08-01
Primary keys may not be unique for variable length regexes

### LINKED ISSUE #2116: Primary keys may not be unique for variable length regexes
### Environment Details
* SDV version: 1.14.0

### Error Description
If a regex has variable length (eg. 3 or 4 characters long with a format such as `'[0-9]{3,4}'`), the primary keys that SDV generates may not always be unique. This is supported by the diagnostic report being less than 100%.

### Steps to reproduce
```python
import pandas as pd

from rdt.transformers.text import RegexGenerator
from sdv.evaluation.single_table import run_diagnostic

data = pd.DataFrame(data={
    'id': [123, 456, 1234, , 5678],
    'col': [23.4, 12.3, 10.0, 90.4, 10.3]
})

    'columns': {
        'id': { 'sdtype': 'id', 'regex_format': '[0-9]{3,4}' },
        'col': { 'sdtype': 'numerical' }
    },
    'primary_key': 'id'
})

# I am specifically assigning the RegexGenerator for illustration purposes
synth.update_transformers({
    'id': RegexGenerator(regex_format='[0-9]{3,4}', enforce_uniqueness=True, generation_order='alphanumeric')
})

synth.fit(data)
synthetic_data = synth.sample(num_rows=10_000)

diagnostic = run_diagnostic(data, synthetic_data, metadata)
diagnostic.get_details('Data Validity')
```
Output:
image

### Expected Fix
This case can only happen when:

- The real data has a primary key (or alternate key) stored as an `int` AND
- The metadata for it includes a Regex AND
- The Regex allows for a `'0'` to be in the first position of the string

If _and only if_ all 3 conditions are met, the synthesizer should throw an error.

```python
synthesizer = HSASynthesizer(metadata)
synthesiser.fit(data)
```
```
SynthesizerInputError: Primary key for table 'users' is stored as an int but the Regex allows it to start with '0'. Please remove the Regex or update it to correspond to valid ints.
```

amontanez24:
resolves #2116 
CU-86b17dk49

### DIFF
--- a/sdv/multi_table/base.py
+++ b/sdv/multi_table/base.py
@@ -374,6 +374,8 @@ class BaseMultiTableSynthesizer:
                         f'Primary key for table "{table_name}" {INT_REGEX_ZERO_ERROR_MESSAGE}'
                     )
 
+                raise e
+
         for table in list_of_changed_tables:
             data[table].columns = self._original_table_columns[table]
 
--- a/sdv/multi_table/base.py
+++ b/sdv/multi_table/base.py
@@ -26,6 +26,7 @@ from sdv.errors import (
     SynthesizerInputError,
 )
 from sdv.logging import disable_single_table_logger, get_sdv_logger
+from sdv.single_table.base import INT_REGEX_ZERO_ERROR_MESSAGE
 from sdv.single_table.copulas import GaussianCopulaSynthesizer
 
 SYNTHESIZER_LOGGER = get_sdv_logger('MultiTableSynthesizer')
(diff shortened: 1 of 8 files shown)


sdv-team:
Task linked: [CU-86b17dk49 SDV - Primary keys may not be unique for variable length regexes #2116](https://app.clickup.com/t/86b17dk49)

tests/unit/single_table/test_base.py
```diff
@@ -600,6 +600,47 @@ def test_validate_raises_invalid_data_for_metadata(self):
         instance._validate_constraints.assert_called_once_with(data)
         instance._validate.assert_not_called()
 
+    def test_validate_int_primary_key_regex_starts_with_0(self):
```
pvk-developer, line 603: can we change `0` to `zero`

sdv/multi_table/base.py
```diff
@@ -363,9 +364,17 @@ def preprocess(self, data):
...
+                synthesizer = self._table_synthesizers[table_name]
+                self._assign_table_transformers(synthesizer, table_name, table_data)
+                processed_data[table_name] = synthesizer._preprocess(table_data)
+            except SynthesizerInputError as e:
+                if INT_REGEX_ZERO_ERROR_MESSAGE in str(e):
+                    raise SynthesizerInputError(
+                        f'Primary key for table "{table_name}" {INT_REGEX_ZERO_ERROR_MESSAGE}'
+                    )
+
+                raise e
```
pvk-developer, line 377: Why do we need to do the validation here?
Shouldn't this already occur during `self.validate(data)` ?

amontanez24: It was requested that in the multi table case we add the table name to the error message

rwedge (approved):

pvk-developer (approved):

PR #2175, opened 2024-08-07
Enable demos are using the new metadata

### LINKED ISSUE #2131: Update demos to use new metadata
### Problem Description

Our demo module should use the new metadata class once #2104 is complete.

### Expected behavior

- Add a json for the new metadata structure.
- Write and run a script to update all demo zips to contain the new json structure
- Update [download_demo](https://github.com/sdv-dev/SDV/blob/8cec1f7c49c4e4242c260653db8bea042ad9f209/sdv/datasets/demo.py#L120) to return the new class

### Additional context
- The script to update the metadata doesn't need to be added to the library
- Updating the metadata follows a different set of steps for each modality:
#### Single table and sequential
- We want to convert the json to that of multi-table metadata. To do this we need to create a dictionary mapping table names to the single table metadata.
    - You can use the csv name as the table name.
- Update the METADATA_SPEC_VERSION

#### Multi table

- Update the METADATA_SPEC_VERSION

lajohn4747:
resolves #2131 
CU-86b19awwy

`download_demo` should now download the json and store into `Metadata` object. New metadata is downloaded as `metadata_v2.json` but kept a check path `metadata_v1.json` just in case I missed any zip files.


sdv-team:
Task linked: [CU-86b19awwy SDV - Update demos to use new metadata #2131](https://app.clickup.com/t/86b19awwy)

sdv/data_processing/data_processor.py
```diff
@@ -114,6 +115,8 @@ def __init__(
         locales=['en_US'],
     ):
         self.metadata = metadata
+        if isinstance(metadata, Metadata):
+            self.metadata = metadata._convert_to_single_table()
```
amontanez24, line 119: this situation shouldn't ever occur since we directly pass the metadata to the data processor

sdv/datasets/demo.py
```diff
@@ -104,15 +105,18 @@ def _get_data(modality, output_folder_name, in_memory_directory):
...
+        metadata = metadata.load_from_json(metadata_path, dataset_name)
 
     else:
-        metadict = json.loads(in_memory_directory['metadata_v1.json'])
-        metadata = metadata.load_from_dict(metadict)
+        metadata_path = 'metadata_v2.json'
+        if metadata_path not in in_memory_directory:
+            metadata_path = 'metadata_v1.json'
+        metadict = json.loads(in_memory_directory[metadata_path])
+        metadata = metadata.load_from_dict(metadict, dataset_name)
```
amontanez24, line 119: why does it need the dataset_name now? All of the jsons should be in the new format so they should just be loadable

lajohn4747: Not necessary but if it's a single table, that becomes the name of the table unless we are fine with a default table name in the case of a conversion.

lajohn4747: Yep those are the two places that this code is only used in. I have updated all the public buckets so if we wanted to we could leave out option 1

amontanez24: They still have the old one right? I think we need to support people on old versions of sdv

lajohn4747: Not sure if people save our s3 bucket locally? Otherwise people will always download from our s3 public demodataset buckets but I would recommend to support it just in case.

amontanez24: No I mean that their code wouldn't have these changes you made so it will require the old metadata file to be there with the old name

lajohn4747: Yes the existing metadata is still there so all older code will work as well.

sdv/datasets/demo.py
```diff
@@ -104,15 +104,19 @@ def _get_data(modality, output_folder_name, in_memory_directory):
...
 
 
-def _get_metadata(modality, output_folder_name, in_memory_directory):
-    metadata = MultiTableMetadata() if modality == 'multi_table' else SingleTableMetadata()
+def _get_metadata(output_folder_name, in_memory_directory, dataset_name):
+    metadata = Metadata()
     if output_folder_name:
         metadata_path = os.path.join(output_folder_name, METADATA_FILENAME)
-        metadata = metadata.load_from_json(metadata_path)
+        metadata = metadata.load_from_json(metadata_path, dataset_name)
```
amontanez24, line 111: Is passing the `dataset_name` here just a precaution? It shouldn't be required here because there shouldn't ever be a case where the json is in single table format right?

lajohn4747: Yeah it's a precaution, not necessary and it defaults to None. For multi-table it does not affect anything

amontanez24 (approved):
LGTM!

sdv/datasets/demo.py
```diff
@@ -104,15 +104,19 @@ def _get_data(modality, output_folder_name, in_memory_directory):
...
 
     else:
-        metadict = json.loads(in_memory_directory['metadata_v1.json'])
-        metadata = metadata.load_from_dict(metadict)
+        metadata_path = 'metadata_v2.json'
+        if metadata_path not in in_memory_directory:
+            warnings.warn(f'Metadata for {dataset_name} is missing updated version v2.')
+            metadata_path = 'metadata_v1.json'
```
pvk-developer, line 117: can we add a breakline here

sdv/utils/utils.py
```diff
@@ -91,6 +92,8 @@ def get_random_sequence_subset(
             - random: Randomly choose n rows to keep within the sequence. It is important to keep
             the randomly chosen rows in the same order as they appear in the original data.
     """
+    if isinstance(metadata, Metadata):
+        metadata = metadata._convert_to_single_table()
```
pvk-developer, line 96: breakline here

pvk-developer (approved):

PR #2178, opened 2024-08-09
Update tests to use Metadata

### LINKED ISSUE #2151: Move all tests over to use Metadata instead of SingleTableMetadata and MultiTableMetadata
### Problem Description

Move all tests over to use Metadata instead of SingleTableMetadata and MultiTableMetadata. When we introduce Metadata we should make sure all tests get switched over except for compatibility tests. Metadata should have been introduced while tests covered the legacy metadata class, but once they are confirmed to work we should move over all tests (besides compatibility tests) to use Metadata instead of doubling our test suite and slowing down our testing automation jobs.
 
### Expected behavior

Once Metadata has been merged, move over existing tests to use `Metadata` instead of `SingleTableMetadata` and `MultiTableMetadata`

lajohn4747:
resolves #2151 
CU-86b1efup0

Moves tests to use Metadata.
This does two things:
- Confirms the Metadata interface is working fine with our library
- Helps users/developers to transition over to Metadata. Keeps conformity within the library.

### DIFF
--- a/sdv/metadata/metadata.py
+++ b/sdv/metadata/metadata.py
@@ -1,7 +1,5 @@
 """Metadata."""
 
-from pathlib import Path
-
 from sdv.metadata.errors import InvalidMetadataError
 from sdv.metadata.multi_table import MultiTableMetadata
 from sdv.metadata.single_table import SingleTableMetadata
--- a/sdv/metadata/metadata.py
+++ b/sdv/metadata/metadata.py
@@ -76,7 +76,7 @@ class Metadata(MultiTableMetadata):
         """Get the table name if there is only one table.
 
         Checks to see if the metadata contains only a single table, if so
-        return the name. Otherwise raise an error.
+        return the name. Otherwise warn the user and return None.
 
         Args:
             metadata (dict):
@@ -88,6 +88,7 @@ class Metadata(MultiTableMetadata):
                 'This metadata does not contain only a single table. Could not determine '
                 'single table name and will return None.'
             )
+            return None
 
         return next(iter(self.tables), None)
 
--- a/sdv/metadata/metadata.py
+++ b/sdv/metadata/metadata.py
@@ -1,5 +1,7 @@
 """Metadata."""
 
+import warnings
+
 from sdv.metadata.errors import InvalidMetadataError
 from sdv.metadata.multi_table import MultiTableMetadata
 from sdv.metadata.single_table import SingleTableMetadata
@@ -10,6 +12,7 @@ class Metadata(MultiTableMetadata):
     """Metadata class that handles all metadata."""
 
     METADATA_SPEC_VERSION = 'V1'
+    DEFAULT_SINGLE_TABLE_NAME = 'default_table_name'
 
     @classmethod
     def load_from_json(cls, filepath, single_table_name=None):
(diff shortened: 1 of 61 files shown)


sdv-team:
Task linked: [CU-86b1efup0 SDV - Move all tests over to use Metadata instead of SingleTableMetadata and MultiTableMetadata #2151](https://app.clickup.com/t/86b1efup0)

pvk-developer (approved):

sdv/single_table/base.py
```diff
@@ -112,7 +112,9 @@ def __init__(
     ):
         self._validate_inputs(enforce_min_max_values, enforce_rounding)
         self.metadata = metadata
+        self.table_name = Metadata.DEFAULT_SINGLE_TABLE_NAME
```
amontanez24, line 115: can we make this a private attribute?
```suggestion
        self._table_name = Metadata.DEFAULT_SINGLE_TABLE_NAME
```

sdv/metadata/metadata.py
```diff
@@ -66,9 +69,28 @@ def _set_metadata_dict(self, metadata, single_table_name=None):
...
+        return the name. Otherwise raise an error.
+
+        Args:
+            metadata (dict):
+                Python dictionary representing a ``MultiTableMetadata`` or
+                ``SingleTableMetadata`` object.
+        """
+        if len(self.tables) != 1:
+            warnings.warn(
+                'This metadata does not contain a single table. Could not determine '
```
amontanez24, line 88: This reads as there are no tables. Maybe change to
```suggestion
                'This metadata contains more than one table. Could not determine '
```

sdv/single_table/base.py
```diff
@@ -112,7 +112,9 @@ def __init__(
...
         if isinstance(metadata, Metadata):
+            self.table_name = metadata._get_single_table_name()
```
amontanez24, line 117: I'm confused about this scenario. If multiple tables are in the metadata, then we error in the single table case so we don't need the warning. Maybe we should also just error if the user doesn't pass a table name in instead of replacing with a default

sdv/metadata/metadata.py
lajohn4747: Updated to this:
`'This metadata does not contain only a single table. Could not determine '
                'single table name and will return None.'`
                
  Because it could also throw this warning with 0 tables

sdv/single_table/base.py
lajohn4747: This scenario is if there is a single table, we want to make sure `get_metadata` returns the single table object with the correct name attached to the table (same one that is passed in) if there isn't a name (e.g. legacy `SingleTableMetadata` passed in) we can use the default name.

amontanez24: If it's a legacy metadata then it won't enter this branch of the if right? And if it's the legacy metadata why do we need a name?

lajohn4747: You're right, legacy metadata will just have the default name so it's not the best example.

When we call `get_metadata` that comes out with the class `Metadata`. That table should match whatever was put into the class so the user can refer to the table using the same table name.

sdv/metadata/metadata.py
```diff
@@ -66,9 +69,28 @@ def _set_metadata_dict(self, metadata, single_table_name=None):
...
+
+        Args:
+            metadata (dict):
+                Python dictionary representing a ``MultiTableMetadata`` or
+                ``SingleTableMetadata`` object.
+        """
+        if len(self.tables) != 1:
+            warnings.warn(
+                'This metadata does not contain only a single table. Could not determine '
+                'single table name and will return None.'
```
amontanez24, line 89: won't the code below return the next element if there are 2 tables instead of None?

amontanez24 (approved):

PR #2184, opened 2024-08-14
Metadata Detection Fails with new Data Type

### LINKED ISSUE #2182: Metadata Detection Fails with new Data Type
### Error Description

Hardcoded logic in our software causes `metadata.detect_from_dataframe` to fail in detecting new data types beyond the primitive ones. There is code that focuses only on primitive data types, ignoring others and raises an error that we only support those.

### Steps to reproduce

```python
import pandas as pd
data = pd.DataFrame({
    'UInt8': pd.Series([1, 2, None, 3], dtype='UInt8'),
    'UInt16': pd.Series([1, 2, None, 3], dtype='UInt16'),
    'UInt32': pd.Series([1, 2, None, 3], dtype='UInt32'),
    'UInt64': pd.Series([1, 2, None, 3], dtype='UInt64'),
})

sdv.metadata.errors.InvalidMetadataError: Unsupported data type for column 'UInt8' (kind: u).The valid data types are: 'object', 'int', 'float', 'datetime', 'bool'.
```
### Expected behavior

We should be able to detect the following data types:
* UInt(8, 16, 32, 64) -- Pandas
* uint(8, 16, 32, 64) -- Numpy

## Additional Context

This is the current code that we use to detect data types. This has to be enhanced to support more data types.
```python
if dtype in self._DTYPES_TO_SDTYPES:
    sdtype = self._DTYPES_TO_SDTYPES[dtype]
elif dtype in ['i', 'f']:
    sdtype = self._determine_sdtype_for_numbers(column_data)

elif dtype == 'O':
    sdtype = self._determine_sdtype_for_objects(column_data)

if sdtype is None:
    raise InvalidMetadataError(
        f"Unsupported data type for column '{field}' (kind: {dtype})."
        "The valid data types are: 'object', 'int', 'float', 'datetime', 'bool'."
    )

```
In order to support a wider range of dtypes, we have to enhance this logic by adding more dtype mappings. With the given configuration, we are not supporting Pandas nullable ints and other data types that are not the primitive ones loaded with a numpy backend.

R-Palazzo:
CU-86b1nhh6a
Resolve #2182

Open question from this PR:
Do we want to cast the column to a `pandas integer` when it fails to cast back to a `numpy integer` due to NaNs (The input was numpy integer dtype)? Currently, the column remains as `float` in this scenario. @amontanez24 @pvk-developer


sdv-team:
Task linked: [CU-86b1nhh6a SDV - Metadata Detection Fails with new Data Type #2182](https://app.clickup.com/t/86b1nhh6a)

pyproject.toml
```diff
@@ -35,7 +35,7 @@ dependencies = [
     'copulas>=0.11.0',
     'ctgan>=0.10.0',
     'deepecho>=0.6.0',
-    'rdt @ git+https://github.com/sdv-dev/RDT@main',
+    'rdt @ git+https://github.com/sdv-dev/RDT@issue-865-uint-hypertransformer',
```
R-Palazzo, line 38: I will update to point to the main branch once this [PR](https://github.com/sdv-dev/RDT/pull/867) is merged.
Or if the RDT release is done, I could point to the latest RDT version

tests/integration/data_processing/test_data_processor.py
```diff
@@ -401,3 +401,133 @@ def test_update_transformers_id_generator(self):
         )
         with pytest.raises(SynthesizerInputError, match=error_msg):
             dp.update_transformers({'user_id': UniformEncoder()})
+
+    def test_numerical_dtype_handling(self):
```
gsheni, line 405: Do you think `pytest.mark.parametrize` is better to use for these? You could make the dtype the input

pvk-developer (approved):

pvk-developer:
> Open question from this PR: Do we want to cast the column to a `pandas integer` when it fails to cast back to a `numpy integer` due to NaNs? Currently, the column remains as `float` in this scenario. @amontanez24 @pvk-developer

If your input was a `numpy.dtype` or `numpy.integer` that doesn't support `NaN` values then it is fine to still fall back to `np.float`, the reason being that your `pd.DataFrame` that you provided is `float`. If you do provide a `Int64` you with `NaNs` which is `pd.Int64` then we should return `pd.Int64` since this does support `NaN's`. But we shouldn't change anything from the input, our output should be the same as the input.

tests/integration/data_processing/test_data_processor.py
R-Palazzo: Yes this is good advice, thanks! Done in [5f4a918](https://github.com/sdv-dev/SDV/pull/2184/commits/5f4a918cc4a1fd8b76b21756572b5864cded84af)

tests/integration/data_processing/test_data_processor.py
```diff
@@ -401,3 +401,55 @@ def test_update_transformers_id_generator(self):
...
+        ('UInt16', pd.Series([1, 2, 3, pd.NA], dtype='UInt16'), 'UInt16'),
+        ('UInt32', pd.Series([1, 2, 3, pd.NA], dtype='UInt32'), 'UInt32'),
+        ('UInt64', pd.Series([1, 2, 3, pd.NA], dtype='UInt64'), 'UInt64'),
+        ('Float32', pd.Series([1.1, 2.2, 3.3, pd.NA], dtype='Float32'), 'Float32'),
+        ('Float64', pd.Series([1.1, 2.2, 3.3, pd.NA], dtype='Float64'), 'Float64'),
+        ('uint8', pd.Series([1, 2, 3, 4], dtype='uint8'), 'float'),
+        ('uint16', pd.Series([1, 2, 3, 4], dtype='uint16'), 'float'),
+        ('uint32', pd.Series([1, 2, 3, 4], dtype='uint32'), 'float'),
+        ('uint64', pd.Series([1, 2, 3, 4], dtype='uint64'), 'float'),
+        ('float', pd.Series([1.1, 2.2, 3.3, 4.4], dtype='float'), 'float'),
```
gsheni, line 420:
```suggestion
        ('float', pd.Series([1.1, 2.2, 3.3, 4.4], dtype='float8'), 'float'),
        ('float', pd.Series([1.1, 2.2, 3.3, 4.4], dtype='float32'), 'float'),
        ('float', pd.Series([1.1, 2.2, 3.3, 4.4], dtype='float64'), 'float'),
```

gsheni (approved):

sdv/data_processing/data_processor.py
```diff
@@ -902,6 +903,8 @@ def reverse_transform(self, data, reset_keys=False):
             reversed_data[column_name] = column_data[column_data.notna()]
             try:
                 reversed_data[column_name] = reversed_data[column_name].astype(dtype)
+            except IntCastingNaNError:  # Keep the column as float dtype
+                continue
```
amontanez24, line 907: maybe we can add a logger.debug statement here with a little info about the error

amontanez24 (commented):
One small comment but after that :shipit:

sdv/data_processing/data_processor.py
```diff
@@ -903,20 +903,23 @@ def reverse_transform(self, data, reset_keys=False):
...
-                continue
-            except ValueError as e:
+            except (IntCastingNaNError, ValueError) as e:
+                message = (
+                    f"The real data in '{column_name}' was stored as '{dtype}' but the "
+                    'synthetic data could not be cast back to this type. If this is a '
+                    'problem, please check your input data and metadata settings.'
+                )
+                if isinstance(e, IntCastingNaNError):
+                    LOGGER.info(message)
```
amontanez24, line 913: I think this one can be debug level

amontanez24 (approved):

PR #2185, opened 2024-08-14
Using FixedCombinations constraint with an integer constraint column causes sampling to fail

### LINKED ISSUE #2183: Using FixedCombinations constraint with an integer constraint column causes sampling to fail
### Environment Details

* SDV version: 1.15.0 (Latest)

### Problem Description

When using FixedCombinations, if the underlying dtype of a constraint column is `int`, then an error is thrown during sampling.

### Error Description

Error seems to start with this line of code: https://github.com/sdv-dev/SDV/blob/9301b964504ec53df977f1db5eab28b5b2e2c352/sdv/data_processing/data_processor.py

Screenshot 2024-08-12 at 4 50 57 PM

### Steps to reproduce

Dataset: [stock_missingcol.csv](https://github.com/user-attachments/files/16591774/stock_missingcol.csv)

Code:
```
from datetime import datetime

import pandas as pd
from sdv.evaluation.single_table import evaluate_quality, run_diagnostic

constrains_column = ['Quantity', 'Total Price']
real_data = pd.read_csv('stock_missingcol.csv')
real_data.head()
metadata.detect_from_dataframe(real_data)
metadata.update_columns(column_names=constrains_column, sdtype='categorical')

synthesizer = CTGANSynthesizer(metadata,epochs=2,verbose=True)
my_constraint = {
    'constraint_class': 'FixedCombinations',
    'constraint_parameters': {
        'column_names': constrains_column
    }
}

synthesizer.add_constraints(constraints=[
    my_constraint
])

synthesizer.fit(real_data)
synthetic_data = synthesizer.sample(num_rows=1000)
```

Internal Colab Notebook: https://colab.research.google.com/drive/1XUL42Wa13NQ2t0qewCoyCDYHHxmRO6ku?authuser=1#scrollTo=UGo3e-QJFk0a

#### Workaround

I was able to avoid this error by casting all int columns to float, then re-fitting, then sampling, then finally casting back to int in the sampled data. 

```
# In real data, cast from int -> float
real_data['Total Price'] = real_data['Total Price'].astype(float)

# In synthetic data, cast from float -> int
synthetic_data['Total Price'] = synthetic_data['Total Price'].astype(int) 
```

R-Palazzo:
CU-86b1np6ue
Resolve #2183

### DIFF
--- a/sdv/constraints/utils.py
+++ b/sdv/constraints/utils.py
@@ -204,3 +204,16 @@ def get_datetime_diff(high, low, high_datetime_format=None, low_datetime_format=
     diff_column = diff_column.astype(np.float64)
     diff_column[nan_mask] = np.nan
     return diff_column
+
+
+def get_mappable_combination(combination):
+    """Get a mappable combination of values.
+
+    This function replaces NaN values with None inside the tuple
+    to ensure consistent comparisons when using mapping.
+
+    Args:
+        combination (tuple):
+            A combination of values.
+    """
+    return tuple(None if pd.isna(x) else x for x in combination)
--- a/sdv/constraints/utils.py
+++ b/sdv/constraints/utils.py
@@ -215,5 +215,9 @@ def get_mappable_combination(combination):
     Args:
         combination (tuple):
             A combination of values.
+
+    Returns:
+        tuple:
+            A mappable combination of values.
     """
     return tuple(None if pd.isna(x) else x for x in combination)
--- a/tests/integration/constraints/test_tabular.py
+++ b/tests/integration/constraints/test_tabular.py
@@ -1,0 +1,38 @@
+import numpy as np
+import pandas as pd
+
+from sdv.metadata import SingleTableMetadata
+from sdv.single_table import GaussianCopulaSynthesizer
+
+
+def test_fixed_combinations_with_nans():
+    # Setup
+    data = pd.DataFrame({
+        'A': [1, 2, np.nan, 1, 2, 1],
+        'B': [10, 20, 30, 10, 20, 10],
+    })
+    metadata = SingleTableMetadata().load_from_dict({
+        'columns': {
+            'A': {'sdtype': 'categorical'},
+            'B': {'sdtype': 'categorical'},
+        }
+    })
+
+    synthesizer = GaussianCopulaSynthesizer(metadata)
+    my_constraint = {
+        'constraint_class': 'FixedCombinations',
+        'constraint_parameters': {'column_names': ['A', 'B']},
+    }
+    synthesizer.add_constraints(constraints=[my_constraint])
+
+    # Run
+    synthesizer.fit(data)
+    synthetic_data = synthesizer.sample(1000)
+
+    # Assert
+    assert len(synthetic_data) == 1000
+    pd.testing.assert_frame_equal(
+        synthetic_data.drop_duplicates(ignore_index=True),
+        data.drop_duplicates(ignore_index=True),
+        check_like=True,
+    )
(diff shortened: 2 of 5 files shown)


sdv-team:
Task linked: [CU-86b1np6ue SDV - Using FixedCombinations constraint with an integer constraint column causes sampling to fail #2183](https://app.clickup.com/t/86b1np6ue)

tests/integration/constraints/test_tabular.py
```diff
@@ -0,0 +1,39 @@
+import numpy as np
+import pandas as pd
+
+from sdv.metadata import SingleTableMetadata
+from sdv.single_table import GaussianCopulaSynthesizer
+
+
+def test_fixed_combinations_with_nans():
```
R-Palazzo, line 8: This test is crashing on the `main` branch with the error described in the issue

sdv/constraints/utils.py
```diff
@@ -204,3 +204,16 @@ def get_datetime_diff(high, low, high_datetime_format=None, low_datetime_format=
...
+
+def get_mappable_combination(combination):
+    """Get a mappable combination of values.
+
+    This function replaces NaN values with None inside the tuple
+    to ensure consistent comparisons when using mapping.
+
+    Args:
+        combination (tuple):
+            A combination of values.
```
amontanez24, line 217: can we add a return statement to the docstring

tests/integration/constraints/test_tabular.py
```diff
@@ -0,0 +1,39 @@
...
+    """Test that FixedCombinations constraint works with NaNs."""
+    # Setup
+    data = pd.DataFrame({
+        'A': [1, 2, np.nan, 1, 2, 1],
```
amontanez24, line 12: I don't get how this matches the issue if the issue said the column was an int? The nan will make this column a float

R-Palazzo: The bug turns out to be bigger than what we thought. It's related to FixedCombinations not being able to map combinations with NaN correctly, especially with a column with integer dtype, which causes a crash. The integer column in the test is the column B.

To give you more context, here's what is happening during the test:
- During fitting, the constraint maps the combination (`np.nan`, 30) to something.
- However, in the transformer, since we can't compare `np.nan`, the combination (np.nan, 30) is mapped to `None`.
- The UniformEncoder then learns that there is None in the column and will generate `None` in the same proportion.
- When the constraint reverses the transformation, there is no mapping for None to a combination, resulting in the value being set to `np.nan` for all the columns of the combinations.
- Because B is an integer column, it can't have `np.nan`, causing the data processor to crash when attempting to cast the column back to its dtype.

To fix this issue, we need to change all `np.nan` in a combination to `None`, making the combination mappable. You can test this fix using the following code snippet on the `main` branch. It shouldn't crash this time because B is an object column but there will be no (`np.nan`, ’30’) combination in the synthetic data. 
```python
import numpy as np
import pandas as pd

data = pd.DataFrame({
    'A': [1, 2, np.nan, 1, 2, 1],
    'B': ['10', '20', '30', '10', '20', '10'],
})
metadata = SingleTableMetadata().load_from_dict({
    'columns': {
        'A': {'sdtype': 'categorical'},
        'B': {'sdtype': 'categorical'},
    }
})

my_constraint = {
    'constraint_class': 'FixedCombinations',
    'constraint_parameters': {'column_names': ['A', 'B']},
}
synthesizer.add_constraints(constraints=[my_constraint])

# Run
synthesizer.fit(data)
synthetic_data = synthesizer.sample(1000)
```
Let me know if it makes sense

amontanez24: That does make sense. Can we add a test with two int columns in the combination to sanity check the issue if such a test doesn't already exist?

R-Palazzo: Yes, done in [617d5c4](https://github.com/sdv-dev/SDV/pull/2185/commits/617d5c4bfb78e7ee58bc81ea6fcf368faf6470a6)

amontanez24 (approved):

fealho (approved):

Issue #643, opened 2021-11-21
Use transformed data for multiple models or subsequent fitting runs (pipelining SDV).


xamm:
### Problem Description

When fitting a model, the input data is first transformed, and then the model is trained using the transformed data.
The transformation of the data accounts for a large portion of the actual fitting time for the model. Moreover, this transformation is performed on the CPU.

The transformed inputs are not stored, and for each model fitting, the inputs must be transformed again.
When evaluating multiple models, hyperparameters, or input sizes, this transformation causes a high overhead.

### Expected behavior / Feature request

It would be nice if the processes of data conversion and training of the actual SDV model were split into two parts.
This would allow the data to be transformed first and then the transformed data to be reused and also shared.

### Advantages

- Transformed inputs can be reused / shared
- Larger inputs can be transformed and then sampled for fitting
  - Instead of subsetting inputs, transforming them and then fitting them
- Transforming inputs on the CPU, then training on the GPU
  - could make the fitting process more efficient
  - reduces the cost

### Example Pipeline
![general_pipeline gv](https://user-images.githubusercontent.com/39380924/142760952-e7f99eb5-a11e-4a5b-b430-433f278dc745.gif)

npatki:
Hello @xamm, seems like with the introduction of SDV 1.0 this workflow is now possible.

### Recommended Usage
We recommend using the [RDT library](https://docs.sdv.dev/rdt) library directly to transform your original data into numerical data (and back!). You'll need to choose which transformers to use based on what your data represents.

```python
from rdt import HyperTransformer
from rdt.transformers.categorical import LabelEncoder

# preprocess the data ourselves using the RDT library
ht = HyperTransformer()
ht.detect_initial_config(original_data)
ht.update_transformers(column_name_to_transformer={
    'column_name': LabelEncoder(add_noise=True)
})
numerical_data = ht.fit_transform(df) 
```

Now, you can use the numerical data to fit any of the synthesizers.
```python
from sdv.metadata import SingleTableMetadata
from sdv.single_table import GaussianCopulaSynthesizer

metadata = SingleTableMetadata()
metadata.detect_from_dataframe(numerical_data) # this will detect everything as numerical
synth = GaussianCopulaSynthesizer(metadata)
synth.fit(data)
```
Finally, after sampling synthetic data, remember to reverse transform it back to get the original formatting.
```python
raw_synthetic_data = synth.sample(num_rows=1000)
[...]

PR #2190, opened 2024-08-19
Add `data` parameter to `_detect_relationships` method 

fealho:
CU-86b0zfny6, Resolve https://github.com/datacebo/SDV-Enterprise/issues/643.

### DIFF
--- a/sdv/metadata/multi_table.py
+++ b/sdv/metadata/multi_table.py
@@ -560,25 +560,6 @@ class MultiTableMetadata:
 
         self._detect_relationships(data)
 
-    def detect_table_from_csv(self, table_name, filepath, read_csv_parameters=None):
-        """Detect the metadata for a table from a csv file.
-
-        Args:
-            table_name (str):
-                Name of the table to detect.
-            filepath (str):
-                String that represents the ``path`` to the ``csv`` file.
-            read_csv_parameters (dict):
-                A python dictionary of with string and value accepted by ``pandas.read_csv``
-                function. Defaults to ``None``.
-        """
-        self._validate_table_not_detected(table_name)
-        table = SingleTableMetadata()
-        data = _load_data_from_csv(filepath, read_csv_parameters)
-        table._detect_columns(data)
-        self.tables[table_name] = table
-        self._log_detected_table(table)
-
     def detect_from_csvs(self, folder_name, read_csv_parameters=None):
         """Detect the metadata for all tables in a folder of csv files.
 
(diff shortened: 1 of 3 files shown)


sdv-team:
Task linked: [CU-86b0zfny6 SDV-Enterprise - Improve foreign key detection algorithm #643](https://app.clickup.com/t/86b0zfny6)

sdv/metadata/multi_table.py
```diff
@@ -595,7 +603,12 @@ def detect_from_csvs(self, folder_name, read_csv_parameters=None):
             table_name = csv_file.stem
             self.detect_table_from_csv(table_name, str(csv_file), read_csv_parameters)
 
-        self._detect_relationships()
+        read_csv_parameters = read_csv_parameters or {}
+        data = {
+            csv_file.stem: pd.read_csv(str(csv_file), **read_csv_parameters)
+            for csv_file in csv_files
+        }
```
amontanez24, line 610: should we just add the data to the dict in the for loop above?

sdv/metadata/multi_table.py
```diff
@@ -599,15 +599,13 @@ def detect_from_csvs(self, folder_name, read_csv_parameters=None):
         if not csv_files:
             raise ValueError(f"No CSV files detected in the folder '{folder_name}'.")
 
+        data = {}
+        read_csv_parameters = read_csv_parameters or {}
         for csv_file in csv_files:
             table_name = csv_file.stem
             self.detect_table_from_csv(table_name, str(csv_file), read_csv_parameters)
+            data[csv_file.stem] = pd.read_csv(str(csv_file), **read_csv_parameters)
```
frances-h, line 607: Can we just use the `_load_from_csv` util? Also instead of having to load the CSV twice, would it make more sense to load it and then use `detect_from_dataframe` instead?

fealho: @frances-h Sure. Do you know if the `detect_table_from_csv` method is user facing? Otherwise this implementation removes the only usage of that method, so I would remove it.

frances-h: @amontanez24 thoughts? It's not in the docs but also doesn't have an _ preceding the function name. Maybe leave it for now and come back to it later?

frances-h (approved):

sdv/metadata/multi_table.py
amontanez24: Yeah it's a private method, but you'd want to make sure it doesn't cause anything else to fail. I'm not sure if we use it anywhere else or in enterprise at all

amontanez24 (approved):

sdv/metadata/multi_table.py
fealho: It's not used in SDV, SDV-Enterprise or SDGym.

amontanez24: Then it should be fine to remove

PR #2202, opened 2024-08-29
Hide warnings when using Metadata in PARSynthesizer

lajohn4747:
Fixing issue from bug bash:

- Hide `SingleTableMetadata` future warnings when using PARSynthesizer.
- Hide `Sequential modeling is not yet supported on SDV Multi Table models.` warnings coming from `MultiTableMetadata` unless there is more than one table in `Metadata`

### DIFF
--- a/sdv/metadata/metadata.py
+++ b/sdv/metadata/metadata.py
@@ -110,8 +110,6 @@ class Metadata(MultiTableMetadata):
                 Name of the sequence index column.
         """
         self._validate_table_exists(table_name)
-        if len(self.tables) > 1:
-            warnings.warn('Sequential modeling is not yet supported on SDV Multi Table models.')
         self.tables[table_name].set_sequence_index(column_name)
 
     def set_sequence_key(self, table_name, column_name):
@@ -124,6 +122,4 @@ class Metadata(MultiTableMetadata):
                 Name (or tuple of names) of the sequence key column(s).
         """
         self._validate_table_exists(table_name)
-        if len(self.tables) > 1:
-            warnings.warn('Sequential modeling is not yet supported on SDV Multi Table models.')
         self.tables[table_name].set_sequence_key(column_name)
--- a/sdv/metadata/metadata.py
+++ b/sdv/metadata/metadata.py
@@ -110,6 +110,10 @@ class Metadata(MultiTableMetadata):
                 Name of the sequence index column.
         """
         self._validate_table_exists(table_name)
+        if len(self.tables) > 1:
+            warnings.warn(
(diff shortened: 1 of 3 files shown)


sdv/metadata/metadata.py
```diff
@@ -99,3 +99,31 @@ def _convert_to_single_table(self):
...
+        Args:
+            table_name (str):
+                Name of the table to set the sequence key.
+            column_name (str, tulple[str]):
+                Name (or tuple of names) of the sequence key column(s).
+        """
+        self._validate_table_exists(table_name)
+        if len(self.tables) > 1:
+            warnings.warn('Sequential modeling is not yet supported on SDV Multi Table models.')
+        self.tables[table_name].set_sequence_key(column_name)
```
lajohn4747, line 129: These two methods in `MultiTableMetadata` would always throw the warning when used in `PARSynthesizer`

sdv/metadata/metadata.py
```diff
@@ -99,3 +99,31 @@ def _convert_to_single_table(self):
...
+
+        Args:
+            table_name (str):
+                Name of the table to set the sequence index.
+            column_name (str):
+                Name of the sequence index column.
+        """
+        self._validate_table_exists(table_name)
+        if len(self.tables) > 1:
+            warnings.warn('Sequential modeling is not yet supported on SDV Multi Table models.')
```
amontanez24, line 114: I think you can get rid of the yet as I don't think there is a plan to add support for that

sdv/metadata/metadata.py
```diff
@@ -99,3 +99,31 @@ def _convert_to_single_table(self):
...
+
+        Args:
+            table_name (str):
+                Name of the table to set the sequence key.
+            column_name (str, tulple[str]):
+                Name (or tuple of names) of the sequence key column(s).
+        """
+        self._validate_table_exists(table_name)
+        if len(self.tables) > 1:
+            warnings.warn('Sequential modeling is not yet supported on SDV Multi Table models.')
```
amontanez24, line 128: same here

sdv/metadata/metadata.py
lajohn4747: Should we put a different message instead?

fealho:
@lajohn4747 Why did we decide to remove the warning message?

lajohn4747:
> @lajohn4747 Why did we decide to remove the warning message?

I decided to move it to an error message in PARSynthesizer. Adding keys to Metadata that contains multiple data might be an edge case so I think that only throwing an error when the try to use multi-table Metadata would be a more useful msg.

fealho (approved):

PR #2203, opened 2024-08-30
Add warnings that were suggested from metadata bughunt

lajohn4747:
Handle the remaining usability issues by adding warnings to let the user know that they will get a default table name.

Also add errors to prevent the user from adding invalid `SingleTableMetadata` arguments  into `Metadata`

### DIFF
--- a/sdv/metadata/metadata.py
+++ b/sdv/metadata/metadata.py
@@ -70,6 +70,10 @@ class Metadata(MultiTableMetadata):
         else:
             if single_table_name is None:
                 single_table_name = self.DEFAULT_SINGLE_TABLE_NAME
+                warnings.warn(
+                    'Did not assign a table name to single table. '
+                    f'Assigning name: {single_table_name}'
+                )
             self.tables[single_table_name] = SingleTableMetadata.load_from_dict(metadata)
 
     def _get_single_table_name(self):
--- a/sdv/metadata/metadata.py
+++ b/sdv/metadata/metadata.py
@@ -71,7 +71,7 @@ class Metadata(MultiTableMetadata):
             if single_table_name is None:
                 single_table_name = self.DEFAULT_SINGLE_TABLE_NAME
                 warnings.warn(
-                    'Did not assign a table name to single table. '
+                    'No table name was provided to metadata containing only one table. '
                     f'Assigning name: {single_table_name}'
                 )
             self.tables[single_table_name] = SingleTableMetadata.load_from_dict(metadata)
--- a/sdv/metadata/multi_table.py
+++ b/sdv/metadata/multi_table.py
@@ -409,6 +409,8 @@ class MultiTableMetadata:
             **kwargs:
                 Any key word arguments that describe metadata for the columns.
         """
+        if not isinstance(column_names, list):
+            raise InvalidMetadataError('Please pass in a list to column_names arg.')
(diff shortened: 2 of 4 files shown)


sdv/metadata/multi_table.py
```diff
@@ -409,6 +409,8 @@ def update_columns(self, table_name, column_names, **kwargs):
             **kwargs:
                 Any key word arguments that describe metadata for the columns.
         """
+        if not isinstance(column_names, list):
```
fealho, line 412: This wouldn't work for tuples and other list-like objects, something like `pandas.api.types is_list_like` could handle those.

If that's expected then you can leave as is.

sdv/metadata/multi_table.py
```diff
@@ -843,6 +845,9 @@ def validate_data(self, data):
             A warning is being raised if ``datetime_format`` is missing from a column represented
             as ``object`` in the dataframe and its sdtype is ``datetime``.
         """
+        if not isinstance(data, dict):
+            raise InvalidMetadataError('Please pass in a dictionary mapping tables to dataframes.')
```
fealho, line 849: The docstring says it expects a dataframe.

fealho (approved):
Looks good, I left two minor comments.

sdv/metadata/multi_table.py
lajohn4747: I think that's expected. Don't imagine a user putting in something more complicated than a list of strings and it's easy enough to convert.

sdv/metadata/metadata.py
```diff
@@ -70,6 +70,10 @@ def _set_metadata_dict(self, metadata, single_table_name=None):
         else:
             if single_table_name is None:
                 single_table_name = self.DEFAULT_SINGLE_TABLE_NAME
+                warnings.warn(
+                    'Did not assign a table name to single table. '
```
amontanez24, line 74: maybe we can be slightly more verbose
```suggestion
                    'No table name was provided to metadata containing only one table.  '
```

tests/unit/metadata/test_multi_table.py
```diff
@@ -3064,3 +3064,40 @@ def test_anonymize(self, mock_load):
             'parent_primary_key': 'col1',
             'child_foreign_key': 'col2',
         }
+
+    def test_update_columns_no_list_error(self):
+        # Setup
```
amontanez24, line 3069: minor: can we add small docstrings for both tests here adding a little more detail

sdv/multi_table/base.py
```diff
@@ -77,7 +77,8 @@ def _initialize_models(self):
         with disable_single_table_logger():
             for table_name, table_metadata in self.metadata.tables.items():
                 synthesizer_parameters = self._table_parameters.get(table_name, {})
-                metadata = Metadata.load_from_dict(table_metadata.to_dict())
+                metadata_dict = {'tables': {table_name: table_metadata.to_dict()}}
+                metadata = Metadata.load_from_dict(metadata_dict)
```
amontanez24, line 81: was this crashing?

lajohn4747: No it was causing the single table metadata name to be missing since it gave it a default table (it was in a SingleTableMetadata format)

amontanez24 (approved):

PR #2204, opened 2024-09-05
Fix circular relationships validation

### LINKED ISSUE #2205: The `_validate_circular_relationships` method may fail to detect circular relationships
### Problem Description

The `_validate_circular_relationships` method in `MultiTableMetadata` has an implementation bug which makes it not detect some circular relationships.

### Expected behavior

The issue happens in the following lines:
https://github.com/sdv-dev/SDV/blob/f4be5cfd000fc659e95de985e52e06aba5594a6d/sdv/metadata/multi_table.py#L139-L141

It breaks out of the loop whenever it finds a child table that has already been visited during the traversal, potentially missing circular relationships for the other child tables.

Changing the `break` statement to `continue` avoids this issue.

fealho:
CU-86b21jrjd, Resolve #2205.

### DIFF
--- a/sdv/metadata/multi_table.py
+++ b/sdv/metadata/multi_table.py
@@ -126,10 +126,10 @@ class MultiTableMetadata:
                 )
 
     def _validate_circular_relationships(
-        self, parent, children=None, parents=None, child_map=None, errors=None
+        self, parent, children=None, visited=None, child_map=None, errors=None
     ):
         """Validate that there is no circular relationship in the metadata."""
-        parents = set() if parents is None else parents
+        visited = set() if visited is None else visited
         if children is None:
             children = child_map[parent]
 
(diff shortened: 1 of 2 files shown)


sdv-team:
Task linked: [CU-86b21jrjd SDV - Fix circular relationships validation](https://app.clickup.com/t/86b21jrjd)

frances-h (approved):

sdv/metadata/multi_table.py
```diff
@@ -138,7 +138,7 @@ def _validate_circular_relationships(
 
         for child in children:
             if child in parents:
```
amontanez24, line 140: minor: can we rename parents to visited? I was confused reading this thinking that we were checking the parents of only the current node, when this list actually keeps growing.

amontanez24 (commented):
Good catch! One suggested renaming then :shipit:

amontanez24 (approved):

PR #2228, opened 2024-09-16
For single-table use cases, make it frictionless to update Metadata

### LINKED ISSUE #2213: For single-table use cases, make it frictionless to update Metadata
### Problem Description
In an upcoming version of SDV, we will be consolidating the Multi and Single table metadata objects into a streamlined Metadata object. Effectively, this object is the same as multi-table, so it includes functions that are suited for detecting multiple tables at once.

However, if you have a single table, it should still be usable without adding much friction. Currently, all the Metadata updating functions require you to input a table name, which really isn't necessary when all you have is a single table.

### Expected behavior
For all of the metadata update functions (listed below), make the `table_name` parameter optional. If necessary, this means we should move its position to be _after_ all the required parameters.

Then update the functionality as follows:
- If there are multiple tables in the Metadata, a `table_name` should be provided. If it is not provided, throw an error.
```
Error: There are multiple tables in your metadata. Please specify a 'table_name' to use this function.
```
- But if there is only a single table in the Metadata, then it's ok if `table_name` is missing. Proceed with the desired functionality.

### Additional context
Make the change for the following functions:

- `get_column_names`
- `update_column`
- `update_columns`
- `update_columns_metadata`
- `add_column`
- `add_column_relationship`
- `set_primary_key`
- `remove_primary_key`
- `add_alternate_keys`

R-Palazzo:
Resolve #2213


sdv/metadata/metadata.py
```diff
@@ -173,7 +173,7 @@ def set_sequence_index(self, table_name, column_name):
         self._validate_table_exists(table_name)
         self.tables[table_name].set_sequence_index(column_name)
 
-    def set_sequence_key(self, table_name, column_name):
+    def set_sequence_key(self, *args, **kwargs):
```
amontanez24, line 176: I think for all of these we can just use the signature normally. For this one it would be `set_sequence_key(self, column_name, table_name)`. I don't think we have to support a different order

sdv/metadata/metadata.py
```diff
@@ -206,3 +209,86 @@ def validate_table(self, data, table_name=None):
...
+
+    def get_column_names(self, *args, **kwargs):
+        """Return a list of column names that match the given metadata keyword arguments."""
+        parameters = self._resolve_arguments([], *args, **kwargs)
+        return super().get_column_names(**parameters)
+
+    def update_column(self, *args, **kwargs):
+        """Update an existing column for a table in the ``Metadata``."""
+        parameters = self._resolve_arguments(['column_name'], *args, **kwargs)
+        super().update_column(**parameters)
```
amontanez24, line 259: I don't think we need this. Can we just list the args normally and then pass them in the correct order when calling super?

amontanez24 (changes requested):

lajohn4747 (approved):
lgtm

amontanez24 (approved):
:shipit:

PR #2231, opened 2024-09-18
Update `_is_datetime_type` to handle nanoseconds

fealho:
CU-86b253u85, Resolve https://github.com/datacebo/SDV-Enterprise/issues/815

### DIFF
--- a/sdv/_utils.py
+++ b/sdv/_utils.py
@@ -81,6 +81,7 @@ def _is_datetime_type(value):
             bool(_get_datetime_format([value]))
             or isinstance(value, pd.Timestamp)
             or isinstance(value, datetime)
+            or (isinstance(value, str) and pd.notna(pd.to_datetime(value, errors='coerce')))
         ):
             return False
 
--- a/tests/unit/test__utils.py
+++ b/tests/unit/test__utils.py
@@ -240,6 +240,30 @@ def test__is_datetime_type_with_datetime_str():
     assert is_datetime
 
 
+def test__is_datetime_type_with_datetime_str_nanoseconds():
+    """Test it for a datetime string with nanoseconds."""
+    # Setup
+    value = '2011-10-15 20:11:03.498707'
+
+    # Run
+    is_datetime = _is_datetime_type(value)
+
+    # Assert
+    assert is_datetime
+
+
+def test__is_datetime_type_with_str_int():
+    """Test it for a string with an integer."""
+    # Setup
+    value = '123'
+
+    # Run
+    is_datetime = _is_datetime_type(value)
+
+    # Assert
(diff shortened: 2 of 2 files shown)


sdv-team:
Task linked: [CU-86b253u85 SDV-Enterprise - HSA crashes when using DayZ data with no datetime format #815](https://app.clickup.com/t/86b253u85)

sdv/_utils.py
```diff
@@ -81,6 +81,7 @@ def _is_datetime_type(value):
             bool(_get_datetime_format([value]))
```
amontanez24, line 81: why doesn't guessing the datetime format for this string format work?

fealho: It can't guess times.

amontanez24: What do you mean by times?

amontanez24: It's able to guess the following
```python
a = np.array(['09-12-2020 00:00:00'], dtype=object)
_guess_datetime_format_for_array(a.astype(object))
```
yields
```
'%m-%d-%Y %H:%M:%S'
```

amontanez24: It also successfully guesses
```python
 a = np.array(['09-12-2020 00:00:00.010122'], dtype=object)
_guess_datetime_format_for_array(a.astype(object))
```
yields
```
'%m-%d-%Y %H:%M:%S.%f'
```

fealho: Curious. `'2011-10-15 21:11:03.498707'` and `'2011-10-15 19:11:03.498707'` work, but it fails for `'2011-10-15 20:11:03.498707'`.

fealho: @amontanez24 The issue happens when year and hour:minute have the same number (e.g., 2023 and 20:23). Pandas [addressed this bug 5 months ago](https://github.com/pandas-dev/pandas/pull/57471), but I don't know if they have released it yet (they are in the process of making a release [right now](https://github.com/pandas-dev/pandas/pull/59840), so it's possible this will get fixed in 2.2.3).

amontanez24: Ah makes sense. I guess my follow up question is why use this function at all instead of just `pd.notna(pd.to_datetime(value, errors='coerce')))`? That seems to work for a lot of formats

fealho: We can delete it. I left it to preserve previous behavior - I'm not certain `to_datetime` includes every case that `_get_datetime_format` does - but since it does pass every test case, we can probably delete it.

amontanez24 (approved):

pvk-developer (approved):

pvk-developer:
LGTM!

PR #2251, opened 2024-10-03
Use `Metadata` instead of `SingleTableMetadata` when using `MultiTableSynthesizers`

### LINKED ISSUE #2250: Extraneous metadata warning is printed when customizing HMA Synthesizer
### Environment Details
* SDV version: 1.17.0 (latest)
* Python version: 3.10
* Operating System: Linux (Colab)

### Error Description
I am using HMA and providing the new, streamlined `Metadata` object for it. If I try to customize the synthesizer (eg. by setting distributions for individual columns), then HMA unnecessarily complains that I am using `SingleTableMetadata`. 

```
FutureWarning: The 'SingleTableMetadata' is deprecated. Please use the new 'Metadata' class for synthesizers.
```

I expect that there should be no such warning because I am using the correct `Metadata` object.

### Steps to reproduce

Use the `fake_hotels` demo dataset to replicate. The `download_demo` provides the new, streamlined Metadata object so there should be no warning.

```python

real_data, metadata = download_demo(modality='multi_table', dataset_name='fake_hotels')

custom_synthesizer = HMASynthesizer(metadata, verbose=False)
custom_synthesizer.set_table_parameters(
    table_name='hotels',
    table_parameters={ 'default_distribution': 'truncnorm' })
custom_synthesizer.fit(real_data)
```
```
FutureWarning: The 'SingleTableMetadata' is deprecated. Please use the new 'Metadata' class for synthesizers.
```

### Additional Context
Note that the warning goes away if I just use the default HMA (i.e. I do not apply `set_table_parameters`)
```
default_synthesizer = HMASynthesizer(metadata)
default_synthesizer.fit(data)
```

pvk-developer:
Resolves #2250 
CU-86b2ckcm7

### DIFF
--- a/tests/unit/metadata/test_metadata.py
+++ b/tests/unit/metadata/test_metadata.py
@@ -518,8 +518,7 @@ class TestMetadataClass:
             - Instance of ``Metadata`` with all valid tables and no relationships.
         """
         # Setup
-        metadata = metadata_instance
-        metadata_no_relationships = metadata.to_dict()
+        metadata_no_relationships = metadata_instance.to_dict()
         del metadata_no_relationships['relationships']
         test_metadata = Metadata.load_from_dict(metadata_no_relationships)
 
--- a/tests/unit/metadata/test_metadata.py
+++ b/tests/unit/metadata/test_metadata.py
@@ -1,5 +1,5 @@
 import re
-from unittest.mock import Mock, call, patch
+from unittest.mock import Mock, call, mock_open, patch
 
 import pandas as pd
 import pytest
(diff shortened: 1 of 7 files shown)


sdv-team:
Task linked: [CU-86b2ckcm7 SDV - Extraneous metadata warning is printed when customizing HMA Synthesizer #2250](https://app.clickup.com/t/86b2ckcm7)

lajohn4747 (approved):
lgtm

pvk-developer:
@amontanez24 After testing the `self.tables` attribute with a custom dictionary, I discovered that this approach is inefficient and leads to an infinite loop when modifying the `Metadata` instance.

In our `MultiTableMetadata` code, we continue accessing self.tables for adding new columns or making other updates. However, by implementing a custom dictionary that always returns a `Metadata` object, we create a recursion issue, as the same method is repeatedly called, resulting in an infinite loop.

The most effective solution I see is to override the existing `get_table_metadata` method in the `Metadata` class. This would allow us to retrieve and return the correct `Metadata` instance. This approach would avoid the infinite loop and ensure that we can access the correct object when we want to recreate a synthesizer.

tests/unit/metadata/test_metadata.py
```diff
@@ -1,74 +1,77 @@
...
-            },
-            'payments': {
-                'columns': {
-                    'payment_id': {'sdtype': 'id'},
-                    'user_id': {'sdtype': 'id'},
-                    'date': {'sdtype': 'datetime'},
-                },
-                'primary_key': 'payment_id',
+@pytest.fixture
+def metadata_instance():
```
pvk-developer, line 15: Added this fixture because we had `self.get_multi_table_metadata` and `get_multi_table_metadata` which was very confusing (self returned one instance and the `get_` without self returned another).

tests/unit/metadata/test_metadata.py
```diff
@@ -147,9 +150,10 @@ def test_load_from_json_single_table(self, mock_read_json):
             assert instance.tables[table_name].sequence_index is None
             assert instance.tables[table_name]._version == 'SINGLE_TABLE_V1'
 
+    @patch('builtins.open', new_callable=mock_open)
     @patch('sdv.metadata.utils.Path')
     @patch('sdv.metadata.utils.json')
-    def test_load_from_json_multi_table(self, mock_json, mock_path):
+    def test_load_from_json_multi_table(self, mock_json, mock_path, mock_file):
```
pvk-developer, line 156: This test was failing when run standalone -- With this we fix the issue of running just this file as test.

amontanez24 (approved):
LGTM!

PR #2269, opened 2024-10-24
Add support for numpy 2.0.0

### LINKED ISSUE #2078: Add support for numpy 2.0.0
### Problem Description
- Currently, we restrict numpy < 2 for all python versions.
- We would like to support the newest major release of numpy. 

### Expected behavior
- SDV supports numpy 2.0

### Additional context
- Remove upper bound restriction in pyproject.toml
- Update any associated tests

R-Palazzo:
Resolve #2078
CU-86b0y0uu0


sdv-team:
Task linked: [CU-86b0y0uu0 SDV - Add support for numpy 2.0.0 #2078](https://app.clickup.com/t/86b0y0uu0)

sdv/data_processing/data_processor.py
```diff
@@ -571,7 +572,13 @@ def _create_config(self, data, columns_created_by_constraints):
...
                     bothify_format = 'sdv-id-??????'
                     if is_numeric:
-                        bothify_format = '#########'
+                        column_dtype = str(column_dtype).lower()
+                        if 'int8' in column_dtype:
+                            bothify_format = '##'
+                        elif 'int16' in column_dtype:
+                            bothify_format = '####'
+                        else:
+                            bothify_format = '#########'
```
R-Palazzo, line 581: I had to make this change to fix the dtype benchmarking because NumPy 2.0 raises an overflow error when you want to convert object numbers that exceed the value range of the dtype. Before, the behavior was weird; it was modifying the values of the number to be within the range allowed by the dtype.
Let me know if it makes sense.

amontanez24: This makes sense. The only issue I see is that we're cutting the range of possible values to less than what it can be (eg. [-99, 99] instead of [-128, 127]. Maybe we should handle this in the place where we actually cast the string to an int?

pvk-developer:
The reason why `pa.float32` is failing with `id's` is because of this:
```python
*** pyarrow.lib.ArrowInvalid: Integer value 640058449 not in range: -16777216 to 16777216
```

This also raises a new issue about the benchmarking, to check if the sampled range is the expected one, currently it does not check, the same number or error is generated and therefore the test fails. We fallback to `object` because we couldn't cast it.

tests/benchmark/numpy_dtypes.py
```diff
@@ -61,14 +61,14 @@
     }),
     'np.string': pd.DataFrame({
         'np.string': pd.Series([
-            np.string_('string1'),
-            np.string_('string2'),
-            np.string_('string3'),
+            np.bytes_('string1'),
+            np.bytes_('string2'),
+            np.bytes_('string3'),
```
pvk-developer, line 66: Can we use `str_` instead ? And duplicate this to `np.bytes` or something and add the same strings but with `bytes` instead.

pvk-developer (approved):
Left a minor comment but otherwise it looks good 👍🏻

sdv/constraints/tabular.py
```diff
@@ -334,6 +334,12 @@ def _transform(self, table_data):
             pandas.DataFrame:
                 Transformed data.
         """
+        # To make the NaN to None mapping work for pd.Categorical data, we need to convert
+        # the columns to object before replacing NaNs with None.
+        for column in self._columns:
+            if pd.api.types.is_categorical_dtype(table_data[column]):
+                table_data[column] = table_data[column].astype(object)
+
```
R-Palazzo, line 342: Adding the np.bytes revealed an actual bug in the FixedCombinations with pd.Categorical and NaNs. It was not caught before due to reject sampling; before enough rows were generated, but none with a combination including NaNs in the categorical column. After adding np.bytes and somehow a bit randomly, the synthesizer was able to generate only 9 out of the 10 rows. Changing the number of rows to sample also made the test pass but did not fix the bug, haha. This fixes the bug, and I added a test for it. Let me know if it makes sense

pvk-developer (approved):
Yes, the change about the bug you found makes sense.

sdv/constraints/tabular.py
amontanez24: I'm confused why adding bytes revealed this when the check is for is_categorical_dtype. Bytes are not a categorical dtype

sdv/constraints/tabular.py
```diff
@@ -334,6 +334,14 @@ def _transform(self, table_data):
             pandas.DataFrame:
                 Transformed data.
         """
+        # To make the NaN to None mapping work for pd.Categorical data, we need to convert
+        # the columns to object before replacing NaNs with None.
+        table_data[self._columns] = table_data[self._columns].astype({
+            col: object
+            for col in self._columns
+            if pd.api.types.is_categorical_dtype(table_data[col])
+        })
```
amontanez24, line 343: Based on [this discussion](https://github.com/pandas-dev/pandas/issues/40472), should we just use fillna instead of replace? Then we don't have to convert

amontanez24 (commented):
Looks good, just one comment

sdv/constraints/tabular.py
amontanez24: @R-Palazzo I forgot that fillna cannot be used with None which is probably why we have this line of code in the first place. This solution is good

amontanez24 (approved):

PR #2279, opened 2024-11-05
Add warning when unable to turn off rounding scheme for a column 

### LINKED ISSUE #2266: Unable to turn off rounding scheme for a column (need a warning)
### Environment Details
* SDV version: 1.17.10 (latest)
* Python version: 3.10

### Error Description
I am explicitly trying to turn off rounding for a column, but if I set up `FloatFormatter` with `learn_rounding_scheme=False`, I see that this doesn't have any effect.

### Steps to reproduce

```python
from rdt.transformers.numerical import FloatFormatter

    modality='single_table',
)

    'amenities_fee': FloatFormatter(learn_rounding_scheme=False) # I'm trying to turn off rounding for this column
})

synthesizer.fit(data)
synthesizer.sample(5)
```
image

### Expected Results
The code itself is actually working as intended -- because the rounding is being enforced at a synthesizer level, it's not possible to override it by just updating a transformer. In this case, I expect a warning should be thrown after `update_transformers`.

```
Warning: Unable to turn off rounding scheme for column 'amenities_fee', because the overall synthesizer is
enforcing rounding. We recommend setting the synthesizer's 'enforce_rounding' parameter to False.
```

Note that this warning should only appear if:
- The synthesizer's `enforce_rounding` is set to True (this is also the default) AND
- The user is providing a transformer with `learn_rounding_scheme=False` (they are trying to turn it off)

### Additional Context
Note that the behavior is fine the other way around -- i.e. if the synthesizer's `enforce_rounding=False`, I am still able to turn on rounding for individual columns with using the `FloatFormatter`. So in this particular case, I could do this instead:

```python
# at synthesizer level, do not enforce rounding for any of the columns
synthesizer = GaussianCopulaSynthesizer(metadata, enforce_rounding=False) 

# now, you can selectively turn rounding back ON for any columns you need
    'room_rate': FloatFormatter(learn_rounding_scheme=True)
})

synthesizer.fit(data)
synthesizer.sample(5)
```

fealho:
CU-86b2jarvu, Resolve #2266.


sdv-team:
Task linked: [CU-86b2jarvu SDV - Unable to turn off rounding scheme for a column (need a warning) #2266](https://app.clickup.com/t/86b2jarvu)

sdv/single_table/base.py
```diff
@@ -259,6 +259,20 @@ def _warn_for_update_transformers(self, column_name_to_transformer):
...
 
+    def _warn_unable_to_enforce_rounding(self, column_name_to_transformer):
+        if self.enforce_rounding:
+            for column, transformer in column_name_to_transformer.items():
+                if (
+                    hasattr(transformer, 'learn_rounding_scheme')
+                    and not transformer.learn_rounding_scheme
+                ):
+                    warnings.warn(
+                        f"Unable to turn off rounding scheme for column '{column}', "
```
frances-h, line 270: Should we warn once for all columns instead of re-warning for each?

frances-h (approved):

gsheni (approved):

PR #2282, opened 2024-11-07
Update integer boundaries for ID columns

R-Palazzo:
I set the max boundary to work with pa.float32 so all the benchmark dtypes also work.


frances-h (approved):

pvk-developer (approved):

sdv/data_processing/data_processor.py
```diff
@@ -124,6 +121,10 @@ def __init__(
         self._constraints = []
         self._constraints_to_reverse = []
         self._custom_constraint_classes = {}
+        self._COLUMN_RELATIONSHIP_TO_TRANSFORMER = {
+            'address': 'RandomLocationGenerator',
+            'gps': 'GPSNoiser',
+        }
```
R-Palazzo, line 127: This is to fix the `CompositeKey` PR on enterprise. This happens because I used DayZ to generate data and without this change some integration tests were failing because DayZ was updating the `_COLUMN_RELATIONSHIP_TO_TRANSFORMER` at the DataProcessor class level instead of just for the DataProcessor inside DayZ

amontanez24 (approved):

PR #2292, opened 2024-11-13
 Modularize `validate` method for multi-table synthesizers

### LINKED ISSUE #2291: Modularize `validate` method for multi-table synthesizers
### Problem Description
Currently, we cannot easily extend validation for multi-table synthesizers because most of the functionality is contained in the public `BaseMultiTableSynthesizer.validate` method. We should modularize the `validate` method so it can be extended more easily.

### Expected behavior
In order to support additional validation functionality, we should move the existing validation functionality out of the `validate` method and into the helper method `_validate`. The `validate` method should just call `_validate`.

Additionally, in the `preprocess` method, validation of the data should occur _before_ the `_transform_helper` method is called on the data.

frances-h:
Resolve #2291

### DIFF
--- a/sdv/multi_table/base.py
+++ b/sdv/multi_table/base.py
@@ -240,15 +240,8 @@ class BaseMultiTableSynthesizer:
 
         return errors
 
-    def validate(self, data):
-        """Validate the data.
-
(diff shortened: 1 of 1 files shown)


frances-h:
Out of scope for now, closing

PR #2301, opened 2024-11-18
Raise `UserWarnings` for Unused Numerical Distributions when using `GaussianCopulaSynthesizer`

### LINKED ISSUE #2296: When using GaussianCopula, show a warning if a column in `numerical_distributions` cannot be applied
### Environment Details
* SDV version: 1.17.1

### Error Description
When I instantiate a GaussianCopulaSynthesizer, I am able to supply a parameter called `numerical_distributions` that allows me to map each column name to the distribution I want to use for modeling it.

But in some cases, the column isn't actually a statistical column that the SDV models. This ultimately depends on the transformer that is used. For eg. I may have a credit card column that is auto-assigned to the `AnonymizedFaker` transformer; this transformer just drops the column entirely (does not model it), and then recreates random credit card values.

In these cases, the GaussianCopulaSynthesizer silently ignores the numerical distribution I provide for the column. It would be more helpful to **show a warning, whenever we realize that the column won't be modeled** (this may be during fit).

### Steps to reproduce
Notice how even though I add a distribution for `credit_card_number`, it doesn't actually get modeled. The parameter is ignored and I don't see any resulting distribution learned for this column.
```python

    modality='single_table',
)

synthesizer = GaussianCopulaSynthesizer(
    metadata,
    numerical_distributions={'credit_card_number': 'beta'}
)

synthesizer.fit(data)
synthesizer.get_learned_distributions()
```

### Expected Behavior
Give me a warning whenever it becomes clear that the column I supplied a distribution for won't actually be modeled.

```
Warning: Cannot use distribution 'beta' for column 'billing_address' because the 
column is not statistically modeled.
```

Note that in some cases, updating the transformer might make the warning go away because it will tell the SDV to start modeling portions of that column. In this instance, updating to `PseudoAnonymizedFaker` would mean that we would actually learn a distribution for the `credit_card_number` column.
```python
from rdt.transformers.pii import PseudoAnonymizedFaker

transformer = PseudoAnonymizedFaker(
    provider_name='credit_card',
    function_name='credit_card_number', 
    locales=['en_US'])

    'credit_card_number': transformer
})

synthesizer.fit(data)

# now you'll see credit_card_number modeled as a beta
synthesizer.get_learned_distributions() 
```

pvk-developer:
Resolves #2296
CU-86b2w3cnn

### DIFF
--- a/sdv/single_table/copulagan.py
+++ b/sdv/single_table/copulagan.py
@@ -8,8 +8,8 @@ import rdt
 from sdv.single_table.copulas import GaussianCopulaSynthesizer
 from sdv.single_table.ctgan import CTGANSynthesizer
 from sdv.single_table.utils import (
-    log_numerical_distributions_error,
     validate_numerical_distributions,
+    warn_missing_numerical_distributions,
 )
 
 LOGGER = logging.getLogger(__name__)
@@ -204,10 +204,7 @@ class CopulaGANSynthesizer(CTGANSynthesizer):
             processed_data (pandas.DataFrame):
                 Data to be learned.
         """
-        log_numerical_distributions_error(
-            self.numerical_distributions, processed_data.columns, LOGGER
-        )
-
+        warn_missing_numerical_distributions(self.numerical_distributions, processed_data.columns)
         gaussian_normalizer_config = self._create_gaussian_normalizer_config(processed_data)
         self._gaussian_normalizer_hyper_transformer = rdt.HyperTransformer()
         self._gaussian_normalizer_hyper_transformer.set_config(gaussian_normalizer_config)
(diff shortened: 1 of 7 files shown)


sdv-team:
Task linked: [CU-86b2mq8cy Copulas - Update codecov and add flag for integration tests #433](https://app.clickup.com/t/86b2mq8cy)

sdv-team:
Task linked: [CU-86b2w3cnn SDV - When using GaussianCopula, show a warning if a column in `numerical_distributions` cannot be applied #2296](https://app.clickup.com/t/86b2w3cnn)

amontanez24 (approved):

frances-h (approved):

PR #2320, opened 2024-12-14
Fixing tests for rdt 1.13.2

amontanez24:


fealho (approved):
The failing test can be fixed by sorting the order of the rows for both dataframes.

sdv/constraints/tabular.py
```diff
@@ -299,7 +299,7 @@ def _fit(self, table_data):
         self._uuids_to_combinations = {}
         for combination in self._combinations.itertuples(index=False, name=None):
             mappable_combination = get_mappable_combination(combination)
-            uuid_str = str(uuid.uuid4())
+            uuid_str = str(uuid.uuid5(uuid.NAMESPACE_DNS, str(mappable_combination)))
```
amontanez24, line 302: This works but do we have to worry about the fact that `mappable_combination` is not actually a valid DNS? In the [docs for uuid](https://docs.python.org/3/library/uuid.html#uuid.NAMESPACE_DNS), it says
```
When this namespace is specified, the name string is a fully qualified domain name.
```

fealho: The namespace is any uuid and name is any string, which are then concatenated and hashed. [This](url) is a good discussion on the topic.

amontanez24: The link to the discussion is broken, but honestly either one is fine. We just need to make sure that each combination is mapped to a unique value. 

The more I think about it, we only promise that sampling can be reset to get the same values. Since this happens during fitting, it might be ok to just leave as is and sort in the test.

fealho: True, from a user perspective both solutions work. I'd still lean towards getting rid of the randomness, but either way is fine.

PR #2358, opened 2025-01-23
Surface more detailed error info when detecting metadata from dataframes

### LINKED ISSUE #2327: Surface more detailed error info when detecting metadata from dataframes
### Problem Description
This problem was first observed in #2323.

I may have a pandas DataFrame containing data that is not in the form that SDV expects. For example, I may have a column of dtype `object` in which the column values are dictionaries. This is not really supported from SDV right now, as it only models data that is int, float, strings, etc.

In such a case, the `detect_from_dataframes` function crashes with an error such as:

```
TypeError: unhashable type: 'dict'
```

This isn't really helpful in telling me which particular column or table is having an issue.

### Expected behavior
As we detect metadata from a DataFrame, column by column, it would be useful to wrap the detection code in a try/catch statement. That way, if there is an error, we can print out the table/column name (in addition to the original error).

Eg.
```
InvalidDataError: Unable to detect metadata for table 'users' column 'age' due to an invalid data format.

TypeError: unhashable type: 'dict'
```

It would be nice if the original stack trace were also provided.

R-Palazzo:
CU-86b38xdpr
Resolve #2327

### DIFF
--- a/sdv/metadata/multi_table.py
+++ b/sdv/metadata/multi_table.py
@@ -544,7 +544,7 @@ class MultiTableMetadata:
         """
         self._validate_table_not_detected(table_name)
         table = SingleTableMetadata()
-        table._detect_columns(data)
+        table._detect_columns(data, table_name)
         self.tables[table_name] = table
         self._log_detected_table(table)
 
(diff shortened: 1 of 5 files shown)


sdv-team:
Task linked: [CU-86b38xdpr SDV - Surface more detailed error info when detecting metadata from dataframes #2327](https://app.clickup.com/t/86b38xdpr)

frances-h (approved):

PR #2363, opened 2025-02-06
Add `infer_sdtypes` and `infer_keys` parameters to `detect_from_dataframes` method

### LINKED ISSUE #2341: When detecting metadata from dataframes, allow me the option to turn on/off sdtype and relationship detection
### Problem Description
Currently, our [AI Connectors](https://docs.sdv.dev/sdv/reference/explore-sdv/ai-connectors-bundle) offer some customizability when detecting metadata from a database. All of them allow you to toggle on/off the detection for individual sdtypes and for keys (primary/foreign key relationships). For eg. See [this API](https://docs.sdv.dev/sdv/multi-table-data/data-preparation/loading-data/alloydb#create_metadata) (copied below)

> `infer_sdtypes`: A boolean describing whether to infer the sdtypes of each column
>   - (default) True: Infer the sdtypes based on the data
>   - False: Do not infer the sdtypes. All columns will be marked as unknown, ready for you to manually update

> `infer_keys`: A string describing whether to infer the primary and/or foreign keys. Options are:
>   - (default) 'primary_and_foreign': Infer the primary keys in each table, and the foreign keys in other tables that refer to them
>   - 'primary_only': Infer only the primary keys of each table
>   - None: Do not infer any keys

By contrast, when detecting metadata from DataFrames, we do not allow these options of `infer_sdtypes` and `infer_keys`. This functionality would come in useful when a user just wants to auto-detect _part_ of the metadata (eg. maybe because they have other, explicit info they'd like to input themselves). For example, see issue #2338.

### Expected behavior
It would be nice if the [`detect_from_dataframes` function](https://docs.sdv.dev/sdv/multi-table-data/data-preparation/creating-metadata#detect_from_dataframes) also had the same functionality as AI Connectors. 

Essentially, we'd add the following parameters that would behave the exact same way as AI Connectors:
- `infer_sdtypes`
- `infer_keys`

The single table method `detect_from_dataframe` should similarly be updated.

fealho:
CU-86b3f92p5, Resolve #2341.

A few notes about the PR:
* Setting `infer_sdtypes=False` means primary and foreign keys will not be detected. Primary keys can't be detected because all sdtypes are unknown and foreign keys are not detected because there are no primary keys.
* Setting `infer_sdtypes=False` will set all columns to `pii=True`. This is the default behavior for columns with `unknown` sdtype.
* Support for these parameters is only being added to the `Metadata` class methods, not to `MultiTableMetadata` or `SingleTableMetadata`.

### DIFF
--- a/sdv/metadata/multi_table.py
+++ b/sdv/metadata/multi_table.py
@@ -536,13 +536,11 @@ class MultiTableMetadata:
             raise ValueError("'infer_sdtypes' must be a boolean value.")
 
         if infer_keys not in ['primary_only', None]:
-            raise ValueError(
-                "'infer_keys' must be one of: 'primary_only', None."
-            )
+            raise ValueError("'infer_keys' must be one of: 'primary_only', None.")
 
     def detect_table_from_dataframe(
-            self, table_name, data, infer_sdtypes=True, infer_keys='primary_only'
-        ):
+        self, table_name, data, infer_sdtypes=True, infer_keys='primary_only'
+    ):
         """Detect the metadata for a table from a dataframe.
 
         This method automatically detects the ``sdtypes`` for the given ``pandas.DataFrame``,
--- a/sdv/metadata/multi_table.py
+++ b/sdv/metadata/multi_table.py
@@ -530,7 +530,19 @@ class MultiTableMetadata:
                         )
                         continue
 
-    def detect_table_from_dataframe(self, table_name, data):
+    @staticmethod
+    def _validate_infer_sdtypes_and_keys(infer_sdtypes, infer_keys):
+        if not isinstance(infer_sdtypes, bool):
+            raise ValueError("'infer_sdtypes' must be a boolean value.")
+
+        if infer_keys not in ['primary_only', None]:
+            raise ValueError(
+                "'infer_keys' must be one of: 'primary_only', None."
+            )
+
+    def detect_table_from_dataframe(
+            self, table_name, data, infer_sdtypes=True, infer_keys='primary_only'
+        ):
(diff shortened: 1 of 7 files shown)


sdv-team:
Task linked: [CU-86b3f92p5 SDV - When detecting metadata from dataframes, allow me the option to turn on/off sdtype and relationship detection #2341](https://app.clickup.com/t/86b3f92p5)

frances-h (approved):

sdv/metadata/multi_table.py
```diff
@@ -541,10 +543,20 @@ def detect_table_from_dataframe(self, table_name, data):
...
+            infer_sdtypes (bool):
+                A boolean describing whether to infer the sdtypes of each column.
+                If True it infers the sdtypes based on the data.
+                If False it does not infer the sdtypes and all columns are marked as unknown.
+                Defaults to True.
+            infer_keys (str):
+                A string describing whether to infer the primary keys. Options are:
+                    - 'primary_only': Infer only the primary keys of each table
+                    - None: Do not infer any keys
+                Defaults to 'primary_only'.
```
lajohn4747, line 555: For MultiTableMetadata, should we have the `primary_and_foreign` infer key option?

lajohn4747 (commented):
Looks good just had one comment

sdv/metadata/multi_table.py
fealho: @lajohn4747 This is the single table version of the method, `detect_table_from_dataframe`. There are no foreign keys to be detected.

lajohn4747 (approved):

PR #2400, opened 2025-02-27
Add FixedCombinations CAG pattern + add CAG base class to public

### LINKED ISSUE #2383: Add `FixedCombinations` CAG pattern + add CAG base class to public
### Problem Description

#### CAG Base Class
This CAG should inherit from the CAG base class which can be brought over from enterprise (excluding the `fix_dayz_data` method). We should also add the `is_valid` method to the CAG base class:
`is_valid(self, data)`
- Check that the pattern is fitted, else raise a NotFittedError.
- Call the `self._is_valid(data)` method, which should be defined in the base class to raise a `NotImplemented` error

#### Add the `FixedCombinations` CAG pattern. 
This CAG should copy the functionality of the existing constraint, much of `_fit`, `_transform`, and `_reverse_transform` should be able to be directly copied without significant logical rewrites.

**Args**
`column_names (list[str])`: Same as FixedCombinations.
`table_name (str, optional)`: The name of the table to apply the CAG pattern to.

**Expected Functionality**

`_validate_pattern_with_metadata`
- If no `table_name` is provided, the metadata must only contain a single table (this should be considered the target table)
- Validate that all columns in `column_names` exist in the metadata.
- Validate that all columns in `column_names` have either the boolean or categorical sdtype
- Validate that no column in `column_names` are part of a column relationship.

`_validate_pattern_with_data`
- No data validation needed

`_get_updated_metadata`
- Add a joint combination column (by default can be the column names joined with `#`, as in the constraint)
- Drop all `column_names` from the metadata.

`_fit`
- Same logic as the constraint

`_is_valid`
- Same logic as the constraint

`_transform`
- Same logic as the constraint

`_reverse_transform`
- Same logic as the constraint

### Additional context

frances-h:
Resolve #2383
CU-86b424067

### DIFF
--- a/sdv/cag/_utils.py
+++ b/sdv/cag/_utils.py
@@ -1,0 +1,18 @@
+from sdv.cag._errors import PatternNotMetError
+
+
+def _validate_table_and_column_names(table_name, columns, metadata):
+    """Validate the table and column names for the pattern."""
+    if table_name is None and len(metadata.tables) > 1:
+        raise PatternNotMetError(
+            'Metadata contains more than 1 table but no ``table_name`` provided.'
+        )
+    if table_name is None:
+        table_name = metadata._get_single_table_name()
+    elif table_name not in metadata.tables:
+        raise PatternNotMetError(f"Table '{table_name}' missing from metadata.")
+
+    if not set(columns).issubset(set(metadata.tables[table_name].columns)):
+        missing_columns = columns - set(metadata.tables[table_name].columns)
+        missing_columns = "', '".join(sorted(missing_columns))
+        raise PatternNotMetError(f"Table '{table_name}' is missing columns '{missing_columns}'.")
--- a/sdv/cag/base.py
+++ b/sdv/cag/base.py
@@ -1,0 +1,166 @@
+"""Base Multi-Table Pattern."""
+
+import pandas as pd
+
+from sdv.errors import NotFittedError
+
+
+class BasePattern:
+    """Base CAG Pattern Class."""
+
+    def __init__(self):
+        self.metadata = None
+        self._fitted = False
+        self._single_table = False
+
+    def _get_single_table_name(self, metadata):
+        if not hasattr(self, 'table_name'):
+            raise ValueError('No ``table_name`` attribute has been set.')
+
+        return metadata._get_single_table_name() if self.table_name is None else self.table_name
+
+    def _validate_pattern_with_metadata(self, metadata):
+        raise NotImplementedError()
+
+    def _validate_pattern_with_data(self, data, metadata):
+        raise NotImplementedError()
+
+    def validate(self, data=None, metadata=None):
+        """Validate the data/metadata meets the pattern requirements.
+
+        Args:
+            data (dict[str, pd.DataFrame], optional)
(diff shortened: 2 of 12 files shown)


sdv-team:
Task linked: [CU-86b424067 SDV - Add `FixedCombinations` CAG pattern + add CAG base class to public #2383](https://app.clickup.com/t/86b424067)

tests/unit/cag/test_fixed_combinations.py
```diff
@@ -0,0 +1,327 @@
...
+        instance.fit(data, metadata)
+
+        # Run
+        out = instance.transform(data)
+
+        # Assert
+        assert instance._combinations_to_uuids is not None
+        assert instance._uuids_to_combinations is not None
+        expected_out_a = pd.Series(['a', 'b', 'c'], name='a')
+        pd.testing.assert_series_equal(expected_out_a, out['a'])
```
R-Palazzo, line 248: Could we also check the joint column here?

sdv/cag/_utils.py
```diff
@@ -0,0 +1,18 @@
+from sdv.cag._errors import PatternNotMetError
+
+
+def _validate_table_and_column_names(table_name, columns, metadata):
```
R-Palazzo, line 4: There is also a `_validate_table_and_column_names` in `sdv/cag/utils.py`. Do you think we could keep only one method?

sdv/cag/base.py
```diff
@@ -0,0 +1,166 @@
...
+        if metadata is None:
+            if self.metadata is None:
+                raise NotFittedError('Pattern must be fit before validating without metadata.')
+
+            metadata = self.metadata
+
+        self._validate_pattern_with_metadata(metadata)
+        if data is not None:
+            if self._single_table:
+                data = {self._table_name: data}
```
R-Palazzo, line 47: There are `self._table_name` and `self.table_name`, it seems; should they be the same?

R-Palazzo (commented):
This looks good!
Should we update anything in a `pyproject.toml` or in a `package_config.yaml` file?

sdv/cag/base.py
frances-h: Not necessarily, the base class learns `self._table_name` during fit if it's a single table. `self.table_name` stores whatever the actual `table_name` parameter was. If the input `self.table_name` is `None`, then we don't want `self._table_name` to break the pattern on a refit in case the table name has changed.

sdv/cag/fixed_combinations.py
```diff
@@ -0,0 +1,202 @@
...
+            rel
+            for rel in metadata['tables'][table_name].get('column_relationships', [])
+            if set(rel['column_names']).isdisjoint(column_set)
+        ]
+
+        return Metadata.load_from_dict(metadata)
+
+    def _fit(self, data, metadata):
+        """Fit the pattern."""
+        table_name = self._get_single_table_name(metadata)
```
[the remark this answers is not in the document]
frances-h: I thought we could but actually we can't, since in the multi-table case `self._table_name` won't be set

R-Palazzo (approved):

gsheni (approved):

PR #2401, opened 2025-02-28
Make `GaussianCopula` report the correct distribution name in the case of a fallback

### LINKED ISSUE #2394: GaussianCopula is not reporting the correct distribution name in the case of a fallback
### Environment Details

- SDV version: 1.18.0
- Python version: 3.11.11
- Operating System: Linux (Google Colab)

### Error Description
The GaussianCopulaSynthesizer starts by estimating the shape of each individual column (marginal distribution). Sometimes, for reasons outside of our control, this estimation fails to converge for a particular set of data. In such cases, Copulas has a built-in fallback in place: It tries again using `'norm'` (standard normal distribution); this distribution's parameters can be directly computed via a closed-form formula, so it should not fail.

When this happens, I expected that the [get_learned_distributions function](https://docs.sdv.dev/sdv/single-table-data/modeling/synthesizers/gaussiancopulasynthesizer#get_learned_distributions) will correctly report that the distribution has fallen back to `'norm'`. But in practice, I see that **the distribution name is not correctly updating**. This can cause confusion about what the synthesizer has really learned.

### Steps to reproduce
First, fit a synthesizer with some data that is known to fail for the `'beta'` distribution.

```python
import pandas as pd
import numpy as np

data = pd.DataFrame(data={
    'A': np.concatenate([np.zeros(29), np.ones(21)])
})

    'tables': {
        'table': {
            'columns': {
                'A': { 'sdtype': 'numerical'}}}}})

synthesizer = GaussianCopulaSynthesizer(metadata, default_distribution='beta')
synthesizer.fit(data)
```
Now, you can inspect the internals of the model to see that the `'norm'` distribution was used instead (GaussianUnivariate is the name for normal distribution):
```python
>>> synthesizer._model.to_dict()['univariates']
[{'loc': 0.42,
  'scale': 0.49355850717012256,
  'type': 'copulas.univariate.gaussian.GaussianUnivariate'}]
```

However, notice that the `get_learned_distributions()` method continues to report the distribution name as `'beta'` instead of `'norm'` (as expected). **This is a bug.** 
```python
>>> synthesizer.get_learned_distributions()
{'A': {'distribution': 'beta',
  'learned_parameters': {'loc': 0.42, 'scale': 0.49355850717012256}}}
```
(Note that the learned parameters themselves are correct. Just the name of the distribution is wrong. It should be `'norm'`.)

fealho:
CU-86b42ufzb, Resolve #2394.


rwedge (approved):
LGTM

pvk-developer (approved):

PR #2405, opened 2025-03-04
Add Inequality CAG

fealho:
CU-86b42406t, Resolve #2384.

PR #2407, opened 2025-03-06
Add Range CAG

### LINKED ISSUE #2385: Add `Range` CAG pattern
### Problem Description

### Expected behavior
Add the `Range` CAG pattern. This CAG should inherit from the base CAG class. This CAG should copy the functionality of the existing constraint, much of `_fit`, `_transform`, and `_reverse_transform` should be able to be directly copied without significant logical rewrites.

**Args**
`low_column_name (str)`: Same as `Range` constraint
`middle_column_name (str)`: Same as the `Range` constraint
`high_column_name (str)`: Same as `Range` constraint
`strict_boundaries (str)`: Same as `Range` constraint
`table_name (str, optional)`: The name of the table to apply the CAG pattern to.

**Expected Functionality**

`_validate_pattern_with_metadata`
- If no `table_name` is provided, the metadata must only contain a single table (this should be considered the target table)
- Validate that all input columns exist in the table in the metadata.
- Validate that all columns have the same sdtype, and that it is either `numerical` or `datetime`

`_validate_pattern_with_data`
- Validate that the data meets the Range pattern

`_get_updated_metadata`
- Add the low diff column to the table (by default can be `'{low_column_name}#{middle_column_name}'`)
- Add the high diff column to the table (by default should be `'{middle_column_name}#{high_column_name}'`
- Drop the `middle_column_name` and `high_column_name` columns from the metadata.

`_fit`
- Same logic as the constraint

`_is_valid`
- Same logic as the constraint

`_transform`
- Same logic as the constraint

`_reverse_transform`
- Same logic as the constraint

### Additional context

fealho:
CU-86b424066, Resolve #2385.

### DIFF
--- a/sdv/cag/range.py
+++ b/sdv/cag/range.py
@@ -1,0 +1,374 @@
+"""Inequality CAG pattern."""
+
+import numpy as np
+import pandas as pd
+
+from sdv._utils import _convert_to_timedelta, _create_unique_name
+from sdv.cag._errors import PatternNotMetError
+from sdv.cag._utils import _validate_table_and_column_names
+from sdv.cag.base import BasePattern
+from sdv.constraints.utils import (
+    cast_to_datetime64,
+    compute_nans_column,
+    get_datetime_diff,
+    revert_nans_columns,
+)
+from sdv.metadata import Metadata
+
+
+class Range(BasePattern):
+    """Ensure that the `middle_column_name` is between `low_column_name` and `high_column_name`.
+
+    The transformation strategy works the same as the Inequality constraint but with two
+    columns instead of one. We compute the difference between the `middle_column_name`
+    and the `low_column_name` column and then apply a logarithm to the difference + 1 to ensure
+    that the value stays positive when reverted afterwards using an exponential.
+    We do the same for the difference between the `high_column_name` and `middle_column_name`.
+
+    Args:
+        low_column_name (str):
+            Name of the column which will be the lower bound.
+        middle_column_name (str):
+            Name of the column that has to be between the lower bound and upper bound.
+        high_column_name (str):
+            Name of the column which will be the higher bound.
+        strict_boundaries (bool):
+            Whether the comparison of the values should be strict `>=` or
+            not `>` when comparing them.
+            Defaults to True.
+        table_name (str, optional):
+            The name of the table that contains the columns. Optional if the
+            data is only a single table. Defaults to None.
+    """
+
+    @staticmethod
+    def _validate_init_inputs(
+        low_column_name,
+        middle_column_name,
+        high_column_name,
+        strict_boundaries,
+        table_name,
+    ):
+        if not (
+            isinstance(low_column_name, str)
+            and isinstance(middle_column_name, str)
+            and isinstance(high_column_name, str)
+        ):
+            raise ValueError(
+                '`low_column_name`, `middle_column_name` and `high_column_name` must be strings.'
+            )
+
+        if not isinstance(strict_boundaries, bool):
+            raise ValueError('`strict_boundaries` must be a boolean.')
+
+        if table_name and not isinstance(table_name, str):
+            raise ValueError('`table_name` must be a string or None.')
+
+    def __init__(
+        self,
+        low_column_name,
+        middle_column_name,
+        high_column_name,
+        strict_boundaries=False,
+        table_name=None,
+    ):
+        super().__init__()
+        self._validate_init_inputs(
+            low_column_name,
+            middle_column_name,
+            high_column_name,
+            strict_boundaries,
+            table_name,
+        )
+        self._low_column_name = low_column_name
+        self._middle_column_name = middle_column_name
+        self._high_column_name = high_column_name
+        self._low_diff_column_name = f'{self._low_column_name}#{self._middle_column_name}'
+        self._high_diff_column_name = f'{self._middle_column_name}#{self._high_column_name}'
+        self._operator = np.greater if strict_boundaries else np.greater_equal
+        self.table_name = table_name
+
+        # Set during fit
+        self._is_datetime = None
+        self._dtype = None
+        self._low_datetime_format = None
+        self._middle_datetime_format = None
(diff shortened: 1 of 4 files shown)


sdv-team:
Task linked: [CU-86b424066 SDV - Add `Range` CAG pattern #2385](https://app.clickup.com/t/86b424066)

sdv/cag/range.py
```diff
@@ -0,0 +1,376 @@
...
+    We do the same for the difference between the `high_column_name` and `middle_column_name`.
+
+    Args:
+        low_column_name (str):
+            Name of the column which will be the lower bound.
+        middle_column_name (str):
+            Name of the column that has to be between the lower bound and upper bound.
+        high_column_name (str):
+            Name of the column which will be the higher bound.
+        strict_boundaries (bool):
```
gsheni, line 37:
```suggestion
        strict_boundaries (bool, optional):
```

sdv/cag/range.py
```diff
@@ -0,0 +1,376 @@
...
+        )
+        high_diff_column = _create_unique_name(
+            self._high_diff_column_name, metadata.tables[table_name].columns.keys()
+        )
+
+        metadata = metadata.to_dict()
+        metadata['tables'][table_name]['columns'][low_diff_column] = {'sdtype': 'numerical'}
+        metadata['tables'][table_name]['columns'][high_diff_column] = {'sdtype': 'numerical'}
+        del metadata['tables'][table_name]['columns'][self._high_column_name]
+        del metadata['tables'][table_name]['columns'][self._middle_column_name]
```
gsheni, line 189: The issue says to drop middle_column_name and high_column_name, but I wonder why we would not drop low_column_name

sdv/cag/range.py
```diff
@@ -0,0 +1,376 @@
...
+        low_is_nan = pd.isna(low)
+        mid_is_nan = pd.isna(mid)
+        high_is_nan = pd.isna(high)
+
+        low_lt_middle = self._operator(low, mid) | low_is_nan | mid_is_nan
+        mid_lt_high = self._operator(mid, high) | mid_is_nan | high_is_nan
+        low_lt_high = self._operator(low, high) | low_is_nan | high_is_nan
+
+        valid = low_lt_middle & mid_lt_high & low_lt_high
+        is_valid[table_name] = valid
```
gsheni, line 374: Given the common code here and in `_validate_pattern_with_data`, can we create a function?

tests/integration/cag/test_range.py
```diff
@@ -0,0 +1,290 @@
...
+        diff = (data[col] - reverse_transformed[col]).abs()
+        assert (diff.dt.total_seconds() < 1e-6).all()
+
+
+def test_range_pattern_datetime_nans():
+    """Test that Range pattern works with datetime columns with NaNs."""
+    # Setup
+    data = pd.DataFrame({
+        'A': [
+            np.nan,
```
gsheni, line 167: For datetimes columns, I believe there are 2 possibilities. A `datetime64[ns]` dtype (numpy) and an `object` dtype. 
We should also test `object` dtype: 
```python
pd.Series([pd.Timestamp('2024-01-02'), np.nan], dtype="object")
pd.Series([pd.Timestamp('2024-01-02'), pd.NA], dtype="object")
```

sdv/cag/range.py
fealho: Both `Range` and `Inequality` constraint/cag create a `diff` column during transform. When reversing it adds the `diff` column to the `low` column to get back the middle/high columns. Both constraints need the `low` column for this process, so it's not removed from the data/metadata.

tests/integration/cag/test_range.py
fealho: The `validate` method crashes if the datetimes are passed as strings, since there is no logic to convert them to numerical values.

As I understand it, the goal of this issue is to convert the existing constraint into a CAG. Any additional changes should probably be done in a separate PR. We had a similar [discussion](https://github.com/sdv-dev/SDV/pull/2405#discussion_r1983648731) on the `Inequality` constraint.

It is a good point though @gsheni, I agree this should be addressed.

gsheni: Gotcha. Let's maintain existing functionality then

gsheni (approved):

R-Palazzo (commented):
Looks good!

sdv/cag/range.py
```diff
@@ -0,0 +1,365 @@
...
+
+        # Set during fit
+        self._is_datetime = None
+        self._dtype = None
+        self._low_datetime_format = None
+        self._middle_datetime_format = None
+        self._high_datetime_format = None
+
+        # Set during transform
+        self._nan_column_name = None
```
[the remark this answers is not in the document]
fealho: https://github.com/sdv-dev/SDV/pull/2405#discussion_r1987758358

sdv/cag/range.py
```diff
@@ -0,0 +1,365 @@
...
+            middle_column_name,
+            high_column_name,
+            strict_boundaries,
+            table_name,
+        )
+        self._low_column_name = low_column_name
+        self._middle_column_name = middle_column_name
+        self._high_column_name = high_column_name
+        self._low_diff_column_name = f'{self._low_column_name}#{self._middle_column_name}'
+        self._high_diff_column_name = f'{self._middle_column_name}#{self._high_column_name}'
```
amontanez24, line 89: do we need to check these names for collisions? If we don't do that for the constraint, we don't need to here either

amontanez24 (approved):
One small comment but this LGTM!

sdv/cag/range.py
fealho: These variables are used with the `_create_unique_name` method, like [here](https://github.com/sdv-dev/SDV/pull/2407/files#diff-03e714e20c666f9d2f2ffd5a5ba128b443fe144e1037ccb0aaf2d492925c6f34R183), so there are no collisions.

R-Palazzo (approved):

PR #2417, opened 2025-03-11
Add CAG support to single table synthesizers

### LINKED ISSUE #2389: Add CAG support to single table synthesizers
### Problem Description

### Expected behavior
Add CAG support to all single table synthesizers. This will closely resemble the integration with multi-table synthesizers.

**API**
`add_cag(patterns)`
-  Add the patterns to the synthesizer, calculating the updated metadata 

`get_cag()`
- Returns the list of CAGs applied to the synthesizer

`get_metadata(version='original')`
- Returns the metadata for the synthesizer. If `version` is `'original'`, returns the original metadata. If `version` is `'modified'`, returns the modified metadata after running through all applied CAG patterns.

**Functionality**
- CAGs should be applied as a primary pre-processing step, and should modify the expected metadata for the rest of the downstream synthesizer processes
- Overlapping CAGs
  - If a CAG cannot be applied to the end of the CAG chain due to invalid metadata, but CAN be applied to the original input metadata, it should be fit and saved separately and used exclusively for reject sampling 
- Reject sampling
  - To support reject sampling, the data should be run through the `is_valid` method for all CAGs and invalid rows should be dropped. 
  - The sampled data should first be filtered through the `is_valid` methods for any CAGs that were held out for reject sampling
  - The remaining sampled data should be passed through each CAG. For each CAG: check if the data is valid, drop invalid rows, transform any remaining rows, then pass the transformed to the next CAG in line
  - Try to resample the number of dropped rows, up to a given number of retries (by default, use the max number of retries we currently use for constraints)

### Additional context

fealho:
CU-86b4244n9, Resolve #2389.


sdv-team:
Task linked: [CU-86b4244n9 SDV - Add CAG support to single table synthesizers #2389](https://app.clickup.com/t/86b4244n9)

PR #2420, opened 2025-03-13
Autodetection of object types should default to categorical not unknown

### LINKED ISSUE #2413: When in doubt, metadata auto-detection should mark columns as sdtype `categorical`

### Problem Description
Currently the metadata auto-detection feature ([`detect_from_dataframe`](https://docs.sdv.dev/sdv/single-table-data/data-preparation/creating-metadata#detect_from_dataframe) and [`detect_from_dataframes`](https://docs.sdv.dev/sdv/multi-table-data/data-preparation/creating-metadata#detect_from_dataframes)) will try to detect the sdtypes for each column. One of the sdtypes it's trying to detect is `categorical`.

Currently, the `categorical` sdtype is determined based on the _cardinality_ (# of unique values) that appear in the column. If this threshold is low enough, it detects the sdtype as `categorical` but if it's above the threshold, it might be detected as a different sdtype -- possibilities here are `numerical` (if the values are numbers) or `unknown` (if the values are objects/strings). 

The problem is that **many high-cardinality categorical columns are being detected a `unknown` right now** due to these rules. The consequences of this can be stark: SDV is designed to treat `unknown` values as sensitive, PII values. So users may see completely unexpected results.

### Expected behavior
For high cardinality columns that are represented as objects/strings, detect the sdtype as `categorical` instead of `unknown`. [This line of code](https://github.com/sdv-dev/SDV/blob/main/sdv/metadata/single_table.py#L510) is probably the one to fix -- though there could be other instances of this happening too.

As a result of this change, should be no column auto-detected as `unknown`.

lajohn4747:
resolves #2413 
CU-86b472jdd


fealho (approved):
This may affect the results of the Metadata Detection Report Benchmark in SDV-Enterprise, maybe run that report with this change to make sure the quality stays the same.

frances-h (approved):

lajohn4747:
> This may affect the results of the Metadata Detection Report Benchmark in SDV-Enterprise, maybe run that report with this change to make sure the quality stays the same.

Confirmed!

PR #2422, opened 2025-03-14
Store metadata as `Metadata` for `BaseSynthesizer`

fealho:
CU-86b48qdxh.

### DIFF
--- a/sdv/metadata/metadata.py
+++ b/sdv/metadata/metadata.py
@@ -153,20 +153,6 @@ class Metadata(MultiTableMetadata):
         metadata.detect_table_from_dataframe(table_name, data, infer_sdtypes, infer_keys)
         return metadata
 
-    def validate_data(self, data):
-        """Validate the data matches the metadata.
-
-        Args:
-            data (dict):
-                A dictionary of table names to pd.DataFrames.
-        """
-        if isinstance(data, pd.DataFrame):
-            table_name = self._get_single_table_name()
-            table_metadata = self.tables.get(table_name, SingleTableMetadata())
-            table_metadata.validate_data(data)
-        else:
-            super().validate_data(data)
-
     def _set_metadata_dict(self, metadata, single_table_name=None):
         """Set a ``metadata`` dictionary to the current instance.
 
--- a/sdv/metadata/metadata.py
+++ b/sdv/metadata/metadata.py
@@ -153,6 +153,20 @@ class Metadata(MultiTableMetadata):
         metadata.detect_table_from_dataframe(table_name, data, infer_sdtypes, infer_keys)
         return metadata
 
+    def validate_data(self, data):
+        """Validate the data matches the metadata.
+
+        Args:
+            data (dict):
+                A dictionary of table names to pd.DataFrames.
+        """
+        if isinstance(data, pd.DataFrame):
+            table_name = self._get_single_table_name()
+            table_metadata = self.tables.get(table_name, SingleTableMetadata())
+            table_metadata.validate_data(data)
+        else:
+            super().validate_data(data)
+
     def _set_metadata_dict(self, metadata, single_table_name=None):
         """Set a ``metadata`` dictionary to the current instance.
 
--- a/sdv/metadata/multi_table.py
+++ b/sdv/metadata/multi_table.py
@@ -776,9 +776,11 @@ class MultiTableMetadata:
                     self.tables[table_name].validate_data(table_data, table_sdtype_warnings)
 
             except InvalidDataError as error:
-                error_msg = f"Table: '{table_name}'"
-                for _error in error.errors:
-                    error_msg += f'\nError: {_error}'
+                error_msg = '' if len(self.tables) == 1 else f"Table: '{table_name}'\n"
+                for i, _error in enumerate(error.errors):
+                    if i > 0:
+                        error_msg += '\n'
+                    error_msg += f'Error: {_error}'
 
                 errors.append(error_msg)
 
--- a/sdv/metadata/multi_table.py
+++ b/sdv/metadata/multi_table.py
@@ -776,11 +776,9 @@ class MultiTableMetadata:
                     self.tables[table_name].validate_data(table_data, table_sdtype_warnings)
 
             except InvalidDataError as error:
-                error_msg = '' if len(self.tables) == 1 else f"Table: '{table_name}'\n"
-                for i, _error in enumerate(error.errors):
-                    if i > 0:
-                        error_msg += '\n'
-                    error_msg += f'Error: {_error}'
+                error_msg = f'Errors in {table_name}:'
+                for _error in error.errors:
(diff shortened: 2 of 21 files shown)


sdv-team:
Task linked: [CU-86b48qdxh Store metadata as `Metadata` for `BaseSynthesizer`](https://app.clickup.com/t/86b48qdxh)

sdv/metadata/metadata.py
```diff
@@ -153,6 +153,20 @@ def detect_from_dataframe(
         metadata.detect_table_from_dataframe(table_name, data, infer_sdtypes, infer_keys)
         return metadata
 
+    def validate_data(self, data):
```
amontanez24, line 156: why are we adding this?

sdv/single_table/base.py
```diff
@@ -110,22 +110,22 @@ def __init__(
...
         if isinstance(metadata, Metadata):
             self._table_name = metadata._get_single_table_name()
-            self.metadata = metadata._convert_to_single_table()
-        elif isinstance(metadata, SingleTableMetadata):
+            self.metadata = metadata
+        else:
             warnings.warn(DEPRECATION_MSG, FutureWarning)
+            self._table_name = Metadata.DEFAULT_SINGLE_TABLE_NAME
+            self.metadata = Metadata.load_from_dict(metadata.to_dict(), self._table_name)
+            self.metadata.tables[self._table_name]._updated = metadata._updated
```
amontanez24, line 120: why do we need to set the updated flag here?

sdv/metadata/metadata.py
fealho: `MultiTableMetadata` requires `data` to be a [dictionary](https://github.com/sdv-dev/SDV/blob/391fd824fad082a97207705c522a61b7ae0ca68a/sdv/metadata/multi_table.py#L864-L865). Without this change single table synthesizers don't support taking data as a dataframe.

sdv/single_table/base.py
fealho: Loading from the dict in line 119 doesn't set the `_updated` attribute, we need to do it manually.

amontanez24: but how was it updated? It just loaded from a dict right? Did any attributes change

sdv/metadata/metadata.py
amontanez24: Can we have the single table synthesizers handle this by passing in a dictionary instead?

sdv/single_table/base.py
fealho: Updating a single table metadata may set `_updated = True`, like [here](https://github.com/sdv-dev/SDV/blob/391fd824fad082a97207705c522a61b7ae0ca68a/sdv/metadata/single_table.py#L294). One example is [this](https://github.com/sdv-dev/SDV/blob/391fd824fad082a97207705c522a61b7ae0ca68a/tests/integration/single_table/test_base.py#L719) test where the metadata is updated before initializing the synthesizer.

sdv/sampling/independent_sampler.py
```diff
@@ -98,16 +100,18 @@ def _finalize(self, sampled_data):
         final_data = {}
         for table_name, table_rows in sampled_data.items():
             synthesizer = self._table_synthesizers.get(table_name)
-            metadata = synthesizer.get_metadata()._convert_to_single_table()
+            if isinstance(synthesizer.metadata, Metadata):
+                metadata = synthesizer.metadata.tables[synthesizer._table_name]
+            else:
+                metadata = synthesizer.metadata
             dtypes = synthesizer._data_processor._dtypes
```
frances-h, line 107: Is this for Enterprise compatibility? Can we add an issue for addressing this change in Enterprise as well?

fealho: Yes. What do you mean by addressing the change in Enterprise?

sdv/metadata/multi_table.py
```diff
@@ -776,9 +776,11 @@ def _validate_all_tables(self, data):
                     self.tables[table_name].validate_data(table_data, table_sdtype_warnings)
 
             except InvalidDataError as error:
-                error_msg = f"Table: '{table_name}'"
-                for _error in error.errors:
-                    error_msg += f'\nError: {_error}'
+                error_msg = '' if len(self.tables) == 1 else f"Table: '{table_name}'\n"
```
amontanez24, line 779: From what I remember in the eng meeting, I think we said to change this to something that is fine in both cases. For example
```
Errors in {table_name}:
```
If the above happens to be the default single table case, it will read
```
Errors in table:
```
Which is fine.

sdv/sampling/independent_sampler.py
```diff
@@ -98,16 +100,18 @@ def _finalize(self, sampled_data):
         final_data = {}
         for table_name, table_rows in sampled_data.items():
             synthesizer = self._table_synthesizers.get(table_name)
-            metadata = synthesizer.get_metadata()._convert_to_single_table()
+            if isinstance(synthesizer.metadata, Metadata):
+                metadata = synthesizer.metadata.tables[synthesizer._table_name]
+            else:
+                metadata = synthesizer.metadata
```
amontanez24, line 106: what's wrong with keeping the convert method here?

fealho: Backwards compatibility. When loading a synthesizer from a previous version, the stored metadata will be `SingleTableMetadata`. Because we updated the `get_metadata` method in this PR to simply return the stored metadata, `synthesizer.get_metadata()` will already be a `SingleTableMetadata` object.

sdv/sampling/independent_sampler.py
frances-h: Just a mirror issue to track making any necessary changes to enterprise for compatibility, basically the changes you made in the [draft PR](https://github.com/datacebo/SDV-Enterprise/pull/1130)

frances-h (approved):

sdv/sampling/independent_sampler.py
amontanez24: Ah ok. I think it would be a little less mysterious to put that in the `get_metadata` method instead of here. So in that method we can convert it to `Metadata` if it's an old one.

amontanez24 (approved):

sdv/single_table/base.py
```diff
@@ -99,8 +99,8 @@ def _update_default_transformers(self):
                 self._data_processor._update_transformers_by_sdtypes(sdtype, transformer)
 
     def _check_metadata_updated(self):
-        if self.metadata._updated:
-            self.metadata._updated = False
+        if self.metadata._check_updated_flag():
```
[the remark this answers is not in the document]
amontanez24: Can you file this as a bug with code to reproduce?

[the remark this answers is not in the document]
fealho: @amontanez24 Filed [here](https://github.com/sdv-dev/SDV/issues/2444).

PR #2425, opened 2025-03-14
Feature Branch for PII and ID changes

### LINKED ISSUE #2412: Metadata auto-detection should find `id` columns that are not primary/foreign keys

### Problem Description
Currently the metadata auto-detection feature ([`detect_from_dataframe`](https://docs.sdv.dev/sdv/single-table-data/data-preparation/creating-metadata#detect_from_dataframe) and [`detect_from_dataframes`](https://docs.sdv.dev/sdv/multi-table-data/data-preparation/creating-metadata#detect_from_dataframes)) will try to detect the sdtypes for each column. One of the possible sdtypes is `id`, meaning that the column is an identifier for an entity.

An `id` column need not be a primary or a foreign key (but it can be!). For example in the table below, the `transaction_id` and `product_id` are both `id` columns; however only the transaction is a primary key. The `product_id` is not a primary key (it repeats), nor is it a foreign key (because there is no separate `products` table that is available). It just a column.

| transaction_id | date        | amount | product_id |
|----------------|-------------|--------|------------|
| t_k39jcd       | Mar 9, 2025 | 23.55  | p_39cl1    |
| t_0391ls       | Mar 8, 2025 | 100.90 | p_0ckql    |
| t_93mcp        | Mar 8, 2025 | 101.35 | p_0ckql    |
| ...        | ... | ... | ...   |

Currently, the **metadata auto-detection feature only identifies id columns if they are primary or foreign keys**. This would become a problem for the `product_id` column in the table above.

### Expected behavior
The metadata auto-detection feature should be able to identify columns of sdtype `id` even if they are not primary or foreign keys. We can accomplish this by looking at the column name (just like we do for PII columns): If the column name has a token `'id'` in it, then we should mark it as sdtype id.

- Just like we do for PII columns, the substring should be tokenized before looking for a match. This means that it should only count as a match in cases like `'vendor-id'` or `'Product Id'`. It should not count as a match for strings like `'paid'` or `'uuids'`.
- The token we're looking for (`'id'`) should be case-insensitive. So it should match against `'ID'`, `'Id'`, `'id'`, etc.
- The primary and foreign key relationship detection should happen separately. When the relationships are detected, they should not affect the sdtype for any other column _not_ in that relationship. Eg. Detecting `transaction_id` as a primary key should not affect the detected sdtype of any other column like `product_id`.

### LINKED ISSUE #2413: When in doubt, metadata auto-detection should mark columns as sdtype `categorical`

### LINKED ISSUE #2416: Update transformer assignment for `id` columns

### Problem Description
Many users have now given us feedback that the synthesized data can contain random values that start with `sdv-id` for example `sdv-id-392fl`.  This is happening because the SDV doesn't currently know how to handle columns of sdtype `id` if they are not keys (primary key, foreign key, sequence key, alternate key, etc.). So it just generally creates random values for them.

To better align with user expectations, we should update the transformers that are automatically assigned to such _non-key ID columns_.

### Expected behavior
For _non-key ID columns only_, update the auto-assigned RDT transformers. The new assignment should only be based on whether or not a `regex_format` is present for that ID column in the metadata.

**No regex format present**: If there is no `regex_format` provided in the metadata, we will recycle the ID values that are present in the real data.
- Assign the `UniformEncoder` to this column.
- _Note that `UniformEncoder` is a categorical RDT. [RDT #953](https://github.com/sdv-dev/RDT/issues/953) should be completed and will allow us to use this categorical transformer for ID columns._

**Regex format is present**: If there is a `regex_format` present, then we should create new ID values ~at a realistic scale.~ 
- Assign the `RegexGenerator` with `cardinality_rule='unique'` to this column. 
- _Note that this `'cardinality_rule'` parameter is not currently available in RDT. [RDT #949](https://github.com/sdv-dev/RDT/issues/949) should be completed and will allow us to assign this transformer._

### Additional context
Below is the full list of blocking issues for this one:
- [RDT #953](https://github.com/sdv-dev/RDT/issues/953)
- [RDT #949](https://github.com/sdv-dev/RDT/issues/949)

### LINKED ISSUE #2456: When auto-detecting metadata, add a parameter to control the foreign key detection algorithm
### Problem Description
For multi-table datasets, the metadata auto-detection is done using the [`detect_from_dataframes` method](https://docs.sdv.dev/sdv/multi-table-data/data-preparation/creating-metadata#detect_from_dataframes).

When this method infers the foreign keys, there are different algorithms that are possible. In SDV Community, we offer a simple inference algorithm while SDV Enterprise users have additional options. This difference should be made explicit within the API.

### Expected behavior
Add a parameter called `foreign_key_inference_algorithm` to the multi table auto-detection method. For SDV Community users, there is only one option available: `'column_name_match'`, which matches the foreign key columns based on the name of the column. This should be the default.

```python
metadata = Metadata.detect_from_dataframes(
    data=my_dataframes,
    foreign_key_inference_algorithm='column_name_match'
)
```

(For SDV Enterprise users, there will be additional options and a different default.)

### Additional context
If `infer_keys='primary_only'` or `infer_keys=None`, then the foreign key inference algorithm isn't used at all. This should be self-evident given the name of this parameter.

lajohn4747:
contains #2420 

resolves #2413 
resolves #2412 
resolves #2416 
resolves #2456

### DIFF
--- a/.github/workflows/dtypes.yml
+++ b/.github/workflows/dtypes.yml
@@ -1,9 +1,9 @@
 name: Dtypes Tests
 
 on:
-  push:
-  pull_request:
-    types: [opened, reopened]
+  workflow_dispatch:
+  schedule:
+    - cron: '0 3 * * 1'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/sdv/metadata/metadata.py
+++ b/sdv/metadata/metadata.py
@@ -66,8 +66,47 @@ class Metadata(MultiTableMetadata):
         if not isinstance(infer_sdtypes, bool):
             raise ValueError("'infer_sdtypes' must be a boolean value.")
 
+    @staticmethod
+    def _validate_foreign_key_inference_algorithm(foreign_key_inference_algorithm):
+        if foreign_key_inference_algorithm != 'column_name_match':
+            raise ValueError("'foreign_key_inference_algorithm' must be 'column_name_match'")
+
     @classmethod
-    def detect_from_dataframes(cls, data, infer_sdtypes=True, infer_keys='primary_and_foreign'):
+    def _detect_from_dataframes(
+        cls,
+        data,
+        infer_sdtypes=True,
+        infer_keys='primary_and_foreign',
+        foreign_key_inference_algorithm='column_name_match',
+    ):
+        if not data or not all(isinstance(df, pd.DataFrame) for df in data.values()):
+            raise ValueError('The provided dictionary must contain only pandas DataFrame objects.')
+        if infer_keys not in ['primary_and_foreign', 'primary_only', None]:
+            raise ValueError(
+                "'infer_keys' must be one of: 'primary_and_foreign', 'primary_only', None."
+            )
+        cls._validate_foreign_key_inference_algorithm(foreign_key_inference_algorithm)
+        cls._validate_infer_sdtypes(infer_sdtypes)
+
+        metadata = Metadata()
+        for table_name, dataframe in data.items():
+            metadata.detect_table_from_dataframe(
+                table_name, dataframe, infer_sdtypes, None if infer_keys is None else 'primary_only'
+            )
+
+        if infer_keys == 'primary_and_foreign':
+            metadata._detect_relationships(data, foreign_key_inference_algorithm)
+
+        return metadata
+
+    @classmethod
+    def detect_from_dataframes(
+        cls,
+        data,
+        infer_sdtypes=True,
+        infer_keys='primary_and_foreign',
+        foreign_key_inference_algorithm='column_name_match',
+    ):
         """Detect the metadata for all tables in a dictionary of dataframes.
 
         This method automatically detects the ``sdtypes`` for the given ``pandas.DataFrames``.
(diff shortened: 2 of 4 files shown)


pyproject.toml
```diff
@@ -38,7 +38,7 @@ dependencies = [
     'copulas>=0.12.1',
     'ctgan>=0.11.0',
     'deepecho>=0.7.0',
-    'rdt>=1.14.0',
+    'rdt @ git+https://github.com/sdv-dev/rdt.git@feature/allow-id-categorical-transformers',
```
amontanez24, line 41:
```suggestion
    'rdt>=1.16.0.dev0',
```

pyproject.toml
```diff
@@ -38,7 +38,7 @@ dependencies = [
     'copulas>=0.12.1',
     'ctgan>=0.11.0',
     'deepecho>=0.7.0',
-    'rdt>=1.14.0',
+    'rdt>=1.16.0.dev0',
```
amontanez24, line 41:
```suggestion
    'rdt>=1.16.0',
```

amontanez24 (approved):

frances-h (approved):

PR #2428, opened 2025-03-19
Update transformer assignment for `id` columns

### LINKED ISSUE #2416: Update transformer assignment for `id` columns

frances-h:
Resolve #2416

### DIFF
--- a/.github/workflows/dtypes.yml
+++ b/.github/workflows/dtypes.yml
@@ -1,4 +1,4 @@
-name: Integration Tests
+name: Dtypes Tests
 
 on:
   push:
@@ -10,7 +10,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  integration:
+  dtypes:
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: true
@@ -32,18 +32,9 @@ jobs:
       run: |
           python -m pip install --upgrade pip
           python -m pip install invoke .[test]
-    - name: Run integration tests
+    - name: Run dtypes benchmark tests
       env:
         PYDRIVE_CREDENTIALS: ${{ secrets.PYDRIVE_CREDENTIALS }}
 
       run: |
-        invoke integration
-
-    - if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.13
-      name: Upload integration codecov report
-      uses: codecov/codecov-action@v4
-      with:
-        flags: integration
-        file: ${{ github.workspace }}/integration_cov.xml
-        fail_ci_if_error: true
-        token: ${{ secrets.CODECOV_TOKEN }}
+        invoke benchmark-dtypes
--- a/sdv/data_processing/data_processor.py
+++ b/sdv/data_processing/data_processor.py
@@ -488,7 +488,7 @@ class DataProcessor:
         regex_format = column_metadata.get('regex_format', default_regex_format)
         transformer = rdt.transformers.RegexGenerator(
             regex_format=regex_format,
-            enforce_uniqueness=(column_name in self._keys),
+            cardinality_rule='unique',
             generation_order='scrambled',
         )
 
@@ -572,8 +572,7 @@ class DataProcessor:
                         column, sdtype, column_metadata, is_numeric
                     )
                     sdtypes[column] = 'text'
-
-                else:
+                elif column in self._keys or column_metadata.get('pii'):
                     if is_numeric:
                         function_name = 'random_int'
                         column_dtype = str(column_dtype).lower()
@@ -598,6 +597,9 @@ class DataProcessor:
                     )
 
                     sdtypes[column] = 'pii' if column_metadata.get('pii') else 'text'
+                else:
+                    transformers[column] = rdt.transformers.categorical.UniformEncoder()
+                    sdtypes[column] = 'id'
 
             elif sdtype == 'unknown':
                 sdtypes[column] = 'pii'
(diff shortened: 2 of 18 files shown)


sdv/data_processing/data_processor.py
```diff
@@ -572,8 +572,7 @@ def _create_config(self, data, columns_created_by_constraints):
                         column, sdtype, column_metadata, is_numeric
                     )
                     sdtypes[column] = 'text'
-
-                else:
+                elif column in self._keys or column_metadata.get('pii'):
```
amontanez24, line 575: I don't think the metadata should ever have the 'pii' field if the sdtype is an id. The accepted args for id are
https://github.com/sdv-dev/SDV/blob/6d7d67e43d2d6e56a6e5df686573ec0cd479bf52/sdv/metadata/single_table.py#L43-L50
we raise an error otherwise
https://github.com/sdv-dev/SDV/blob/6d7d67e43d2d6e56a6e5df686573ec0cd479bf52/sdv/metadata/single_table.py#L224

tests/integration/sequential/test_par.py
```diff
@@ -466,3 +466,35 @@ def test_par_categorical_column_represented_by_floats():
     # Assert
     synth.validate(sampled)
     assert sampled['category'].isin(data['category']).all()
+
+
+def test_par_multiple_sequence_keys():
+    """Test the ``PARSynthesizer`` end to end."""
```
amontanez24, line 472: can we add to the description here? I'm curious what the test is going for

sdv/sequential/par.py
```diff
@@ -70,19 +70,20 @@ class PARSynthesizer(LossValuesMixin, BaseSynthesizer):
     _model_sdtype_transformers = {'categorical': None, 'numerical': None, 'boolean': None}
 
     def _get_context_metadata(self):
+        context_metadata_dict = {}
         context_columns_dict = {}
         context_columns = self.context_columns.copy() if self.context_columns else []
         if self._sequence_key:
             context_columns += self._sequence_key
+            context_metadata_dict['primary_key'] = self._sequence_key[0]
```
amontanez24, line 78: The [docs](https://docs.sdv.dev/sdv/reference/metadata-spec/metadata-api#set_sequence_key) make it seem like we only support one sequence key

frances-h: Yeah, PAR stores the sequence key in a list as a relic from when we used to support composite sequence keys. I think we decided to just disallow more than one sequence key and didn't bother refactoring in case we wanted to allow more than one sequence key again in the future. 

Since we only allow one sequence key though, we know that if there's a sequence key it'll just be the first (and only) item in the list.

.github/workflows/dtypes.yml
R-Palazzo, line 1: This workflow is expected to fail currently?

R-Palazzo (approved):

.github/workflows/dtypes.yml
frances-h: Yep, basically it just runs the dtype report and checks the results against the previously saved version. Dtypes that were passing before because they were just being assigned `RegexGenerator` are failing now that they're being assigned `UniformEncoder`. Notably, the dtypes are failing are ones that are already failing for `UniformEncoder`.

Since we're in the process of overhauling the dtype report anyways, it doesn't make sense to fix it here. We can always re-run the dtype benchmark with these changes to update what the latest saved version of the report is so that these pass again.

sdv/multi_table/hma.py
```diff
@@ -52,8 +52,14 @@ def _get_num_data_columns(metadata):
         """
         columns_per_table = {}
         for table_name, table in metadata.tables.items():
+            key_columns = metadata._get_all_keys(table_name)
             columns_per_table[table_name] = sum([
-                1 for col in table.columns.values() if col['sdtype'] != 'id'
+                1
```
amontanez24, line 57: can we update the docstring on line 47 to say all columns that aren't modeled

amontanez24 (approved):

PR #2429, opened 2025-03-20
Metadata auto-detection should find `id` columns that are not primary/foreign keys

### LINKED ISSUE #2412: Metadata auto-detection should find `id` columns that are not primary/foreign keys

amontanez24:
resolves #2412

### DIFF
--- a/sdv/metadata/single_table.py
+++ b/sdv/metadata/single_table.py
@@ -135,7 +135,7 @@ class SingleTableMetadata:
     METADATA_SPEC_VERSION = 'SINGLE_TABLE_V1'
     _DEFAULT_SDTYPES = list(_SDTYPE_KWARGS) + list(SDTYPE_ANONYMIZERS)
     _MIN_ROWS_FOR_PREDICTION = 5
-    _NUMERIC_DTYPES = set('i', 'f', 'u')
+    _NUMERICAL_DTYPES = frozenset(['i', 'f', 'u'])
 
     def _validate_numerical(self, column_name, **kwargs):
         representation = kwargs.get('computer_representation')
@@ -505,8 +505,6 @@ class SingleTableMetadata:
 
             if whole_values and positive_values and unique_lt_categorical_threshold:
                 sdtype = 'categorical'
-            elif unique_values == len(data) and whole_values:
-                sdtype = 'id'
 
         return sdtype
 
@@ -517,14 +515,7 @@ class SingleTableMetadata:
             data (pandas.Series):
                 The data to be analyzed.
         """
-        if len(data) <= self._MIN_ROWS_FOR_PREDICTION:
-            sdtype = 'categorical'
-        else:
-            unique_values = data.nunique()
-            if unique_values == len(data):
-                sdtype = 'id'
-            else:
-                sdtype = 'categorical'
+        sdtype = 'categorical'
 
         with warnings.catch_warnings():
             warnings.simplefilter('ignore', category=UserWarning)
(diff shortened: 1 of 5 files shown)


sdv/metadata/single_table.py
```diff
@@ -536,65 +550,125 @@ def _determine_sdtype_for_objects(self, data):
...
+
+            return sdtype, dtype
+
+        except InvalidMetadataError as e:
+            raise e
+        except Exception as e:
+            self._handle_detection_error(e, column_name, table_name)
+
+    def _select_primary_key(self, infer_sdtypes, pk_candidates, pii_pk_candidates):
+        """Select the primary from a list of candidates.
```
R-Palazzo, line 641: Select the primary from a list of candidates.
```suggestion
        """Select the primary key from a list of candidates.
```

sdv/metadata/single_table.py
```diff
@@ -536,65 +550,125 @@ def _determine_sdtype_for_objects(self, data):
...
+
+        If there ary any non-pii candidates, we select the first one. Otherwise, we select the
```
R-Palazzo, line 643:
```suggestion
        If there are any non-pii candidates, we select the first one. Otherwise, we select the
```

tests/integration/metadata/test_metadata.py
```diff
@@ -160,9 +161,17 @@ def test_detect_from_dataframes_multi_table_without_infer_sdtypes():
...
         },
-        'relationships': [],
+        'relationships': [
+            {
+                'child_foreign_key': 'hotel_id',
+                'child_table_name': 'guests',
+                'parent_primary_key': 'hotel_id',
+                'parent_table_name': 'hotels',
+            }
+        ],
```
R-Palazzo, line 174: Thanks to the changes, we can now detect the relationship in this case?

R-Palazzo (approved):
This looks good!
Thanks for the refactoring 🚀

tests/integration/metadata/test_metadata.py
amontanez24: Yeah because the hotel_id is now being detected as the primary key even if we don't detect the sdtypes, so it's able to find the relationship down the line

frances-h (approved):

PR #2431, opened 2025-03-24
Allow default transformers to have custom arguments

frances-h:
This PR updates the data processor so that parameters used to instantiate the default transformer for an sdtype are used to create additional instances of the transformer.

### DIFF
--- a/sdv/data_processing/data_processor.py
+++ b/sdv/data_processing/data_processor.py
@@ -1,6 +1,5 @@
 """Single table data processing."""
 
-import inspect
 import json
 import logging
 import warnings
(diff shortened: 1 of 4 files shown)


gsheni (approved):

sdv/data_processing/data_processor.py
```diff
@@ -494,6 +495,15 @@ def create_regex_generator(self, column_name, sdtype, column_metadata, is_numeri
...
         return transformer
 
+    @staticmethod
+    def _get_transformer_kwargs(transformer):
+        args = inspect.getfullargspec(transformer.__init__).args[1:]
+        return {
+            key: getattr(transformer, key)
+            for key in args
+            if key != 'model_missing_values' and hasattr(transformer, key)
+        }
```
pvk-developer, line 505: This method doesn't seem to belong here. We are trying to avoid creating static methods overall (following our standard). I feel like this method belongs more to `RDT` than the `DataProcessor` class. If we changed something, lets say, we deprecate one more argument, this change won't propagate here. However, if we were to add this to `rdt` we would require a new release etc.. and that is why I would like to suggest the following:
* Move this function to the `_utils.py` in `sdv` (if possible).
* File an issue in `RDT` regarding adding this functionality to the transformers.
* File another issue in `SDV` to remove this utility function and use the transformer's method instead.

pvk-developer (approved):

PR #2439, opened 2025-03-30
Make single table CAGs backwards compatible

fealho:
CU-86b4exee1.

This PR makes the usage of CAGs backwards compatible.

The feature branch `feature/single-table-CAG` adds two new attributes to `BaseSingleTableSynthesizer`: `_chained_patterns` and `_reject_sampling_patterns`. If a synthesizer from an older version is fitted, it won't have these two attributes and will crash during the `reverse_transform`, when the attributes are called.

This PR fixes the issue by skipping the part of the code which uses these attributes if they are not present. This works because that part of the code is only responsible for handling CAGs, which were not supported for the older versions of the synthesizer.

It also fixes the `get_metadata` method for a similar problem when the `_original_metadata` attribute doesn't exist.

### DIFF
--- a/sdv/single_table/base.py
+++ b/sdv/single_table/base.py
@@ -674,7 +674,7 @@ class BaseSingleTableSynthesizer(BaseSynthesizer):
 
         if hasattr(self, '_original_metadata') and version == 'original':
             return self._original_metadata
-        return self.metadata
+        return super().get_metadata()
 
     def _transform_helper(self, data):
         """Validate and transform all CAG patterns during preprocessing.
--- a/sdv/single_table/base.py
+++ b/sdv/single_table/base.py
@@ -191,6 +191,7 @@ class BaseSynthesizer:
     def _get_table_metadata(self):
         if isinstance(self.metadata, Metadata):
             return self.metadata.tables.get(self._table_name, SingleTableMetadata())
+
         return self.metadata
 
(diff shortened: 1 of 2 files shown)


sdv-team:
Task linked: [CU-86b4exee1 SDV - Make CAG reject sampling backwards compatible](https://app.clickup.com/t/86b4exee1)

tests/unit/single_table/test_base.py
```diff
@@ -264,7 +264,7 @@ def test_get_metadata(self):
         result = instance.get_metadata()
 
         # Assert
-        assert result == metadata
+        assert result.to_dict() == metadata.to_dict()
```
fealho, line 267: Metadata is now a deepcopy of the given object, so the ids are not the same anymore.

pvk-developer (approved):

sdv/single_table/base.py
```diff
@@ -672,7 +677,9 @@ def get_metadata(self, version='original'):
                 f"Unrecognized version '{version}', please use 'original' or 'modified'."
             )
 
-        return self._original_metadata if version == 'original' else self.metadata
+        if hasattr(self, '_original_metadata') and version == 'original':
+            return self._original_metadata
```
frances-h, line 681: Add new line

PR #2443, opened 2025-03-31
Update unit test to work with new RDT `learn_rounding_digits` logic 

fealho:


pyproject.toml
```diff
@@ -38,7 +38,7 @@ dependencies = [
     'copulas>=0.12.1',
     'ctgan>=0.11.0',
     'deepecho>=0.7.0',
-    'rdt>=1.14.0',
+    'rdt @ git+https://github.com/sdv-dev/RDT.git@rounding-dtype-object',
```
amontanez24, line 41: can we bump this to use the release candidate [1.15.1.dev0](https://pypi.org/project/rdt/1.15.1.dev0/)? If it works we can go through with the release

frances-h (commented):
Can you update the test docstring so it's accurately explaining what the test is doing now?

Also since this test is showing that we don't round if we exceed the max number of decimals, should we increase the number of decimal places by 1 so it keeps the old behavior?

frances-h (approved):

amontanez24 (approved):

PR #2458, opened 2025-04-11
When auto-detecting metadata, add a parameter to control the foreign key detection algorithm

### LINKED ISSUE #2456: When auto-detecting metadata, add a parameter to control the foreign key detection algorithm

amontanez24:
resolves #2456

### DIFF
--- a/.github/workflows/dtypes.yml
+++ b/.github/workflows/dtypes.yml
@@ -1,9 +1,8 @@
 name: Dtypes Tests
 
 on:
-  push:
-  pull_request:
-    types: [opened, reopened]
+  schedule:
+    - cron: '0 3 * * 1'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/.github/workflows/dtypes_benchmark.yml
+++ b/.github/workflows/dtypes_benchmark.yml
@@ -1,9 +1,8 @@
 name: Data Types Benchmark
 
 on:
-  push:
-    branches:
-      - main
+  schedule:
+    - cron: '0 3 * * 1'
 
 jobs:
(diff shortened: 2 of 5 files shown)


.github/workflows/dtypes.yml
```diff
@@ -1,9 +1,8 @@
 name: Dtypes Tests
 
 on:
-  push:
-  pull_request:
-    types: [opened, reopened]
+  schedule:
```
gsheni, line 4:
```suggestion
  workflow_dispatch:
  schedule:
```

frances-h (approved):

R-Palazzo (approved):

PR #2468, opened 2025-04-16
Metadata auto-detection no longer reuses the foreign key when creating a relationship

### LINKED ISSUE #2453: Metadata validation does not catch the case where a foreign key is reused
### Environment Details
* SDV version: 1.19.0
* Python version: 3.11
* Operating System: Linux (Colab notebook)

### Error Description
Given a column, it should only be possible to list it as a foreign key _once_ in the `'relationships'` portion of my metadata. It does not make sense for this same column to be listed multiple times as a foreign key to different tables; it is not possible for a single column to simultaneously refer to multiple, different columns.

### Steps to reproduce
The example below creates a metadata that lists the _same_ foreign key column twice in the `'relationships'` section -- once as linking to table `'A'` and once as linking to table `'B'`. This should not be possible, as the same column cannot be simultaneously linking to 2 different parent tables.

Image

Yet, this validates without any errors:

```python

    'tables': {
        'A': {
            'primary_key': 'id',
            'columns': {
                'id': { 'sdtype': 'id'},
            }
        },
        'B': {
            'primary_key': 'id',
            'columns': {
                'id': { 'sdtype': 'id'},
            }
        },
        'C': {
            'columns': {
                'parent_id': { 'sdtype': 'id'},
            }
        }
    },
    'relationships': [{
        'parent_table_name': 'A',
        'parent_primary_key': 'id',
        'child_table_name': 'C',
        'child_foreign_key': 'parent_id'
    },{
        'parent_table_name': 'B',
        'parent_primary_key': 'id',
        'child_table_name': 'C',
        'child_foreign_key': 'parent_id'
    }]
})

metadata.validate()
```

### Expected Behavior
The metadata `validate` should make sure that a given column only appears as a foreign key _once_ (within all the relationships listed). If this is not true, I expect an `InvalidMetadataError` to be thrown.

The error can describe which relationships are having issues. It can list the offending instance(s), when we detect that the foreign key is being re-used.

```
Relationships:
Relationship between tables (B, C) uses a foreign key column ('parent_id') that is already used in another relationship.
```

### Additional Context
Such a relationship would imply that table A's primary key and table B's primary key are completely in-sync. The recommended solution is to then list a [PrimaryToPrimaryKey pattern](https://docs.sdv.dev/sdv/reference/constraints/constraint-augmented-generation-cag/primarytoprimarykey) between table's A and B. Meanwhile, C should only link to _one of them_.

Currently, when you give SDV the (incorrect) schema: 
- HMA will ultimately crash during fitting with `KeyError: 'parent_id'`
- HSA will not crash and will successfully sample data. However, the relationship validity will only be guaranteed for _one_ of the relationships listed (not both). The diagnostic score will not be 1.0.

### LINKED ISSUE #2454: Metadata auto-detection should not be creating a schema where a foreign key column is reused
### Environment Details
* SDV version: 1.19.0
* Python version: 3.11

### Error Description
The SDV Enterprise's metadata detection can sometimes create as schema where the same column is listed as a foreign key multiple times (as part of multiple relationships).

- (a) This does not make sense, because the same column cannot simultaneously be referring to multiple different primary keys and
- (b) Ultimately, our synthesizers cannot effectively model such a schema anyways. HMA will crash

I expect that the auto-detect method should only allow a column to be listed once as a foreign key (considering all the relationships of the metadata).

### Steps to reproduce
```python
import pandas as pd
import numpy as np

# create a schema with 3 tables: A, B, and C
A = pd.DataFrame(data={
    'id': ['id-' + str(i) for i in range(100)],
    'col1': [round(i, 2) for i in np.random.uniform(low=0, high=10, size=100)]
})

B = pd.DataFrame(data={
    'id': ['id-' + str(i) for i in range(100)],
    'col2': [round(i, 2) for i in np.random.uniform(low=0, high=10, size=100)]
})

C = pd.DataFrame(data={
    'id': ['id-' + str(i) for i in np.random.randint(0, 100, size=1000)],
    'col3': [round(i, 2) for i in np.random.uniform(low=0, high=10, size=1000)]
})

data = {
    'A': A,
    'B': B,
    'C': C
}

# auto-detect the metadata
metadata = Metadata.detect_from_dataframes(data)
metadata.visualize()
```

Observe that it lists `C.id` as a foreign key twice: First to refer to `A.id`, and then again to refer to `B.id`. It should not be possible for a column to simultaneously refer to two different primary key columns.

Image

Next, try passing it into HMA. It will crash during fit because the schema is ultimately not valid.

```python
synthesizer.fit(data)
```
```
KeyError: 'id'
```

### Expected Behavior
The auto-detect should only list a column as a foreign key once within the metadata.

In this case, it doesn't really matter whether it detects `C.id` as referring to `A.id` or `B.id` (either is a reasonable option). But whatever it chooses, it should only list it once.

### Additional Context
`metadata.validate()` should also be failing on such a metadata. That is covered in a separate issue #2453

pvk-developer:
Resolves #2454 
Resolves #2453
CU-86b4kbb27

### DIFF
--- a/sdv/metadata/multi_table.py
+++ b/sdv/metadata/multi_table.py
@@ -174,7 +174,7 @@ class MultiTableMetadata:
         self, parent_table_name, parent_primary_key, child_table_name, child_foreign_key
     ):
         for relationship in self.relationships:
-            foreign_key_already_exists = (
+            foreign_key_already_used = (
                 relationship['child_table_name'] == child_table_name
                 and relationship['child_foreign_key'] == child_foreign_key
             )
@@ -182,7 +182,7 @@ class MultiTableMetadata:
                 relationship['parent_table_name'] == parent_table_name
                 and relationship['parent_primary_key'] == parent_primary_key
             )
-            if foreign_key_already_exists and not parent_matches:
+            if foreign_key_already_used and not parent_matches:
                 raise InvalidMetadataError(
                     f'Relationship between tables ({parent_table_name}, {child_table_name}) uses '
                     f"a foreign key column ('{child_foreign_key}') that is already used in another "
--- a/sdv/metadata/multi_table.py
+++ b/sdv/metadata/multi_table.py
@@ -304,9 +304,6 @@ class MultiTableMetadata:
         self._validate_relationship_does_not_exist(
             parent_table_name, parent_primary_key, child_table_name, child_foreign_key
         )
-        self._validate_foreign_key_is_not_reused(
-            parent_table_name, parent_primary_key, child_table_name, child_foreign_key
-        )
         self._validate_child_map_circular_relationship(child_map)
 
         self.relationships.append({
--- a/sdv/metadata/multi_table.py
+++ b/sdv/metadata/multi_table.py
@@ -180,8 +180,18 @@ class MultiTableMetadata:
                 and relationship['child_table_name'] == child_table_name
                 and relationship['child_foreign_key'] == child_foreign_key
             )
+            foreign_key_already_exists = (
+                relationship['child_table_name'] == child_table_name
+                and relationship['child_foreign_key'] == child_foreign_key
+            )
             if already_exists:
                 raise InvalidMetadataError('This relationship has already been added.')
+
+            if foreign_key_already_exists:
+                raise InvalidMetadataError(
+                    f"The foreign key '{child_foreign_key}' for table '{child_table_name}' is "
+                    'already being used in a relationship.'
+                )
 
     def _validate_relationship(
         self, parent_table_name, child_table_name, parent_primary_key, child_foreign_key
--- a/sdv/metadata/multi_table.py
+++ b/sdv/metadata/multi_table.py
@@ -170,6 +170,25 @@ class MultiTableMetadata:
                 'with a non-primary key.'
             )
 
+    def _validate_foreign_key_is_not_reused(
+        self, parent_table_name, parent_primary_key, child_table_name, child_foreign_key
+    ):
+        for relationship in self.relationships:
+            foreign_key_already_exists = (
+                relationship['child_table_name'] == child_table_name
+                and relationship['child_foreign_key'] == child_foreign_key
+            )
+            parent_matches = (
+                relationship['parent_table_name'] == parent_table_name
+                and relationship['parent_primary_key'] == parent_primary_key
+            )
+            if foreign_key_already_exists and not parent_matches:
+                raise InvalidMetadataError(
+                    f'Relationship between tables ({parent_table_name}, {child_table_name}) uses '
+                    f"a foreign key column ('{child_foreign_key}') that is already used in another "
+                    'relationship.'
+                )
+
     def _validate_relationship_does_not_exist(
         self, parent_table_name, parent_primary_key, child_table_name, child_foreign_key
     ):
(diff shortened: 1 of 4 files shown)


sdv-team:
Task linked: [CU-86b4kbb27 SDV - Metadata auto-detection should not be creating a schema where a foreign key column is reused #2454](https://app.clickup.com/t/86b4kbb27)

frances-h (approved):

sdv/metadata/multi_table.py
```diff
@@ -180,9 +180,19 @@ def _validate_relationship_does_not_exist(
                 and relationship['child_table_name'] == child_table_name
                 and relationship['child_foreign_key'] == child_foreign_key
             )
+            foreign_key_already_exists = (
```
amontanez24, line 183: minor: can we call this `foreign_key_used_in_other_relationship` or `foreign_key_already_used`?

amontanez24 (commented):
The issue says that sometimes a foreign key was detected for multiple columns, but I don't see any changes to the detect method

pvk-developer:
> The issue says that sometimes a foreign key was detected for multiple columns, but I don't see any changes to the detect method

There is no need to handle that during detection itself, since we are not supporting this schema (as per issue #2453 ) as long as we validate this during `add_relationship` we are covered in a wider variety of cases.
Here is the code for detecting: https://github.com/sdv-dev/SDV/blob/3155834c07e631647482251de6ad10a96c057b5e/sdv/metadata/multi_table.py#L526

This change will also propagate to `SDV-Enterprise` without a need to update the code base there.

sdv/metadata/multi_table.py
```diff
@@ -282,6 +304,9 @@ def add_relationship(
         self._validate_relationship_does_not_exist(
             parent_table_name, parent_primary_key, child_table_name, child_foreign_key
         )
+        self._validate_foreign_key_is_not_reused(
+            parent_table_name, parent_primary_key, child_table_name, child_foreign_key
+        )
```
frances-h, line 309: I don't think we need to call `_validate_foreign_key_is_not_reused` again here because we already called `_validate_relationships`. Also do you know if there's a specific reason why `_validate_relationship_does_not_exist` isn't called in `_validate_relationships`?

pvk-developer: Yes, because we use it during `validate` which validates each relationship one by one. If we called this `_validate_relationship_does_not_exist` this would raise errors saying that the relationships exist.

sdv/metadata/multi_table.py
```diff
@@ -170,6 +170,25 @@ def _validate_foreign_child_key(self, child_table_name, parent_table_name, child
                 'with a non-primary key.'
             )
 
+    def _validate_foreign_key_is_not_reused(
+        self, parent_table_name, parent_primary_key, child_table_name, child_foreign_key
+    ):
+        for relationship in self.relationships:
+            foreign_key_already_used = (
```
amontanez24, line 177: won't this cause the detection to error if it accidentally tries to add a relationship with a foreign key that's already used? I don't think it should crash. Instead I think the detection should just not detect a foreign key it already has seen

amontanez24: nvm the add_relationship is wrapped in a try/catch

amontanez24 (approved):

frances-h (approved):

R-Palazzo (approved):

PR #2477, opened 2025-04-22
Fix: Error message for reused foreign key in relationships was showing the valid relationship too

pvk-developer:
As part of #2454 and #2453 we've updated the behavior of the validate command when handling reused foreign keys in relationships.

### Previous behavior:
When multiple relationships used the same foreign key, the validation output would display all of them, even if the first one was valid. For example:
```python
Relationships:
Relationship between tables (A1, A3) uses a foreign key column ('fk3_A1_A2') that is already used in another relationship.
Relationship between tables (A2, A3) uses a foreign key column ('fk3_A1_A2') that is already used in another relationship.
```

### New behavior:
With the latest implementation, the validation will now retain the first valid relationship and only raise errors for subsequent conflicting relationships using the same foreign key. Example output:

```
Relationships:
Relationship between tables (A2, A3) uses a foreign key column ('fk3_A1_A2') that is already used in another relationship.
```

### DIFF
--- a/sdv/metadata/multi_table.py
+++ b/sdv/metadata/multi_table.py
@@ -190,8 +190,12 @@ class MultiTableMetadata:
                 )
 
     def _validate_foreign_key_uniqueness_across_relationships(
-        self, parent_table_name, parent_primary_key, child_table_name,
-        child_foreign_key, seen_foreign_keys
+        self,
+        parent_table_name,
+        parent_primary_key,
+        child_table_name,
+        child_foreign_key,
+        seen_foreign_keys,
     ):
         key = (child_table_name, child_foreign_key)
         current_relationship = (parent_table_name, parent_primary_key)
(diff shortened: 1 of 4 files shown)


sdv/metadata/multi_table.py
```diff
@@ -783,8 +800,12 @@ def validate(self):
         """
         errors = []
         self._validate_single_table(errors)
+        self._seen_foreign_keys = {}
```
amontanez24, line 803: can we just make this a variable and not store it on the class

pvk-developer: Done in [81a46ae](https://github.com/sdv-dev/SDV/pull/2477/commits/81a46ae0f4abdf8a8dfbb4eb23db068b6b56fc1a)

amontanez24 (approved):

frances-h (approved):

Issue #961, opened 2022-08-18
Foreign Key can't be in Unique constraint


LiFaytheGoblin:
### Environment Details

Please indicate the following details about the environment in which you found the bug:

* SDV version: 0.16.0
* Python version: 3.8
* Operating System: Ubuntu 20.04.4

### Error Description

I tried to create relational data with 2 tables:
- a table of `sections` with their id, rank, and amount of elements in a section and 
- a table of `elements` with their id, which section they belong to, rank within the section, and type of element

For `elements` I added a Unique constraint for the combination of the columns section and rank, so that the rank is unique per section.

However, now `model.sample()` returns the error: `UserWarning: Unique cannot be transformed because columns: ['section'] were not found. Using the reject sampling approach instead. on the line model.fit(data).`

I do not receive any new data.

### Steps to reproduce
I use the following code:
```
from sdv.metadata.dataset import Metadata
from sdv.relational import HMA1

md = Metadata("test-data/metadata-test-2.json")
data = md.load_tables()
model = HMA1(md)
model.fit(data)
new_data = model.sample()
```

An extract from elements-test-2.csv:

```
element_id,section,rank,type
1,58964,1,label
2,58964,2,forum
3,58967,2,page
4,58967,1,book
```

An extract from sections-test-2.csv:
```
section_id,rank,elements_amount
58964,1,2
58967,2,4
```

My metadata is as follows:
```
{
  "tables": {
    "sections": {
        "fields": {
          "section_id": { "type": "id", "subtype": "integer" },
          "rank": {"type": "numerical", "subtype": "integer" },
          "elements_amount": {"type": "numerical", "subtype": "integer" }
        },
        "path": "sections-test-2.csv",
        "primary_key": "section_id",
        "constraints": [
          {
            "constraint": "sdv.constraints.Unique",
            "column_names": ["rank"]
          }
        ]
    },
    "elements": {
        "fields": {
          "element_id": { "type": "id", "subtype": "integer" },
          "rank": {"type": "numerical", "subtype": "integer" },
          "type": { "type": "categorical" },
          "section": {
            "type": "id",
            "subtype": "integer",
            "ref": {
              "table": "sections",
              "field": "section_id"
            }
          }
        },
        "path": "elements-test-2.csv",
        "primary_key": "element_id",
        "constraints": [
          {
            "constraint": "sdv.constraints.Unique",
            "column_names": ["rank", "section"]
          }
        ]
    }
  }
}
```

The problematic part seems to be
[...]

PR #2485, opened 2025-04-25
Add method to validate CAG pattern for synthetic data 

### LINKED ISSUE #2471: Add synthesizer.validate_cag method
### Problem Description

_As a user, I'd like to see if the synthetic data I generate matches the CAG patterns I added._

### Expected behavior

- Add a new method to the [BaseSynthesizer](https://github.com/sdv-dev/SDV/blob/main/sdv/single_table/base.py) class called `validate_cag(self, synthetic_data)`.
    - Inputs
        - `synthetic_data`: DataFrame in single table case.
    - Returns: If there are any invalid rows, this method should raise an error and print the first 5 rows it's failing on for each CAG that has failures.
- Add a new method to the [BaseMultiTableSynthesizer](https://github.com/sdv-dev/SDV/blob/main/sdv/multi_table/base.py) class called `validate_cag(self, synthetic_data)`.
    - Inputs
        - `synthetic_data`: Dictionary mapping table name to DataFrames.
    - Returns: If there are any invalid rows, this method should raise an error and print the first 5 rows it's failing on for each CAG that has failures.

### Additional context

- Internally, this method should use the is_valid method for each CAG. The data may need to be transformed after validating each CAG, since some CAGs will have been learned on a different data schema (the CAGs before it altered the data and metadata)
https://github.com/sdv-dev/SDV/blob/da8a83801b7d1fe7bafcaddc6d5ccb464323b32c/sdv/cag/base.py#L148

gsheni:
- Fixes #2471


sdv/single_table/base.py
```diff
@@ -674,6 +675,34 @@ def get_cag(self):
...
+                else:
+                    valid = pattern.is_valid(data=transformed_data)
+
+                if not valid.all():
+                    invalid_rows_str = _get_invalid_rows(valid)
+                    pattern_name = _convert_to_snake_case(pattern.__class__.__name__)
+                    pattern_name = pattern_name.replace('_', ' ')
+                    msg = f'The {pattern_name} requirement is not met '
+                    msg += f'for row indices: {invalid_rows_str}.'
+                    raise PatternNotMetError(msg)
```
gsheni, line 702: @frances-h The issue says to raise an error for each failing CAG. However, since a pattern may transforms the data such that another pattern relies upon it, I opted to just raise the error message for the first failing CAG.

sdv/multi_table/base.py
```diff
@@ -186,6 +188,33 @@ def get_cag(self):
...
+                if not valid_data[table_name].all():
+                    invalid_rows_str = _get_invalid_rows(valid_data[table_name])
+                    pattern_name = _convert_to_snake_case(pattern.__class__.__name__)
+                    pattern_name = pattern_name.replace('_', ' ')
+                    msg = (
+                        f"Table '{table_name}': The {pattern_name} requirement is not "
+                        f'met for row indices: {invalid_rows_str}.'
+                    )
+                    raise PatternNotMetError(msg)
+            else:
```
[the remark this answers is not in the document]
gsheni: Removed the else:
https://github.com/sdv-dev/SDV/pull/2485/commits/a5cdeb910751e0eafc2a472c97b8fbd19f6620e3#diff-f825c52e821d110ec1626fbf595d32602644f0f0ece86e637572ba3d34caf0d4R149

sdv/cag/_utils.py
```diff
@@ -60,40 +61,6 @@ def _validate_table_name_if_defined(table_name):
...
-
-    Returns:
-        (sdv.metadata.Metadata): The new Metadata, with the columns removed.
-    """
-    if isinstance(metadata, Metadata):
-        metadata = metadata.to_dict()
-    for column in columns_to_drop:
-        primary_key = metadata['tables'][table_name].get('primary_key')
-        if primary_key and primary_key == column:
-            raise ValueError('Cannot remove primary key from Metadata')
```
amontanez24, line 86: do we still have to worry about the primary key being removed?

amontanez24 (commented):
This looks good! One question

sdv/cag/_utils.py
gsheni: I don't believe so. In CAGs, we remove columns that are one of the arguments of the CAG (ex - `high_column_name` for Inequality). Those user-defined columns have to pass the initial checks for the CAG and then only would they be removed.
A primary key has to be an ID or PII type. So there should not be overlap between user-defined columns in CAG and primary keys.

gsheni: That's a good point. Looking at the [predefined constraints list](https://docs.sdv.dev/sdv/concepts/constraints/predefined-constraints), I can see supporting these for primary keys. Indeed, there is an outstanding feature request for supporting constraints with primary keys:
- https://github.com/sdv-dev/SDV/issues/961

Therefore, I added the back the function `_remove_columns_from_metadata`.

amontanez24 (approved):

fealho (approved):

PR #2501, opened 2025-05-01
Dtypes benchmark includes missing values

### LINKED ISSUE #2494: Dtypes benchmark should include missing values for any dtypes that support it
### Problem Description
Our new testing framework tests to see which data representation types (dtypes) work with SDV's sdtypes. The benchmarking tests a variety of dtypes for every sdtype. For example for sdtype `categorical` it tests dtypes such as `pd.object`, `np.float64`, etc.

For all of the dtypes that support `null` values, we should make sure to add `null` values in our benchmarking framework as well. 

### Expected behavior
Go through all the dtypes in the [numpy_dtypes.py](https://github.com/sdv-dev/SDV/blob/main/tests/benchmark/numpy_dtypes.py) and in the [pandas_dtypes.py](https://github.com/sdv-dev/SDV/blob/main/tests/benchmark/pandas_dtypes.py). If any of the dtypes support null values, then add null values into the data that we're testing.

At a first glance, here are a few dtypes that support null values (that are not being tested):
- `np.float64`, `np.float32`, etc. all support null values 
- `np.complex64`, `np.complex128`, etc. all support null values
- `np.object` supports null values
- `np.string` supports null values (this one needs to be specifically cast to `dtype='string'`)
- `pd.datetime64` supports null values
- etc.

**The above is not an exhaustive list! We should go through each one and add null values if the dtype supports it.**

rwedge:
Resolves #2494 

CU-86b4uwr1q

### DIFF
--- a/.github/workflows/dtypes_benchmark.yml
+++ b/.github/workflows/dtypes_benchmark.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - issue-2494-dtypes-benchmark-nulls
 
 jobs:
   run_dtypes_benchmark:
--- a/.github/workflows/dtypes_benchmark.yml
+++ b/.github/workflows/dtypes_benchmark.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - main
-      - issue-2494-dtypes-benchmark-nulls
 
 jobs:
   run_dtypes_benchmark:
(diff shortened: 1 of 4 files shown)


sdv-team:
Task linked: [CU-86b4uwr1q SDV - Dtypes benchmark should include missing values for any dtypes that support it #2494](https://app.clickup.com/t/86b4uwr1q)

frances-h (approved):

pvk-developer (approved):

PR #2503, opened 2025-05-05
Add version parameter to SingleTableSynthesizer

### LINKED ISSUE #2484: Add 'version' parameter to SingleTableSynthesizer.get_metadata
### Problem Description

_As a user, I'd like to see how the CAG patterns I use modify the metadata so that I know what columns exist in the data that gets modeled._

https://github.com/sdv-dev/SDV/blob/28b24957b8f328be41f87e19a6e8f85323bcb60a/sdv/single_table/base.py#L331

We have a `get_metadata` function on the base class for single table synthesizers. This lets users see the state of the metadata. When CAG patterns, the metadata might change before modeling. This is because the patterns sometimes transform tables and columns. Users and engineers may want to see this changed metadata state, so we need a way to expose it.
 
### Expected behavior

- Add a parameter called `version` to the `get_metadata` function.
- **Description**: `version` (str, optional):
                The version of metadata to return, must be one of 'original' or 'modified'. If
                'original', will return the original metadata used to instantiate the
                synthesizer. If 'modified', will return the modified metadata after applying this
                synthesizer's data patterns. Defaults to 'original'.

### Additional context

- [This PR](https://github.com/sdv-dev/SDV/pull/2472/files) add an attribute called _original_metadata that can be used to access the original instead of the modified metadata.

pvk-developer:
Resolves #2484 
CU-86b4tehv8

This pull request adds a new private attribute to the `BaseSynthesizer`, the `_input_metadata`. This attribute holds reference to the input metadata of the user. This allows us to determine wether or not they have updated the metadata after some steps and they forgot to re-create a synthesizer.

This conserves the current `self.metadata` and `self._original_metadata` as two different objects. The `_original_metadata` being the unmodified metadata of the original input, while `self.metadata` is the mutable version that could be altered by CAG's or other processes.

### DIFF
--- a/sdv/multi_table/base.py
+++ b/sdv/multi_table/base.py
@@ -200,13 +200,10 @@ class BaseMultiTableSynthesizer:
             error_msg = f"Unrecognized version '{version}', please use 'original' or 'modified'."
             raise ValueError(error_msg)
 
-        if version == 'original':
-            original = (
-                self._original_metadata if hasattr(self, '_original_metadata') else self.metadata
-            )
-            return original
-        else:
-            return self.metadata
+        if version == 'original' and hasattr(self, '_original_metadata'):
+            return Metadata.load_from_dict(self._original_metadata.to_dict())
+
+        return Metadata.load_from_dict(self.metadata.to_dict())
 
     def _transform_helper(self, data):
         """Validate and transform all CAG patterns during preprocessing.
--- a/sdv/single_table/base.py
+++ b/sdv/single_table/base.py
@@ -103,7 +103,11 @@ class BaseSynthesizer:
 
     def _check_input_metadata_updated(self):
         if not hasattr(self, '_input_metadata'):
-            unified_metadata = Metadata.load_from_dict(self._original_metadata.to_dict())
+            if hasattr(self, '_original_metadata'):
+                unified_metadata = Metadata.load_from_dict(self._original_metadata.to_dict())
+            else:
+                unified_metadata = Metadata.load_from_dict(self.metadata.to_dict())
+
             setattr(self, '_input_metadata', unified_metadata)
 
         if isinstance(self._input_metadata, Metadata):
--- a/sdv/single_table/base.py
+++ b/sdv/single_table/base.py
@@ -152,7 +152,7 @@ class BaseSynthesizer:
         # Points to the input metadata object and allows us to check if user has changed it
         self._input_metadata = metadata
 
-        # Points to a dynamic metadata object that could be re-created by CAG's
+        # Points to a dynamic metadata object that could be modified by constraints
         self.metadata = metadata
 
         self._table_name = Metadata.DEFAULT_SINGLE_TABLE_NAME
--- a/sdv/single_table/base.py
+++ b/sdv/single_table/base.py
@@ -158,7 +158,7 @@ class BaseSynthesizer:
         self.metadata.validate()
 
         # Points to a metadata object that conserves the initialized status of the synthesizer
-        self._original_metadata = deepcopy(metadata)
+        self._original_metadata = deepcopy(self.metadata)
         self._check_metadata_updated()
         self.enforce_min_max_values = enforce_min_max_values
         self.enforce_rounding = enforce_rounding
--- a/sdv/single_table/base.py
+++ b/sdv/single_table/base.py
@@ -101,16 +101,16 @@ class BaseSynthesizer:
             for sdtype, transformer in self._model_sdtype_transformers.items():
                 self._data_processor._update_transformers_by_sdtypes(sdtype, transformer)
 
-    def _check_original_metadata_updated(self):
-        if not hasattr(self, '_original_metadata'):
+    def _check_input_metadata_updated(self):
+        if not hasattr(self, '_input_metadata'):
             unified_metadata = Metadata.load_from_dict(self.metadata.to_dict())
-            setattr(self, '_original_metadata', unified_metadata)
+            setattr(self, '_input_metadata', unified_metadata)
 
-        if isinstance(self._original_metadata, Metadata):
-            metadata = self._original_metadata._convert_to_single_table()
+        if isinstance(self._input_metadata, Metadata):
+            metadata = self._input_metadata._convert_to_single_table()
 
         else:
-            metadata = self._original_metadata
+            metadata = self._input_metadata
 
         if metadata._updated:
             warnings.warn(
@@ -139,7 +139,13 @@ class BaseSynthesizer:
         self, metadata, enforce_min_max_values=True, enforce_rounding=True, locales=['en_US']
     ):
         self._validate_inputs(enforce_min_max_values, enforce_rounding)
-        self._original_metadata = self.metadata = metadata
+
+        # Points to the input metadata object and allows us to check if user has changed it
+        self._input_metadata = metadata
+
+        # Points to a dynamic metadata object that could be re-created by CAG's
+        self.metadata = metadata
+
         self._table_name = Metadata.DEFAULT_SINGLE_TABLE_NAME
         if isinstance(metadata, Metadata):
             self._table_name = metadata._get_single_table_name()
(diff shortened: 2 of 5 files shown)


sdv/single_table/base.py
```diff
@@ -101,16 +101,16 @@ def _update_default_transformers(self):
...
+            setattr(self, '_input_metadata', unified_metadata)
 
-        if isinstance(self._original_metadata, Metadata):
-            metadata = self._original_metadata._convert_to_single_table()
+        if isinstance(self._input_metadata, Metadata):
+            metadata = self._input_metadata._convert_to_single_table()
 
         else:
-            metadata = self._original_metadata
+            metadata = self._input_metadata
```
fealho, line 113: This approach doesn't resolve [this](https://github.com/sdv-dev/SDV/issues/2463) issue (if you run the code snippet the warning doesn't surface).

The reason is that it checks `if _input_metadata._updated` is True in line 115, but that may not be the case since only `medatada._updated` gets set to True by the user-facing methods.

sdv/single_table/base.py
```diff
@@ -150,6 +156,9 @@ def __init__(
             self.metadata.tables[self._table_name]._updated = metadata._updated
 
         self.metadata.validate()
+
+        # Points to a metadata object that conserves the initialized status of the synthesizer
+        self._original_metadata = deepcopy(self.metadata)
```
fealho, line 161: This needs to be set in line 162. `_check_metadata_updated` sets `self.metadata._updated` to False, which is how the `_original_metadata` attribute should be initialized.

tests/integration/single_table/test_base.py
```diff
@@ -653,18 +653,24 @@ def test_metadata_updated_warning_detect(mock__fit):
...
+    )
 
     # Run
     with warnings.catch_warnings(record=True) as record:
         instance = BaseSingleTableSynthesizer(metadata)
         instance.fit(data)
 
     # Assert
-    assert len(record) == 2
+    assert len(record) == 3
```
fealho, line 667: You shouldn't update this, the warning shouldn't be surfacing here. Moving the position of the `original_metadata` attribute to line 162 as I described above should solve this.

fealho (changes requested):

sdv/single_table/base.py
```diff
@@ -343,12 +352,26 @@ def get_parameters(self):
...
+                synthesizer's CAG patterns. Defaults to 'original'.
+        """
+        if version not in ('original', 'modified'):
+            raise ValueError(
+                f"Unrecognized version '{version}', please use 'original' or 'modified'."
+            )
+
+        table_name = getattr(self, '_table_name', None)
+        if hasattr(self, '_original_metadata') and version == 'original':
+            return Metadata.load_from_dict(self._original_metadata.to_dict(), table_name)
```
fealho, line 372: If `_original_metadata` is already a `Metadata` object this is unnecessary.

sdv/single_table/base.py
```diff
@@ -343,12 +352,26 @@ def get_parameters(self):
...
+
+        return Metadata.load_from_dict(self.metadata.to_dict(), table_name)
```
fealho, line 374: Same here.

sdv/single_table/base.py
pvk-developer: Yeah this makes sense!

sdv/single_table/base.py
pvk-developer: `get_metadata` has to return a copy.

tests/integration/single_table/test_base.py
pvk-developer: This doesn't have to do with the `original_metadata` but the `_input_metadata` which is the object that the user has currently in the running environment. I do see however that we should probably set the flag to `False` once we created the instance of this synthesizer.

Added it here: https://github.com/sdv-dev/SDV/pull/2503/commits/f8b3ef7708ac3500ee6b4dfaa1f8ceff2cc34c20

sdv/single_table/base.py
pvk-developer: image
I'm not sure if I understand your concern here , but I think it works as intended.

sdv/single_table/base.py
fealho: That's a good catch. In that case though, returning a deepcopy of the object makes the intention clearer. Either way, this change should probably be replicated on the other two `get_metadata` methods [here](https://github.com/sdv-dev/SDV/blob/feature/single-table-CAG/sdv/multi_table/base.py#L203-L209) and [here](https://github.com/datacebo/SDV-Enterprise/blob/eb3ac78b1a6a4b685756251a85abb9a371be64fe/sdv_enterprise/sdv/multi_table/cag/base.py#L55-L61).

tests/integration/single_table/test_base.py
fealho: Yes, this is what I meant.

sdv/single_table/base.py
fealho: This gets fixed with your latest commit. My only concern would be if the user changes what metadata instance they are using - by saving and loading the metadata for example - but maybe we don't care about such cases?

sdv/single_table/base.py
pvk-developer: Good catch, I will add it to those files asap.

pvk-developer: Done in [8cceac7](https://github.com/sdv-dev/SDV/pull/2503/commits/8cceac7a3149dc487035a4f51c8bba3b1488a3a2)

sdv/multi_table/base.py
```diff
@@ -200,13 +200,10 @@ def get_metadata(self, version='original'):
...
 
-        if version == 'original':
-            original = (
-                self._original_metadata if hasattr(self, '_original_metadata') else self.metadata
-            )
-            return original
-        else:
-            return self.metadata
+        if version == 'original' and hasattr(self, '_original_metadata'):
+            return deepcopy(self._original_metadata)
```
fealho, line 204: Is it possible this is a `SingleTableMetadata` for the backwards compatibility case? If so, should it be converted to `Metadata`? Same for line 206.

fealho (commented):
Looks good, I left one final comment.

sdv/multi_table/base.py
pvk-developer: I guess thats why we had it with `Metadata.load_from_dict` ? 😓 I will just re-vert this changes (not using deepcopy).

frances-h (approved):

sdv/single_table/base.py
```diff
@@ -101,16 +101,16 @@ def _update_default_transformers(self):
             for sdtype, transformer in self._model_sdtype_transformers.items():
                 self._data_processor._update_transformers_by_sdtypes(sdtype, transformer)
 
-    def _check_original_metadata_updated(self):
-        if not hasattr(self, '_original_metadata'):
-            unified_metadata = Metadata.load_from_dict(self.metadata.to_dict())
-            setattr(self, '_original_metadata', unified_metadata)
+    def _check_input_metadata_updated(self):
+        if not hasattr(self, '_input_metadata'):
+            unified_metadata = Metadata.load_from_dict(self._original_metadata.to_dict())
```
fealho, line 106: You need to check that the `_original_metadata` attribute exists for backwards compatibility.

amontanez24 (commented):
I think this looks good!

fealho (approved):
Great work! :shipit:

sdv/single_table/base.py
```diff
@@ -139,7 +148,13 @@ def __init__(
         self, metadata, enforce_min_max_values=True, enforce_rounding=True, locales=['en_US']
     ):
         self._validate_inputs(enforce_min_max_values, enforce_rounding)
-        self._original_metadata = self.metadata = metadata
+
+        # Points to the input metadata object and allows us to check if user has changed it
+        self._input_metadata = metadata
+
+        # Points to a dynamic metadata object that could be re-created by CAG's
```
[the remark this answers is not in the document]
pvk-developer: Done in [0564306](https://github.com/sdv-dev/SDV/pull/2503/commits/0564306310bae42c16964d77516aa7652175917b)

PR #2516, opened 2025-05-12
Add `ProgrammableConstraint` and `ProgrammableSingleTableConstraint`

### LINKED ISSUE #2513: Add `ProgrammableConstraint` and `ProgrammableSingleTableConstraint`
### Problem Description
As we shift to the CAG framework, it would be helpful to provide a way to allow users to write and use their own constraints. We should create two base classes, `ProgrammableConstraint` and `ProgrammableSingleTableConstraint` that a user can use as a parent class to define the necessary methods to work with the CAG framework.

### Expected behavior
Create the two base classes, `ProgrammableConstraint` and `ProgrammableSingleTableConstraint`. The base class should be set up with the following methods, which by default should raise a `NotImplementedError`:
- fit(self, data, metadata)
- transform(self, data)
- get_updated_metadata(self, metadata)
- reverse_transform(self, transformed_data)
- is_valid(self, synthetic_data)

The following methods should also be defined but are optional. The validation methods can return None (by default providing no validation), and `fix_data` can simply return the input data.
- validate(self, metadata)
- validate_input_data(self, data)
- fix_data(self, synthetic_data)

Additionally, create a `ProgrammableConstraintHarness` class. This class should inherit from the `BasePattern` and take a `ProgrammableConstraint` or `ProgrammableSingleTableConstraint` as input. The `ProgrammableConstraintHarness` should act as an interface between the CAG framework and the programmable constraint class. When a `ProgrammableConstraint` is added to a synthesizer, it should be wrapped in a harness class. Additionally, when calling `get_cag` on the synthesizer, the original `ProgrammableConstraint` should be returned, not the harness instance.

### Additional context
See doc [here](https://docs.google.com/document/d/1VwepqwwHAFU-AYCpqtfo1IbFgrXlA0kEMjPDpgbmKVk/edit?tab=t.0#heading=h.brf9n840qwp1)

frances-h:
Resolve #2513

### DIFF
--- a/sdv/cag/base.py
+++ b/sdv/cag/base.py
@@ -13,6 +13,27 @@ class BasePattern:
         self._fitted = False
         self._single_table = False
 
+    def _convert_data(self, data, metadata, copy=False):
+        """Helper to handle converting single dataframes into dictionaries.
+
+        This method takes in data, metadata, and, optionally, a flag indiciating if the
+        returned data should be a copy of the original input data. If the data is a single
+        dataframe, it converts it into a dictionary of dataframes.
+        """
+        if isinstance(data, pd.DataFrame):
+            if copy:
+                data = data.copy()
+
+            if self._single_table:
+                data = {self._table_name: data}
+            else:
+                table_name = self._get_single_table_name(metadata)
+                data = {table_name: data}
+        elif copy:
+            data = {table_name: table_data.copy() for table_name, table_data in data.items()}
+
+        return data
+
     def _get_single_table_name(self, metadata):
         if not hasattr(self, 'table_name'):
             raise ValueError('No ``table_name`` attribute has been set.')
@@ -43,14 +64,8 @@ class BasePattern:
 
         self._validate_pattern_with_metadata(metadata)
 
-        if isinstance(data, pd.DataFrame):
-            if self._single_table:
-                data = {self._table_name: data}
-            else:
-                table_name = self._get_single_table_name(metadata)
-                data = {table_name: data}
-
         if data is not None:
+            data = self._convert_data(data, metadata)
             self._validate_pattern_with_data(data, metadata)
 
     def _get_updated_metadata(self, metadata):
@@ -82,7 +97,7 @@ class BasePattern:
         if isinstance(data, pd.DataFrame):
             self._single_table = True
             self._table_name = self._get_single_table_name(metadata)
-            data = {self._table_name: data}
+            data = self._convert_data(data, metadata)
 
         self._validate_pattern_with_data(data, metadata)
         self._fit(data, metadata)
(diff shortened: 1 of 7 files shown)


sdv/cag/base.py
```diff
@@ -13,6 +13,27 @@ def __init__(self):
         self._fitted = False
         self._single_table = False
 
+    def _convert_data(self, data, metadata, copy=False):
```
amontanez24, line 16: minor: can we call `convert_data_to_dictionary` or `dataframe_to_dict` or `dataframe_to_multi_table` or something along those lines?

sdv/cag/programmable_constraint.py
```diff
@@ -0,0 +1,196 @@
...
+                The data to validate.
+
+        Returns:
+            dict[pd.Series]:
+                For each table, a boolean Series indicating whether or not
+                the corresponding row for the table is valid for this constraint.
+        """
+        raise NotImplementedError()
+
+    def fix_data(self, synthetic_data):
```
amontanez24, line 108: will we rename the `_fix_dayz_data` methods in enterprise to match this?

sdv/cag/programmable_constraint.py
```diff
@@ -0,0 +1,196 @@
...
+            raise PatternNotMetError(
+                'SingleTableProgrammableConstraint cannot be used with multi-table metadata. '
+                'Please use the ProgrammableConstraint base class instead.'
+            )
+
+        self.programmable_constraint.validate(metadata)
+
+    def _validate_pattern_with_data(self, data, metadata):
+        if self._is_single_table:
+            data = data[self._get_single_table_name(metadata)]
```
amontanez24, line 152: why is `self._table_name` not used here?

amontanez24 (commented):
Looking good! All minor comments

sdv/cag/programmable_constraint.py
frances-h: It's not guaranteed to have been set here. We only set `self._table_name` during `fit`, so we have to use the metadata here. This only ever gets called after validating with metadata though, so we will have already caught if the user is trying to pass multiple tables to the single table programmable constraint.

sdv/cag/programmable_constraint.py
frances-h: I don't think so. Currently the CAG framework isn't integrated with DayZ so the users have to use the `fix_dayz_data` method, but ideally in the future a user can just add the constraint to DayZ directly and DayZ will call the `fix_dayz_data` method for all the constraints. 

For programmable constraints, all that matters is that the harness's `_fix_dayz_data` method calls the programmable constraint's `fix_data` method.

sdv/cag/programmable_constraint.py
amontanez24: but it's available in the _fit below? Is it called by the base fit before _fit is called?

amontanez24 (approved):

pvk-developer (approved):

PR #2480, opened 2025-04-24
Add CAG validation to synthesizer.validate

R-Palazzo:
CU-86b4pmjph
Resolve #2470

PR #2521, opened 2025-05-13
`ValueError` if conditionally sampling on a column dropped by constraints

### LINKED ISSUE #2519: `ValueError` if conditionally sampling on a column dropped by constraints
### Environment Details

* SDV version:
* Python version:
* Operating System:

### Error Description
When I try to conditionally sample a column that was dropped by a CAG constraint, I get a `ValueError: Unexpected column name 'checkout_date'. Use a column name that was present in the original data.`

The new CAG constraints should follow the old constraint behavior and fall back to reject sampling if a conditional column was dropped by a constraint.

### Steps to reproduce

```python
condition = Condition({'checkout_date': '29 Dec 2020'}, 10)
constraint = Inequality(low_column_name='checkin_date', high_column_name='checkout_date')
synthesizer.add_cag([constraint])
synthesizer.fit(data)
synthesizer.sample_from_conditions([condition])
```

frances-h:
Resolve #2519 

This PR also enables conditionally sampling columns that are dropped by the data processor by forcing the generated values to match the condition.

Also fixes the Range constraint erroring with NaN values and converts many tests that were using the old style constraints to the new style constraints.

Currently some tests are blocked by #2480

### DIFF
--- a/sdv/cag/inequality.py
+++ b/sdv/cag/inequality.py
@@ -127,7 +127,7 @@ class Inequality(BasePattern):
         data = data[table_name]
         low, high = self._get_data(data)
         is_datetime = self._get_is_datetime(metadata, table_name)
-        if is_datetime and data[self._high_column_name].dtypes == 'O':
+        if is_datetime and data[self._low_column_name].dtypes == 'O':
             low_format = self._get_datetime_format(metadata, table_name, self._low_column_name)
             high_format = self._get_datetime_format(metadata, table_name, self._high_column_name)
             low = cast_to_datetime64(low, low_format)
--- a/sdv/cag/inequality.py
+++ b/sdv/cag/inequality.py
@@ -2,6 +2,7 @@
 
 import numpy as np
 import pandas as pd
+from pandas.api.types import is_object_dtype
 
 from sdv._utils import _convert_to_timedelta, _create_unique_name
 from sdv.cag._errors import PatternNotMetError
@@ -127,7 +128,7 @@ class Inequality(BasePattern):
         data = data[table_name]
         low, high = self._get_data(data)
         is_datetime = self._get_is_datetime(metadata, table_name)
-        if is_datetime and data[self._low_column_name].dtypes == 'O':
+        if is_datetime and is_object_dtype(data[self._low_column_name]):
             low_format = self._get_datetime_format(metadata, table_name, self._low_column_name)
             high_format = self._get_datetime_format(metadata, table_name, self._high_column_name)
             low = cast_to_datetime64(low, low_format)
@@ -285,7 +286,7 @@ class Inequality(BasePattern):
             diff_column = _convert_to_timedelta(diff_column)
 
         low = table_data[self._low_column_name].to_numpy()
-        if self._is_datetime and self._dtype == 'O':
+        if self._is_datetime and is_object_dtype(self._dtype):
             low = cast_to_datetime64(low)
 
         table_data[self._high_column_name] = pd.Series(diff_column + low).astype(self._dtype)
@@ -310,7 +311,7 @@ class Inequality(BasePattern):
         is_valid = _get_is_valid_dict(data, table_name)
         table_data = data[table_name]
         low, high = self._get_data(table_data)
-        if self._is_datetime and self._dtype == 'O':
+        if self._is_datetime and is_object_dtype(self._dtype):
             low = cast_to_datetime64(low, self._low_datetime_format)
             high = cast_to_datetime64(high, self._high_datetime_format)
 
--- a/sdv/cag/range.py
+++ b/sdv/cag/range.py
@@ -90,6 +90,12 @@ class Range(BasePattern):
         self._high_column_name = high_column_name
         self._low_diff_column_name = f'{self._low_column_name}#{self._middle_column_name}'
         self._high_diff_column_name = f'{self._middle_column_name}#{self._high_column_name}'
+        joined_columns = '#'.join([
+            self._low_column_name,
+            self._middle_column_name,
+            self._high_column_name,
+        ])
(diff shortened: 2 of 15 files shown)


sdv/cag/inequality.py
```diff
@@ -127,7 +127,7 @@ def _validate_pattern_with_data(self, data, metadata):
         data = data[table_name]
         low, high = self._get_data(data)
         is_datetime = self._get_is_datetime(metadata, table_name)
-        if is_datetime and data[self._high_column_name].dtypes == 'O':
+        if is_datetime and data[self._low_column_name].dtypes == 'O':
```
gsheni, line 130:
```suggestion
        from pandas.api.types import is_object_dtype
        if is_datetime and is_object_dtype(data[self._low_column_name].dtypes):
```

sdv/cag/range.py
```diff
@@ -140,21 +143,23 @@ def _validate_pattern_with_metadata(self, metadata):
...
-        high = data[self._high_column_name].to_numpy()
-        return low, mid, high
-
     def _get_valid_table_data(self, table_data):
-        low, mid, high = self._get_data(table_data)
+        low = table_data[self._low_column_name]
+        mid = table_data[self._middle_column_name]
+        high = table_data[self._high_column_name]
+
+        if self._is_datetime and self._dtype == 'O':
```
gsheni, line 151:
```suggestion
        if self._is_datetime and is_object_dtype(self._dtype):
```

sdv/single_table/base.py
```diff
@@ -1335,14 +1336,29 @@ def _sample_with_conditions(
...
+                try:
+                    transformed_condition = self._transform_helper(condition_df)
+                    transformed_condition = self._data_processor.transform(
+                        transformed_condition, is_condition=True
+                    )
+                except PatternNotMetError:
+                    raise PatternNotMetError(
+                        'Provided conditions are not valid for the given constraints.'
+                    )
+                except Exception:
```
gsheni, line 1349: Is it possible to catch a more specific exception?

sdv/single_table/base.py
```diff
@@ -752,6 +752,7 @@ def add_cag(self, patterns):
             try:
                 self.metadata = pattern.get_updated_metadata(self.metadata)
                 self._chained_patterns.append(pattern)
+                self._fitted = False
```
frances-h, line 755: Discovered this while upgrading some tests, `preprocess` was erroring after a fitted synthesizer added a new constraint because the constraint wasn't fitted. By setting `_fitted` to False here, we force refitting the constraints during `preprocess`.

tests/integration/multi_table/test_hma.py
```diff
@@ -311,37 +296,6 @@ def test_hma_custom_constraint_2_tables(self):
         assert all(sampled['child']['numerical_col_2'] > 1)
         assert not all(sampled['child']['numerical_col'] > 1)
 
-    def test_hma_custom_constraint_loaded_from_file(self):
```
frances-h, line 314: Test no longer relevant

sdv/single_table/base.py
frances-h: Not really, we basically want to try to transform the conditions using the constraints but since we only have a subset of the columns it's likely to crash with either a key error or indexing error from pandas when we try to access a column that doesn't exist. We're only guaranteed to get a PatternNotMetError if (1) all necessary columns are present for data validation and (2) data validation fails.

frances-h: If it fails to transform it's fine since we just fall back to reject sampling for the constraints.

gsheni: Gotcha

sdv/single_table/base.py
```diff
@@ -1335,14 +1336,29 @@ def _sample_with_conditions(
...
-                transformed_condition = self._data_processor.transform(
-                    condition_df, is_condition=True
-                )
-            except ConstraintsNotMetError as error:
-                raise ConstraintsNotMetError(
-                    'Provided conditions are not valid for the given constraints.'
-                ) from error
+            if hasattr(self, '_chained_patterns'):
+                try:
+                    transformed_condition = self._transform_helper(condition_df)
```
amontanez24, line 1341: just marking that this will have collisions with this PR https://github.com/sdv-dev/SDV/pull/2480/files

sdv/multi_table/base.py
```diff
@@ -169,20 +170,33 @@ def add_cag(self, patterns):
...
+            added_patterns.append(pattern)
 
         self.metadata = metadata
-        self.patterns += patterns
+        self.patterns += added_patterns
+        self._constraints_fitted = False
         self._initialize_models()
 
     def get_cag(self):
         """Get a list of constraint-augmented generation patterns applied to the synthesizer."""
```
pvk-developer, line 187: We should say that this returns a copy.

sdv/single_table/base.py
```diff
@@ -1342,14 +1343,29 @@ def _sample_with_conditions(
...
             condition_df = dataframe.iloc[0].to_frame().T
-            try:
-                transformed_condition = self._data_processor.transform(
-                    condition_df, is_condition=True
-                )
-            except ConstraintsNotMetError as error:
-                raise ConstraintsNotMetError(
-                    'Provided conditions are not valid for the given constraints.'
-                ) from error
+            if hasattr(self, '_chained_patterns'):
```
pvk-developer, line 1346: Can we move this to its won function like:
```python
def _transform_chained_constraints(self, condition_df, ...):
    ... # your code
    return transformed_condition
```

pvk-developer (approved):
Just a some minor comments but looks good. If there is time to address them it would be good if not we can always file an issue to modularize the sampling process of SDV. I think that would be something to think about.

gsheni (approved):

PR #2531, opened 2025-05-15
Enable single-table constraint reject sampling with multi-table synthesizers

### LINKED ISSUE #2512: Enable single-table constraint reject sampling with multi-table synthesizers
### Problem Description
The new constraint framework with multi-table synthesizers does not support overlapping constraints that must fall back to reject sampling. 

### Expected behavior
Constraints should now have a flag set indicating whether they act on multiple tables or a single table. All public constraints should be set indicating they only act on a single table.

When adding a constraint to a multi-table synthesizer, any multi-table constraints must be added first. These will be set on the multi-table synthesizer, matching the current functionality.

However, once a user adds a constraint that is flagged for single-table use, that constraint should instead be added to the **single-table synthesizer** for the table. Once a single table constraint has been added in this way, no new multi-table constraints may be added. If a user tries to add a multi-table constraint after this point, we should error with: `SynthesizerInputError: Cannot apply multi-table constraint after single-table constraint has been applied`.

### Additional context

This should be merged to the `feature/single-table-CAG` branch.

R-Palazzo:
Resolve #2512
CU-86b4zc0aj

### DIFF
--- a/sdv/cag/base.py
+++ b/sdv/cag/base.py
@@ -174,12 +174,14 @@ class BasePattern:
         """
         data = self._convert_data_to_dictionary(data, self.metadata, copy=True)
         reverse_transformed = self._reverse_transform(data)
+        missing_columns = []
         for table_name, table in reverse_transformed.items():
             required_columns = self._original_data_columns[table_name]
-            # Add missing columns not sampled yet (foreign key in HMA for instance)
+            # Temporarily add missing columns not sampled yet
             for col in required_columns:
                 if col not in table.columns:
                     table[col] = 0
+                    missing_columns.append(col)
 
             table = table[self._original_data_columns[table_name]]
             try:
@@ -188,6 +190,7 @@ class BasePattern:
                 # iterate over the columns and cast individually
                 self._table_as_type_by_col(reverse_transformed, table, table_name)
 
+            table = table.drop(columns=missing_columns)
         if self._single_table:
             return reverse_transformed[self._table_name]
 
--- a/sdv/cag/base.py
+++ b/sdv/cag/base.py
@@ -174,23 +174,20 @@ class BasePattern:
         """
         data = self._convert_data_to_dictionary(data, self.metadata, copy=True)
         reverse_transformed = self._reverse_transform(data)
-        missing_columns = []
         for table_name, table in reverse_transformed.items():
-            required_columns = self._original_data_columns[table_name]
-            # Temporarily add missing columns not sampled yet
-            for col in required_columns:
-                if col not in table.columns:
-                    table[col] = 0
-                    missing_columns.append(col)
-
-            table = table[self._original_data_columns[table_name]]
+            valid_columns = [
+                column
+                for column in self._original_data_columns[table_name]
+                if column in table.columns
+            ]
+            dtypes = {col: self._dtypes[table_name][col] for col in valid_columns}
+            table = table[valid_columns]
             try:
-                reverse_transformed[table_name] = table.astype(self._dtypes[table_name])
+                reverse_transformed[table_name] = table.astype(dtypes)
             except pd.errors.IntCastingNaNError:
                 # iterate over the columns and cast individually
                 self._table_as_type_by_col(reverse_transformed, table, table_name)
 
-            table = table.drop(columns=missing_columns)
         if self._single_table:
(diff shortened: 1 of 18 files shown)


sdv/cag/base.py
```diff
@@ -175,6 +175,12 @@ def reverse_transform(self, data):
         data = self._convert_data_to_dictionary(data, self.metadata, copy=True)
         reverse_transformed = self._reverse_transform(data)
         for table_name, table in reverse_transformed.items():
+            required_columns = self._original_data_columns[table_name]
+            # Add missing columns not sampled yet (foreign key in HMA for instance)
+            for col in required_columns:
+                if col not in table.columns:
+                    table[col] = 0
```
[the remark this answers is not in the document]
frances-h: Instead of adding the dummy column, can we drop the missing column from the order list? Or add in the dummy column, reorder the table columns, then drop the dummy columns. I don't think we should be modifying the data here beyond changing the column order, since it could have an unintended consequence somewhere further down the line and this would be pretty difficult to debug.

sdv/multi_table/base.py
```diff
@@ -375,6 +414,7 @@ def validate(self, data):
         errors = []
         constraints_errors = []
         metadata = self._original_metadata
+        data = {table: df.copy(deep=True) for table, df in data.items()}
```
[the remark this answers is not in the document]
frances-h: Where in validate is the data being modified? I wouldn't expect any of these calls to make in-place modifications to the data dict.

I think it's probably better in general to avoid in-place updates, and if we have to make a copy doing it as close to the in-place modification as possible. This should help prevent unexpected data modification if the method ever gets reused.

sdv/cag/base.py
```diff
@@ -174,14 +174,23 @@ def reverse_transform(self, data):
...
         data = self._convert_data_to_dictionary(data, self.metadata, copy=True)
         reverse_transformed = self._reverse_transform(data)
+        missing_columns = []
         for table_name, table in reverse_transformed.items():
+            required_columns = self._original_data_columns[table_name]
+            # Temporarily add missing columns not sampled yet
+            for col in required_columns:
+                if col not in table.columns:
+                    table[col] = 0
+                    missing_columns.append(col)
```
amontanez24, line 184: why do we need to add this if we're just removing it after trying to cast? Can't we just skip those columns when casting?

sdv/multi_table/base.py
```diff
@@ -162,6 +163,29 @@ def set_address_columns(self, table_name, column_names, anonymization_level='ful
...
+    def _detect_single_table_cag(self, patterns):
+        """Filter out single table CAG patterns from the list of patterns.
+
+        Args:
+            patterns (list):
+                A list of CAG patterns to filter.
+        """
+        idx_single_table = None
+        seen_single_table = False
+        for idx, pattern in enumerate(patterns):
```
amontanez24, line 175: I know you're going to fix this in the end but might as well call them constraints

sdv/multi_table/base.py
```diff
@@ -162,6 +163,29 @@ def set_address_columns(self, table_name, column_names, anonymization_level='ful
...
+            if self._has_seen_single_table_constraint and pattern._is_single_table is False:
+                raise SynthesizerInputError(
+                    'Cannot apply multi-table constraint after single-table constraint has '
+                    'been applied.'
+                )
+
+            if pattern._is_single_table and not seen_single_table:
+                self._has_seen_single_table_constraint = True
+                seen_single_table = True
```
amontanez24, line 184: why do we need both of these?

sdv/cag/base.py
R-Palazzo: Yes done in [202e8fb](https://github.com/sdv-dev/SDV/pull/2531/commits/202e8fb25c0dcf53d7c97c1198b2f0109d35e5b5)

sdv/multi_table/base.py
R-Palazzo: I updated it in [69e0b63](https://github.com/sdv-dev/SDV/pull/2531/commits/69e0b6356458e286a1aa5385e87d8cc399df2567) so we keep the class attribute only.

frances-h (approved):

amontanez24 (approved):

PR #2537, opened 2025-05-16
Add an API for removing a column from the Metadata

### LINKED ISSUE #2526: Add an API for removing a column from the Metadata
### Problem Description
When writing a custom constraint, I may be supplying a `transform` function that modifies the original data (so as to meet the constraint logic). This `transform` function is allowed to do anything to the data -- such as adding or removing columns. Therefore after performing the transformation, the resulting data does not necessarily match the original metadata.

With the new CAG framework, I am required to specify how the metadata will be changing. In order to do that, I have to provide exactly what the updated metadata will look like. Currently, this requires me to convert my metadata to a dictionary, make updates to the dictionary, and then convert it back to the metadata.

This is what it looks like when I have to delete a column from the metadata:
```python
metadata_dict = metadata.to_dict()
del metadata_dict['tables']['table']['columns']['MY_COLUMN_NAME']
metadata = Metadata.load_from_dict(metadata_dict)
```

It will be helpful if I am able to programmatically do this via the Metadata API instead.

### Expected behavior
Create a new function in the Metadata API called `remove_column`. It will have the following parameters:
- (required) `column_name`: A string with the name of the column to remove
- `table_name`: A string with the name of the table. _This is required if you have multiple tables._

```python
metadata.remove_column(column_name='MY_COLUMN_NAME')
```

As a result of calling this function, the following should change in the metadata:
- The column name should be removed from the metadata
- The column name should also be removed if it appears anywhere else in the metadata. This includes:
    - It if appears as a key (primary key, alternate key, sequence key) within the table definition, remove it
    - If it appears as a sequence index of the table, remove it
    - If it appears as a primary or foreign key within the relationships part of the metadata, remove the relationship
    - If it appears in a column relationship, remove the relationship

amontanez24:
resolves #2526

### DIFF
--- a/sdv/metadata/metadata.py
+++ b/sdv/metadata/metadata.py
@@ -377,6 +377,7 @@ class Metadata(MultiTableMetadata):
 
         table_metadata = self.tables[table_name]
         table_metadata._validate_column_exists(column_name)
+
         # Remove relationships
         self._remove_matching_relationships(
             column_name, ['parent_primary_key', 'child_foreign_key']
--- a/sdv/metadata/metadata.py
+++ b/sdv/metadata/metadata.py
@@ -324,6 +324,16 @@ class Metadata(MultiTableMetadata):
         table_name = self._handle_table_name(table_name)
         super().add_column(table_name, column_name, **kwargs)
 
+    def _remove_matching_relationships(self, element, keys):
+        """Remove relationships where the element matches the keys to check."""
+        updated_relationships = []
+        for relationship in self.relationships:
+            matching_keys = [relationship[key] for key in keys]
+            if element not in matching_keys:
+                updated_relationships.append(relationship)
+
+        self.relationships = updated_relationships
+
     def remove_table(self, table_name):
         """Remove a table from the metadata.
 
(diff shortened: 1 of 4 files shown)


sdv/metadata/metadata.py
```diff
@@ -337,18 +347,63 @@ def remove_table(self, table_name):
...
+
+        # Remove keys and special columns
+        if table_metadata.primary_key == column_name:
+            table_metadata.remove_primary_key()
+
+        if column_name in table_metadata.alternate_keys:
+            table_metadata.alternate_keys.remove(column_name)
+
+        if column_name == table_metadata.sequence_key:
+            table_metadata.set_sequence_key(None)
```
pvk-developer, line 400: How come we don't have `remove_sequence_key` ?

pvk-developer (approved):

sdv/metadata/metadata.py
amontanez24: `set_sequence_key` let me set it as None while `set_sequence_index` did not which is why I added the method. I could add it though. What do you think?

pvk-developer: I wouldn't add anything out of the scope of this issue because this changes user facing api. I was just wondering that we have all the other 'remove's but not this one.

frances-h (approved):

PR #2547, opened 2025-05-20
Detect numerical columns post-preprocessing for PARSynthesizer

### LINKED ISSUE #2482: PARSynthesizer is not aware of the sdtypes produced after pre-processing
### Environment Details

* SDV version: 1.20.0
* Python version: 3.11

### Error Description
Given how PARSynthesizer's algorithm works, I know that the fitting time for numerical data can be faster than the fitting time for _high cardinality categorical data_ (eg. a categorical column with 1000s of possible values).

Based on this info, I tried to update the transformer to my PARSynthesizer in the hopes of pre-converting my high cardinality categorical column into numerical data. This should save me some time because when the data is all numerical, PARSynthesizer can be much faster at modeling the data.

But in practice, it seems like PARSynthesizer is not aware that column is being converted from numerical to categorical. As a result, the PARSynthesizer is treating my now-numerical column as categorical, which is taking _much_ longer than before and sometimes crashing my RAM.

### Steps to reproduce

```python
import time
import pandas as pd
import numpy as np

from sdv.sequential import PARSyntheizer
from rdt.transformers.categorical import UniformEncoder

# create some data and metadata with a high-cardinality categorical column
data = pd.DataFrame(data={
    'sequence_key': ['key-' + str(int(i/100)) for i in range(5000)],
    'column': np.random.choice(['value-' + str(i) for i in range(100)], size=5000)})

    'tables': {
        'table': {
            'sequence_key': 'sequence_key',
            'columns': {
                'sequence_key': { 'sdtype': 'id' },
                'column': { 'sdtype': 'categorical'}}}}})

# update the columns pre-processing to UniformEncoder to save the time
synthesizer = PARSynthesizer(metadata)
    'column': UniformEncoder()
})

# observe that this now takes a very long time and sometimes crashes the RAM
start = time.time()
synthesizer.fit(data)
end = time.time()

print('Fit time:', round(end-start, 2), 'seconds')
```

For reference:
1. The fit time for the above code took 10+ minutes
2. If I do not do `update_transformers` (keeping the column as categorical), then the fit time takes 28 seconds
3. If I preprocess the column into a numerical column myself (outside of the PARSynthesizer), then the fit time is further reduced to 20 seconds

I would expect that the time for (1) and (3) should be the same.

fealho:
CU-86b4tc2jj, Resolve #2482.


sdv-team:
Task linked: [CU-86b4tc2jj SDV - PARSynthesizer is not aware of the sdtypes produced after pre-processing #2482](https://app.clickup.com/t/86b4tc2jj)

sdv/sequential/par.py
```diff
@@ -423,11 +423,16 @@ def _fit_sequence_columns(self, timeseries_data):
             kind = dtype.kind
             if kind in ('i', 'f'):
                 data_type = 'continuous'
-                # Check if metadata overrides this data type
-                if self.metadata.columns.get(field, {}).get('sdtype', None) == 'categorical':
-                    data_type = 'categorical'
+                sdtype = self.metadata.columns.get(field, {}).get('sdtype', None)
+                if sdtype == 'categorical':  # Check if metadata overrides this data type
+                    transformers = self._data_processor._hyper_transformer.field_transformers
+                    transformer = transformers.get(field)
```
pvk-developer, line 429:
```suggestion
                    transformers = self.get_transformers()
                    transformer = transformers.get(field)
```

sdv/sequential/par.py
```diff
@@ -423,11 +423,16 @@ def _fit_sequence_columns(self, timeseries_data):
             kind = dtype.kind
             if kind in ('i', 'f'):
                 data_type = 'continuous'
-                # Check if metadata overrides this data type
-                if self.metadata.columns.get(field, {}).get('sdtype', None) == 'categorical':
-                    data_type = 'categorical'
+                sdtype = self.metadata.columns.get(field, {}).get('sdtype', None)
```
pvk-developer, line 426: Should we raise an issue to update `Par` to use the new unified metadata ?

pvk-developer (approved):
Minor suggestion but otherwise looks good.

sdv/sequential/par.py
fealho: @amontanez24 What do you think?

gsheni (approved):

PR #2569, opened 2025-06-04
Update python set up step in workflows to use latest python version

### LINKED ISSUE #2281: Update python setup step in workflows to use latest python version
### Problem Description

In some of our workflows, we use a hardcoded Python version. This means that over time, we have to manually update it.

### Expected behavior
It would be nice to update the python set-up step with something like below so we're always up-to-date
```yml
 - name: Set up latest Python version
      uses: actions/setup-python@v5
      with:
        python-version: 'latest'
```

### Additional context
Here is a list of the worfklows to update:
- static_code_analysis.yml
- release_notes.yml
- lint.yml
- docs.yml
- dependency_checker.yml

frances-h:
Resolve #2281

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
-    - name: Set up lastest Python
+    - name: Set up latest Python
       uses: actions/setup-python@v5
       with:
         python-version-file: 'pyproject.toml'


.github/workflows/docs.yml
```diff
@@ -8,10 +8,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
-    - name: Python
+    - name: Set up lastest Python
```
gsheni:
```suggestion
    - name: Set up latest Python
```

gsheni (approved):

PR #2572, opened 2025-06-05
Metadata should not pass validation if there are unrecognized keys

### LINKED ISSUE #2548: Metadata should not pass validation if there are unrecognized keys
### Environment Details
* SDV version: 1.21.0 (latest)
* Python version: 3.11

### Error Description
In some cases, I may have metadata that has extra keys in a table's definition. This may be left as an artifact from using older versions of SDV, following external tutorials that are not up-to-date, or using AI coding software that makes mistakes. Currently, the `metadata.validate` function allows such extra artifacts to exists in the metadata, even though they don't confirm to the metadata specification. 

I expect that the `metadata.validate()` would fail if there are unrecognized keys from a table's metadata definition. Otherwise, this can cause confusion about what the extra artifact actually does.

### Steps to reproduce
In this example, we'll use the demo data but we'll add an extra, unrecognized key to the `guests` table's definition. Observe that the `metadata.validate()` function still passes (and even the modeling/sampling still work.)

```python

    modality='multi_table',
    dataset_name='fake_hotels'
)

# add an extra, invalid key to the guests table's metadata
metadata_dict = metadata.to_dict()
metadata_dict['tables']['guests']['invalid_key'] = {
    'value1': True,
    'value2': False
}
new_metadata = Metadata.load_from_dict(metadata_dict)

# observe that validation still passes
new_metadata.validate()
new_metadata.validate_data(data)
```

R-Palazzo:
Resolve #2548
CU-86b530nju


sdv-team:
Task linked: [CU-86b530nju SDV - Metadata should not pass validation if there are unrecognized keys #2548](https://app.clickup.com/t/86b530nju)

fealho (approved):

amontanez24 (approved):

Issue #1376, opened 2023-04-14
PytestConfigWarning: Unknown config option: collect_ignore


amontanez24:
### Problem Description

When running pytest we get this warning:
```bash
...\sdv-310\lib\site-packages\_pytest\config\__init__.py:1302: PytestConfigWarning: Unknown config option: collect_ignore

    self._warn_or_fail_if_strict(f"Unknown config option: {key}\n")
```
It seems to be because pytest removed the `collect_ignore` option that we use here:
https://github.com/sdv-dev/SDV/blob/33f9574af1a1bd5482cf1dc82e42baa0cd625ac8/setup.cfg#L49-L50

We should remove this and replace with
```
[tool:pytest]
addopts = --ignore=setup.py
```

PR #2578, opened 2025-06-13
Allow is_valid to be called on unfit constraints

frances-h:
CU-86b586xqm

Adds metadata as an optional parameter to `is_valid` (and a required parameter to the constraint method `_is_valid`) so that the method can be called on an unfit constraint when running DayZ.

### DIFF
--- a/sdv/cag/_utils.py
+++ b/sdv/cag/_utils.py
@@ -106,6 +106,11 @@ def _get_is_valid_dict(data, table_name):
         dict:
             Dictionary of table names to boolean Series of True values.
     """
+    if table_name is None:
+        return {
+            table: pd.Series(True, index=table_data.index) for table, table_data in data.items()
+        }
+
     return {
         table: pd.Series(True, index=table_data.index)
         for table, table_data in data.items()
--- a/sdv/cag/programmable_constraint.py
+++ b/sdv/cag/programmable_constraint.py
@@ -187,7 +187,7 @@ class ProgrammableConstraintHarness(BaseConstraint):
 
         return reverse_transformed
 
-    def _is_valid(self, data):
+    def _is_valid(self, data, metadata):
         if self._is_single_table:
             data = data[self._table_name]
 
--- a/tests/integration/cag/test_programmable_constraint.py
+++ b/tests/integration/cag/test_programmable_constraint.py
@@ -99,6 +99,7 @@ def single_table_programmable_constraint():
             self.table_name = table_name
             self._joint_column = '#'.join(self.column_names)
             self._combinations = None
+            self._fitted = False
 
         def _get_single_table_name(self, metadata):
             # Have to define this so that we can re-use existing methods on the constraint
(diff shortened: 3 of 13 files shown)


sdv-team:
Task linked: [CU-86b586xqm SDV-Enterprise - Add ability to add constraints to single table DayZSynthesizer #1376](https://app.clickup.com/t/86b586xqm)

sdv/cag/_utils.py
```diff
@@ -106,6 +106,11 @@ def _get_is_valid_dict(data, table_name):
         dict:
             Dictionary of table names to boolean Series of True values.
     """
+    if table_name is None:
+        return {
```
amontanez24: can we combine these and just add 
```python
if table != table_name or table_name is None
``` 
to the condition?

amontanez24 (approved):
One suggestion but looks good!

sdv/cag/programmable_constraint.py
```diff
@@ -187,13 +187,14 @@ def _reverse_transform(self, data):
...
         if self._is_single_table:
-            data = data[self._table_name]
+            table_name = self._get_single_table_name(metadata)
+            data = data[table_name]
 
         is_valid = self.programmable_constraint.is_valid(data)
 
         if self._is_single_table:
-            return {self._table_name: is_valid}
+            return {table_name: is_valid}
```
R-Palazzo, line 198: Here, should we update the if condition to be the same as in the base cag class:
`if isinstance(data, pd.DataFrame) or self._single_table:`

tests/integration/cag/test_programmable_constraint.py
```diff
@@ -99,6 +99,7 @@ def __init__(self, column_names, table_name):
             self.table_name = table_name
             self._joint_column = '#'.join(self.column_names)
             self._combinations = None
+            self._fitted = False
```
R-Palazzo, line 102: Why do we need this?

sdv/cag/programmable_constraint.py
frances-h: No, this is a different check - the `_is_single_table` attribute on the harness controls whether we expect a single dataframe or a dataframe dict from the programmable constraint (based on whether it's a `SingleTableProgrammableConstraint` or a `ProgrammableConstraint`).

tests/integration/cag/test_programmable_constraint.py
frances-h: We need it so we can use the `FixedCombinations` methods on this class. The `_is_valid` method for `FixedCombinations` now checks whether or not the constraint has been fit, so we need it set on this class for the method to actually work without erroring.

sdv/cag/_utils.py
frances-h: It is currently technically possible to have multi-table metadata with `None` as a table name, so I kept them separate for that edge case. I'm not sure if that's intended or not.

amontanez24: I guess my thinking was if it's None, you add all tables. If it's not None, you only add the tables that don't match the table name. So effectively it's the same as an or right? You can leave them separate, I'm not sure if it was intentional or not but I assume it was

PR #2584, opened 2025-06-18
Add support for conditional sampling null values

### LINKED ISSUE #2581: Add support for conditional sampling null values
### Problem Description
Currently, if a user passes a condition containing a null value to `sample_with_conditions`, we raise an error stating that null values are not supported with conditional sampling. We'd like to add support for conditional sampling null values so that we aren't restricting users abilities to conditionally sample their data.

### Expected behavior
Allow null values to be passed in to `sample_with_conditions`. If a null value is provided, we should:
1. If the column uses `'from_column'` or a similar option to generate null values, we should ensure that the null flag column is correctly generated and passed down to the model for conditional sampling.
2. During sampling, if the data is not already null force it to be null so that the validation check passes (similar to what we currently do with PII values).

We should not have to check the transformer to conditionally sample null values. Additionally, it is ok to force a value to be null even if the synthesizer would not normally ever generate a null value for the column (e.g. the data did not originally contain null values, the transformer does not add missing values, etc.).

### Additional context
It is likely that we already have support for creating the flag column and passing it to the model in the `from_column` case, but we should still validate that it is working as expected.

frances-h:
CU-86b5c0e6q
Resolve #2581

### DIFF
--- a/sdv/single_table/base.py
+++ b/sdv/single_table/base.py
@@ -1324,13 +1324,6 @@ class BaseSingleTableSynthesizer(BaseSynthesizer):
                         f'Use a column name that was present in the original data.'
                     )
 
-    @staticmethod
-    def _raise_condition_with_nans():
-        raise SynthesizerInputError(
-            'Missing values are not yet supported for conditional sampling. '
-            'Please include only non-null values in your Condition objects.'
-        )
-
     def _validate_conditions(self, conditions):
         """Validate the user-passed conditions."""
         for condition_dataframe in conditions:


sdv-team:
Task linked: [CU-86b5c0e6q SDV - Add support for conditional sampling null values #2581](https://app.clickup.com/t/86b5c0e6q)

sdv/single_table/base.py
```diff
@@ -797,7 +797,9 @@ def _filter_conditions(sampled, conditions, float_rtol):
         """
         for column, value in conditions.items():
             column_values = sampled[column]
-            if column_values.dtype.kind == 'f':
+            if pd.isna(value):
+                sampled = sampled[column_values.isna()]
+            elif column_values.dtype.kind == 'f':
```
gsheni:
```suggestion
            from pandas.api.types import is_float_dtype
            
            if pd.isna(value):
                sampled = sampled[column_values.isna()]
            elif is_float_dtype(column_values.dtype):
```

gsheni (approved):

sdv/single_table/base.py
```diff
@@ -1332,8 +1335,6 @@ def _validate_conditions(self, conditions):
         """Validate the user-passed conditions."""
         for condition_dataframe in conditions:
             self._validate_conditions_unseen_columns(condition_dataframe)
-            if condition_dataframe.isna().any().any():
-                self._raise_condition_with_nans()
```
R-Palazzo, line 1336: Should we get rid of this method?:
`self._raise_condition_with_nans()`

R-Palazzo (approved):

PR #2587, opened 2025-06-23
Improve CAG validation error message

### LINKED ISSUE #2496: Improve CAG validation error message
### Problem Description
The new CAG constraints have less descriptive validation errors than the existing constraints. For the old Inequality constraint, passing in invalid data would result in a nice error message with specific details about what rows are invalid:
```
Constraint: 
ConstraintsNotMetError: 
Data is not valid for the 'Inequality' constraint:
  checkout_date checkin_date
0   29 Dec 2020  27 Dec 2020
1   02 Jan 2021  30 Dec 2020
2   18 Sep 2020  17 Sep 2020
3   31 Dec 2020  28 Dec 2020
4   08 Jan 2021  05 Apr 2020
+495 more
```

The new CAG constraints have a similar error, however there are much fewer details provided:
```PatternNotMetError: The inequality requirement is not met for row indices: [0, 1, 2, 3, 4, +495 more]```

### Expected behavior
We should try to improve the CAG constraint error messages so they contain a similar level of detail to the original constraint error methods.

Instead of just showing the indices of the invalid rows, we should show what specific values from each row are invalid. This will make it easier for users to debug what about their data is invalid according to the CAG constraint.

### PR DESCRIPTION
**Pull request by R-Palazzo**
Resolve #2496
CU-86b4v53zt

### CODE REVIEW

**Comment by sdv-team** [comment]
Task linked: [CU-86b4v53zt SDV - Improve CAG validation error message #2496](https://app.clickup.com/t/86b4v53zt)

File: sdv/cag/fixed_increments.py [hunk 1]
```diff
@@ -112,12 +111,9 @@ def _validate_constraint_with_data(self, data, metadata):
...
         if not valid.all():
-            invalid_rows_str = _get_invalid_rows(valid)
-            raise ConstraintNotMetError(
-                'The fixed increments requirement has not been met because the data is not '
-                f"evenly divisible by '{self.increment_value}' for row indices: "
-                f'[{invalid_rows_str}]'
-            )
+            table_name = self._get_single_table_name(metadata)
+            invalid_rows = data[table_name].loc[~valid, [self.column_name]]
+            self._format_error_message_constraint(invalid_rows, table_name)
```
frances-h: Could we leave the `raise ConstraintNotMetError` here and have `_format_error_message_constraint` just return the formatted string? I think it's clearer to have the error message raised here than have it hidden in the helper method. Alternatively, including `raise` in the method name to make it clear that this helper raises the error.

rwedge (approved):

PR #2591, opened 2025-06-26
Fix: Synthesizer failing for timezone aware data

### LINKED ISSUE #2589: Unable to sample synthetic data when using timezone extraction
### Environment Details
_This is a blocking bug that was found when testing an upcoming version of SDV Enterprise._
* SDV-Enterprise version: 0.29.1.dev0 (zip file used during bug hunt)

### Error Description
In SDV Enterprise, we are adding a new feature that allows the UnixTimestampEncoder to extract timezone information and model it. This is useful if the original data contains mixed timezones (eg. some timezones are EST and some are PST); the synthesizer should be able to learn the proportions of each timezone as well as correlations between the timezone and other columns.

However in practice, I'm unable to apply the timezone extraction-based transformers to my synthesizer. The synthesizer errors during sampling with:
```
AttributeError: Can only use .dt accessor with datetimelike values
```

### Steps to reproduce
```python

import random
import pandas as pd
import numpy as np

# create data and metadata with mixed timezones
data, metadata = download_demo('single_table', 'fake_hotel_guests')

def random_timezone(row):
    if pd.isna(row):
        return row
    return f"{row} 00:00:00 {np.random.choice(['+00:00', '+01:00', '-05:00', '-08:00', '+03:00'])}"

metadata.update_columns(['checkin_date', 'checkout_date'], datetime_format='%d %b %Y %H:%M:%S %z')

np.random.seed(42)
data['checkin_date'] = data['checkin_date'].apply(random_timezone)
np.random.seed(42)
data['checkout_date'] = data['checkout_date'].apply(random_timezone)

# create a synthesizer and manually update its transformers to extract timezones

synth.update_transformers({
    'checkin_date': UnixTimestampEncoder(datetime_format='%d %b %Y %H:%M:%S %z', enforce_min_max_values=True, extract_timezone=True),
    'checkout_date': UnixTimestampEncoder(datetime_format='%d %b %Y %H:%M:%S %z', enforce_min_max_values=True, extract_timezone=True),
})

synth.fit(data)

# observe that there is an error in sampling
synth.sample(10)
```
```
AttributeError: Can only use .dt accessor with datetimelike values
```
[stack_trace.txt](https://github.com/user-attachments/files/20909085/stack_trace.txt)

### Additional Context
Chatting with @pvk-developer and @amontanez24, we think the root cause is that  [the format_data function](https://github.com/sdv-dev/SDV/blob/cef58d1382ea2dc0df5e79147bb707c669065d27/sdv/data_processing/datetime_formatter.py#L52) does not support timezones.

pvk-developer:
Resolves #2589 
CU-86b5frbme

### DIFF
--- a/tests/unit/data_processing/test_datetime_formatter.py
+++ b/tests/unit/data_processing/test_datetime_formatter.py
@@ -126,17 +126,3 @@ class TestDatetimeFormatter:
         # Assert
         expected = pd.Series(['2024/06/01', '2025/06/01'])
         pd.testing.assert_series_equal(result, expected)
-
-    def test_format_data_fallback_value_error_then_success(self):
-        """Test fallback parsing when initial format fails but fallback succeeds."""
-        # Setup
-        formatter = DatetimeFormatter('%Y/%m/%d')
-        formatter._dtype = 'O'
-        column = pd.Series(['2023-01-01', '2023-01-02'])  # doesn't match %Y/%m/%d
-
-        # Run
-        result = formatter.format_data(column)
-
-        # Assert
-        expected = pd.Series(['2023/01/01', '2023/01/02'])
-        pd.testing.assert_series_equal(result, expected)
--- a/tests/unit/data_processing/test_datetime_formatter.py
+++ b/tests/unit/data_processing/test_datetime_formatter.py
@@ -86,3 +86,43 @@ class TestDatetimeFormatter:
             '20220929111311000000',
         ])
         pd.testing.assert_series_equal(result, expected)
+
+    def test_format_data_datetime_dtype_returns_early(self):
+        """Test that if dtype is datetime64 and matches column, it returns early."""
+        # Setup
+        formatter = DatetimeFormatter('%Y-%m-%d')
+        column = pd.Series(pd.to_datetime(['2021-01-01', '2022-01-01']))
+        formatter._dtype = column.dtype
+
+        # Run
+        result = formatter.format_data(column)
+
+        # Assert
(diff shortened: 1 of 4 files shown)


sdv-team:
Task linked: [CU-86b5frbme SDV - Unable to sample synthetic data when using timezone extraction #2589](https://app.clickup.com/t/86b5frbme)

tests/unit/data_processing/test_datetime_formatter.py
```diff
@@ -86,3 +86,57 @@ def test_format_datetime_does_not_match_format(self):
...
+        formatter = DatetimeFormatter('%Y/%m/%d')
+        formatter._dtype = 'O'
+        column = pd.Series(['2023-01-01', '2023-01-02'])  # doesn't match %Y/%m/%d
+
+        # Run
+        result = formatter.format_data(column)
+
+        # Assert
+        expected = pd.Series(['2023/01/01', '2023/01/02'])
+        pd.testing.assert_series_equal(result, expected)
```
R-Palazzo: Out of curiosity, what is the difference between these two tests? I see they don't have the same column values, but I wasn't sure which case each one is checking.

R-Palazzo (approved):
LGTM!

Is the code for the issue running now?

tests/unit/data_processing/test_datetime_formatter.py
pvk-developer: Good catch; I saw that the codecov here complained nad and wrote one more test but I see that i already had it!

PR #2596, opened 2025-06-27
Refactor `_create_config` method

### LINKED ISSUE #2593: Refactor _create_config method
### Problem Description

_As an engineer, I'd like to be able to easily creates subclasses that enhance methods without needing to copy over large amounts of code._

In the DataProcessor we have [this function](https://github.com/sdv-dev/SDV/blob/3eac48ed1967f98d09211ea957fc75ad22145f14/sdv/data_processing/data_processor.py#L349-L355) that is responsible for creating the HyperTransformer config that a synthesizer will use. The issue is that we want to make subclasses that do slightly different logic for certain parts of the config, but not the whole thing. Since everything is in one big function, the only option is to copy most of that code over to the new class and change the small portion that needs changing. This can lead to bugs since if any other part of the function needs to be fixed, it will now need to be fixed in two places.

https://github.com/sdv-dev/SDV/blob/3eac48ed1967f98d09211ea957fc75ad22145f14/sdv/data_processing/data_processor.py#L349-L355

### Expected behavior

- Refactor the `_create_config` method into smaller methods that are each responsible for a portion of the config
    - A good place to start would be pulling each if statement block into its own function that can be overridden.

### Additional context

- One example use case for this is we may want to make a class that changes which datetime transformers get assigned. This is something that we should make sure will be easily supported after the refactor.

fealho:
CU-86b5g9jnm, Resolve #2593.


sdv-team:
Task linked: [CU-86b5g9jnm SDV - Refactor _create_config method #2593](https://app.clickup.com/t/86b5g9jnm)

sdv/data_processing/data_processor.py
```diff
@@ -346,115 +346,234 @@ def _get_transformer_instance(self, sdtype, column_metadata):
...
+                Column name.
+            data (pandas.DataFrame):
+                The input data.
+
+        Returns:
+            tuple:
+                (sdtype, transformer)
+        """
+        column_metadata = self.metadata.columns.get(column)
+        sdtype = column_metadata.get('sdtype')
```
pvk-developer: This is repeated in every `handle` function, maybe we can pass it from the `_get_column_config`.

sdv/data_processing/data_processor.py
```diff
@@ -346,115 +346,234 @@ def _get_transformer_instance(self, sdtype, column_metadata):
...
+            if is_numeric:
+                function_name = 'random_int'
+                column_dtype = str(column_dtype).lower()
+                function_kwargs = {'min': 0, 'max': 16777216}
+                if 'int8' in column_dtype:
+                    function_kwargs['max'] = 127
+                elif 'int16' in column_dtype:
+                    function_kwargs['max'] = 32767
+            else:
+                function_kwargs = {'text': 'sdv-id-??????'}
```
pvk-developer: This can be its own function.

sdv/data_processing/data_processor.py
```diff
@@ -346,115 +346,234 @@ def _get_transformer_instance(self, sdtype, column_metadata):
...
+            transformer = self._get_transformer_instance(sdtype, column_metadata)
+        else:
+            transformer = self.create_anonymized_transformer(
+                sdtype=sdtype,
+                column_metadata=column_metadata,
+                cardinality_rule='unique',
+                locales=self._locales,
+            )
+
+        return sdtype, transformer
```
pvk-developer: The main reason for refactoring this code was to split it into smaller parts so it's easier to override the default values (based on the issue context).  

The idea is to have separate functions like:
- `_get_datetime_transformer`
- `_get_categorical_transformer`  
...and so on.

Right now, I would still need to override `_get_transformer_instance`, which feels the same as before—still too general.

By breaking it down, we’ll be able to change specific defaults more easily. For example, we can replace `_get_datetime_transformer` to return the premium default from `SDV-Enterprise`, without affecting other parts.

```python

    def get_transformer(self, sdtype):
        transformer_map = {
            'a': self._get_transformer_a,
            'b': self._get_transformer_b,
            # add more as needed
        }

        transformer_fn = transformer_map.get(sdtype)
        if transformer_fn:
            return transformer_fn()
        else:
            raise ValueError(f"Unsupported sdtype: {sdtype}")
```

sdv/data_processing/data_processor.py
```diff
@@ -346,115 +346,234 @@ def _get_transformer_instance(self, sdtype, column_metadata):
 
         return deepcopy(transformer)
 
-    def _create_config(self, data):
-        sdtypes = {}
-        transformers = {}
+    def _handle_multi_column(self, column, data):
```
amontanez24: can we rename all of these to `_get_{descriptor}_config`? So this would be `_get_multi_column_config`

sdv/data_processing/data_processor.py
amontanez24: we may want to be able to override the functionality in these two sections

amontanez24 (commented):
This is looking good! I agree with @pvk-developer . We need to be able to easily change which transformer gets selected for each condition. Some of these functions could probably be broken up even more

sdv/data_processing/data_processor.py
```diff
@@ -323,138 +323,433 @@ def create_regex_generator(self, column_name, sdtype, column_metadata, is_numeri
 
         return transformer
 
+    def _get_other_transformer(self, kwargs, transformer):
```
pvk-developer: I think this should be `_get_transformer_instance` mainly because this returns the actual instance of the transformer that will be used, right ?

sdv/data_processing/data_processor.py
```diff
@@ -323,138 +323,433 @@ def create_regex_generator(self, column_name, sdtype, column_metadata, is_numeri
...
         transformer = self._transformers_by_sdtype[sdtype]
-        if isinstance(transformer, AnonymizedFaker):
-            is_lexify = transformer.function_name == 'lexify'
-            is_baseprovider = transformer.provider_name == 'BaseProvider'
-            if is_lexify and is_baseprovider:  # Default settings
-                return self.create_anonymized_transformer(
-                    sdtype, column_metadata, None, self._locales
-                )
+        if transformer is None:
+            return None
```
pvk-developer, line 424: This is for when we set the categorical transformer to `None` in CTGAN ?

sdv/data_processing/data_processor.py
```diff
@@ -323,138 +323,433 @@ def create_regex_generator(self, column_name, sdtype, column_metadata, is_numeri
...
+            'categorical': self._get_categorical_transformer,
+            'boolean': self._get_boolean_transformer,
+        }
+        transformer_fn = transformer_map.get(sdtype)
+        if transformer_fn:
+            return transformer_fn(kwargs, transformer)
+
+        return self._get_other_transformer(kwargs, transformer)
+
     def _get_transformer_instance(self, sdtype, column_metadata):
```
pvk-developer, line 421: This can be renamed to `_get_transformer_for_sdtype` or something similar to be coherent with the other suggested change.

pvk-developer (commented):
Looks great! I just have two more nits, but I think that this now the expected (basically we can change things from enterprise by overriding the kwargs in the smaller functions and that should work).

**(Maybe out of scope)**: One other improvement that could be done in this pull request is to move all the functions that do not use `self` into a different file, something like `rdt_utilities.py` or similar. This is just my thought so first check with @amontanez24 if this is something doable or in the scope of this issue.

sdv/data_processing/data_processor.py
fealho: The transformer input is already an instance, so the name could be a bit confusing. I renamed it to `_get_transformer_with_kwargs` instead.

sdv/data_processing/data_processor.py
fealho: I'm not sure. This logic was already in the code so I kept it.

sdv/data_processing/data_processor.py
fealho: I kept this as is and renamed the other method.

sdv/data_processing/data_processor.py
```diff
@@ -323,53 +323,261 @@ def create_regex_generator(self, column_name, sdtype, column_metadata, is_numeri
...
-    def _get_transformer_instance(self, sdtype, column_metadata):
-        transformer = self._transformers_by_sdtype[sdtype]
-        if isinstance(transformer, AnonymizedFaker):
-            is_lexify = transformer.function_name == 'lexify'
-            is_baseprovider = transformer.provider_name == 'BaseProvider'
-            if is_lexify and is_baseprovider:  # Default settings
-                return self.create_anonymized_transformer(
-                    sdtype, column_metadata, None, self._locales
-                )
+    def _get_transformer_with_kwargs(self, kwargs, transformer):
```
amontanez24: can we replace kwargs with params since kwargs is usually used like **kwargs?

pvk-developer (approved):

sdv/data_processing/data_processor.py
pvk-developer: Yeah I think is because of `CTGAN` :)

PR #2612, opened 2025-07-14
[Fix] auto_assign_transformers crashes if the datetime column is represented as Timestamps 

### LINKED ISSUE #2594: auto_assign_transformers crashes if the datetime column is represented as Timestamps
### Environment Details

* SDV version: 1.23.0

### Error Description

If I create a column of Timestamps with dtype object, then the synthesizer crashes when I attempt to run `auto_assign_transformers`. This seems to be because pd.to_datetime fails to run on a column with mixed timezones and that is being run here.
https://github.com/sdv-dev/SDV/blob/3eac48ed1967f98d09211ea957fc75ad22145f14/sdv/_utils.py#L125-L142

### Steps to reproduce

```python
from sdv.cag import Inequality

data, metadata = download_demo('single_table', 'fake_hotel_guests')
timezones = [
    "+03:00",
    "-03:00",
    "+08:00",
    "+05:00",
    "-05:00"
]
import numpy as np
from zoneinfo import ZoneInfo
import random
import pandas as pd
from datetime import timedelta, timezone

random_timezones = np.random.choice(timezones, size=len(data))

def parse_offset(offset_str):
    # Remove colon for easier parsing (e.g. +0300)
    offset_str_clean = offset_str.replace(":", "")

    sign = 1 if offset_str_clean.startswith("+") else -1
    hours = int(offset_str_clean[1:3])
    minutes = int(offset_str_clean[3:5])

    delta = timedelta(hours=hours, minutes=minutes)
    return timezone(sign * delta)
# Apply timezone conversion row by row
data['checkin_date'] = data.apply(
    lambda row: pd.to_datetime(row['checkin_date']).replace(tzinfo=parse_offset(random.choice(timezones))),
    axis=1
)
data['checkout_date'] = data.apply(
    lambda row: pd.to_datetime(row['checkout_date']).replace(tzinfo=parse_offset(random.choice(timezones))),
    axis=1
)
new_metadata = Metadata.detect_from_dataframe(data)
date_constraint = Inequality(
    low_column_name='checkin_date',
    high_column_name='checkout_date'
)
new_metadata.update_columns(['checkin_date', 'checkout_date'], sdtype='datetime', datetime_format='%Y-%m-%d %H:%M:%S%z')
synth = GaussianCopulaSynthesizer(new_metadata)
```

gsheni:
- Fixes #2594 
- CU-86b5g9twz


sdv-team:
Task linked: [CU-86b5g9twz SDV - auto_assign_transformers crashes if the datetime column is represented as Timestamps #2594](https://app.clickup.com/t/86b5g9twz)

sdv/_utils.py
```diff
@@ -136,7 +136,9 @@ def _validate_datetime_format(column, datetime_format):
             A set of values from the column that do not match the datetime format.
     """
     pandas_datetime_format = datetime_format.replace('%-', '%')
-    datetime_column = pd.to_datetime(column, errors='coerce', format=pandas_datetime_format)
+    datetime_column = pd.to_datetime(
+        column, errors='coerce', format=pandas_datetime_format, utc=True
```
amontanez24: What if we decide to allow different timezones in the future?

gsheni: This function will work with a column with mixed timezones, the same timezones, and no timezones.
I changed the `utc=True` only if the `datetime_format` contains `%z`. 
One thing to keep in mind is that this function does not modify the user data. It just check that datetime columns contains values that matches the given `datetime_format`.

tests/unit/test__utils.py
```diff
@@ -850,10 +853,325 @@ def test__check_regex_format(mock_strings_from_regex):
...
+
+        # Run
+        invalid_values = _validate_datetime_format(column, bad_format)
+
+        # Assert
+        if Version(pd.__version__) >= Version('2.0.0'):
+            assert len(invalid_values) == 3
+            assert sorted(invalid_values) == sorted(column.tolist()[:3])
+        else:
+            pytest.skip('Datetimes are only parsed with a consistent format with pandas >= 2.0.0')
```
amontanez24, line 1001: are you able to skip this at the beginning of the test?

amontanez24 (commented):
One comment but otherwise looks good!

tests/unit/test__utils.py
gsheni: It's better to skip at the end of the test because it verifies that `_validate_datetime_format` does not crash.

amontanez24 (approved):

PR #2614, opened 2025-07-17
Fix PARSynthesizer sampling for all-null column in the input data

### LINKED ISSUE #2473: PARSynthesizer crashes during sample if there was an all-null column in the input data
### Environment Details
* SDV version: 1.20.0
* Python version: 3.11

### Error Description
If my input data (training data) has a column that is all-null, then the PARSynthesizer crashes during `sample` with a `KeyError`. This happens whenever the sdtype of this column indicates that it should be statistically modeled -- for eg. `numerical` or `categorical`.

### Steps to reproduce

```python
import numpy as np
import pandas as pd

data = pd.DataFrame(data={
    'sequence_key': ['sequence-' + str(int(i/5)) for i in range(100)],
    'numerical_col': np.random.randint(low=0, high=100, size=100),
    'categorical_col': np.random.choice(['A', 'B', 'C'], size=100),
    'all_null_col': [np.nan]*100
})

    'tables': {
        'table': {
            'columns': {
                'sequence_key': { 'sdtype': 'id' },
                'numerical_col': { 'sdtype': 'numerical' },
                'categorical_col': { 'sdtype': 'categorical' },
                'all_null_col': { 'sdtype': 'numerical' }
            },
            'sequence_key': 'sequence_key'
        }
    }
})

synthesizer = PARSynthesizer(metadata, epochs=1)
synthesizer.fit(data)
synthesizer.sample(num_sequences=2)
```
```
KeyError: "['all_null_col'] not in index"
```
Stack trace provided below.

[stack_trace.txt](https://github.com/user-attachments/files/19813258/stack_trace.txt)

### Expected Behavior
I expect that -- no matter what the sdtype of the original column -- PARSynthesizer should be able to understand that the column is all-null and therefore produce synthetic data where the column is all-null.

This is the behavior for all our other single-table and multi-table synthesizers.

### Workaround
Until we fix this bug, the recommended workaround is to replace the missing values in the column with a static, numerical value such as `0`.
 
```python
data['all_null_col'] = data['all_null_col'].fillna(0)
```

Now PARSynthesizer will no longer crash. The synthetic data column will contain a static value (`0`) that you can change back to null if you'd like.

fealho:
CU-86b4pwwjk, Resolve #2473.

### DIFF
--- a/sdv/sequential/par.py
+++ b/sdv/sequential/par.py
@@ -548,6 +548,12 @@ class PARSynthesizer(LossValuesMixin, MissingModuleMixin, BaseSynthesizer):
             output.append(sequence_df)
 
         output = pd.concat([df.dropna(axis=1, how='all') for df in output])
+
+        # Handle all-null columns that might have been dropped
+        missing_columns = set(self._output_columns) - set(output.columns)
+        for col in missing_columns:
+            output[col] = np.nan
+
         output = output[self._output_columns].reset_index(drop=True)
 
         return output
--- a/tests/integration/sequential/test_par.py
+++ b/tests/integration/sequential/test_par.py
@@ -534,3 +534,41 @@ def test___init___without_torch(mock_import_error):
     # Run and Assert
     with pytest.raises(ModuleNotFoundError, match=msg):
         PARSynthesizer(metadata)
+
+
+def test_par_with_all_null_column():
+    """Test that the method handles all-null columns correctly."""
+    # Setup
+    data = pd.DataFrame(
+        data={
+            'sequence_key': ['sequence-' + str(int(i / 5)) for i in range(100)],
+            'numerical_col': np.random.randint(low=0, high=100, size=100),
+            'categorical_col': np.random.choice(['A', 'B', 'C'], size=100),
+            'all_null_col': [np.nan] * 100,
+        }
+    )
+
+    metadata = Metadata.load_from_dict({
+        'tables': {
+            'table': {
+                'columns': {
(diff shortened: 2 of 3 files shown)


sdv-team:
Task linked: [CU-86b4pwwjk SDV - PARSynthesizer crashes during sample if there was an all-null column in the input data #2473](https://app.clickup.com/t/86b4pwwjk)

sdv/sequential/par.py
```diff
@@ -548,6 +548,12 @@ def _sample_from_par(self, context, sequence_length=None):
             output.append(sequence_df)
 
         output = pd.concat([df.dropna(axis=1, how='all') for df in output])
+
+        # Handle all-null columns that might have been dropped
+        missing_columns = set(self._output_columns) - set(output.columns)
```
amontanez24, line 553: isn't this just undoing the line above?

frances-h (approved):

sdv/sequential/par.py
fealho: Correct. However, combining the lines as `output = pd.concat(output)` can raise this warning:

```
The behavior of DataFrame concatenation with empty or all-NA entries is deprecated. In a future version, this will no longer exclude empty or all-NA columns when determining the result dtypes. To retain the old behavior, exclude the relevant entries before the concat operation.
```

This code removes the nans before concatenation to avoid this warning.

amontanez24: ah, that's annoying

PR #2627, opened 2025-07-28
Add test showing PARSynthesizer doesn't create different sequences with the same sequence key (uniqueness is enforced)

### LINKED ISSUE #2467: PARSynthesizer can create different sequences with the same sequence key (uniqueness is not enforced)
### Environment Details
* SDV version: 1.20.0
* Python version: 3.11

### Error Description
When using the PARSynthesizer, I expect that the synthetic sequence key value can be used to uniquely identify each sequence. However in practice, I find that PARSynthesizer can create multiple sequences with the _same_ sequence key -- because it is not enforcing uniqueness in the way it generates the sequence keys. 

This is especially a problem if I supply a regex format that does not allow for many unique values. After a while, PARSynthesizer starts repeating values in new sequences.

### Steps to reproduce
```python
import pandas as pd
import numpy as np

# create data with 'sequence_key' as my sequence key
data = pd.DataFrame(data={
    'sequence_key': ['seq-0']*5 + ['seq-1']*2 + ['seq-2']*3,
    'column_A': np.random.randint(low=0, high=10, size=10),
    'column_B': np.random.choice(['Yes', 'No', 'Maybe'], size=10)
})

# for the sequence key, supply a regex that doesn't offer that many possibilities
    'tables': {
        'table': {
            'sequence_key': 'sequence_key',
            'columns': {
                'sequence_key': {'sdtype': 'id', 'regex_format': 'seq-[0-9]'},
                'column_A': { 'sdtype': 'numerical' },
                'column_B': { 'sdtype': 'categorical'}
            }
        }
    }
})

# fit and sample more sequences than the regex allows for
synthesizer = PARSynthesizer(metadata, epochs=1)
synthesizer.fit(data)
synthetic_data = synthesizer.sample(num_sequences=20)
```

Observe that after the synthesizer exhausts the 10 possible regex values, it then starts repeating those values. I would expect that it would instead venture to generate new values (just like what happens for single table synthesizers such as GaussianCopula).

### Root Cause
It seems like the transformer that is assigned to my sequence does not have the `cardinality_rule` set. 

```python
synthesizer._context_synthesizer.get_transformers()['sequence_key']
```
```
RegexGenerator(regex_format='seq-[0-9]', generation_order='random')
```
I expect that it should have `cardinality_rule='unique'` set, as the sequence keys cannot repeat.

fealho:
CU-86b4nm299, Resolve #2467.

### DIFF
--- a/tests/integration/sequential/test_par.py
+++ b/tests/integration/sequential/test_par.py
@@ -605,3 +605,5 @@ def test_par_unique_sequence_key_with_regex():
 
     # Assert
     assert sample['sequence_key'].nunique() == 20
+    transformer = synthesizer._context_synthesizer.get_transformers()['sequence_key']
+    transformer.cardinality_rule == 'unique'
--- a/tests/integration/sequential/test_par.py
+++ b/tests/integration/sequential/test_par.py
@@ -572,3 +572,36 @@ def test_par_with_all_null_column():
     assert 'all_null_col' in result.columns
     assert result['all_null_col'].isna().all()
     assert len(result) > 0
+
+
+def test_par_unique_sequence_key_with_regex():
+    """Test that the method handles unique sequence key with regex correctly."""
+    # Setup
+    data = pd.DataFrame(
+        data={
+            'sequence_key': ['seq-0'] * 5 + ['seq-1'] * 2 + ['seq-2'] * 3,
+            'column_A': np.random.randint(low=0, high=10, size=10),
+            'column_B': np.random.choice(['Yes', 'No', 'Maybe'], size=10),
+        }
+    )
+
+    metadata = Metadata.load_from_dict({
+        'tables': {
+            'table': {
+                'sequence_key': 'sequence_key',
+                'columns': {
+                    'sequence_key': {'sdtype': 'id', 'regex_format': 'seq-[0-9]'},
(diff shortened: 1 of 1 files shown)


sdv-team:
Task linked: [CU-86b4nm299 SDV - PARSynthesizer can create different sequences with the same sequence key (uniqueness is not enforced) #2467](https://app.clickup.com/t/86b4nm299)

tests/integration/sequential/test_par.py
```diff
@@ -572,3 +572,36 @@ def test_par_with_all_null_column():
...
+        }
+    })
+
+    # Run
+    synthesizer = PARSynthesizer(metadata, epochs=1)
+    synthesizer.fit(data)
+    sample = synthesizer.sample(num_sequences=20)
+
+    # Assert
+    assert sample['sequence_key'].nunique() == 20
```
pvk-developer, line 607:
```suggestion
    assert sample['sequence_key'].nunique() == 20
    transformer = synthesizer._context_synthesizer.get_transformers()['sequence_key']
    transformer.cardinality_rule == 'unique'
```

pvk-developer (approved):
Just a small suggestion for the test to be more precise, otherwise LGTM!

PR #2657, opened 2025-08-27
Ensure OneHotEncoding creates floating point numbers

### LINKED ISSUE #2649: [OneHotEncoding Constraint] For higher quality, ensure the model creates floating point numbers
### Problem Description
The [OneHotEncoding constraint](https://docs.sdv.dev/sdv/concepts/constraint-augmented-generation-cag/predefined-constraints/onehotencoding) is meant to be used when you already have several one-hot encoded columns in your data. The constraint guarantees that the one-hot scheme will also be carried over into the synthetic data -- i.e. within each row, exactly 1 of the columns should contain a `1` while the rest should be `0`.

The constraint is supposed to work by allowing the underlying ML model to generate floating point values for each column within the [0, 1] scale. When creating synthetic data, the constraint is supposed to select the _largest_ of these values, and mark it as a `1` (and the rest as 0).

Unfortunately with the way it's being set up right now, the underlying ML model will _not_ generate floating point values within the [0, 1] scale; rather, it will explicitly generate 0 or 1. The constraint is then forced to randomly select a column to be `1`, which can lead to worse data quality overall. As an example of this, see #2631.

### Expected behavior
The constraint should be set up in a way that the underlying ML model will have to generate floating point values within the [0, 1] scale.

1. If the original 0/1 columns are marked as sdtype `categorical`, the constraint should delete them/create new columns that are floating point values, marked as sdtype `numerical` in the update metadata.
2. Even when the columns are `numerical`, we need to make sure the underlying ML model doesn't learn to round the numbers back to 0/1.
    - The constraint could slightly perturb (noise) the data to have floats with many decimal points. For example a `0` can be perturbed to `0.0018291` and a `1` can be perturbed to `0.9938177`
    - The constraint should _not_ be touching the underlying transformer's settings (that's breaking an abstraction we set)

fealho:
CU-86b695bp4, Resolve #2649.

### DIFF
--- a/sdv/cag/one_hot_encoding.py
+++ b/sdv/cag/one_hot_encoding.py
@@ -11,7 +11,7 @@ from sdv.cag._utils import (
 )
 from sdv.cag.base import BaseConstraint
 
-EPSILON = np.finfo(np.float32).eps
+EPSILON = float(np.finfo(np.float32).eps)
 
 
 class OneHotEncoding(BaseConstraint):
@@ -111,11 +111,7 @@ class OneHotEncoding(BaseConstraint):
         """
         table_name = self._get_single_table_name(self.metadata)
         one_hot_data = data[table_name][self._column_names]
-        one_hot_data = np.where(
-            one_hot_data == 0,
-            np.random.uniform(0, EPSILON, size=one_hot_data.shape),
-            1 - np.random.uniform(0, EPSILON, size=one_hot_data.shape),
-        )
+        one_hot_data = np.where(one_hot_data == 0, EPSILON, 1 - EPSILON)
         data[table_name][self._column_names] = one_hot_data
 
         return data
--- a/sdv/cag/one_hot_encoding.py
+++ b/sdv/cag/one_hot_encoding.py
@@ -1,5 +1,7 @@
 """One Hot Encoding constraint."""
 
+from copy import deepcopy
+
 import numpy as np
 
 from sdv.cag._errors import ConstraintNotMetError
@@ -92,6 +94,7 @@ class OneHotEncoding(BaseConstraint):
 
     def _get_updated_metadata(self, metadata):
         table_name = self._get_single_table_name(metadata)
+        metadata = deepcopy(metadata)
         for column in self._column_names:
             if metadata.tables[table_name].columns[column]['sdtype'] == 'categorical':
                 metadata.tables[table_name].columns[column]['sdtype'] = 'numerical'
--- a/sdv/cag/one_hot_encoding.py
+++ b/sdv/cag/one_hot_encoding.py
@@ -96,7 +96,7 @@ class OneHotEncoding(BaseConstraint):
         table_name = self._get_single_table_name(metadata)
         metadata = deepcopy(metadata)
         for column in self._column_names:
-            if metadata.tables[table_name].columns[column]['sdtype'] == 'categorical':
+            if metadata.tables[table_name].columns[column]['sdtype'] in ['categorical', 'boolean']:
                 metadata.tables[table_name].columns[column]['sdtype'] = 'numerical'
 
         return metadata
--- a/sdv/cag/one_hot_encoding.py
+++ b/sdv/cag/one_hot_encoding.py
@@ -11,6 +11,8 @@ from sdv.cag._utils import (
 )
 from sdv.cag.base import BaseConstraint
 
+EPSILON = np.finfo(np.float32).eps
+
 
 class OneHotEncoding(BaseConstraint):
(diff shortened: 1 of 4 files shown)


sdv-team:
Task linked: [CU-86b695bp4 SDV - [OneHotEncoding Constraint] For higher quality, ensure the model creates floating point numbers #2649](https://app.clickup.com/t/86b695bp4)

sdv/cag/one_hot_encoding.py
```diff
@@ -88,6 +90,14 @@ def _fit(self, data, metadata):
         """
         pass
 
+    def _get_updated_metadata(self, metadata):
+        table_name = self._get_single_table_name(metadata)
+        for column in self._column_names:
+            if metadata.tables[table_name].columns[column]['sdtype'] == 'categorical':
+                metadata.tables[table_name].columns[column]['sdtype'] = 'numerical'
```
frances-h, line 100: Instead of modifying the input metadata directly, we should either deepcopy it before doing this assignment or convert it to a dict, update the metadata, then return a new metadata object.

sdv/cag/one_hot_encoding.py
```diff
@@ -88,6 +90,14 @@ def _fit(self, data, metadata):
...
+
```
R-Palazzo: Sometimes both categorical and boolean follow the same logic, so should we apply the same logic to boolean here? @frances-h

frances-h (changes requested):

sdv/cag/one_hot_encoding.py
```diff
@@ -99,6 +109,11 @@ def _transform(self, data):
             dict[str, pd.DataFrame]:
                 Transformed data.
         """
+        table_name = self._get_single_table_name(self.metadata)
+        one_hot_data = data[table_name][self._column_names]
+        one_hot_data = np.where(one_hot_data == 0, EPSILON, 1 - EPSILON)
```
R-Palazzo, line 117: Is it better to keep the noise constant? Doing so might cause the ML model to generate just two float values, but that might be okay.

tests/integration/cag/test_one_hot_encoding.py
```diff
@@ -138,3 +139,33 @@ def test_end_to_end_multi_raises(data_multi, metadata_multi):
...
+        constraint = OneHotEncoding(column_names=columns)
+
+        # Run
+        synthesizer.add_constraints([constraint])
+        synthesizer.fit(df)
+        samples = synthesizer.sample(100)
+
+        # Assert
+        for col in columns:
+            assert sorted(samples[col].unique().tolist()) == [0, 1]
```
R-Palazzo, line 172: Could you also check that the columns all add up to 1?

sdv/cag/one_hot_encoding.py
fealho: Good catch

sdv/cag/one_hot_encoding.py
fealho: Making the noise random won't work with PAR, since context columns need to be constant for each sequence.

sdv/cag/one_hot_encoding.py
frances-h: Yeah good catch, we should update boolean columns in a similar way.

frances-h:
> According to the [docs](https://docs.sdv.dev/sdv/concepts/constraint-augmented-generation-cag/predefined-constraints/onehotencoding) the columns have to be `numerical` sdtype. Should we update the metadata validation?

Correct me if I'm wrong, but if we only allow `'numerical'` then this fix isn't even necessary, right? I'm inclined to allow `'boolean'` and `'categorical'` since it's not an inaccurate way to describe the encoding. Does metadata autodetection detect a one hot encoded column as numerical or categorical?

fealho:
> > According to the [docs](https://docs.sdv.dev/sdv/concepts/constraint-augmented-generation-cag/predefined-constraints/onehotencoding) the columns have to be `numerical` sdtype. Should we update the metadata validation?
> 
> Correct me if I'm wrong, but if we only allow `'numerical'` then this fix isn't even necessary, right? I'm inclined to allow `'boolean'` and `'categorical'` since it's not an inaccurate way to describe the encoding. Does metadata autodetection detect a one hot encoded column as numerical or categorical?

1. This PR affects both 'numerical' and 'categorical' sdtypes
2. Metadata autodetection detects integer columns as categorical by default
3. I will add support for booleans as well

frances-h (approved):

R-Palazzo (approved):
LGTM!

Issue #2663, opened 2025-08-29
Add get_source and get_readme functions for demo datasets


amontanez24:
### Problem Description

Certain datasets have licensing information or citations that can be shared. We would like to add functions that allow users to get and view this information.

We are also adding a readme file with a description of the dataset, how it was created, any interesting modifications to source datasets, etc.

### Expected behavior

- Add a function called `get_source`
```python
def get_source(modality, dataset_name, output_filepath=None):
"""
    Args:
        modality (str): The modality of the dataset.
        dataset_name (str): The name of the dataset to get the source information for.
        output_filepath (str): Optional. Where to save the file.
    Returns:
        String that is the contents of the source file. Can return None.
"""
```
```python
from sdv.datasets.demo import get_source
source_text = get_source(
      modality='single_table',
      dataset_name='ing_SEPA_data',
      output_filepath='my_dataset/license.txt'
)
print(source_text)
```
```
License name: Database Contents License (DbCL) v1.0
Source URL: https://www.database-source.com/...

Citation:
@misc{
        stone,
        m._2025,
        title={Australian Football League (AFL) Database},
        url={https://www.kaggle.com/dsv/10710877},
        DOI={10.34740/KAGGLE/DSV/10710877},
        publisher={Kaggle},
        author={Stone,
        M.},
        year={2025}
}
```
- Add a function `get_readme`
```python
def get_readme(modality, dataset_name, output_filepath=None):
"""
    Args:
        modality (str): The modality of the dataset.
        dataset_name (str): The name of the dataset to get the source information for.
        output_filepath (str): Optional. Where to save the file.
    Returns:
        String that is the contents of the readme file. Can return None.
"""
```
  

### Additional context

- Both of these functions may return None as certain datasets might not have these files
- If they exist, these files will exist in the zip as README.txt and SOURCE.txt

PR #2669, opened 2025-09-05
Update the download_demo and get_available_demos functions

### LINKED ISSUE #2661: Update the download_demo  and get_available_demos functions
### Problem Description

_As a user, I want to have easy access to the most up-to-date demo datasets available._

We are currently updating the datasets we use for demo and the location of these datasets. These functions need to be updated to download from these new locations.

### Expected behavior

- Update the [download_demo ](https://github.com/sdv-dev/SDV/blob/bca136e5d653a922a50f2a91cc2f090d73416c80/sdv/datasets/demo.py#L125)function to download from the new `sdv-datasets-public` bucket.
- Update the [get_available_demos](https://github.com/sdv-dev/SDV/blob/bca136e5d653a922a50f2a91cc2f090d73416c80/sdv/datasets/demo.py#L160C5-L160C24) function to list the datasets in the  `sdv-datasets-public` bucket.
    - This version of the function should be much more robust. Our current code expects the items in the bucket to all follow a certain format and doesn't safely handle the situations when they don't. For example, this code will crash if an item in the bucket doesn't have enough items to unpack because it doesn't contain a `/` character in the key. The contents of the bucket are subject to change so we should be intentional on what types of files we are looking for. When listing the contents of the bucket and creating the dataframe that gets returned, we should catch any errors as opposed to failing.  https://github.com/sdv-dev/SDV/blob/bca136e5d653a922a50f2a91cc2f090d73416c80/sdv/datasets/demo.py#L181-L183
 - Create a new error type (`DemoResourceNotFoundError`) to be raised whenever we cannot find the demo dataset or details about it (such as the license, readme, etc.)

### Additional context

- The new structure of the bucket will be as follows. We will now use these yaml files to get the dataset size and number of tables.
```
sdv-datasets-public
|--- single_table
     |--- dataset1.zip
     |--- dataset2.zip
     |--- ...
     |--- meta_info/
          |--- dataset1.yaml
          |--- dataset2.yaml
          |--- ...
|--- multi_table
     |--- dataset3.zip
     |--- dataset4.zip
     |--- …
     |--- meta_info/
          |--- dataset3.yaml
          |--- dataset4.yaml
          |--- ...
|--- sequential
     |--- dataset5.zip
     |--- dataset6.zip
     |--- …
     |--- meta_info/
          |--- dataset5.yaml
          |--- dataset6.yaml
          |--- ...
```
- The structure of the yaml file will look like
```yaml
dataset-name: census
dataset-bucket: sdv-demo-datasets
num-tables: 1
dataset-size-mb: 23.32
source: EXTERNAL
source-url: unknown
```
- Both functions should be very error-proof! If something is not found as expected in the S3 bucket. We should look for specific file/filetypes in the code, and try/catch any operations on files in case the structure ever changes.
    - Within a modality (like single-table), don't assume that it will only contain zip files. If there is a non-zip file, skip over it
    - When looping through the contents of the bucket, if you are looking for a specific file type or name, safely check for it. DO NOT assume every item in the bucket will be of a certain type or name format. The contents of a bucket are subject to change.
    - If there are multiple json files, read each one and choose the first one that has "METADATA_SPEC_VERSION": "V1". This will make it future-proof, in case we decide to update the spec version later.

fealho:
CU-86b6fugf3, Resolve #2661.

### DIFF
--- a/sdv/datasets/demo.py
+++ b/sdv/datasets/demo.py
@@ -4,7 +4,6 @@ import io
 import json
 import logging
 import os
-import warnings
 from collections import defaultdict
 from pathlib import Path
 from zipfile import ZipFile
@@ -12,15 +11,17 @@ from zipfile import ZipFile
 import boto3
 import numpy as np
 import pandas as pd
+import yaml
 from botocore import UNSIGNED
 from botocore.client import Config
 from botocore.exceptions import ClientError
 
+from sdv.errors import DemoResourceNotFoundError
 from sdv.metadata.metadata import Metadata
 
 LOGGER = logging.getLogger(__name__)
-BUCKET = 'sdv-demo-datasets'
-BUCKET_URL = 'https://sdv-demo-datasets.s3.amazonaws.com'
+BUCKET = 'sdv-datasets-public'
+BUCKET_URL = 'https://sdv-datasets-public.s3.amazonaws.com'
 SIGNATURE_VERSION = UNSIGNED
 METADATA_FILENAME = 'metadata.json'
 
@@ -39,27 +40,121 @@ def _validate_output_folder(output_folder_name):
         )
 
 
-def _get_data_from_bucket(object_key):
+def _create_s3_client():
     session = boto3.Session()
-    s3 = session.client('s3', config=Config(signature_version=SIGNATURE_VERSION))
+    return session.client('s3', config=Config(signature_version=SIGNATURE_VERSION))
+
+
+def _get_data_from_bucket(object_key):
+    s3 = _create_s3_client()
     response = s3.get_object(Bucket=BUCKET, Key=object_key)
     return response['Body'].read()
 
 
+def _list_objects(prefix):
+    s3 = _create_s3_client()
+    contents = []
+    continuation_token = None
+    while True:
+        kwargs = {'Bucket': BUCKET, 'Prefix': prefix}
+        if continuation_token:
+            kwargs['ContinuationToken'] = continuation_token
+        response = s3.list_objects_v2(**kwargs)
+        contents.extend(response.get('Contents', []) or [])
+        if response.get('IsTruncated'):
+            continuation_token = response.get('NextContinuationToken')
+        else:
+            break
+    return contents
+
+
+def _find_data_zip_key(contents, dataset_prefix):
+    for item in contents:
+        key = item.get('Key', '')
+        if not key.startswith(dataset_prefix):
+            continue
+        if key.lower().endswith('data.zip'):
+            return key
+    return None
+
+
+def _collect_json_keys(contents, dataset_prefix):
+    json_keys = []
+    for item in contents:
+        key = item.get('Key', '')
+        if not key.startswith(dataset_prefix):
+            continue
+        if key.lower().endswith('.json'):
+            json_keys.append(key)
+    return json_keys
+
+
+def _select_metadata_v1_bytes(json_keys):
+    for key in json_keys:
+        try:
+            candidate = _get_data_from_bucket(key)
+            try:
+                metadict = json.loads(candidate)
+            except Exception:
+                metadict = json.loads(candidate.decode('utf-8'))
+            if isinstance(metadict, dict) and metadict.get('METADATA_SPEC_VERSION') == 'V1':
+                return candidate
+        except Exception:
+            continue
+    return None
+
+
+def _iter_metainfo_yaml_entries(contents, modality):
+    for item in contents:
+        key = item.get('Key') or ''
+        if not key.lower().endswith('metainfo.yaml'):
+            continue
+        parts = key.split('/')
+        if len(parts) < 3 or parts[0] != modality:
+            continue
+        yield parts[1], key
+
+
 def _download(modality, dataset_name):
-    dataset_url = f'{BUCKET_URL}/{modality.upper()}/{dataset_name}.zip'
-    object_key = f'{modality.upper()}/{dataset_name}.zip'
+    dataset_prefix = f'{modality}/{dataset_name}/'
+    dataset_url = f'{BUCKET_URL}/{dataset_prefix}'
     LOGGER.info(f'Downloading dataset {dataset_name} from {dataset_url}')
+
     try:
-        file_content = _get_data_from_bucket(object_key)
-    except ClientError:
-        raise ValueError(
-            f"Invalid dataset name '{dataset_name}'. "
-            'Make sure you have the correct modality for the dataset name or '
-            "use 'get_available_demos' to get a list of demo datasets."
+        contents = _list_objects(dataset_prefix)
+    except ClientError as exc:
+        raise DemoResourceNotFoundError(
+            f"Failed to list resources for dataset '{dataset_name}' in '{modality}'."
+        ) from exc
+    if not contents:
+        raise DemoResourceNotFoundError(
+            f"Dataset '{dataset_name}' under modality '{modality}' was not found."
         )
 
-    return io.BytesIO(file_content)
+    data_zip_key = _find_data_zip_key(contents, dataset_prefix)
+    json_keys = _collect_json_keys(contents, dataset_prefix)
+
+    if not data_zip_key:
+        raise DemoResourceNotFoundError(
+            f"Could not find 'data.zip' for dataset '{dataset_name}' in modality '{modality}'."
+        )
+
+    metadata_bytes = _select_metadata_v1_bytes(json_keys)
+
+    if metadata_bytes is None:
+        raise DemoResourceNotFoundError(
+            f"Could not locate a metadata JSON with METADATA_SPEC_VERSION 'V1' for dataset "
+            f"'{dataset_name}' in modality '{modality}'."
+        )
+
+    try:
+        data_zip_bytes = _get_data_from_bucket(data_zip_key)
+    except ClientError as exc:
+        raise DemoResourceNotFoundError(
+            f"Failed to download 'data.zip' for dataset '{dataset_name}'."
+        ) from exc
+
+    return io.BytesIO(data_zip_bytes), metadata_bytes
 
 
 def _extract_data(bytes_io, output_folder_name):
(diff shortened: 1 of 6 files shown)


sdv-team:
Task linked: [CU-86b6fugf3 SDV - Add get_source and get_readme functions for demo datasets #2663](https://app.clickup.com/t/86b6fugf3)

sdv/datasets/demo.py
```diff
@@ -39,27 +39,113 @@ def _validate_output_folder(output_folder_name):
...
+        if not (key_lower.startswith(prefix_lower) and key_lower.endswith('.json')):
+            continue
+
+        try:
+            raw = _get_data_from_bucket(key)
+            metadict = json.loads(raw)
+            if isinstance(metadict, dict) and metadict.get('METADATA_SPEC_VERSION') == 'V1':
+                return raw
+
+        except Exception:
```
gsheni, line 156: Could we change these blanket exceptions (`except Exception`) to be more specific?

sdv/datasets/demo.py
```diff
@@ -170,23 +269,41 @@ def get_available_demos(modality):
...
+                size_mb = float(size_mb_val) if size_mb_val is not None else np.nan
+            except Exception:
+                size_mb = np.nan
+
+            num_tables_val = info.get('num-tables')
+            try:
+                num_tables = int(num_tables_val)
+            except Exception:
+                try:
+                    num_tables = int(float(num_tables_val))
```
gsheni: double casting?

sdv/datasets/demo.py
```diff
@@ -170,23 +269,41 @@ def get_available_demos(modality):
...
+            tables_info['size_MB'].append(size_mb)
+            tables_info['num_tables'].append(num_tables)
+
+        except Exception:
+            continue
 
     df = pd.DataFrame(tables_info)
-    df['num_tables'] = pd.to_numeric(df['num_tables'])
+    if not df.empty:
+        df['num_tables'] = pd.to_numeric(df['num_tables'], errors='coerce')
```
gsheni: Is this just casting to an integer?

sdv/datasets/demo.py
```diff
@@ -39,27 +39,113 @@ def _validate_output_folder(output_folder_name):
...
+
+    Returns:
+        list[dict]:
+            A list of object summaries.
+    """
+    client = _create_s3_client()
+    contents = []
+    paginator = client.get_paginator('list_objects_v2')
+    for resp in paginator.paginate(Bucket=BUCKET, Prefix=prefix):
+        contents.extend(resp.get('Contents') or [])
```
gsheni:
```suggestion
        contents.extend(resp.get('Contents', []))
```

sdv/datasets/demo.py
```diff
@@ -39,27 +39,113 @@ def _validate_output_folder(output_folder_name):
...
+        dataset_prefix (str):
+            Prefix like 'single_table/dataset/'.
+
+    Returns:
+        str:
+            The key to the data zip if found.
+    """
+    prefix_lower = dataset_prefix.lower()
+    for entry in contents:
+        key = entry.get('Key') or ''
```
gsheni:
```suggestion
        key = entry.get('Key', '')
```

sdv/datasets/demo.py
```diff
@@ -39,27 +39,113 @@ def _validate_output_folder(output_folder_name):
...
+    Scans S3 listing `contents` and, for any JSON file directly under the dataset prefix,
+    downloads and returns its bytes if it contains METADATA_SPEC_VERSION == 'V1'.
+
+    Returns:
+        bytes:
+            The bytes of the first V1 metadata JSON.
+    """
+    prefix_lower = dataset_prefix.lower()
+    for entry in contents:
+        key = entry.get('Key') or ''
```
gsheni:
```suggestion
        key = entry.get('Key', '')
```

sdv/datasets/demo.py
```diff
@@ -170,23 +269,41 @@ def get_available_demos(modality):
...
+            info = yaml.safe_load(raw) or {}
+            name = info.get('dataset-name') or dataset_name
+
+            size_mb_val = info.get('dataset-size-mb')
+            try:
+                size_mb = float(size_mb_val) if size_mb_val is not None else np.nan
+            except Exception:
+                size_mb = np.nan
+
+            num_tables_val = info.get('num-tables')
```
gsheni:
```suggestion
            num_tables_val = info.get('num-tables', np.nan)
```

sdv/datasets/demo.py
```diff
@@ -70,10 +156,13 @@ def _extract_data(bytes_io, output_folder_name):
...
-                os.path.join(output_folder_name, 'metadata_v1.json'),
-                os.path.join(output_folder_name, METADATA_FILENAME),
-            )
+
+            metadata_v1_filepath = os.path.join(output_folder_name, 'metadata_v1.json')
+            if os.path.isfile(metadata_v1_filepath):
+                os.rename(
+                    metadata_v1_filepath,
+                    os.path.join(output_folder_name, METADATA_FILENAME),
+                )
```
amontanez24: I think all of the files in the new bucket should be named `metadata.json`, so I'm not sure if we need to worry about this

sdv/datasets/demo.py
```diff
@@ -149,14 +220,42 @@ def download_demo(modality, dataset_name, output_folder_name=None):
...
+        if len(parts) < 3:
+            continue
+        if parts[0].lower() != modality_lower:
+            continue
+        filename = parts[-1]
+        if filename.lower() != 'metainfo.yaml':
+            continue
+        dataset_name = parts[1]
+        if not dataset_name:
+            continue
```
amontanez24: if a key is formatted `single_table/metainfo.yaml`, this will think that it's valid with a dataset_name of `metainfo.yaml` no?

sdv/datasets/demo.py
```diff
@@ -170,23 +269,41 @@ def get_available_demos(modality):
...
+            tables_info['num_tables'].append(num_tables)
+
+        except Exception:
+            continue
 
     df = pd.DataFrame(tables_info)
-    df['num_tables'] = pd.to_numeric(df['num_tables'])
+    if not df.empty:
+        df['num_tables'] = pd.to_numeric(df['num_tables'], errors='coerce')
+        df['size_MB'] = pd.to_numeric(df['size_MB'], errors='coerce')
```
amontanez24: why is this necessary?

sdv/datasets/demo.py
fealho: I changed 3 of them to be more specific. The 2 remaining I believe should stay as broad Exceptions, the code should not raise regardless of what the error happens.

sdv/datasets/demo.py
fealho: Wouldn't it be skipped in line 245, since it doesn't have 3 parts? Maybe I'm not understanding the question.

sdv/datasets/demo.py
fealho: Yes, otherwise entries like '3.0' would raise an error.

sdv/datasets/demo.py
```diff
@@ -4,23 +4,23 @@
...
-from botocore.exceptions import ClientError
 
+from sdv.errors import DemoResourceNotFoundError
 from sdv.metadata.metadata import Metadata
 
 LOGGER = logging.getLogger(__name__)
-BUCKET = 'sdv-demo-datasets'
-BUCKET_URL = 'https://sdv-demo-datasets.s3.amazonaws.com'
+BUCKET = 'sdv-datasets-public'
+BUCKET_URL = 'https://sdv-datasets-public.s3.amazonaws.com'
```
frances-h: If this URL is always using the bucket name, should we just change this to an f-string so that we don't have to remember to modify it in both places if it ever changes again?

sdv/datasets/demo.py
amontanez24: why does just casting to an int not suffice?

sdv/datasets/demo.py
```diff
@@ -282,16 +272,16 @@ def get_available_demos(modality):
             size_mb_val = info.get('dataset-size-mb')
             try:
                 size_mb = float(size_mb_val) if size_mb_val is not None else np.nan
-            except Exception:
+            except (ValueError, TypeError):
```
amontanez24, line 308: can we add logging for these cases?

sdv/datasets/demo.py
fealho: `int(str(3.0))` raises `ValueError: invalid literal for int() with base 10: '3.0'`

amontanez24: why would it be a string though?

fealho: No idea, I'm just adding support in case that's the format in the file. I can delete it if you think it's unnecessary.

amontanez24: I would say if we're doing it for that reason, we can be more explicit like:
```python
if isinstance(num_tables_val, str):
    try:
        num_tables_val = float(num_tables_val)
    except ValueError as e:
        logger.info('Could not cast num_tables_val)

num_tables = int(num_tables_val)
```
Unless it's always a string right now in which case we can leave as is

sdv/datasets/demo.py
```diff
@@ -39,41 +39,120 @@ def _validate_output_folder(output_folder_name):
...
+    Returns:
+        str:
+            The key to the data zip if found.
+    """
+    prefix_lower = dataset_prefix.lower()
+    for entry in contents:
+        key = entry.get('Key', '')
+        key_lower = key.lower()
+        if key_lower.startswith(prefix_lower) and key_lower.endswith('/data.zip'):
+            return key
```
frances-h: Should this be an exact check for {modality}/{dataset_name}/data.zip instead?

sdv/datasets/demo.py
```diff
@@ -39,41 +39,120 @@ def _validate_output_folder(output_folder_name):
...
+    Returns:
+        bytes:
+            The bytes of the first V1 metadata JSON.
+    """
+    prefix_lower = dataset_prefix.lower()
+    for entry in contents:
+        key = entry.get('Key', '')
+        key_lower = key.lower()
+        if not (key_lower.startswith(prefix_lower) and key_lower.endswith('.json')):
+            continue
```
frances-h: Should this be an exact check for {modality}/{dataset_name}/*.json instead?

sdv/datasets/demo.py
```diff
@@ -39,41 +39,126 @@ def _validate_output_folder(output_folder_name):
...
+    Returns:
+        str:
+            The key to the data zip if found.
+    """
+    prefix_lower = dataset_prefix.lower()
+    for entry in contents:
+        key = entry.get('Key', '')
+        key_lower = key.lower()
+        if key_lower == f'{prefix_lower}data.zip':
+            return key
```
frances-h: Seems like we do something similar to this in 2 other places, `_get_first_v1_metadata_bytes` and `iter_metainfo_yaml_entries`. Should we create one helper function that searches the contents for a given pattern and returns a list of all matches and use that instead?

sdv/datasets/demo.py
```diff
@@ -39,41 +39,153 @@ def _validate_output_folder(output_folder_name):
...
+            The bytes of the first V1 metadata JSON.
+    """
+    prefix_lower = dataset_prefix.lower()
+
+    def is_direct_json_under_prefix(key):
+        key_lower = key.lower()
+        return (
+            key_lower.startswith(prefix_lower)
+            and key_lower.endswith('.json')
+            and key_lower.count('/') == prefix_lower.count('/')
```
amontanez24, line 144: should we check that metadata is also in the name? What if we add a different json in the future

PR #2670, opened 2025-09-05
Allow users to estimate parameters for DayZSynthesizer

### LINKED ISSUE #2666: Allow users to estimate parameters for DayZSynthesizer
### Problem Description
The [DayZSynthesizer](https://docs.sdv.dev/sdv/multi-table-data/modeling/synthesizers/dayzsynthesizer) (both multi-table and single-table) is a useful starting point for creating synthetic data from scratch. Although the DayZSynthesizer is only available for SDV Enterprise users, we at DataCebo will sometimes create DayZ data for some interested users, for things like [estimating performance](https://docs.sdv.dev/sdv/multi-table-data/modeling/performance-estimates).

The problem is that to create realistic data from scratch, DayZ can be customized with additional parameters. These parameters right now can only be added via the functionality of DayZSynthesizer, which is only available to SDV Enterprise customers.

### Expected behavior
Create a class-level function called `DayZSynthesizer.create_parameters` that is accessible in SDV Community. This function should be added for both single- and multi-table synthesizers.

Parameters:
- (required) `data`: The data for which you want to create parameters
   - For the single-table modality, this is the pandas dataframe
   - For the multi-table modality, this is a dictionary that maps each table name to the pandas dataframe
- (required): `metadata`, the Metadata that describes the data
- `output_filename`: A string with the name of the file. The DayZ parameters will be written to the file. This must end in `.json`.

Eg.

```python
from sdv.multi_table import DayZSynthesizer

my_parameters = DayZSynthesizer.create_parameters(
  data=my_data,
  metadata=my_metadata,
  output_filename='dayz_parameters.json'
)
```

At this point, SDV Enterprise users should then be able to set all the parameters at once. SDV Community users can share these parameters with DataCebo (along with the metadata), to get a realistic estimate.

### Parameter Format
The parameter format mirrors metadata format, but it only contains additional parameters that are useful for the DayZ synthesizer.

By table:
- The number of rows

By columns:
- ID/PII columns: The proportion of missing values
- Numerical columns: The proportion of missing values, min value, max value, and number of decimal digits
- Datetime columns: The proportion of missing values, start timestamp and end timestamp
- Categorical columns: The proportion of missing value, a list of all the (non-null) categorical values

By relationship (for multi-table):
- The parent and child table names, and the child foreign key (used to identify the relationship)
- The min and max cardinality

These correspond to the settings that are available in the [DayZ API](https://docs.sdv.dev/sdv/multi-table-data/modeling/synthesizers/dayzsynthesizer#making-the-data-more-realistic).

This example showcases a column of each different sdtype in the desired format (with the desired keywords, etc.).

```python
{
  'DAYZ_SPEC_VERSION': 'V1',
  'tables': {
    'guests': {
      'num_rows': 250,
      'columns': {
        'guest_id': {
          'missing_values_proportion': 0.0
        },
        'city': {
          'category_names': [ 'A', 'B', 'C', 'D', 'E' ],
          'missing_values_proportion': 0.25,
        },
        'age': { # NUMERICAL
          'min_value': 0,
          'max_value': 100,
          'num_decimal_digits': None, # will be added 
        },
        'checkout_date': {
          'start_timestamp': '2000-01-01 00:00:00.000000',
          'end_timestamp': '2022-12-31 23:59:59.999999',
          'missing_values_proportion': 0.0, 
        }
      }
    },
    'hotels': {
      'num_rows': 100,
      'columns': {
        ...
      }
    }
   },
  'relationships': [{
    'child_table_name': 'guests',
    'child_foreign_key': 'hotel_id',
    'min_cardinality': 1,
    'max_cardinality': None
  }],
}
```

R-Palazzo:
Resolve #2666
CU-86b6h4f4w

### DIFF
--- a/tests/integration/multi_table/test_dayz.py
+++ b/tests/integration/multi_table/test_dayz.py
@@ -1,0 +1,105 @@
+"""Integration tests for DayZ parameter detection."""
+
+import pandas as pd
+
+from sdv.datasets.demo import download_demo
+from sdv.multi_table import DayZSynthesizer
+
+
+class TestDayZSynthesizer:
+    def test_create_parameters_end_to_end(self):
+        """Test the `create_parameters` method end to end."""
+        # Setup
+        data, metadata = download_demo(modality='multi_table', dataset_name='fake_hotels')
+
+        # Run
+        parameters = DayZSynthesizer.create_parameters(data, metadata)
+
+        # Assert
+        expected_results = {
+            'DAYZ_SPEC_VERSION': 'V1',
+            'tables': {
+                'guests': {
+                    'num_rows': 658,
+                    'columns': {
+                        'guest_email': {'missing_value_proportion': 0.0},
+                        'hotel_id': {'missing_value_proportion': 0.0},
+                        'has_rewards': {
+                            'category_values': [False, True],
+                            'missing_value_proportion': 0.0,
+                        },
+                        'room_type': {
+                            'category_values': ['BASIC', 'DELUXE', 'SUITE'],
+                            'missing_value_proportion': 0.0,
+                        },
+                        'amenities_fee': {
+                            'num_decimal_digits': 2,
+                            'min_value': 0.0,
+                            'max_value': 46.64,
+                            'missing_value_proportion': 0.07598784194528875,
+                        },
+                        'checkin_date': {
+                            'start_timestamp': pd.Timestamp('2020-01-03 00:00:00'),
+                            'end_timestamp': pd.Timestamp('2021-01-05 00:00:00'),
+                            'missing_value_proportion': 0.0,
+                        },
+                        'checkout_date': {
+                            'start_timestamp': pd.Timestamp('2020-01-04 00:00:00'),
+                            'end_timestamp': pd.Timestamp('2021-01-07 00:00:00'),
+                            'missing_value_proportion': 0.04559270516717325,
+                        },
+                        'room_rate': {
+                            'num_decimal_digits': 2,
+                            'min_value': 48.33,
+                            'max_value': 481.61,
+                            'missing_value_proportion': 0.0,
+                        },
+                        'billing_address': {'missing_value_proportion': 0.0},
+                        'credit_card_number': {'missing_value_proportion': 0.0},
+                    },
+                },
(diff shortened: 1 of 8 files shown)


sdv-team:
Task linked: [CU-86b6h4f4w SDV - Allow users to estimate parameters for DayZSynthesizer #2666](https://app.clickup.com/t/86b6h4f4w)

frances-h (commented):
Looks good! We'll need to update Enterprise as well since we now have a public DayZ class (right now Enterprise overwrites this change).

We might also want to add a default `__init__` method to public that errors out saying that DayZ is not available. Currently trying to instantiate a `DayZSynthesizer` with arguments results in `TypeError: DayZSynthesizer() takes no arguments`.

tests/integration/multi_table/test_dayz.py
```diff
@@ -0,0 +1,105 @@
...
+                        },
+                        'amenities_fee': {
+                            'num_decimal_digits': 2,
+                            'min_value': 0.0,
+                            'max_value': 46.64,
+                            'missing_value_proportion': 0.07598784194528875,
+                        },
+                        'checkin_date': {
+                            'start_timestamp': pd.Timestamp('2020-01-03 00:00:00'),
+                            'end_timestamp': pd.Timestamp('2021-01-05 00:00:00'),
```
frances-h: Shouldn't we expect this to be a string with the correct datetime format instead of a timestamp?

sdv/single_table/_dayz_utils.py
```diff
@@ -0,0 +1,95 @@
...
+            column_parameters[column_name] = {
+                'start_timestamp': start_timestamp,
+                'end_timestamp': end_timestamp,
+            }
+        elif sdtype in ['categorical', 'boolean']:
+            column_parameters[column_name] = {
+                'category_values': data[column_name].dropna().unique().tolist()
+            }
+
+        column_parameters[column_name]['missing_value_proportion'] = (
```
fealho, line 71: The issue uses plural values: `missing_values_proportion`

fealho: The issue mentions missing_values_proportion. Plural values.

PR #2671, opened 2025-09-09
Allow users to validate the DayZ parameters

### LINKED ISSUE #2667: Allow users to validate the DayZ parameters
### Problem Description
Issue #2666 allows users to estimate all the DayZ parameters at once as a Python dictionary. In this issue, we should provide logic that validates the format of those parameters, ensuring that they can be used properly in conjunction with the metadata to create data from scratch.

### Expected behavior
Create a class-level function in SDV Community called `DayZSynthesizer.validate_parameters`. This should be available for both single-table and multi-table DayZ.

Parameters:
- (required) `parameters`: A Python dictionary containing the parameters, whose format is defined in #2666
- (required) `metadata`: A metadata object that corresponds to the parameters

Returns: (None) Raises a `SynthesizerProcessingError` if any of the parameters are not valid.

```python
from sdv.multi_table import DayZSynthesizer

my_parameters = DayZSynthesizer.create_parameters(
  metadata=my_metadata,
  data=my_data,
  output_filename='dayz_parameters.json'
)

DayZSynthesizer.validate_parameters(
  metadata=my_metadata,
  parameters=my_parameters
)
```

### Validation Checks
Unlike the metadata, the parameters do not have to be complete -- that is to say, certain tables, columns, relationships, and parameters are allowed to be removed. So we should only check the values that are there. However, whatever is provided should be consistent with the metadata.

(Prerequisite): First make sure that the metadata itself is valid. If it's not valid, then error out immediately. (We cannot validate parameters against invalid metadata.)

Parameter checks:
- Make sure that all tables, columns, and relationships that are provided in the parameters are actually present in the metadata
- Within each column, make sure that the parameters are valid for the sdtype that is denoted in the metadata
- Make sure that all the parameters are the correct type (eg. `num_rows` should be an integer)
- Make sure that the num rows (per table) align with the cardinality. For example, let's say table A has 100 rows and table B has 200 rows, but the cardinality stipulates that every row of A must be connected to at least 3 rows of B -- that isn't valid!
    - If the num rows for a table isn't provided, assume it's 1000 (default)
    - If the cardinality for a relationship isn't provided, assume a minimum of 0 and maximum of `None` (infinity) 
- Make sure that there are no unexpected, additional keywords present

frances-h:
Resolve #2667 
CU-86b6h4n3f

### DIFF
--- a/sdv/multi_table/dayz.py
+++ b/sdv/multi_table/dayz.py
@@ -18,6 +18,35 @@ RELATIONSHIP_PARAMETER_KEYS = REQUIRED_RELATIONSHIP_KEYS + [
 DEFAULT_NUM_ROWS = 1000
 
 
+def _validate_min_cardinality(relationship):
+    min_cardinality = relationship['min_cardinality']
+    if not isinstance(min_cardinality, int) or min_cardinality < 0:
+        msg = (
+            f"Invalid 'min_cardinality' {min_cardinality}. 'min_cardinality' "
+            'must be an integer greater than or equal to zero.'
+        )
+        raise SynthesizerProcessingError(msg)
+
+
+def _validate_max_cardinality(relationship):
+    max_cardinality = relationship['max_cardinality']
+    if not isinstance(max_cardinality, int) or max_cardinality <= 0:
+        msg = (
+            f"Invalid 'max_cardinality' {max_cardinality}. 'max_cardinality' "
+            'must be an integer greater than zero.'
+        )
+        raise SynthesizerProcessingError(msg)
+
+
+def _validate_cardinality_bounds(relationship):
+    if relationship['min_cardinality'] > relationship['max_cardinality']:
+        msg = (
+            "Invalid cardinality, 'min_cardinality' must be less than or "
+            "equal to 'max_cardinality'."
+        )
+        raise SynthesizerProcessingError(msg)
+
+
 def _validate_relationship_structure(dayz_parameters):
     if not isinstance(dayz_parameters.get('relationships', []), list):
         raise SynthesizerProcessingError("The 'relationships' value must be a list.")
@@ -35,29 +64,11 @@ def _validate_relationship_structure(dayz_parameters):
             raise SynthesizerProcessingError(msg)
 
         if 'min_cardinality' in relationship:
-            min_cardinality = relationship['min_cardinality']
-            if not isinstance(min_cardinality, int) or min_cardinality < 0:
-                msg = (
-                    f"Invalid 'min_cardinality' {min_cardinality}. 'min_cardinality' "
-                    'must be an integer greater than or equal to zero.'
-                )
-                raise SynthesizerProcessingError(msg)
+            _validate_min_cardinality(relationship)
         if 'max_cardinality' in relationship:
-            max_cardinality = relationship['max_cardinality']
-            if not isinstance(max_cardinality, int) or max_cardinality <= 0:
-                msg = (
-                    f"Invalid 'max_cardinality' {max_cardinality}. 'max_cardinality' "
-                    'must be an integer greater than zero.'
-                )
-                raise SynthesizerProcessingError(msg)
-
-        if 'max_cardinality' in relationship and 'min_cardinality' in relationship:
-            if relationship['min_cardinality'] > relationship['max_cardinality']:
-                msg = (
-                    "Invalid cardinality, 'min_cardinality' must be less than or "
-                    "equal to 'max_cardinality'."
-                )
-                raise SynthesizerProcessingError(msg)
+            _validate_max_cardinality(relationship)
+        if 'min_cardinality' in relationship and 'max_cardinality' in relationship:
+            _validate_cardinality_bounds(relationship)
 
 
 def _validate_cardinality(relationship_parameters, parent_num_rows, child_num_rows):
--- a/sdv/multi_table/dayz.py
+++ b/sdv/multi_table/dayz.py
@@ -22,8 +22,8 @@ def _validate_min_cardinality(relationship):
     min_cardinality = relationship['min_cardinality']
     if not isinstance(min_cardinality, int) or min_cardinality < 0:
         msg = (
-            f"Invalid 'min_cardinality' {min_cardinality}. 'min_cardinality' "
-            'must be an integer greater than or equal to zero.'
+            f"Invalid 'min_cardinality' parameter ({min_cardinality}). The "
+            "'min_cardinality' parameter must be an integer greater than or equal to zero."
         )
         raise SynthesizerProcessingError(msg)
 
@@ -32,8 +32,8 @@ def _validate_max_cardinality(relationship):
     max_cardinality = relationship['max_cardinality']
     if not isinstance(max_cardinality, int) or max_cardinality <= 0:
         msg = (
-            f"Invalid 'max_cardinality' {max_cardinality}. 'max_cardinality' "
-            'must be an integer greater than zero.'
+            f"Invalid 'max_cardinality' parameter ({max_cardinality}). The "
+            "'max_cardinality' parameter must be an integer greater than zero."
         )
         raise SynthesizerProcessingError(msg)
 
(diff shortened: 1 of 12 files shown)


sdv-team:
Task linked: [CU-86b6h4n3f SDV - Allow users to validate the DayZ parameters #2667](https://app.clickup.com/t/86b6h4n3f)

tests/unit/multi_table/test_dayz.py
```diff
@@ -0,0 +1,273 @@
...
+    def test___init__(self):
+        """Test trying to initialize DayZSynthesizer errors."""
+        # Run and Assert
+        expected_error_msg = 'DayZSynthesizer is not available.'
+        with pytest.raises(SynthesizerProcessingError, match=expected_error_msg):
+            DayZSynthesizer(metadata)
+
+    @patch('sdv.multi_table.dayz._validate_parameters')
+    def test_validate_parameters(self, mock__validate_parameters, metadata):
+        # Setup
```
rwedge, line 297: Add docstring describing test

tests/unit/multi_table/test_dayz.py
```diff
@@ -0,0 +1,273 @@
...
+@patch('sdv.multi_table.dayz._validate_tables_parameter')
+@patch('sdv.multi_table.dayz._validate_relationship_parameters')
+def test__validate_parameters(
+    mock__validate_relationship_parameters,
+    mock__validate_tables_parameter,
+    mock__validate_relationship_structure,
+    mock__validate_parameter_structure,
+    metadata,
+):
+    # Setup
```
rwedge, line 217: Add docstring for test

sdv/multi_table/dayz.py
```diff
@@ -0,0 +1,143 @@
...
+def _validate_relationship_structure(dayz_parameters):
+    if not isinstance(dayz_parameters.get('relationships', []), list):
+        raise SynthesizerProcessingError("The 'relationships' value must be a list.")
+
+    for relationship in dayz_parameters.get('relationships', []):
+        unknown_relationship_parameters = relationship.keys() - set(RELATIONSHIP_PARAMETER_KEYS)
+        if unknown_relationship_parameters:
+            unknown_relationship_parameters = "', '".join(unknown_relationship_parameters)
+            msg = f"Relationship contains unexpected key(s) '{unknown_relationship_parameters}'."
+            raise SynthesizerProcessingError(msg)
```
amontanez24, line 62: Shouldn't these all be `SynthesizerInputError`s @npatki ? The docs say we use those for "Invalid parameters when creating a synthesizer, updating its transformers, etc." It seems like in this case we're getting bad input parameters to validate

sdv/multi_table/dayz.py
```diff
@@ -0,0 +1,143 @@
...
+        if 'max_cardinality' in relationship:
+            max_cardinality = relationship['max_cardinality']
+            if not isinstance(max_cardinality, int) or max_cardinality <= 0:
+                msg = (
+                    f"Invalid 'max_cardinality' {max_cardinality}. 'max_cardinality' "
+                    'must be an integer greater than zero.'
+                )
+                raise SynthesizerProcessingError(msg)
+
+        if 'max_cardinality' in relationship and 'min_cardinality' in relationship:
```
amontanez24: can we break all these ifs into smaller functions in case we ever need to override any of them?

sdv/single_table/dayz.py
```diff
@@ -0,0 +1,202 @@
...
+CATEGORICAL_PARAMETER_KEYS = COLUMN_PARAMETER_KEYS + ['category_values']
+DATETIME_PARAMETER_KEYS = COLUMN_PARAMETER_KEYS + ['start_timestamp', 'end_timestamp']
+SDTYPE_TO_PARAMETERS = {
+    'numerical': NUMERICAL_PARAMETER_KEYS,
+    'categorical': CATEGORICAL_PARAMETER_KEYS,
+    'datetime': DATETIME_PARAMETER_KEYS,
+}
+
+
+def _validate_parameter_structure(dayz_parameters):
```
amontanez24: can we break this function up as well? No need to change any of the tests

sdv/single_table/dayz.py
```diff
@@ -0,0 +1,202 @@
...
+            try:
+                pd.to_datetime(column_parameters[param], format=datetime_format)
+            except Exception as e:
+                if datetime_format:
+                    msg = (
+                        f"The '{param}' for {column_table_msg} is not a valid datetime string "
+                        f'or does not match the date time format ({datetime_format}).'
+                    )
+                else:
+                    msg = f"The '{param}' for {column_table_msg} is not a valid datetime string."
```
amontanez24: do we have a function for validating a string matches the datetime format already?

sdv/single_table/dayz.py
```diff
@@ -0,0 +1,202 @@
...
+        if unknown_table_keys:
+            unknown_table_keys = "', '".join(unknown_table_keys)
+            msg = (
+                f"DayZ parameters contain unexpected key(s) '{unknown_table_keys}' "
+                f"for table '{table}'."
+            )
+            raise SynthesizerProcessingError(msg)
+
+
+def _validate_column_parameters(table, column, column_metadata, column_parameters):
```
amontanez24, line 147: can we break this function up as well?

sdv/multi_table/dayz.py
```diff
@@ -1,16 +1,149 @@
...
+    'min_cardinality',
+    'max_cardinality',
+]
+
+DEFAULT_NUM_ROWS = 1000
+
+
+def _validate_min_cardinality(relationship):
+    min_cardinality = relationship['min_cardinality']
+    if not isinstance(min_cardinality, int) or min_cardinality < 0:
```
R-Palazzo, line 23: Should we accept an integer represented as a float like '3.0'?

sdv/multi_table/dayz.py
```diff
@@ -1,16 +1,149 @@
...
+
+    min_child_size = min_cardinality * parent_num_rows
+    max_child_size = max_cardinality * parent_num_rows if max_cardinality else None
+
+    if child_num_rows < min_child_size:
+        msg = (
+            f'Invalid cardinality for relationship {relationship_parameters}. '
+            f'Minimum cardinality requires child table to be at least {min_child_size} rows.'
+        )
+        raise SynthesizerProcessingError(msg)
```
R-Palazzo, line 95: Add a blank line

sdv/single_table/dayz.py
```diff
@@ -1,16 +1,232 @@
...
+        )
+
+
+def _validate_datetime_parameters(column_parameters, column_metadata, column_table_msg):
+    datetime_format = column_metadata.get('datetime_format')
+    for param in ['start_timestamp', 'end_timestamp']:
+        if param not in column_parameters:
+            continue
+
+        if param in column_parameters and not isinstance(column_parameters[param], str):
```
R-Palazzo: You don't need to check `if param in column_parameters` here

sdv/single_table/dayz.py
```diff
@@ -1,16 +1,232 @@
...
+            msg = f"'{param}' for {column_table_msg} must be a string."
+            raise SynthesizerProcessingError(msg)
+
+        if not _datetime_string_matches_format(column_parameters[param], datetime_format):
```
R-Palazzo: We should check if datetime_format is None before calling `_datetime_string_matches_format`. Otherwise we're getting:

```python
TypeError: strftime() argument 1 must be str, not None
```

sdv/single_table/dayz.py
```diff
@@ -1,16 +1,232 @@
...
+            msg = (
+                f"Invalid parameters for {column_table_msg}. The 'start_timestamp' must be "
+                "less than the 'end_timestamp'."
+            )
+            raise SynthesizerProcessingError(msg)
+    elif 'start_timestamp' in column_parameters or 'end_timestamp' in column_parameters:
+        raise SynthesizerProcessingError(
+            f'Invalid parameters for {column_table_msg}. Both a '
+            "'start_timestamp' and an 'end_timestamp' must be set."
+        )
```
R-Palazzo: It's not possible to set only one of the two?

sdv/single_table/dayz.py
```diff
@@ -1,16 +1,232 @@
...
+        _validate_missing_value_parameters(column_parameters, column_table_msg)
+
+
+def _validate_table_parameters(table, table_metadata, table_parameters):
+    missing_cols = table_parameters.get('columns', {}).keys() - table_metadata.columns.keys()
+    if missing_cols:
+        missing_cols = "', '".join(missing_cols)
+        msg = (
+            f"Invalid DayZ parameters provided, column(s) '{missing_cols}' are missing from "
+            f"table '{table}'."
```
R-Palazzo:
```suggestion
            f"Invalid DayZ parameters provided, column(s) '{missing_cols}' are missing from "
            f"table '{table}' inside the metadata."
```

Or something similar

sdv/single_table/dayz.py
```diff
@@ -26,3 +242,15 @@ def create_parameters(cls, data, metadata, output_filename=None):
             dict: The created parameters.
         """
         return create_parameters(data, metadata, output_filename)
+
+    @staticmethod
+    def validate_parameters(metadata, my_parameters):
```
R-Palazzo:
```suggestion
    def validate_parameters(metadata, parameters):
```

sdv/single_table/dayz.py
```diff
@@ -26,3 +242,15 @@ def create_parameters(cls, data, metadata, output_filename=None):
...
+        """Validate a DayZSynthesizer parameters dictionary.
+
+        Args:
+            metadata (sdv.Metadata):
+                Metadata for the data.
+            my_parameters (dict):
```
R-Palazzo:
```suggestion
            parameters (dict):
```

R-Palazzo (commented):
Looking good!

I did some testing and most of it is working.
- The datetime validation logic is a bit confusing; maybe we could add some documentation describing what is checked.
- I would update some error message to be a bit more clear that it's `parameters` (currently my_parameters) parameter that is incorrect (it could also be the metadata). For instance, I got:

```python
SynthesizerProcessingError: The column 'numerical' in table 'table' contains unexpected key(s) 'min_values'.
```

sdv/single_table/dayz.py
frances-h: Not according to the DayZ methods - it makes sense though since we shouldn't assume that the defaults are compatible with a provided value.

sdv/multi_table/dayz.py
frances-h: Currently the DayZ `set_table_sizes` validation only accepts ints, so I have the validation set up to match that.

sdv/single_table/dayz.py
```diff
@@ -1,16 +1,242 @@
...
+    for param in ['min_value', 'max_value']:
+        if param in column_parameters and not _is_numerical(column_parameters[param]):
+            msg = f"The '{param}' parameter for {column_table_msg} must be a float."
+            raise SynthesizerProcessingError(msg)
+
+    if 'min_value' in column_parameters and 'max_value' in column_parameters:
+        if column_parameters['min_value'] > column_parameters['max_value']:
+            msg = (
+                f"Invalid parameters for {column_table_msg}. The 'min_value' "
+                "must be less than the 'max_value'"
```
R-Palazzo:
```suggestion
                "must be less than or equal to the 'max_value'"
```

PR #2715, opened 2025-10-10
[DayZ Parameters] Make create_parameters use a fall back to default parameters if parameters cannot be detected

### LINKED ISSUE #2702: [DayZ Parameters] `create_parameters` should fall back to default parameters if parameters cannot be detected
### Environment Details

* SDV version:
* Python version:
* Operating System:

### Error Description
Currently, the `create_parameters` method will add invalid parameter values in cases like an entirely null column (`'min_value'` and `'max_value'` both get set to `None`).

The `create_parameters` method should always return valid parameters - if a parameter can not be detected for a column for any reason, we should fall back to the expected default values.

Cases to check:
- Creating parameters for data with 0 rows
- Creating parameters for an entirely null column (both numerical and datetime)

### Steps to reproduce

```python
data = pd.DataFrame({'col': []})
metadata = Metadata().detect_from_dataframe(data)
DayZSynthesizer.create_parameters(data, metadata)
```

fealho:
CU-86b6yznjz, Resolve #2702.

### DIFF
--- a/sdv/single_table/_dayz_utils.py
+++ b/sdv/single_table/_dayz_utils.py
@@ -82,7 +82,7 @@ def _detect_datetime_column_parameters(series, column_metadata):
     return params
 
 
-def _detect_categorical_or_boolean_column_parameters(series):
+def _detect_categorical_column_parameters(series):
     """Detect categorical/boolean parameters."""
     categorical_values = series.dropna().unique()
     if len(categorical_values) == 0:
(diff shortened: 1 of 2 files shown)


sdv-team:
Task linked: [CU-86b6yznjz SDV - [DayZ Parameters] `create_parameters` should fall back to default parameters if parameters cannot be detected #2702](https://app.clickup.com/t/86b6yznjz)

tests/unit/single_table/test_dayz.py
```diff
@@ -351,3 +352,93 @@ def test_validate_parameters(
...
+                        'col': {'missing_values_proportion': 0.0},
+                    },
+                    'num_rows': 0,
+                },
+            },
+            'DAYZ_SPEC_VERSION': 'V1',
+        }
+
+    def test_create_parameters_all_null_categorical_column(self):
+        """Categorical column with all nulls should produce empty category_values."""
```
R-Palazzo:
```suggestion
        """Categorical column with all nulls should not have the category_values key parameter."""
```

sdv/single_table/_dayz_utils.py
```diff
@@ -18,6 +18,79 @@ def detect_table_parameters(data):
...
+    if non_na.empty:
+        return params
+
+    start_dt = non_na.min()
+    end_dt = non_na.max()
+    if datetime_format:
+        params['start_timestamp'] = start_dt.strftime(datetime_format)
+        params['end_timestamp'] = end_dt.strftime(datetime_format)
+    else:
+        params['start_timestamp'] = str(start_dt)
```
frances-h: We should use the actual default format here since casting a date to string will end in `'%H:%M:%S'` and fail when we try to cast back to datetime using `'%Y-%m-%d %H:%M:%S.%f'`.

PR #2731, opened 2025-10-21
Remove support for Python 3.8

### LINKED ISSUE #2681: Remove support for Python 3.8
### Problem Description

Python 3.8 hit its end-of-life 10-07-2024. We are nearing the end of life for python 3.9 as well. Supporting older versions makes it difficult to use new features from libraries we depend on, support bug fixes from those libraries and increases time for testing since we test each version. We should remove support for python 3.8 now.

### Expected behavior

- Official support for python 3.8 will be dropped in the next release
- We remove 3.8 from all existing workflows
- We should add a new workflow that runs the tests on deprecated python versions. This will let us know if the old version is still supported by luck.
    - This workflow should only run once a week and have a manual trigger.
- We can remove any special code blocks or if statements made to accommodate 3.8 specifically

### Additional context

- The new workflow does not need to pass. It is just for us to easily see what features are no longer supported on old versions in case we ever have to try and support it
- If we bump the lower bound for our project to 3.9, then we will not be able to install it for the tests. What we can do is have a script that removes the lower bound before installing in that workflow. This is for the workflow that tests deprecated versions.

fealho:
CU-86b6wkh5u, Resolve #2681.


sdv-team:
Task linked: [CU-86b6wkh5u SDV - Remove support for Python 3.8 #2681](https://app.clickup.com/t/86b6wkh5u)

pvk-developer (approved):

sarahmish (approved):

PR #2680, opened 2025-09-25
When using PyTorch, enable GPU usage for MacOS

R-Palazzo:
Resolve #2672
CU-86b6m8uuw

PR #2740, opened 2025-11-14
When using PyTorch, enable GPU usage for MacOS (#2680)

### LINKED ISSUE #2672: When using PyTorch, enable GPU usage for MacOS
### Problem Description
SDV currently offers several different neural network-based synthesizers that use PyTorch under-the-hood:
- CTGANSynthesizer
- TVAESynthesizer
- CopulaGANSynthesizer
- PARSynthesizer

Because these are neural network-based, the synthesizers may be slower to run using a CPU. SDV currently offers a solution for this using the `cuda` parameter, which you can set during initialization. When set to True (default), SDV will use the GPU whenever it's possible.

Unfortunately, this is not available for MacOS users (only for Windows/Linux).

### Expected behavior
Rather than CUDA, MacOS users can use [MPS](https://developer.apple.com/metal/pytorch/).

API-wise, this means:
- Deprecating the `cuda` parameter for all of the above synthesizers. (For backwards compatibility, throw a `FutureWarning` if it is supplied)
- Add a new parameter called `enable_gpu`, which is `True` by default. This parameter enable GPU usage with whatever technology is available for the platform.
   - For Linux/Windows users, this parameter behaves identically to the cuda parameter
   - For MacOS users, it will use MPS instead

### Additional context
This will only be possible in SDV after making the changes to underlying libraries with the ML models -- CTGAN and DeepEcho. Issues will be filed there as well.

R-Palazzo:
Resolve #2672

### DIFF
--- a/sdv/single_table/base.py
+++ b/sdv/single_table/base.py
@@ -311,7 +311,7 @@ class BaseSynthesizer:
 
     def _resolve_gpu_parameters(self, parameters):
         if parameters.get('cuda') is not None and parameters.get('enable_gpu') is None:
-            parameters.pop('enable_gpu', None)  # Ensure backward-compatibilty
+            parameters.pop('enable_gpu', None)  # Ensure backward-compatibility
         elif 'cuda' in parameters:  # Removed because deprecated
             del parameters['cuda']
 
--- a/sdv/single_table/base.py
+++ b/sdv/single_table/base.py
@@ -309,15 +309,23 @@ class BaseSynthesizer:
             msg = 'For this change to take effect, please refit the synthesizer using `fit`.'
             warnings.warn(msg, RefitWarning)
 
+    def _resolve_gpu_parameters(self, parameters):
+        if parameters.get('cuda') is not None and parameters.get('enable_gpu') is None:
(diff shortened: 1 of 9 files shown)


sdv/single_table/base.py
```diff
@@ -309,15 +309,23 @@ def update_transformers(self, column_name_to_transformer):
             msg = 'For this change to take effect, please refit the synthesizer using `fit`.'
             warnings.warn(msg, RefitWarning)
 
+    def _resolve_gpu_parameters(self, parameters):
+        if parameters.get('cuda') is not None and parameters.get('enable_gpu') is None:
+            parameters.pop('enable_gpu', None)  # Ensure backward-compatibilty
```
fealho: typo, compatibility

fealho (commented):
Does this change also need to be done for PAR?

R-Palazzo:
> Does this change also need to be done for PAR?

@fealho no we decided not to update PAR in the end because we noticed a slowdown in its performance. There is more info [here](https://github.com/sdv-dev/SDV/pull/2680).

fealho (approved):

amontanez24 (approved):

Chris-debuggs (commented):
Well written

PR #2786, opened 2026-01-24
When reading in CSV data, do not cut off any leading 0s

### LINKED ISSUE #2780: When reading in CSV data, do not cut off any leading 0s
### Problem Description
Some data might look like numbers that contain leading 0s. The leading 0s are frequently indicative that the data is not numerical, but rather some other data type. For example:
- Regexes might have a specific # of characters that includes leading zeros. For example a 6-digit code `093418` represented by Regex `[0-9]{6}`.
- Datetimes might be stored in a DDMMYYYY format, which uses leading 0s to pad the day. For example Nov 5, 2025 would be `05112025`
- Postal codes might start with a 0, for example `02116` which serves Boston, MA

This data may be stored in a CSV file correctly. However, when reading in the data using the [CSVHandler](https://docs.sdv.dev/sdv/multi-table-data/data-preparation/loading-data/csv), the leading 0s are stripped out. This leads to invalid data and an incorrect interpretation of what the data means. (For example `05112025` is no longer Nov 11, 2025 but a number `5,112,025`). This makes for incorrect metadata detection and synthetic data creation down the line.

### Expected behavior
Add a parameter to the `CSVHandler.read` called `keep_leading_zeros`. By default, this parameter should be set to `True`.

```python
data = connector.read(
    folder_name='project/data/',
    file_names=['users.csv', 'transactions.csv', 'sessions.csv'],
    keep_leading_zeros=True,
    read_csv_parameters={
        'parse_dates': False,
        'encoding':'latin-1'
    }
)
```

If this parameter is `True` then the connector should be able to automatically detect the columns with leading 0s and handle them appropriately.
- The returned data should always keep the leading 0s in any column that would otherwise be cutoff. The cutoffs generally happen for `float` or `int` type. Leading zero columns should instead be loaded in as an `object` (`string`)
- For all other columns (without leading 0s), the loading behavior should be unaffected 

Internally, we use the pandas [`read_csv` function](https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html), which does not have this capability out-of-the-box. (It allows us to set a dtype for all columns, or to specifically set column names to be loaded in as specific types.) So instead, we have to find some other way to detect leading 0s and accomplish this. For example, we should read everything in first and see what pandas converts to `int`. Then go back through those columns only in the CSV file, and see if there are any leading 0s. If so, re-load those columns in as `string`.

However, we should make sure there is not too much of a hit on performance.

fealho:
CU-86b85x0qf, Resolve #2780

### DIFF
--- a/.github/workflows/dtypes_benchmark.yml
+++ b/.github/workflows/dtypes_benchmark.yml
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13', '3.14']
 
     steps:
       - name: Checkout code
--- a/.github/workflows/install.yaml
+++ b/.github/workflows/install.yaml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        python_version: ["3.9", "3.13"]
+        python_version: ["3.9", "3.14"]
     runs-on: ubuntu-latest
     steps:
       - name: Set up python ${{ matrix.python_version }}
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -15,13 +15,13 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13', '3.14']
         os: [ubuntu-latest, windows-latest]
         include:
           - os: macos-latest
             python-version: '3.9'
           - os: macos-latest
-            python-version: '3.13'
+            python-version: '3.14'
     steps:
     - uses: actions/checkout@v4
     - name: Set up Python ${{ matrix.python-version }}
@@ -39,7 +39,7 @@ jobs:
       run: |
         invoke integration
 
-    - if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.13
+    - if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.14
       name: Upload integration codecov report
       uses: codecov/codecov-action@v4
       with:
--- a/.github/workflows/minimum.yml
+++ b/.github/workflows/minimum.yml
@@ -15,13 +15,13 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13', '3.14']
         os: [ubuntu-latest, windows-latest]
         include:
           - os: macos-latest
             python-version: '3.9'
           - os: macos-latest
-            python-version: '3.13'
+            python-version: '3.14'
     steps:
     - uses: actions/checkout@v4
     - name: Set up Python ${{ matrix.python-version }}
--- a/.github/workflows/prepare_release.yml
+++ b/.github/workflows/prepare_release.yml
@@ -28,7 +28,7 @@ jobs:
         run: |
             python -m pip install --upgrade pip
             python -m pip install requests==2.31.0
-            python -m pip install bandit==1.7.7
+            python -m pip install bandit==1.9.2
             python -m pip install packaging
             python -m pip install .[test]
 
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -15,13 +15,13 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13', '3.14']
         os: [ubuntu-latest, windows-latest]
         include:
           - os: macos-latest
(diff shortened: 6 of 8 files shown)


sdv-team:
Task linked: [CU-86b85x0qf SDV - When reading in CSV data, do not cut off any leading 0s #2780](https://app.clickup.com/t/86b85x0qf)

tests/unit/io/local/test_local.py
```diff
@@ -107,6 +108,74 @@ def test_read(self, mock_read_csv, mock_glob):
...
+        out = handler.read(tmpdir, file_names=['users.csv'])
+
+        # Assert
+        pd.testing.assert_frame_equal(out['users'], data)
+
+    def test_read_keep_leading_zeros_multiple_files_mixed_types(self, tmpdir):
+        """Test leading zeros with multiple files and mixed dtypes."""
+        # Setup
+        file_path = Path(tmpdir)
+        users = pd.DataFrame({
```
gsheni, line 134: Could we specify the dtypes in these pandas DataFrames for this test?

tests/unit/io/local/test_local.py
```diff
@@ -107,6 +108,74 @@ def test_read(self, mock_read_csv, mock_glob):
...
+            'user_id': [1, 2, None],
+            'zip_code': ['00123', '98765', np.nan],
+            'age': [30, 25, None],
+            'is_active': [True, False, None],
+            'joined_at': ['2024-01-01', '2024-01-02', None],
+        })
+        orders = pd.DataFrame({
+            'order_id': [10, 20, np.nan],
+            'tracking_code': ['000045', '123450', None],
```
gsheni, line 150: Just curious, does this work with padded integers (spaces in front of the integer)? I think its out of scope for the issue but just wondering

fealho: No, it won't match the regex.

sdv/io/local/local.py
```diff
@@ -123,7 +171,12 @@ def read(self, folder_name, file_names=None, read_csv_parameters=None):
 
         for file_path in file_paths:
             table_name = file_path.stem  # Remove file extension to get table name
-            data[table_name] = pd.read_csv(file_path, **read_csv_parameters)
+            table_data = pd.read_csv(file_path, **read_csv_parameters)
+
+            if keep_leading_zeros:
+                table_data = self._keep_leading_zeros(file_path, table_data, read_csv_parameters)
+
```
pvk-developer, line 207: I think this is very 'expensive', we are reading the data twice. When I encountered this issue I was thinking that maybe we can read the first 5 rows or something and check if we have leading zeros there and build the columns that way.

For example, something like this:
```python

In [1]: data = pd.DataFrame({'unique_id': ['0001', '0002', '0003']})

In [2]: data.to_csv('demo_zero.csv', index=False)

In [3]: read_data = pd.read_csv('demo_zero.csv')

In [4]: read_data
Out[4]: 
   unique_id
0          1
1          2
2          3

In [5]: data
Out[5]: 
  unique_id
0      0001
1      0002
2      0003

In [6]: import csv

In [7]: with open('demo_zero.csv', newline='', encoding='utf-8') as data_file:
   ...:     reader = csv.reader(data_file)
   ...:     columns = next(reader)
   ...:     for row in reader:
   ...:         for idx, value in enumerate(row):
   ...:             v = value.strip()
   ...:             print(v)
   ...: 
0001
0002
0003
```
Either you build the columns with that or you actually load the data with the csv reader. Because if we have a very large data file I don't think this is good.

pvk-developer (changes requested):
You are still reading the file twice;

I think we can do something like this:
```python
df = pd.read_csv(path, dtype=str, low_memory=False)
has_leading_zero = (
    df.apply(
        lambda col: col.str.match(r"^0\d+").any()
    )
)
for col in df.columns:
    if not has_leading_zero[col]:
        converted = pd.to_numeric(df[col], errors="ignore")
        if not converted.equals(df[col]):
            df[col] = converted
```

I tested this with 1M rows * 100 columns and it takes ~1minute 10 sec covering all records, while your approach took ~1minute 10 sec and covered only 100k rows per column, which for some edge cases did not work.

Here is the data I used:
```python
import numpy as np
import pandas as pd

def generate_mixed_zero_df(
    n_rows=1_000_000,
    n_cols=100,
    width=8,
    last_n_padded=100_000,
    n_mixed_cols=10,
    seed=42
):
    rng = np.random.default_rng(seed)
    base_vals = np.arange(1, n_rows + 1) % (10**width)

    data = {}

    mixed_cols = set(rng.choice(n_cols, size=n_mixed_cols, replace=False))

    for i in range(n_cols):
        col_name = f"col_{i}"
        vals = base_vals.copy()
        if i in mixed_cols:
            first_part = np.char.ljust(
                vals[: n_rows - last_n_padded].astype(str),
                width,
                fillchar="0"
            )
            last_part = np.char.zfill(
                vals[n_rows - last_n_padded :].astype(str),
                width
            )

            data[col_name] = np.concatenate([first_part, last_part])
        else:
            data[col_name] = np.char.zfill(vals.astype(str), width)

    df = pd.DataFrame(data)

    return df, sorted(mixed_cols)


df, mixed_columns = generate_mixed_zero_df(
    n_mixed_cols=15
)

print("Mixed columns:", mixed_columns)
print(df.head())
print(df.tail())
```

Also, I think this approach could be used then in the `download_demo` functionality to load the leading zeros from the demo datasets.

PR #2792, opened 2026-01-30
Add support for auto-detect when a primary key is also a foreign key in Metadata

### LINKED ISSUE #2788: Metadata should be able to auto-detect cases where a primary key is also a foreign key (`column_name_match` algorithm)
### Problem Description
In [SDV 2779](https://github.com/sdv-dev/SDV/issues/2779), we'll be allowing schemas where the primary key of a table is also the foreign key into a different table. This type of data pattern is typically used when there is a 1-to-1 relationship between the tables.

In that issue, we'll make sure the metadata will validate and that we can make synthetic data for it down-the-line. But it would be nice if the metadata were able to auto-detect this as well.

### Expected behavior
For SDV Community, only the `'column_name_match'` algorithm is support when inferring foreign keys ([see API](https://docs.sdv.dev/sdv/multi-table-data/data-preparation/creating-metadata#detect_from_dataframes)). For this algorithm, we should ensure that a primary key also become a candidate to be a foreign key into another table.

Note that we should be careful in making sure that our updated algorithm does not create any cycles in the graph, as that is no longer valid.

gsheni:
- Fixes #2788

### DIFF
--- a/sdv/metadata/multi_table.py
+++ b/sdv/metadata/multi_table.py
@@ -545,6 +545,8 @@ class MultiTableMetadata:
         for parent_candidate in sorted_tables:
             parent_meta = self.tables[parent_candidate]
             primary_key = parent_meta.primary_key
+            if primary_key is None:
+                continue
             for child_candidate in sorted_tables:
                 if child_candidate == parent_candidate:
                     continue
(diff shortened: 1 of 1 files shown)


pvk-developer (approved):

sdv-team:

PR #2798, opened 2026-02-04
Allow primary key to also be a foreign key 

### LINKED ISSUE #2455: Allow a column to be both a primary key and foreign key at the same time
### Problem Description
In certain cases, it is possible to have a schema in which a column is both (a) a primary key of the table, and (b) a foreign key linking to a different table at the same time. This generally happens when there is a "0 or 1" linkage between two tables.

For example, consider a table **Users** that contains all users of a service:
| User ID  | Date Joined  | Subscription Type |
|----------|--------------|-------------------|
| id_03234 | June 2, 2024 | Normal            |
| id_93813 | Apr 4, 2023  | Normal            |
| id_9192  | Dec 7, 2024  | Premium           |
| ...  | ...  | ...           |

Then consider another table **Survey Responses** that contains optional survey responses submitted by those users. Not all users filled out the survey, and the ones who did could only submit it 1 time.
| User ID  | Age | Satisfaction       |
|----------|-----|--------------------|
| id_03234 | 43  | Very Satisfied     |
| id_9192  | 23  | Slightly Satisfied |

In this sense, the Users the Survey Responses have a "0 or 1" mapping. 

In this schema, the `Survey Responses.User ID` column is special:
- It is the primary key of `Survey Responses.User ID` because it is uniquely identifying each response
- It is also the foreign key into `Users.User ID` because it is referring to the same concept of a user

However, the metadata definition won't currently allow me to specify the column as both a primary and foreign key. The metadata validation fails:
```
Relationships:
Invalid relationship between table 'Users' and table 'Survey Responses'. A relationship must connect a primary key with a non-primary key.
```

### Expected behavior
I expect that I should be able to add such a relationship to my metadata and that the synthesizers would create valid synthetic data ensuring that:
- The values in the column are all unique (because it's a primary key) and
- The values in the column are referencing values in the parent table's primary key too (because it's a foreign key)

### Workaround
The simple workaround for anyone running into this is: **Only supply the foreign key designation in the metadata for now** (and leave out the primary key designation).

In this example: Do not designate that `Survey Responses.User ID` is a primary key, but _do_ designate it as a foreign key into `Users.User ID`.

This will create fully valid synthetic data that contains fully unique values for `Survey Responses.User ID`. Even though it's not designated as a primary key, all SDV synthesizers respect the cardinality logic so the synthetic data will be valid.

### LINKED ISSUE #2779: Allow a primary key to also be a foreign key (in the metadata and also for modeling)
### Problem Description
Currently, SDV does not allow schemas where a primary key of a table is also a foreign key into a different table. This is a valid schema pattern, frequently used when there the table contains additional information for the main entity. For example, consider a table called `Users` and another table called `Supplemental Info`. The `Supplemental Info` table can have a primary key that is _also_ the foreign key to the `Users` table. This indicates that each user may be associated with at most 1 supplemental info entry.

**Workaround**: Right now, SDV will actually correctly model this schema _if_ the primary key designation of the table is left out. For example, if you just denote that there's a foreign key in `Supplemental Info` without also mentioning that it's a primary key of that table. The synthetic data will be valid because SDV always preserves the cardinality of the table -- in this case, it will preserve the fact that there can be at most 1 supplemental info entry associated with a given user.

### Expected behavior
The metadata should allow the case where a primary key of a table is _also_ a foreign key to a different table. Making this change requires updating the following areas:

1. Updating the metadata validation to allow for this case
2. Allowing this type of schema to be specified via the metadata API. This would mean updating the validation for [`set_primary_key`](https://docs.sdv.dev/sdv/concepts/metadata/metadata-api#set_primary_key) and [`add_relationship`](https://docs.sdv.dev/sdv/concepts/metadata/metadata-api#add_relationship).

In addition to this we should double-check that the rest of the functionality works as-expected:

- The metadata visualization should correctly show the tables as being connected
- There are no bugs when modeling and sampling this type of data with any of our multi-table synthesizers (HMA, HSA, Independent). The synthetic data should correctly show either an exact 1-to-1 relationship or a 1-to-(0 or 1) relationship (depending on what the real data has).

gsheni:
- Fixes #2455
  - CU-86b4kbzf2 
- Fixes #2779 
  - CU-86b84tfje

### DIFF
--- a/tests/integration/metadata/test_metadata.py
+++ b/tests/integration/metadata/test_metadata.py
@@ -1378,39 +1378,9 @@ def test_validate_pk_to_pk(primary_key_to_primary_key):
     # Setup
     data, metadata_instance = primary_key_to_primary_key
 
-    # Run
+    # Run and Assert
     metadata_instance.validate()
     metadata_instance.validate_data(data)
-
-    # Assert
-    expected_metadata = {
-        'tables': {
-            'tableA': {
-                'columns': {
-                    'table_A_primary_key': {'sdtype': 'id'},
-                    'column_1': {'sdtype': 'categorical'},
-                },
-                'primary_key': 'table_A_primary_key',
-            },
-            'tableB': {
-                'columns': {
-                    'table_B_primary_key': {'sdtype': 'id'},
-                    'column_2': {'sdtype': 'categorical'},
-                },
-                'primary_key': 'table_B_primary_key',
-            },
-        },
-        'relationships': [
-            {
-                'parent_table_name': 'tableA',
-                'parent_primary_key': 'table_A_primary_key',
-                'child_table_name': 'tableB',
-                'child_foreign_key': 'table_B_primary_key',
-            }
-        ],
-        'METADATA_SPEC_VERSION': 'V1',
-    }
-    assert metadata_instance.to_dict() == expected_metadata
 
 
 def test_validate_pk_to_pk_email():
@@ -1464,38 +1434,10 @@ def test_validate_pk_to_pk_email():
             'column_2': ['A', 'B', 'B', 'C', 'C'],
         }),
     }
-    # Run
+
+    # Run and Assert
     metadata_instance.validate()
     metadata_instance.validate_data(data)
-
-    # Assert
-    assert metadata_instance.to_dict() == {
-        'tables': {
-            'tableA': {
-                'columns': {
-                    'table_A_primary_key': {'sdtype': 'email'},
-                    'column_1': {'sdtype': 'categorical'},
-                },
-                'primary_key': 'table_A_primary_key',
-            },
-            'tableB': {
-                'columns': {
-                    'table_B_primary_key': {'sdtype': 'email'},
-                    'column_2': {'sdtype': 'categorical'},
-                },
-                'primary_key': 'table_B_primary_key',
-            },
-        },
-        'relationships': [
-            {
-                'parent_table_name': 'tableA',
-                'parent_primary_key': 'table_A_primary_key',
-                'child_table_name': 'tableB',
-                'child_foreign_key': 'table_B_primary_key',
-            }
-        ],
-        'METADATA_SPEC_VERSION': 'V1',
-    }
 
 
 def test_set_primary_key_pk_to_pk():


sdv-team:
Task linked: [CU-86b89n0hm SDV - Metadata should be able to auto-detect cases where a primary key is also a foreign key (`column_name_match` algorithm) #2788](https://app.clickup.com/t/86b89n0hm)
Task linked: [CU-86b84tfje SDV - Allow a primary key to also be a foreign key (in the metadata and also for modeling) #2779](https://app.clickup.com/t/86b84tfje)
Task linked: [CU-86b4kbzf2 SDV - Allow a column to be both a primary key and foreign key at the same time #2455](https://app.clickup.com/t/86b4kbzf2)

tests/integration/metadata/test_metadata.py
```diff
@@ -1411,3 +1371,265 @@ def test_validate_empty_metadata():
...
+            {
+                'parent_table_name': 'tableA',
+                'parent_primary_key': 'table_A_primary_key',
+                'child_table_name': 'tableB',
+                'child_foreign_key': 'table_B_primary_key',
+            }
+        ],
+        'METADATA_SPEC_VERSION': 'V1',
+    }
+    assert metadata_instance.to_dict() == expected_metadata
```
pvk-developer: This is not needed. Here you are just asserting that the dictionary that I can copy from the conftest matches the metadata_instance.to_dict(), expecting that the calls of `validate` or `validate_data` will alter it. I think that the above runs are the `Run and Assert` for this test.

tests/integration/metadata/test_metadata.py
```diff
@@ -1411,3 +1371,265 @@ def test_validate_empty_metadata():
...
+            ],
+            'column_2': ['A', 'B', 'B', 'C', 'C'],
+        }),
+    }
+    # Run
+    metadata_instance.validate()
+    metadata_instance.validate_data(data)
+
+    # Assert
+    assert metadata_instance.to_dict() == {
```
pvk-developer: Same as before. This assert is not needed.

tests/integration/metadata/test_metadata.py
```diff
@@ -1411,3 +1371,265 @@ def test_validate_empty_metadata():
...
+        },
+    }
+    assert metadata_instance.to_dict()['relationships'] == [
+        {
+            'parent_table_name': parent_table_name,
+            'child_table_name': child_table_name,
+            'parent_primary_key': parent_primary_key,
+            'child_foreign_key': child_foreign_key,
+        }
+    ]
```
pvk-developer, line 1577: Should we add a test to `detect` from the data and see if it matches the metadata ? Something like this:
```python
def test_autodetection_primary_to_primary_key(pk_to_pk_dataset):
     # Setup
    data, metadata = pk_to_pk_dataset

    # Run
    detected_metadata = Metadata.detect_from_dataframes(data)
    # Assert
    assert metadata.to_dict() == detected_metadata.to_dict()

tests/integration/metadata/test_metadata.py
gsheni: Addressed:
https://github.com/sdv-dev/SDV/pull/2798/commits/f607bad780bc9709128959e0bd5c9a4d17c77463

tests/integration/metadata/test_metadata.py
gsheni: Addressed:
https://github.com/sdv-dev/SDV/pull/2798/commits/f607bad780bc9709128959e0bd5c9a4d17c77463

tests/integration/metadata/test_metadata.py
gsheni: No, I don't think we should. The changes to `detect_from_dataframes` will be addressed in a follow-up PR. I want to keep those changes separate. 
For now `detect_from_dataframes` should be unchanged and will not detect PK to PK relationships.

PR #2827, opened 2026-02-25
[Metadata] Raise a warning if the metadata and data columns differ.

### LINKED ISSUE #2822: Metadata validation should produce a warning if the metadata and data have columns in different orders
### Problem Description
The metadata is meant to accurately reflect all the tables and columns in the dataset -- this includes the order that the columns appear within each table (from left to right). If there is a mismatch in column order between the data and metadata, there may be some inconsistencies down-the-line when we create synthetic data. See https://github.com/sdv-dev/SDV/issues/2803, as an example.

### Expected behavior
When [validating metadata](https://docs.sdv.dev/sdv/concepts/metadata/metadata-api#validate_data), there should be a warning if the order of the columns within each table is different between the metadata and data.

```python
metadata.validate_data(data)

Warning: The metadata lists columns in a different order than the data. This may result in the synthetic data having a different order.
Affected tables: 'users', 'transactions', 'sections'
```

### Additional context
- If a user auto-detects metadata (or imports it from an AI Connector), then the metadata should be in the same order as the data so this warning shouldn't really appear in practice. However, our demo datasets are out-of-sync right now. See #2821 for fixing this.
- The order of the tables (within the data dictionary) doesn't matter for now. Just check the order of the columns within each table.

pvk-developer:
Resolves #2822
CU-86b8mey9j

### DIFF
--- a/sdv/metadata/multi_table.py
+++ b/sdv/metadata/multi_table.py
@@ -1300,12 +1300,12 @@ class MultiTableMetadata:
             filepath (str):
                 String that represents the ``path`` to the ``json`` file.
 
+        Returns:
+            A ``MultiTableMetadata`` instance.
+
         Raises:
             - An ``Error`` if the path does not exist.
             - An ``Error`` if the ``json`` file does not contain the ``METADATA_SPEC_VERSION``.
-
-        Returns:
-            A ``MultiTableMetadata`` instance.
         """
         metadata = read_json(filepath)
         return cls.load_from_dict(metadata)
@@ -1358,11 +1358,11 @@ class MultiTableMetadata:
             filepath (str):
                 String that represents the ``path`` to the old metadata ``json`` file.
 
-        Raises:
-            Raises a ``ValueError`` if the filepath does not exist.
-
         Returns:
             A ``MultiTableMetadata`` instance.
+
+        Raises:
+            Raises a ``ValueError`` if the filepath does not exist.
         """
         old_metadata = read_json(filepath)
         tables_metadata = {}
--- a/sdv/metadata/metadata.py
+++ b/sdv/metadata/metadata.py
@@ -24,12 +24,12 @@ class Metadata(MultiTableMetadata):
             filepath (str):
                 String that represents the ``path`` to the ``json`` file.
 
+        Returns:
+            A ``Metadata`` instance.
+
         Raises:
             - An ``Error`` if the path does not exist.
             - An ``Error`` if the ``json`` file does not contain the ``METADATA_SPEC_VERSION``.
-
-        Returns:
-            A ``Metadata`` instance.
         """
         metadata = read_json(filepath)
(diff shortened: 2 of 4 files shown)


sdv-team:
Task linked: [CU-86b8mey9j SDV - Metadata validation should produce a warning if the metadata and data have columns in different orders #2822](https://app.clickup.com/t/86b8mey9j)

sdv/metadata/multi_table.py
```diff
@@ -835,11 +835,14 @@ def _validate_all_tables(self, data):
         """Validate every table of the data has a valid table/metadata pair."""
         errors = []
         warning_dataframes = []
+        tables_with_missmatching_columns_order = []
```
fealho: typo, mismatching

tests/unit/metadata/test_single_table.py
```diff
@@ -3508,3 +3554,46 @@ def test_anonymize(self):
...
+        metadata.detect_from_dataframe(data)
+
+        # Run
+        result = metadata._check_data_columns_order(data.columns)
+
+        # Assert
+        assert result is True
+
+    def test__check_data_columns_order_mismatches(self):
+        """Test that `_check_data_columns_order` returns `False` when columns miss match."""
```
fealho: typo: mismatch

sdv/sampling/hierarchical_sampler.py
```diff
@@ -285,6 +285,35 @@ def _finalize(self, sampled_data):
...
+        Args:
+            sampled_data (dict[str, pandas.DataFrame]):
+                Mapping of table names to sampled dataframes.
+
+        Returns:
+            dict[str, pandas.DataFrame]:
+                The same mapping with columns reordered to match the metadata.
+        """
+        for table_name, table in sampled_data.items():
+            metadata = self.get_metadata()
```
fealho: Move this outside the loop

sdv/metadata/multi_table.py
frances-h:
```suggestion
        tables_with_mismatching_columns_order = []
```

sdv/sampling/hierarchical_sampler.py
```diff
@@ -341,5 +371,10 @@ def _sample(self, scale=1.0):
                 )
                 added_relationships.add((parent_name, child_name))
 
+        # Reorder the columns to match the metadata-defined order before calling
+        # `_reverse_transform_constraints`. During that step,
+        # `drop_unknown_references` triggers `metadata.validate(data)`, which now
+        # warns if the column order in the data does not match the metadata.
+        sampled_data = self._align_columns_with_metadata_order(sampled_data)
```
fealho: Is this call just to suppress the warning? Because _finalize already sorts the columns, no? If so, couldn't we just suppress that warning?

pvk-developer: 👍🏻  Yeah actually that would be better I think. I was using `ignore` all but now filtering over the first portion of the message seems to work.

PR #2873, opened 2026-04-22
Add verbose parameter to metadata auto-detection (`detect_from_dataframes` and `detect_from_dataframe`)

### LINKED ISSUE #2871: Add verbosity to metadata auto-detection
### Problem Description
When auto-detecting metadata, the algorithm makes many nuanced decisions about the columns, sdtypes, and other properties. Right now, it is not possible to easily view properties such as datetime or regex formats. The visualize function doesn't display this information; the only other alternative is to print out the JSON object, which is hard to parse.

### Expected behavior
For all the metadata detection functions (multi-table `detect_from_dataframes` and single-table `detect_from_dataframe`), add a parameter called `verbose`:
- (default) `False`: Do not print out any information about what is detected
- `True`: Print out information about what is detected

If set to `True`, the metadata auto-detection should print out what it's doing. For example:

```python
metadata = Metadata.detect_from_dataframes(data, verbose=True)
```
```
Detecting table 'hotels':
- Column 'hotel_id': sdtype='id', regex_format='HID_[0-9]{3}'
- Column 'city': sdtype='categorical'
- Column 'rating': sdtype='numerical', computer_representation='Float'

Detecting table 'guests':
- ...

Detecting primary keys:
- Table 'hotels': primary_key='hotel_id'
- Table 'guests': primary_key=None

Detecting foreign keys:
- Column 'guests.hotel_id' refers to column 'hotels.hotel_id' (updating sdtype to 'id')
- ...
```

### Additional context
- Depending on the parameters, not all sections may be present
- Some of the sdtypes may change based on the foreign key detection (as the foreign key and primary key's sdtype should match)

gsheni:
- Fixes #2871

### DIFF
--- a/sdv/metadata/multi_table.py
+++ b/sdv/metadata/multi_table.py
@@ -3,6 +3,7 @@
 import datetime
 import json
 import logging
+import sys
 import warnings
 from collections import defaultdict
 from copy import deepcopy
@@ -547,7 +548,7 @@ class MultiTableMetadata:
                 f'The relationships in the dataset are disjointed. {table_msg}'
             )
 
-    def _detect_foreign_keys_by_column_name(self, data):
+    def _detect_foreign_keys_by_column_name(self, data, verbose=False):
         """Detect the foreign keys based on if a column name matches a primary key.
 
         If a column name (a child table) is a primary key, it will also be considered
@@ -557,7 +558,14 @@ class MultiTableMetadata:
             data (dict):
                 Dictionary of table names to dataframes.
                 NOTE: this is only used in SDV-Enterprise.
+            verbose (bool):
+                A boolean that determines if information should be printed regarding detection.
+                If True, it prints out information about what is detected.
+                If False, it does not print out any information about what is detected.
+                Defaults to False.
         """
+        if verbose:
+            sys.stdout.write('\nDetecting foreign keys:\n')
         for parent_candidate in self.tables.keys():
             primary_key = self.tables[parent_candidate].primary_key
             if primary_key is None:
@@ -573,15 +581,24 @@ class MultiTableMetadata:
                         continue
 
                     try:
+                        sdtype_updated = False
                         if pk_sdtype == 'id' and original_fk_sdtype != 'id':
                             self.update_column(
                                 table_name=child_candidate,
                                 column_name=primary_key,
                                 sdtype='id',
                             )
+                            sdtype_updated = True
                         self.add_relationship(
                             parent_candidate, child_candidate, primary_key, primary_key
                         )
+                        if verbose:
+                            child_col = f"'{child_candidate}.{primary_key}'"
+                            parent_col = f"'{parent_candidate}.{primary_key}'"
+                            suffix = " (updating sdtype to 'id')" if sdtype_updated else ''
+                            sys.stdout.write(
+                                f'- Column {child_col} refers to column {parent_col}{suffix}\n'
+                            )
 
                     except InvalidMetadataError:
                         # circular relationship
@@ -593,7 +610,9 @@ class MultiTableMetadata:
                             )
                         continue
 
-    def _detect_relationships(self, data=None, foreign_key_inference_algorithm='column_name_match'):
+    def _detect_relationships(
+        self, data=None, foreign_key_inference_algorithm='column_name_match', verbose=False
+    ):
         """Automatically detect relationships between tables.
 
         Args:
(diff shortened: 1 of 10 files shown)


sdv/metadata/single_table.py
```diff
@@ -666,22 +669,55 @@ def _select_primary_key(self, infer_sdtypes, pk_candidates, pii_pk_candidates):
...
+            sys.stdout.write('\nDetecting primary key:\n')
+        chosen_pk = None
+        sdtype_updated = False
+        pii_removed = False
+
         if pk_candidates:
             selected_pk = pk_candidates[0]
             self.primary_key = selected_pk
             self.columns[self.primary_key]['sdtype'] = 'id'
+            chosen_pk = self.primary_key
```
fealho, line 698: Don't you also need to set sdtype_updated = True if sdtype was not 'id'?

sdv/metadata/single_table.py
```diff
@@ -728,13 +773,18 @@ def _detect_columns(self, data, table_name=None, infer_sdtypes=True, infer_keys=
                 column_dict['pii'] = True
 
             column_dict['sdtype'] = sdtype
+            if verbose and infer_sdtypes:
+                parts = [f"{k}='{v}'" for k, v in column_dict.items()]
```
fealho: it should be {v} for bools/None, '{v}' for others. Otherwise you get `pii='True'`, when it should be `pii=True`.

sdv/metadata/single_table.py
```diff
@@ -699,7 +735,16 @@ def _detect_columns(self, data, table_name=None, infer_sdtypes=True, infer_keys=
...
                     - None: Do not infer any keys.
                 Defaults to 'primary_only'.
+            verbose (bool):
+                A boolean that determines if information should be printed regarding detection.
+                If True, it prints out information about what is detected.
+                If False, it does not print out any information about what is detected.
+                Defaults to False.
         """
+        if verbose and infer_sdtypes:
+            table_str = f" table '{table_name}'" if table_name else ' table None'
```
fealho: if table_name=None, message is confusing.

sdv/metadata/multi_table.py
```diff
@@ -557,7 +558,14 @@ def _detect_foreign_keys_by_column_name(self, data):
...
                 Dictionary of table names to dataframes.
                 NOTE: this is only used in SDV-Enterprise.
+            verbose (bool):
+                A boolean that determines if information should be printed regarding detection.
+                If True, it prints out information about what is detected.
+                If False, it does not print out any information about what is detected.
+                Defaults to False.
         """
+        if verbose:
+            sys.stdout.write('\nDetecting foreign keys:\n')
```
fealho, line 569: This will print even if there are no foreign keys, which is confusing. Either this should only print if fks are found, or print something like "No foreign keys found" at the end

fealho (commented):
Should `detect_from_csvs` be changed?

sdv/metadata/multi_table.py
gsheni: Fixed. I added a `-No foreign keys found` print statement at the end. 
I also checked if this applies ~to PKs and~ sdtypes (it does not). 
https://github.com/sdv-dev/SDV/pull/2873/commits/9923c656a2df54fbb5a2a9e722b24e36b1898b27

sdv/metadata/single_table.py
gsheni: Fixed. I now only print out the table name if it is defined:
https://github.com/sdv-dev/SDV/pull/2873/commits/9923c656a2df54fbb5a2a9e722b24e36b1898b27

sdv/metadata/single_table.py
gsheni: Fixed. I added a util function to handle the formatting:
https://github.com/sdv-dev/SDV/pull/2873/commits/9923c656a2df54fbb5a2a9e722b24e36b1898b27

sdv/metadata/single_table.py
gsheni: Fixed. I added a check to compare the original sdtype for the PK.
https://github.com/sdv-dev/SDV/pull/2873/commits/9923c656a2df54fbb5a2a9e722b24e36b1898b27

gsheni:
@fealho I don't believe that `detect_from_csvs` should be changed. The issue specifies that
> For all the metadata detection functions (multi-table `detect_from_dataframes` and single-table `detect_from_dataframe`), add a parameter called verbose

I believe `verbose` should only be added to those user-facing functions.

fealho (approved):

sdv/metadata/single_table.py
```diff
@@ -666,22 +674,58 @@ def _select_primary_key(self, infer_sdtypes, pk_candidates, pii_pk_candidates):
...
+            table_name (str):
+                The name of the table to be analyzed. Defaults to ``None``.
+            verbose (bool):
+                A boolean that determines if information should be printed regarding detection.
+                If True, it prints out information about what is detected.
+                If False, it does not print out any information about what is detected.
+                Defaults to False.
         """
+        if verbose:
+            sys.stdout.write('\nDetecting primary key:\n')
```
frances-h: In the verbose output, it's odd to me that this gets printed after every table. It seems like the expected detection behavior  in the issue is that primary keys for tables are all detected at the same time after the column metadata is detected, while we actually detect them table by table immediatly after detecting the table metadata.

Based on this, I think it would be clearer to either:
1) Include the table name, so the header becomes `"Detecting primary key for table 'table'"` or
2) Indent it instead of it being a new header so that it's clear that it's part of the detection for the table

We should probably also add a "No primary key detected" fallback in case no primary key gets detected.

frances-h (commented):
Looking good, just some feedback on the formatting

sdv/metadata/single_table.py
gsheni: Fair points. I went with option 1 and added the table name for the PK header. I also added a `- No primary key found` fallback.

frances-h (approved):

PR #2878, opened 2026-05-12
Allow single table ProgrammableConstraints with multi-table synthesizers if `table_name` is set

frances-h:
Required for datacebo/sdv-enterprise#2036

PR #2887, opened 2026-05-19
Remove `SingleTableProgrammableConstraint` class

### LINKED ISSUE #2882: Remove `SingleTableProgrammableConstraint` class
### Problem Description
Currently, we offer the `ProgrammableConstraint` and `SingleTableProgrammableConstraint` classes as convenient base classes for creating new constraints. We'd like to simplify the process of creating a new custom constraint by having a single base class that always expects Metadata and data dictionaries for the method inputs. 

### Expected behavior
Remove the `SingleTableProgrammableConstraint` class. The `ProgrammableConstraint` API should remain the same, except that, by default, the `_is_single_table` flag should be set to `True`.

Update the `ProgrammableConstraintHarness` so that it always passes the entire input metadata and/or data dictionaries to the programmable constraint methods, instead of extracting the dataframe in the single table case.

frances-h:
Resolve #2882

### DIFF
--- a/sdv/cag/programmable_constraint.py
+++ b/sdv/cag/programmable_constraint.py
@@ -130,7 +130,7 @@ class ProgrammableConstraintHarness(BaseConstraint):
     def __init__(self, programmable_constraint):
         super().__init__()
         self.programmable_constraint = programmable_constraint
-        self.table_name = getattr(programmable_constraint, 'table_name', None)
+        self.table_name = getattr(self.programmable_constraint, 'table_name', None)
         self._is_single_table = self.programmable_constraint._is_single_table
 
     def _validate_constraint_with_metadata(self, metadata):
--- a/sdv/cag/programmable_constraint.py
+++ b/sdv/cag/programmable_constraint.py
@@ -2,14 +2,13 @@
 
 from copy import deepcopy
 
-from sdv.cag._errors import ConstraintNotMetError
 from sdv.cag.base import BaseConstraint
 
 
 class ProgrammableConstraint:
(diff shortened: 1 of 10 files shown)


sdv/cag/programmable_constraint.py
```diff
@@ -125,79 +124,33 @@ def fix_data(self, synthetic_data):
...
-
-
 class ProgrammableConstraintHarness(BaseConstraint):
     """Constraint interface class for programmable constraints."""
 
     def __init__(self, programmable_constraint):
         super().__init__()
         self.programmable_constraint = programmable_constraint
-        self.table_name = getattr(self.programmable_constraint, 'table_name', None)
+        self.table_name = getattr(programmable_constraint, 'table_name', None)
```
R-Palazzo: Do we need this change?

R-Palazzo (approved):
LGTM!

sdv/cag/programmable_constraint.py
frances-h: I tried to revert [this](https://github.com/sdv-dev/SDV/pull/2878/changes#diff-070bd763006a9883993555c98cf81b29a08636f2677f9212520397b07fd0b416R140) change, but there's a [check in multi-table](https://github.com/sdv-dev/SDV/blob/760c5084becc5f67285327a23861649979e4e1bc/sdv/multi_table/base.py#L219) that forces single-table constraints to have the `table_name` attribute set.

I'll revert this change here back to what it was, I just wanted to comment on why I tried to change it in the first place.

amontanez24 (approved):

PR #2894, opened 2026-05-27
Update to newer versions of actions to address Node.js 20 deprecation

### LINKED ISSUE #2893: Use Node.js 24 compatible github actions
In many of our github workflows, we receive warnings like these when running the workflow:

Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/checkout@v4, actions/setup-python@v5, actions/upload-artifact@v4. Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

We should update to newer versions of actions so our workflows are running on Node.js 24.

gsheni:
- Fixes #2893

### DIFF
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -23,9 +23,9 @@ jobs:
           - os: macos-latest
             python-version: '3.14'
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v6
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@v6
       with:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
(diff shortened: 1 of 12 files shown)


.github/workflows/unit.yml
```diff
@@ -36,9 +36,9 @@ jobs:
       run: invoke unit
     - if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.14
       name: Upload unit codecov report
-      uses: codecov/codecov-action@v4
+      uses: codecov/codecov-action@v6
```
gsheni, line 39: > `file` (this has been deprecated in favor of `files`)

https://github.com/codecov/codecov-action#migration-guide

frances-h (approved):

pvk-developer (approved):

PR #2903, opened 2026-06-05
Add load_constraints function to load constraints from filepath

### LINKED ISSUE #2902: Add `load_constraints` utility function
### Problem Description
Currently, constraints saved to a JSON file can only be loaded by using the `set_constraints` method directly from a synthesizer. To better align with other saving/loading concepts, like metadata, we should add a utility function that will return a list of the loaded constraints.

Since loading may now be handled separately from a synthesizer instance, we should also deprecate the `set_constraints` method.

### Expected behavior
Add a utility function `load_constraints` to the `sdv.utils` module. The function should take in the filepath where the constraints are stored, and return a list of the loaded constraints. Additionally, since loading constraints will now be handled through this utility function, we should deprecate the `set_constraints` method on the synthesizers.

#### API
`load_constraints(filepath)`
**Args**
- `filepath (str)`: The filepath where the constraints are stored.

**Returns**
A list of instantiated constraints loaded from the constraint JSON file. 

### Additional context
```python
from sdv.utils import load_constraints

constraints = load_constraints(
  filepath='my_constraints.json'
)

synthesizer.add_constraints(constraints)
```

gsheni:
- Fixes #2902

### DIFF
--- a/sdv/_utils.py
+++ b/sdv/_utils.py
@@ -507,6 +507,16 @@ def warn_load_deprecated():
     )
 
 
+def warn_set_constraints_deprecated():
+    """Warn that the `set_constraints` function on synthesizer is deprecated."""
+    deprecation_msg = (
+        'Warning: The `set_constraints` method is deprecated. '
+        'Please use the `load_constraints` utility function to load constraints from a file '
+        'and add them to the synthesizer with the `add_constraints` method.'
+    )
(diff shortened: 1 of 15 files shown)


frances-h (approved):
Looks good! I think there's some tests in Enterprise that need to be updated now that we're deprecating `set_constraints`

fealho (approved):

PR #2910, opened 2026-06-18
Add backwards compatibility for deprecated programmable constraint

### LINKED ISSUE #2909: Add backwards compatibility for deprecated programmable constraint
### Problem Description
The single table programmable constraint was recently deprecated in favor of using the multi-table functionality. However, the deprecation did not include backwards compatibility.

Currently if a user has a synthesizer with a single table programmable constraint that was created on a previous version of SDV, then they are unable to load it in an sample from it in the latest version of SDV. When sampling, there is an error for unsupported data types.

### Expected behavior
We should add a patch so that the backwards compatibility works for single table programmable constraints.

We can add a check on the harness and pass in just the dataframe if it's the older SingleTableProgrammableConstraint class.

### Additional context
Note that for any future constraints, we recommend customers talk to us directly so that we can provide them sandbox constraints. This is best for handling any complex logic; we are able to write it in the way that maintains synthetic data quality.

For simpler constraints, users can program in their logic using the new ColumnFormula constraint (in experimental state).

frances-h:
Resolve #2909

This PR restores the `SingleTableProgrammableConstraint` class and adds checks in the harness to revert to the old behavior if the programmable constraint  is a `SingleTableProgrammableConstraint`.

### DIFF
--- a/sdv/cag/programmable_constraint.py
+++ b/sdv/cag/programmable_constraint.py
@@ -3,6 +3,7 @@
 import inspect
 from copy import deepcopy
 
+from sdv.cag._errors import ConstraintNotMetError
 from sdv.cag.base import BaseConstraint
 
 
@@ -130,6 +131,12 @@ class ProgrammableConstraint:
         return synthetic_data
 
 
+class SingleTableProgrammableConstraint(ProgrammableConstraint):
+    """Single table variant of the base programmable constraint class."""
+
+    _is_single_table = True
+
+
 class ProgrammableConstraintHarness(BaseConstraint):
     """Constraint interface class for programmable constraints."""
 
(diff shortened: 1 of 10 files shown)


amontanez24 (approved):

sdv/cag/programmable_constraint.py
```diff
@@ -3,13 +3,14 @@
...
 
+from sdv.cag._errors import ConstraintNotMetError
 from sdv.cag.base import BaseConstraint
 
 
 class ProgrammableConstraint:
     """Base class to create programmable constraints."""
 
-    _is_single_table = True
+    _is_single_table = False
```
frances-h, line 13: We have to revert this so older multi-table `ProgrammableConstraint` work as expected.

sarahmish (approved):

PR #2929, opened 2026-07-21
Update HMA `PerformanceAlert` to be an error

### LINKED ISSUE #2918: Update HMA `PerformanceAlert` to be an error
### Problem Description
The HMASynthesizer that is part of SDV Community is meant to be a general education resource for showcasing synthetic data for very small, simple datasets. It not really designed to handle complex data schemas, such as schemas with more than 1 level of depth and more than 5 tables. For this, it is recommended to purchase SDV Enterprise, which provides access to synthesizers that scale better by _orders of magnitude_ (eg. what takes a couple hours for HMA takes a few minutes with SDV Enterprise).

Currently if you try to use HMASynthesizer with a complex schema, then SDV generates a `PerformanceAlert` explaining the details. For example:
```
PerformanceAlert: Using the HMASynthesizer on this metadata schema is not recommended. To model this data, HMA will generate a large number of columns. (10659 columns)


       Table Name  # Columns in Metadata  Est # Columns
bill_of_materials                      1              1
     comp_adaptor                     14             14
...

We recommend simplifying your metadata schema using 'sdv.utils.poc.simplify_schema'.
If this is not possible, please visit datacebo.com and reach out to us for enterprise solutions.
```

But it is easy for this to go unnoticed. Users may try fitting it only to find that it takes an unreasonable amount of time, when the solution all along is to upgrade to SDV Enterprise. 

Furthermore, it is not always feasible to simplify the schema, especially for complex datasets where all tables are important/necessary.

### Expected behavior
Instead of showing a PerformanceAlert, we should just turn it into an error outright. The error message should describe that SDV Community is not meant for complex schemas (re-iterating the above).

```python
```

```
SynthesizerInputError: HMASynthesizer is not designed to handle a schema with many tables and connections. 
Please use SDV Enterprise to model this schema. 

SDV Enterprise provides access to synthesizers that can easily scale with the amount of data and complexity of your schema.

For more information, visit datacebo.com
```

### Additional context
The trigger for this error should be the same.

frances-h:
Resolve #2918 

After discussion, we decided to simplify the error condition to any schema with more than five tables or a depth greater than 2.

### DIFF
--- a/sdv/multi_table/hma.py
+++ b/sdv/multi_table/hma.py
@@ -10,7 +10,7 @@ import pandas as pd
 from rdt.transformers import FloatFormatter
 from tqdm import tqdm
 
-from sdv._utils import _get_max_child_depth, _get_root_tables
+from sdv._utils import _get_root_tables
 from sdv.errors import SynthesizerInputError
 from sdv.multi_table.base import BaseMultiTableSynthesizer
 from sdv.sampling import BaseHierarchicalSampler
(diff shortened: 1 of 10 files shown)


sdv/multi_table/hma.py
```diff
@@ -184,6 +184,29 @@ def _estimate_num_columns(cls, metadata, distributions=None):
             table_name: sum(columns_list) for table_name, columns_list in columns_per_table.items()
         }
 
+    @staticmethod
+    def _get_max_schema_depth(metadata):
```
amontanez24: should we move this to the base?

tests/integration/multi_table/test_hma.py
```diff
@@ -655,162 +655,22 @@ def test_progress_bar_print(self, capsys):
...
-            'relationships': [
-                {
-                    'parent_table_name': 'parent',
-                    'parent_primary_key': 'parent_id',
-                    'child_table_name': 'child',
-                    'child_foreign_key': 'parent_id',
-                }
-            ],
-        })
+        metadata = get_multi_table_metadata()
```
amontanez24, line 661: which rule does this dataset break? Can we add two explicit tests for depth > 2 and num tables > 5? These tests might alreaady exist so maybe we can rename them

amontanez24 (approved):

pvk-developer (approved):

PR #2931, opened 2026-07-22
Update `save_resource` to handle downloading resources in nested folders

### LINKED ISSUE #2917: The `save_resource` for demo datasets should be able to handle filepaths (not just filenames)
### Problem Description
The [`save_resource` function](https://docs.sdv.dev/sdv/integration/utils/demo#save_resource) is a generic function that will allow a user to save any information that is associated with a given demo dataset. This function indexes into a particular dataset (for a given modality), and then retrieves files within that dataset's folder.

The problem is that it only retrieves single files for that dataset (eg. `README.txt`). But if a file is nested inside a folder, then it wouldn't work (eg. `examples/other_file.txt`).

### Expected Behavior
We should deprecate the current `resource_filename` parameter and replace it with a `resource_filepath` parameter. This new parameter should be able accept both filenames and filepaths. It is assumed that the user is already indexed into the dataset, so the filename/filepath would be read from that dataset.

```python
from sdv.datasets.demo import save_resource

save_source(
    modality='multi_table',
    dataset_name='Bupa',
    resource_filepath='examples/other_file.txt', # file is nested in a folder
    output_filepath='example_file.txt'
)

save_source(
    modality='multi_table',
    dataset_name='Bupa',
    resource_filepath='README.txt', # file is in the top-level space for the dataset
    output_filepath='README.txt'
)
```

### Additional context
The older `resource_filename` parameter should be backwards compatible and emit a `FutureWarning` if used.

gsheni:
- Fixes #2917 
- CU-86baqh1bb

### DIFF
--- a/sdv/datasets/demo.py
+++ b/sdv/datasets/demo.py
@@ -860,10 +860,11 @@ def get_readme(
 def save_resource(
     modality,
     dataset_name,
-    resource_filename,
-    output_filepath,
+    resource_filepath=None,
+    output_filepath=None,
     s3_bucket_name='sdv-datasets-public',
     credentials=None,
+    resource_filename=None,
 ):
     """Save the resource to disk.
 
@@ -871,9 +872,10 @@ def save_resource(
         modality (str):
             The modality of the dataset: ``'single_table'``, ``'multi_table'``, ``'sequential'``.
         dataset_name (str):
-            The name of the dataset to get the README for.
-        resource_filename (str):
-            The name of the file to download from S3.
+            The name of the dataset to get the resource for.
+        resource_filepath (str):
+            The location of the file to download from S3. This can be a filename or a filepath.
+            The location of the file should be under the dataset.
         output_filepath (str or None):
             Optional path where to save the file.
         s3_bucket_name (str, optional):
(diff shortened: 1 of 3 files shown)


sdv-team:
Task linked: [CU-86baqh1bb SDV - The `save_resource` for demo datasets should be able to handle filepaths (not just filenames) #2917](https://app.clickup.com/t/86baqh1bb)

pvk-developer (approved):

sdv/datasets/demo.py
```diff
@@ -860,20 +874,22 @@ def get_readme(
 def save_resource(
     modality,
     dataset_name,
-    resource_filename,
-    output_filepath,
+    resource_filepath=None,
+    output_filepath=None,
```
frances-h, line 878: ~Why are we defaulting these arguments to `None` now if they're both still required?~
Never mind, it's needed for backwards compatibility

sdv/datasets/demo.py
```diff
@@ -860,20 +874,22 @@ def get_readme(
...
         dataset_name (str):
-            The name of the dataset to get the README for.
-        resource_filename (str):
-            The name of the file to download from S3.
+            The name of the dataset to get the resource for.
+        resource_filepath (str):
+            The location of the file to download from S3. This can be a filename or a filepath.
+            The location of the file should be under the dataset.
         output_filepath (str or None):
             Optional path where to save the file.
```
frances-h: This docstring should be updated since we're not allowing `output_filepath` to be `None`

frances-h (approved):

sdv/datasets/demo.py
gsheni: yep, I did it for backwards compatibility and to keep the arguments in the same order.

sdv/datasets/demo.py
gsheni: Fixed:
https://github.com/sdv-dev/SDV/pull/2931/commits/816ca10743ccd5bf503a7db2f4bdfd5e6ab213df

PR #2938, opened 2026-08-04
Issue 2921 and 2920 overlaping check

### LINKED ISSUE #2920: Add a check for overlapping PII values between real and synthetic data
### Problem Description
If a value is listed as PII, we expect that that the synthetic data does not re-use the same values as the real data. It should be inventing new values instead.

We should write a function that explicitly measures this.

### Expected behavior
Write a function called `get_pii_overlap` that computes the overlap of PII values between the real and synthetic data. (I.e. values that appear in both datasets).

```python
from sdv.evaluation.utils import get_pii_overlap

overlap_amt = get_pii_overlap(
    real_data=real_data,
    table_name=table_name,
    pii_column_name=column_name,
    verbose=True
)
```
```
Number of common data points: 0 (0.0%)
✅ The synthetic data does not contain any PII values from the real data
```

Parameters:
- (required) `real_data`: A dictionary mapping a table name to a pandas DataFrame containing real data
- (required) `synthetic_data`: A dictionary mapping a table name to a pandas DataFrame containing synthetic dat
- (required) `table_name`: The name of the table that contains the PII column to check
- (required) `pii_column_name`: The name of the column that contains PII values to check
- `verbose`: Whether to print out the interpretation of the results

Returns: The # of overlapping PII values

Prints: If verbose, it should print: "Number of common data points: [number] ([percent])" The percent is defined as  (# overlapping values)/(# of unique values across all the real and synthetic data). After this, it should print out the interpretation:
- If there are 0 overlapping values, then print: "✅ The synthetic data does not contain any PII values from the real data"
- If there are only a few overlapping values (<2%), then print: "⚠️ The synthetic data contains a few PII values from the real data. This might be due to random chance."
- If there are more than a few overlapping values (>2%), then print: "❌ The synthetic data contains a significant number of the same PII values of as the real data. This might be due to a small number of possible PII values, a large sample of synthetic data, or a misconfiguration in your synthesizer."

### LINKED ISSUE #2921: Add a check for overlapping combinations of data values between the real and synthetic data
### Problem Description
Even if a column isn't explicitly PII, a combination of different values might be enough to de-identify the original data. For example, date of birth, zipcode, and gender.

SDV should offer a function that checks whether combinations of columns overlap between the real and synthetic data.

### Expected behavior
Write a function called get_combination_overlap that computes the overlap of combinations of column values between the real and synthetic data. (I.e. combinations that appear exactly the same in both datasets).

```python
from sdv.evaluation.utils import get_combination_overlap

overlap_amt = get_combination_overlap(
    real_data=real_data,
    table_name=table_name,
    column_names=['date_of_birth', 'zipcode', 'gender'],
    verbose=True
)
```
```
Number of common combinations: 0 (0.0%)
✅ The synthetic data does not contain any of the same combinations from the real data
```

Parameters:
- (required) `real_data`: A dictionary mapping a table name to a pandas DataFrame containing real data
- (required) `synthetic_data`: A dictionary mapping a table name to a pandas DataFrame containing synthetic dat
- (required) `table_name`: The name of the table that contains the columns to check
- (required) `column_names`: A list of strings represent the column names to check. Combinations of these columns will be checked.
- `verbose`: Whether to print out the interpretation of the results

Returns: The # of overlapping combinations

Prints: If verbose, it should print: "Number of common combinations: [number] ([percent])" The percent is defined as (# unique overlapping combinations)/(# of unique combinations across both real and synthetic data). After this, it should print out the interpretation:
- If there are 0 overlapping combinations, then print: "✅ The synthetic data does not contain any of the same combinations from the real data"
- If there are only a few overlapping combinations (<2%), then print: "⚠️ The synthetic data contains a few of the same combinations as the real data. This might be due to random chance."
- If there are more than a few overlapping combinations (>2%), then print: "❌ The synthetic data contains a significant number of the same combinations as the real data. This might be due to a small number of possible combinations, a large sample of synthetic data, or a misconfiguration in your synthesizer."

nemanja-vujic:
Add utils.py with two functions that measure how much the synthetic data reuses values from the real data.

get_combination_overlap counts combinations of column values that appear in both datasets

get_pii_overlap does the same for the values of a single PII column.

Resolves #2921
Resolves #2920

### DIFF
--- a/sdv/evaluation/utils.py
+++ b/sdv/evaluation/utils.py
@@ -4,9 +4,12 @@ import sys
 import warnings
 
 import pandas as pd
+from pandas.api.types import is_datetime64_any_dtype, is_numeric_dtype
 
 from sdv._utils import _cast_to_iterable, _check_is_dict_of_dataframes
 from sdv.metadata import Metadata
+
+MISSING_VALUE_PLACEHOLDER = '__sdv_missing_value__'
 
 
 def _validate_referential_integrity_inputs(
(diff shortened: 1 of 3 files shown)


sdv/evaluation/utils.py
```diff
@@ -0,0 +1,196 @@
...
+    'This might be due to random chance.'
+)
+SIGNIFICANT_PII_OVERLAP_MESSAGE = (
+    '❌ The synthetic data contains a significant number of the same PII values of as the '
+    'real data. This might be due to a small number of possible PII values, a large sample '
+    'of synthetic data, or a misconfiguration in your synthesizer.'
+)
+
+
+def _validate_data(real_data, synthetic_data, table_name, column_names):
```
sarahmish, line 160: maybe add a validation to check:
* `table_name` is a string
* `column_names` is a list of strings

sdv/evaluation/utils.py
```diff
@@ -0,0 +1,196 @@
...
+        ValueError:
+            If the table or any of the columns is missing from the data.
+    """
+    _validate_data(real_data, synthetic_data, table_name, column_names)
+    num_common, percent = _compute_overlap(real_data, synthetic_data, table_name, column_names)
+
+    if verbose:
+        print(f'Number of common combinations: {num_common} ({percent}%)')  # noqa: T201
+        if num_common == 0:
+            print(NO_OVERLAP_MESSAGE)  # noqa: T201
```
sarahmish: you can use `sys.stdout.write` instead of print

sdv/evaluation/utils.py
```diff
@@ -0,0 +1,196 @@
+"""Utility methods to compare the real and synthetic data."""
+
+import pandas as pd
+from pandas.api.types import is_datetime64_any_dtype, is_numeric_dtype
+
+MISSING_VALUE_PLACEHOLDER = '__sdv_missing_value__'
+
+NO_OVERLAP_MESSAGE = (
```
sarahmish: nit: since these messages are only used once, I would prefer to move them inside the function that uses them instead of creating a constant variable.

sdv/evaluation/utils.py
```diff
@@ -0,0 +1,196 @@
...
+            The column names to combine.
+
+    Returns:
+        tuple[int, float]:
+            The number of shared combinations and their percentage of all combinations.
+    """
+    real_values = real_data[table_name][column_names].copy()
+    synthetic_values = synthetic_data[table_name][column_names].copy()
+    for column_name in column_names:
+        real_values[column_name], synthetic_values[column_name] = _align_dtypes(
```
sarahmish, line 242: is this step necessary given that all the columns will be casted as `'object'` in `_get_combinations`?

tests/integration/evaluation/test_utils.py
```diff
@@ -0,0 +1,125 @@
...
+        synthetic_data={'customer': synthetic_data},
+        table_name='customer',
+        column_names=column_names,
+        verbose=False,
+    )
+
+    # Assert
+    real_combinations = set(real_data[column_names].itertuples(index=False, name=None))
+    assert isinstance(result, int)
+    assert 0 <= result <= len(real_combinations)
```
sarahmish: isn't the upper limit here the union of the unique combinations of real and synthetic?

tests/unit/evaluation/test_utils.py
```diff
@@ -0,0 +1,221 @@
...
+    get_pii_overlap,
+)
+
+
+def _wrap(real_table, synthetic_table):
+    """Return the real and synthetic data as single-table dictionaries."""
+    return {'table': real_table}, {'table': synthetic_table}
+
+
+class TestGetCombinations:
```
sarahmish: If we are testing a function, then the test will be written as a function. If we are testing a class, then the test will be written as a class.

In this case, `_get_combinations` is a function and the test function definition should be something like `test_{function_name}_{purpose_of_test}`.

nemanja-vujic:
@sarahmish all should be addressed in the previous commit

tests/unit/evaluation/test_utils.py
```diff
@@ -0,0 +1,273 @@
...
+    'This might be due to random chance.'
+)
+SIGNIFICANT_PII_OVERLAP_MESSAGE = (
+    '❌ The synthetic data contains a significant number of the same PII values of as the real '
+    'data. This might be due to a small number of possible PII values, a large sample of '
+    'synthetic data, or a misconfiguration in your synthesizer.'
+)
+
+
+def _wrap(real_table, synthetic_table):
```
pvk-developer: minor: I’d rename this to `_pandas_to_table_dicts` or something similar. `wrap` can be misleading in Python because it often suggests wrapping another function, e.g. with decorators or closures.

pvk-developer (approved):

sarahmish (approved):

PR #2946, opened 2026-08-14
Issue 2919 add a manual check for referential integrity

### LINKED ISSUE #2919: Add a manual check for referential integrity
### Problem Description
For multi-table datasets, it's very important for the synthetic data to maintain _referential integrity_, meaning that a foreign key column should contain no unknown/broken references. Currently, the [Diagnostic Evaluation](https://docs.sdv.dev/sdv/evaluation/diagnostic) runs a check for referential integrity.

However, it would be useful to add a function that runs through a few rows and provides another form of manual confirmation that referential integrity is met. This is useful for educational/illustrative purposes.

### Expected behavior
Add a function called `print_referential_integrity` that manually checks a few rows and prints what it has checked.

```python
from sdv.evaluation.utils import print_referential_integrity

print_referential_integrity(
  metadata=metadata,
  table_name='transaction',
  foreign_key_name='user_id',
  num_rows=10
)
```
```
Picking random transaction row: 1468732365
✅ Found user row! user_id: 86-25-19730100-15 

Picking random transaction row: 1988317645
✅ Found user row! user_id: 0-35-79964375-26 

Picking random transaction row: 1507984147
❌ Unable to find the linked user row
```

Parameters:
- (required) `metadata`: The metadata object
- (required) `synthetic_data`: A dictionary that maps each table name to a dataframe containing the synthetic data for it
- (required) `table_name`: A string containing the table name that has the foreign key to check
- (required) ~`foreign_key_column_name`~ `foreign_key_name`: A string with the column of the foreign key to check. For composite keys, this is a tuple of strings. _Edited on Jul 24 to account for the case of composite keys._
- `num_rows`: An int containing the number of columns to check; defaults to 10.

Returns: Nothing.

Behavior: This function should run a basic lookup and take the following steps:
1. Pick a random row (without replacement) from the table name provided. Print "Picking random <table_name> row". If the table name has a primary key, then print out the primary key's value of the chosen row.
2. For the given foreign key, perform a lookup to see if it exists in in the parent table (refer to the metadata to find the parent table).        
    - If it is found, print: " ✅ Found <parent_table> row! <parent_primary_key_name>: <parent_primary_key_value>"
    - If it is not found, print: "❌ Unable to find the linked <parent_table> row"
    - If the foreign key is null, print: "✅ Foreign key is null; no linked parent row expected"

### Additional context
For a comprehensive calculation of referential integrity across all rows and all foreign keys across a multi-table schema, we recommend running the [diagnostic evaluation](https://docs.sdv.dev/sdv/evaluation/diagnostic) instead of using this function. This function is for eduational/illustrative purposes only.

nemanja-vujic:
Resolves #2919

### DIFF
--- a/sdv/evaluation/utils.py
+++ b/sdv/evaluation/utils.py
@@ -61,17 +61,19 @@ def _get_parent_relationship(metadata, table_name, foreign_key_names):
             The columns making up the foreign key to look up.
 
     Returns:
-        tuple[str, list[str]]:
-            The parent table name and the columns making up its primary key.
+        tuple[str, list[str], list[str]]:
+            The parent table name, the columns making up its primary key, and the foreign key
+            columns in the order the relationship defines them.
     """
     for relationship in metadata.relationships:
         child_foreign_key = _cast_to_iterable(relationship['child_foreign_key'])
-        if table_name == relationship['child_table_name'] and child_foreign_key == list(
+        if table_name == relationship['child_table_name'] and set(child_foreign_key) == set(
             foreign_key_names
         ):
             return (
                 relationship['parent_table_name'],
                 _cast_to_iterable(relationship['parent_primary_key']),
+                child_foreign_key,
             )
 
     foreign_key = "', '".join(foreign_key_names)
@@ -118,7 +120,7 @@ def print_referential_integrity(
     foreign_key_names = _validate_referential_integrity_inputs(
         metadata, synthetic_data, table_name, foreign_key_name, num_rows
     )
-    parent_table_name, parent_primary_keys = _get_parent_relationship(
+    parent_table_name, parent_primary_keys, foreign_key_names = _get_parent_relationship(
         metadata, table_name, foreign_key_names
     )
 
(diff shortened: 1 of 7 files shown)


sdv-team:

sdv/evaluation/utils.py
```diff
@@ -0,0 +1,152 @@
...
+
+    Returns:
+        tuple[str, list[str]]:
+            The parent table name and the columns making up its primary key.
+    """
+    for relationship in metadata.relationships:
+        child_foreign_key = _cast_to_iterable(relationship['child_foreign_key'])
+        if table_name == relationship['child_table_name'] and child_foreign_key == list(
+            foreign_key_names
+        ):
```
amontanez24: will this fail if the foreign key names aren't in the same order? For example ['colA', 'colB', 'colC'] instead of ['colB', 'colC', 'colA'] ? I think those should still match

amontanez24 (commented):
@nemanja-vujic Can you add "resolves #2919" to the description of this PR. That will close the issue when the PR is merged

sdv/evaluation/utils.py
nemanja-vujic: Right, switched to comparing with sets and uses column order for the lookup. Added a test for this

amontanez24 (approved):

frances-h (approved):

PR #2964, opened 2026-08-25
SDV 2.0: Update metadata to include range and null info

### LINKED ISSUE #2952: SDV 2.0: Update metadata to include range and null info
### Problem Description
Currently, the metadata does not describe the range of possible values within the column. As a result, SDV assumes the range based on what is observed during the training data. 

The range information belongs in the metadata because it represents factual information about what the values are allowed to be.

### Expected behavior
For SDV 2.0, we should update the metadata with the following:

- Remove the `computer_representation` flag that is allowed in `numerical` today; it is no longer needed
- Allow the input of range information for different sdtypes. The range info is all optional. What is allowable is based on the sdtype:
    - For numerical and datetime columns: `range_min` and `range_max` should describe the lowest possible and highest possible value for that concept. For datetimes, the values should be in the same format as the datetime format (if provided).
    - For categorical columns: `range_values` should be a list of values that are possible within the column (excluding missing values)
    - For all columns: `range_is_nullable` is a boolean describing whether missing values are allowed in the column
- Add a field called `decimal_places` that is only available for numerical sdtypes. The value should be an integer describing the number of decimal places.
   - Note that `0` decimal places can either refer to integers or whole numbers (eg. a float column with `2.0`, `3.0`, etc.)
- When validating the metadata against the data: Make sure that the values in the data adhere to the range boundaries that are provided; if not, it is an invalid metadata error.
- When fitting a synthesizer: If the ranges are broader than what the training data has (eg. the min/max are lower/higher; the range values have some unseen categories; or the nullability is beyond what is observed), then show a warning. Extrapolating ranges is only available within the SDV Enterprise Targeted Sampling add-on.
- Update the metadata spec version to `V2`

Example:
```json
{
    "METADATA_SPEC_VERSION": "V2",
    "tables": {
        "users": {
            "age": { 
                "sdtype": "numerical", 
                "range_min": 0, 
                "range_max": 120,
                "decimal_places": 0
             },
            "tax_status": {
                "sdtype": "categorical", 
                "range_values": ["married_filing_jointly", "married_filing_separately", "head_of_household", "single", ... ],
                "range_is_nullable": true,
        }
    }
```

```python
synthesizer.fit(data)
```
```
Warning: The training data does not cover the full range. Synthetic data will be based on the training data.
To extrapolate ranges for full coverage, please use the Targeted Sampling bundle.
```

Other notes:
- API-wise: The existing functions should allow the users to input the range info programmatically. Eg.
```python
metadata.update_column(
  column_name='age',
  table_name='users',
  range_max=100
)
```
- If a synthesizer is instantiated with the `V1` version of the metadata (rather than V2), it is ok as all of the info is basically compatible. We should just throw a warning.
- _Don't change the metadata detection algorithm._ This will be a separate issue.

### Additional context
No backwards compatibility is needed since this is for SDV 2.0.

Merge into a feature branch for SDV 2.0, as there will be other issues that are also needed for this version.

frances-h:
Resolve #2952

### DIFF
--- a/sdv/data_processing/numerical_formatter.py
+++ b/sdv/data_processing/numerical_formatter.py
@@ -45,10 +45,15 @@ class NumericalFormatter:
     _rounding_digits = None
 
     def __init__(
-        self, enforce_rounding=False, enforce_min_max_values=False, computer_representation='Float'
+        self,
+        enforce_rounding=False,
+        enforce_min_max_values=False,
+        decimal_places=None,
+        computer_representation='Float',
     ):
-        self.enforce_rounding = enforce_rounding
+        self.enforce_rounding = enforce_rounding or (decimal_places is not None)
         self.enforce_min_max_values = enforce_min_max_values
+        self.decimal_places = decimal_places
         self.computer_representation = computer_representation
 
     def learn_format(self, column):
@@ -63,7 +68,8 @@ class NumericalFormatter:
             self._min_value = column.min()
             self._max_value = column.max()
 
-        if self.enforce_rounding:
+        self._rounding_digis = self.decimal_places
+        if self.enforce_rounding and self.decimal_places is None:
             self._rounding_digits = learn_rounding_digits(column)
 
     def format_data(self, column):
--- a/sdv/data_processing/numerical_formatter.py
+++ b/sdv/data_processing/numerical_formatter.py
@@ -9,16 +9,6 @@ from rdt.transformers.utils import learn_rounding_digits
 LOGGER = logging.getLogger(__name__)
 
 MAX_DECIMALS = sys.float_info.dig - 1
-INTEGER_BOUNDS = {
-    'Int8': (-(2**7), 2**7 - 1),
-    'Int16': (-(2**15), 2**15 - 1),
-    'Int32': (-(2**31), 2**31 - 1),
-    'Int64': (-(2**63), 2**63 - 1),
-    'UInt8': (0, 2**8 - 1),
-    'UInt16': (0, 2**16 - 1),
-    'UInt32': (0, 2**32 - 1),
-    'UInt64': (0, 2**64 - 1),
-}
 
 
 class NumericalFormatter:
@@ -33,10 +23,6 @@ class NumericalFormatter:
             Whether or not to clip the data returned by ``reverse_transform`` to the min and
             max values seen during ``fit``.
             Defaults to ``False``.
-        computer_representation (dtype):
-            Accepts ``'Int8'``, ``'Int16'``, ``'Int32'``, ``'Int64'``, ``'UInt8'``, ``'UInt16'``,
-            ``'UInt32'``, ``'UInt64'``, ``'Float'``.
-            Defaults to ``'Float'``.
     """
 
     _dtype = None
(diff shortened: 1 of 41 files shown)


sdv-team:

sdv/data_processing/numerical_formatter.py
```diff
@@ -45,10 +45,15 @@ class NumericalFormatter:
     _rounding_digits = None
 
     def __init__(
-        self, enforce_rounding=False, enforce_min_max_values=False, computer_representation='Float'
+        self,
+        enforce_rounding=False,
+        enforce_min_max_values=False,
+        decimal_places=None,
+        computer_representation='Float',
```
amontanez24: should we just get rid of this?

sdv/datasets/demo.py
```diff
@@ -192,11 +192,13 @@ def is_data_zip(key):
     )
 
 
-def _get_first_v1_metadata_bytes(contents, dataset_prefix, bucket, client):
+def _get_first_v2_metadata_bytes(contents, dataset_prefix, bucket, client):
     """Find and return bytes of the first V1 metadata JSON under `dataset_prefix`.
```
amontanez24:
```suggestion
    """Find and return bytes of the first V2 metadata JSON under `dataset_prefix`.
```

sdv/datasets/demo.py
```diff
@@ -429,7 +432,13 @@ def _get_metadata(metadata_bytes, dataset_name, output_folder_name=None):
     """
     try:
         metadict = json.loads(metadata_bytes)
-        metadata = Metadata().load_from_dict(metadict, dataset_name)
+        with warnings.catch_warnings():
+            warnings.filterwarnings(
+                'ignore',
+                category=FutureWarning,
+                module='sdv.metadata._single_table',
```
amontanez24, line 439: why is this raising a warning? I think we can remove any futurewarnings raised in the `_SingleTableMetadata`

sdv-team:

sdv/data_processing/numerical_formatter.py
sarahmish: should `computer_representation` be removed?

sdv/metadata/_single_table.py
```diff
@@ -1594,6 +1732,27 @@ def _valdiate_no_extra_keys_metadata_dict(self, metadata_dict):
...
+        sdtype = column_dict.get('sdtype')
+        if sdtype != 'numerical':
+            return column_dict
+
+        if 'computer_representation' in column_dict:
+            warnings.warn(
+                f"`computer_representation` key for numerical column '{column_name}' "
+                'is deprecated and will be ignored.',
+                category=FutureWarning,
+            )
```
amontanez24, line 1750: I thought we were dropping computer_representation from the metadata. Why do we need to support it

tests/unit/test__utils.py
```diff
@@ -1223,3 +1387,98 @@ def test__validate_boolean_parameter():
...
+            True,
+        ),
+        (
+            pd.Series([None] * 10),
+            {'sdtype': 'categorical', 'range_values': ['a', 'b', 'c']},
+            True,
+        ),
+    ],
+)
+def test__column_range_exceeds_real(column, col_meta, expected):
```
sarahmish, line 1447: can you add one more where `range_max` is defined but min isn't

sarahmish (approved):
LGTM!

sdv/metadata/_single_table.py
frances-h: We are, this handles the case where someone is loading V1 metadata with columns that have `computer_representation` set. Otherwise, if we try to load the demos, for example, then validation will fail since `computer_representation` isn't a valid parameter anymore.

sdv/datasets/demo.py
frances-h: V1 metadata might have `computer_representation`, which will error now with V2. For V1 compatibility, I just had the load method ignore it and throw a warning instead.

sdv-team:

amontanez24 (approved):

PR #2974, opened 2026-08-31
Add the necessary formatting when downloading the adventure-works demo dataset

### LINKED ISSUE #2971: Add the necessary formatting when downloading the `adventure-works` demo dataset
### Problem Description
Adventure works is a publicly available multi-table demo dataset that showcases an example of a complex schema. Unfortunately if you try to download the data/metadata for this dataset, it appears be inconsistent/invalid for modeling. 

```python

    modality='multi_table',
    dataset_name='adventure-works')

metadata.validate_data(data)
```
```
Errors in CountryRegion:
Error: Key column 'CountryRegionCode' contains missing values.

Errors in ProductListPriceHistory:
Error: Invalid values found for numerical column 'ListPrice': ['1000,4375', '1003,9100', '101,2400', '+ 117 more'].
...
```

The problem occurs because the original data (CSV) is present in a slightly different format that the loading doesn't handle right now. For example it parses the country code `NA` (for Namibia) as null, and it parses European-encoded decimals as strings.

This issue is for fixing this problem directly within the `download_demo` function -- but doing so in a _graceful_ way that will continue to work in the future, when we fix the underlying data (or data loading mechanism).

### Expected behavior
The `download_demo` should apply formatting updates for the `adventure-works` dataset in a graceful way.

Proposed solution: Download the dataset and metadata normally. If the dataset name is `adventure-works` and the modality is `multi_table`, then go through the following steps:

1.  Attempt to validate the data against the metadata (or a subset of it if that yields better performance). If it validates, then just return the data/metadata as-is. _This captures the case in the future when when we eventually fix the dataset._
2. Assuming it does not validate: Run a function to fix the data. This function might just run some commands to modify the data itself, or it might re-download the data with different parameters (such as parsing decimals as `,`). Essentially it runs whatever it needs to in order to try to make the data valid. Then, re-validate the data/metadata. _Put this function in a try/catch in case there are any errors._
3. If the function in (2) results in an error OR if the function in (2) still produces invalid data/metadata, then return the original (invalid) metadata/data with a warning to the user saying that this won't work.
4. Otherwise, return the now valid data/metadata.

Proposed warning text for step (3):
```
Invalid data warning: This downloaded data and metadata are not compatible with each other. They cannot be modeled using  
SDV without some modifications.
```

This solution is designed to work even in future cases where the adventure-works dataset is ultimately fixed and does not need any formatting. It is also designed to be resilient to any bad modification attempts.

### Additional context
The changes we are making are formatting and parsing modifications only. SDV Enterprise should be able to run on this complex dataset.

amontanez24:
resolves #2971

### DIFF
--- a/sdv/datasets/_utils.py
+++ b/sdv/datasets/_utils.py
@@ -1,0 +1,116 @@
+import warnings
+
+from pandas.core.dtypes.common import is_object_dtype
+
+
+class InvalidDataWarning(Warning):
+    pass
+
+
+def _process_adventure_works(data):
+    data = data.copy()
+    data['CountryRegion'].loc[data['CountryRegion']['Name'] == 'Namibia', 'CountryRegionCode'] = (
+        'NA'
+    )
+    integer_comma_columns = {
+        'PurchaseOrderDetail': [
+            'ReceivedQty',
+            'RejectedQty',
+            'StockedQty',
+        ],
+        'BillOfMaterials': ['PerAssemblyQty'],
+        'TransactionHistoryArchive': [],
+    }
+    float_comma_columns = {
+        'Product': ['StandardCost', 'ListPrice', 'Weight'],
+        'ProductListPriceHistory': ['ListPrice'],
+        'ProductCostHistory': ['StandardCost'],
+        'ProductVendor': ['StandardPrice', 'LastReceiptCost'],
+        'SalesOrderDetail': ['UnitPrice', 'UnitPriceDiscount', 'LineTotal'],
+        'SalesOrderHeader': ['SubTotal', 'TaxAmt', 'Freight', 'TotalDue'],
+        'PurchaseOrderDetail': [
+            'UnitPrice',
+            'LineTotal',
+        ],
+        'PurchaseOrderHeader': ['SubTotal', 'TaxAmt', 'Freight', 'TotalDue'],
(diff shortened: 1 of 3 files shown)


frances-h (approved):

R-Palazzo:
@amontanez24 could we also include this fix in this PR so we can remove it from the test on the feature branch?:

https://github.com/sdv-dev/SDV/blob/774acf963f37c9a9a099d14ab3a0d263acf1b665/tests/integration/multi_table/test_hma.py#L699

amontanez24:
> @amontanez24 could we also include this fix in this PR so we can remove it from the test on the feature branch?:
> 
> https://github.com/sdv-dev/SDV/blob/774acf963f37c9a9a099d14ab3a0d263acf1b665/tests/integration/multi_table/test_hma.py#L699

I don't understand, that fix is already on main @R-Palazzo

amontanez24:
> @amontanez24 could we also include this fix in this PR so we can remove it from the test on the feature branch?:
> 
> https://github.com/sdv-dev/SDV/blob/774acf963f37c9a9a099d14ab3a0d263acf1b665/tests/integration/multi_table/test_hma.py#L699

@R-Palazzo I will leave this PR as is, but once it is merged, we can add a line to cast that column as ordinal in the SDV 2.0 feature branch

sarahmish (approved):

